PR #532, opened 2014-09-04
Add app.ready() and Model ready hooks

ritch:
@bajtos @raymondfeng @fabien @kraman 

This adds a set of ready hooks to `Model` classes. Override these hooks to implement custom behavior that depends on your app being bootstrapped:

 - models attached to datasources, methods mixed in
 - relations, remoting, and other metadata defined
 - boot scripts run, `loopback-boot` complete
 - etc.

## Docs

**Do not call this method** if you are using `loopback-boot` to bootstrap your application.
`loopback-boot` will call this method for you! Otherwise you must call `app.ready()` to run
the `Model.ready()` hooks.

Calling `ready()` will call `ready()` on all models attached to the `app`. Override
the `ready()` method on a `Model` class to ensure you have access to a bootstrapped application.

```js
module.exports = function(MyModel) {
  MyModel.setup = function() {
    // setup is called when a model is extended
    // you must call `base.setup()` to extend the base class properly
    this.base.setup();

    // add or remove remote methods and otherwise modify the `Model`
    this.remoteMethod('myMethod');
  }

  MyModel.beforeReady = function(app, cb) {
    // async setup, runs after all models are `setup()`
    // and before `ready()` is called
    setTimeout(cb, 100);
  }

  MyModel.ready = function(app) {
    // MyModel and other classes can be used
    // you should not modify any classes in this method
    console.log(this.sharedClass.methods());
  }
}
```

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -10,6 +10,7 @@ var DataSource = require('loopback-datasource-juggler').DataSource
   , _ = require('underscore')
   , RemoteObjects = require('strong-remoting')
   , stringUtils = require('underscore.string')
+  , async = require('async')
   , path = require('path');
 
 /**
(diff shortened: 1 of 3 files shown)


ritch:
This was a dead end... Closing for now.

PR #905, opened 2014-12-09
Fixes duplicate Id on bulkUpdate.

BerkeleyTrue:
When using the replication algorithm, data changes with updates would cause the server to crash with duplicate id error. This PR fixes this issue.


ritch:
@BerkeleyTrue Sorry for the confusion... "test please" is a message our CI looks for. This message just triggers CI to run the tests again. I don't think we need to add a test for this change.

BerkeleyTrue:
Couldn't get this to pass tests, please check this one https://github.com/strongloop/loopback/pull/973 based on newer version of master.

Issue #931, opened 2014-12-17
suggestion: check emailVerified after valid password


greaterweb:
I would suggest moving the User model [`emailVerificationRequired` check](https://github.com/strongloop/loopback/blob/cb2f40bb86905439434a358debbd2e044a2de1ca/common/models/user.js#L186-L194) after [successful matching](https://github.com/strongloop/loopback/blob/cb2f40bb86905439434a358debbd2e044a2de1ca/common/models/user.js#L199) of the password (supplied to remote `/login` request). This would require that the user supply a valid username/password combination before the email verification check is made. 

For the scenario where an email is not verified, leaving it as is makes it possible to check if a username or email address exists. A password doesn't have to be supplied to the remote `/login` call, just either the `username` or `email`.

This may seem like trivial suggestion but it's an easy fix to prevent this type of behavior.

PR #936, opened 2014-12-18
Require valid login credentials before verified email check.

greaterweb:
Require valid login credentials before verified email check is made. 

PR for #931


ritch:
I think this change makes sense. Any thoughts @raymondfeng @bajtos ? Otherwise I will merge this.

bajtos:
@slnode ok to test

@ritch no objections

Issue #251, opened 2014-05-08
Password validators operate on hashed password, making `length` validators useless


STRML:
Modifying the User model in the following way does not trigger a validation error when attempting to create a user with an invalid password:

``` javascript
var User = app.models.User;
User.validatesLengthOf('password', {min: 6, message: {min: 'Password is too short'}});
var u = new User({email: "foo@bar.com", password: "short"});
assert(u.isValid()) // Error
```

Tracing it with `node-inspector` shows that the length is being compared to the hashed password, which is obviously always quite long.

bajtos:
> Password validators operate on hashed password, making `length` validators useless

@STRML is this still an issue for you? I see we landed https://github.com/strongloop/loopback/pull/941 which allows users to implement a workaround. Is that good enough?

@raymondfeng @ritch do you remember if we have any idea what a proper solution for this problem would look like?

Then there is https://github.com/strongloop/loopback/pull/941#issuecomment-98756858 mentioning this issue should be easy to fix/rework once we have the "persist" hook in place, which has happened quite some time ago - see https://github.com/strongloop/loopback-datasource-juggler/issues/559.

Perhaps that's the way to go? To keep the password plaintext all the way until the "persist" hook that is called after any validations were passed.

Thoughts?

av-k:
For example you can use context for temp data.

`  User.beforeRemote('create' , function(ctx, user, next) {
    var context = User.app.loopback.getCurrentContext();

    if (context) {
      context.set('currentPassword', ctx.req.body.password);
    }

    next();
  });`

And after validate it:
`  User.validateAsync('password', function (err, done) {
    var context = User.app.loopback.getCurrentContext(),
          currentPassword = context.get('currentPassword');

    if (currentPassword.length < config.validations.password.min) {
      err();
    }

    done();
  }, {message: 'Password is too short (minimum 6 symbols)'});`

JohnPhoto:
Is there any updates on this issue, what is current best practice?

I am not feeling overly confident in reimplementing a proper hashing myself and @av-k solution feels outdated now that `getCurrentContext()` is deprecated AFAIK (?)

akapaul:
I've found another way to do this. In common model **User** there is a method called **validatePassword**. If we extend our **UserModel** from **User**, we can redefine this method in JS, like following:

```javascript
var g = require('loopback/lib/globalize');

module.exports = function(UserModel) {
  UserModel.validatePassword = function(plain) {
    var err,
[...]

PR #941, opened 2014-12-20
Allow User.hashPassword/validatePassword to be overridden

raymondfeng:
/to @ritch @bajtos 

The PR allows User.hashPassword and validatePassword methods to be overridden. It can be used as a workaround to https://github.com/strongloop/loopback/issues/251.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -454,6 +454,24 @@ module.exports = function(User) {
   };
 
   /*!
+   * Hash the plain password
+   */
+  User.hashPassword = function(plain) {
+    this.validatePassword(plain);
+    var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
+    return bcrypt.hashSync(plain, salt);
+  };
+
+  User.validatePassword = function(plain) {
+    if (typeof plain === 'string' && plain) {
+      return true;
+    }
+    var err =  new Error('Invalid password: ' + plain);
+    err.statusCode = 422;
+    throw err;
+  };
+
+  /*!
    * Setup an extended user model.
    */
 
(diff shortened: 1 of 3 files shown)


bajtos:
What is the scenario where users should override `User.hashPassword`? I am concerned that most developers don't understand security implications of different hashing algorithms and thus can easily introduce a serious vulnerability.

common/models/user.js
```diff
@@ -454,6 +454,22 @@ module.exports = function(User) {
...
+    this.validatePassword(plain);
+    var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
+    return bcrypt.hashSync(plain, salt);
+  };
+
+  User.validatePassword = function(plain) {
+    if (typeof plain === 'string' && plain) {
+      return true;
+    }
+    throw new Error('Invalid password: ' + plain);
```
bajtos: How `throw new Error` going to integrate with the existing validations? The request setting invalid password should return HTTP status 422 and a description of the validation error in the usual format. This code will IMO return 500 without enough details.

bajtos:
LGTM.

bajtos:
please check why the CI builds are failing before you land this patch.

cleas:
User.create({...}, function (err,user){...} ) will cause an uncaught exception as  
validatePassword throws an error if the supplied password is invalid. This is a change in behavior , I would expect the error to be returned in the callback function.

raymondfeng:
Using the setter to hash passwords is pretty evil. We should find a better solution. Maybe the validatePassword should be set up as a validator on the User model.

bajtos:
Once we have "persist" hook in place (https://github.com/strongloop/loopback-datasource-juggler/issues/559), it should be very easy to rework the User model to hash the password in this hook.

Let's keep the scope of this discussion around the change in behaviour introduced by this pull request. @cleas can you post an example that works in `loopback@2.8` and fails in `loopback@2.9`?

cleas:
example : https://github.com/cleas/loopback-sandbox

bajtos:
@cleas can you please open a new github issue to track this problem? I'll post some ideas how to fix it there.

Issue #622, opened 2014-10-07
Rework built-in models to use `Model.remoteMethod`


bajtos:
Rework all built-in models to register remoting metadata via `Model.remoteMethod` instead of `loopback.remoteMethod` and pass method name instead of method function as the first argument.

That way it will be easier to override methods in subclassing models - see #619.

loay:
@bajtos 
I see the PR already merged. Can I close this one?
Thanks.

bajtos:
@loay I am proposing to mark `loopback.remoteMethod` as deprecated. The deprecation message should include instructions what to do instead - either call `Model.remoteMethod` or even better, place the metadata into "methods" in the model JSON file.

Thoughts?

bajtos:
Other than that (or after that is landed?), this issue can be closed, indeed.

loay:
I misunderstood, I thought the PR merged before is what was required.

PR #943, opened 2014-12-22
Use User.remoteMethod instead of loopbacks method

BerkeleyTrue:
This is needed for loopback-connector-remote authorization.
Addresses https://github.com/strongloop/loopback/issues/622.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -319,7 +319,7 @@ module.exports = function(User) {
       options.port +
       options.restApiRoot +
       userModel.http.path +
-      userModel.confirm.http.path +
+      userModel.sharedClass.find('confirm', true).http.path +
       '?uid=' +
       options.user.id +
       '&redirect=' +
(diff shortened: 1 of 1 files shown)


common/models/user.js
```diff
@@ -319,7 +319,7 @@ module.exports = function(User) {
       options.port +
       options.restApiRoot +
       userModel.http.path +
-      userModel.confirm.http.path +
+      '/confirm' +
```
bajtos: This is changing configurable `confirm.http.path` to hard-coded `/confirm`, isn't that a breaking change?

BerkeleyTrue: Maybe, but there isn't a way that I am aware of to get the information dynamically. Using Model.remoteMethod does not make these properties available like loopback.remoteMethod does.

bajtos: > get the information dynamically. Using Model.remoteMethod does not make these properties available like loopback.remoteMethod does.

You need to call `sharedClass.find` to get the method metadata:

```
userModel.sharedClass.find('confirm', true).http.path
```

BerkeleyTrue: Sweet! Good to know. I'll make the changes tomorrow morning.

ritch:
triggering the CI tests:

test please

bajtos:
@slnode test please

@ritch IIRC Jenkins looks at the first line of the comment only

PR #955, opened 2014-12-27
Fix Change.getCheckpointModel() giving new models each call

alFReD-NSH:
This was a huge memory leak in our app... About more than 120MB per hour!


ritch:
LGTM

Issue #685, opened 2014-10-23
Cover the "ttl" property?


Spark-star:
I created a model "Account" which extended from "User", I defined the property "ttl" like this:

```
{

  "name": "Account",
  "plural": "accounts",
  "base": "User",
  "Model": {
    "Master": "master"
  },
  "ttl": 31449600,
  "properties": {
    "cellphone": {
      "type": "string",
      "required": true
    }
  },
"relations": {
    "accessTokens": {
      "type": "hasMany",
      "model": "AccessToken",
      "foreignKey": "accountId"
    }
}
...
}
```

But when I use "Account.login" with given username and password , the responses' "ttl" has not been set to   31449600,  which is still 1209600.

Is it a bug?

The max value the ttl can set is 31556926.

loay:
@Spark-star 
This PR should have fixed the issue:
https://github.com/strongloop/loopback/pull/956

Thanks

loay:
Hi @Spark-star 
The issue is closed for now since the PR should have fixed it. If you are still running into problems, feel free to leave a comment and I will reopen the issue.
Thanks.

shaheero:
@loay using Loopback 3.6, this is still an issue.

PR #956, opened 2014-12-28
fix User.settings.ttl can't be overridden in sub model

clarkorz:
see https://github.com/strongloop/loopback/issues/685

Signed-off-by: Clark Wang <clark.wangs@gmail.com>


ritch:
LGTM

PR #962, opened 2015-01-01
fix nestRemoting is nesting hooks from other relations

clarkorz:
Signed-off-by: Clark Wang <clark.wangs@gmail.com>


raymondfeng:
ok to test

PR #972, opened 2015-01-05
Update README.md

ranacseruet:
I was trying to build the code and it failed. then I discovered that, current commit status is unstable. Thus just adding the build status so that other developers can easily discover the state directly on repo homepage.


slnode:

bajtos:
Hi @ranacseruet, we are intentionally not including build status badge in the README. The build occasionally fails on Travis due to timing issues, even though the tests correctly pass. The master branch should be always build-able and pass all tests, or at least that's our goal.

PR #973, opened 2015-01-05
Fix duplicate Id on bulkUpdate.

BerkeleyTrue:
When using the replication algorithm, data changes with updates would cause the server to crash with duplicate id error. This PR fixes this issue.

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -943,9 +943,11 @@ PersistedModel.bulkUpdate = function(updates, callback) {
   updates.forEach(function(update) {
     switch (update.type) {
       case Change.UPDATE:
+        tasks.push(function(cb) {
+          Model.upsert(update.data, cb);
+        });
+        break;
       case Change.CREATE:
-        // var model = new Model(update.data);
-        // tasks.push(model.save.bind(model));
         tasks.push(function(cb) {
           var model = new Model(update.data);
           model.save(cb);


bajtos:
@BerkeleyTrue please add a unit test verifying the fix.

@ritch could you please review? The change looks innocent to me, though I am concerned about possible side effects, as I am not that familiar with this part of the replication implementation.

BerkeleyTrue:
@bajtos Can you point me to the proper place to put a unit test for this?

bajtos:
> Can you point me to the proper place to put a unit test for this?

I'd say `test/replication.test.js`. Can you confirm, @ritch?

Another candidate is `test/model.test.js`.

ritch:
replication.test.js

ritch:
@BerkeleyTrue if you don't want to add the test LMK and I'll create a new PR with the fix + test.

BerkeleyTrue:
@ritch Yes please. Not sure when I can come back around to this.

bajtos:
@BerkeleyTrue what's the scenario to trigger the bug you have discovered? At least when using a memory connector, updates are handled correctly for me - see e.g. [test/replication.test.js](https://github.com/strongloop/loopback/blob/8bfa3ba8d7489d04839e9c653ef747d35b889ed7/test/replication.test.js#L767-L782).

BerkeleyTrue:
We are using the postgres connector. I have had many issue before that are not covered by using the memory connector.

BerkeleyTrue:
@bajtos I can verify this is still an issue with newer version of loopback.

loopback 2.14
loopback datasource juggler 2.19.2
loopback connector postgres 1.6.1

using loopback replication

bajtos:
I am reworking this part of the code in https://github.com/strongloop/loopback/pull/1214. Would you mind testing it out in your app?

BerkeleyTrue:
I will give it a try when app is back to previous working state. Currently chasing down a number of bugs from upgrade.

bajtos:
> We are using the postgres connector. I have had many issue before that are not covered by using the memory connector.

https://github.com/strongloop/loopback-datasource-juggler/pull/510 is fixing the memory connector to reject duplicate CREATE in the same way as most databases do, this will hopefully allow us to catch more problems like you have described right in loopback test suite.

If there is any other edge case where the memory connector behaves differently than PostgreSQL, then please let us know. Ideally, the memory connector should be as close to real databases as possible.

I am closing this pull request in favour of #1214. Please open a new GH or send a new patch if there is still a bug even after #1214 was landed.

bajtos:
Oh, one more thing: #1214 requires changes in the postgresql connector, see https://github.com/strongloop/loopback/issues/1167#issuecomment-83930450

BerkeleyTrue:
got it.

Issue #416, opened 2014-07-29
After remote hook not called when user confirms registration


johnsoftek:
user.beforeRemote('confirm', beforeConfirm) 
    user.afterRemote('confirm', afterConfirm) 

beforeConfirm is called.

afterConfirm is not called following successful remote call to User.confirm.

I am using user model based on User.

bajtos:
Pull request in review: https://github.com/strongloop/loopback/pull/1066

bajtos:
This should have been fixed by https://github.com/strongloop/loopback/commit/8766d4a68d014402dd8ed70295eb3b9d8a635c0d since `loopback@2.13.0`.

PR #976, opened 2015-01-07
Allow user to hook the confirm method

socketwiz:
If a user overrides the User model, they might want to hook the confirm method so they can perform some action in addition to the redirect like maybe send a confirmation email.


slnode:

BerkeleyTrue:
Does this work for you? Because it seems like the redirect would cause the user's browser to make a new request and anything done after the confirm would not be able to affect the client application.

bajtos:
See the discussion in #416. The correct solution is to remove the redirect (the whole afterRemote hook). To preserve backwards compatibility, this new behaviour must be explicitly enabled by a flag.

PR #978, opened 2015-01-07
Allow accessType per remote method

raymondfeng:
/to @ritch 

It depends on https://github.com/strongloop/strong-remoting/pull/147.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -314,6 +314,25 @@ Model._getAccessTypeForMethod = function(method) {
 
   var ACL = Model._ACL();
 
+  // Check the explicit setting of accessType
+  if (method.accessType) {
+    assert(method.accessType === ACL.READ ||
+      method.accessType === ACL.WRITE ||
+      method.accessType === ACL.EXECUTE, 'invalid accessType ' +
+      method.accessType +
+      '. It must be "READ", "WRITE", or "EXECUTE"');
+    return method.accessType;
+  }
+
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -314,6 +314,25 @@ Model._getAccessTypeForMethod = function(method) {
 
   var ACL = Model._ACL();
 
+  // Check the explicit setting of accessType
+  if (method.accessType) {
+    assert(method.accessType === ACL.READ ||
```
ritch: Can we avoid hard coding these? I know we have the enum... but that is already a bit static. Custom access types might be nice in the future.

lib/model.js
```diff
@@ -314,6 +314,25 @@ Model._getAccessTypeForMethod = function(method) {
...
+    return method.accessType;
+  }
+
+  // Default GET requests to READ
+  var verb = method.http && method.http.verb;
+  if (typeof verb === 'string') {
+    verb = verb.toUpperCase();
+  }
+  if (verb === 'GET' || verb === 'HEAD') {
+    return ACL.READ;
```
ritch: Maybe we should allow the user to specify this via `config.json`. You could just check `app.get('access').defaultAccessTypes[verb]`.

ritch:
Feel free to ignore my suggestions. LGTM

raymondfeng:
Good suggestions for extensibility of access types. I'll leave the enhancements to future work.

PR #980, opened 2015-01-08
Added object type for push settings

glesage:
After looking at suggestion from @raymondfeng in https://github.com/strongloop/loopback/pull/946
I was realizing that the pushSettings member of Application model was converted to varchar for some reason... not sure why but setting the type to "object" fixed it.

Are there other cleanups to do in the declaration file?


slnode:

glesage:
It does not seem like pushSettings is typed as a model originally... when logging the type conversion in mysql.js's various functions (and looking straight in the DB) it is saves as a varchar(4096). Which is too small to include all the SSL certificates for APN & GCM.

However, when explicitely setting it's type to "object" it is then saved properly in the DB as "TEXT"

glesage:
Where can I see how the object is typed as an anonymous model?

glesage:
To expand on this, with the current version of Loopback, when creating my data:
```
mysql> describe Application;
+-----------------------+---------------+------+-----+---------+-------+
| Field                 | Type          | Null | Key | Default | Extra |
+-----------------------+---------------+------+-----+---------+-------+
| id                    | varchar(255)  | NO   | PRI | NULL    |       |
| realm                 | varchar(512)  | YES  |     | NULL    |       |
| name                  | varchar(512)  | NO   |     | NULL    |       |
| description           | varchar(512)  | YES  |     | NULL    |       |
| icon                  | varchar(512)  | YES  |     | NULL    |       |
| owner                 | varchar(512)  | YES  |     | NULL    |       |
| collaborators         | varchar(4096) | YES  |     | NULL    |       |
| email                 | varchar(512)  | YES  |     | NULL    |       |
| emailVerified         | tinyint(1)    | YES  |     | NULL    |       |
| url                   | varchar(512)  | YES  |     | NULL    |       |
| callbackUrls          | varchar(4096) | YES  |     | NULL    |       |
| permissions           | varchar(4096) | YES  |     | NULL    |       |
| clientKey             | varchar(512)  | YES  |     | NULL    |       |
| javaScriptKey         | varchar(512)  | YES  |     | NULL    |       |
| restApiKey            | varchar(512)  | YES  |     | NULL    |       |
| windowsKey            | varchar(512)  | YES  |     | NULL    |       |
| masterKey             | varchar(512)  | YES  |     | NULL    |       |
| pushSettings          | varchar(4096) | YES  |     | NULL    |       |
| authenticationEnabled | tinyint(1)    | YES  |     | NULL    |       |
| anonymousAllowed      | tinyint(1)    | YES  |     | NULL    |       |
| authenticationSchemes | varchar(4096) | YES  |     | NULL    |       |
| status                | varchar(512)  | YES  |     | NULL    |       |
| created               | datetime      | YES  |     | NULL    |       |
| modified              | datetime      | YES  |     | NULL    |       |
+-----------------------+---------------+------+-----+---------+-------+
24 rows in set (0.00 sec)
```

pushSettings is still varchar(4096)

glesage:
actually according to Mysql module, the default is VARCHAR:
https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L1038

Should I switch that and make:
```
case 'JSON':
case 'Object':
case 'Any':
case 'Text':
	dt = columnType(p, 'TEXT');
	dt = stringOptionsByType(p, dt);
	break;
```
the default?

raymondfeng:
The default is the same as `string`. I think we should fix it. If the type is a custom type (a LoopBack model), it should be default to Any (TEXT). Would you like to submit a patch?

glesage:
Yes, that is what I meant. I will submit a patch ASAP (:

glesage:
https://github.com/strongloop/loopback-connector-mysql/pull/68

PR #996, opened 2015-01-12
status middleware: add `scopeToGetRoot` option

bajtos:
Before this change, the status middleware was implemented as a route,
i.e. it was always handling all requests.

This commit introduces a new config option `scopeToGetRoot`. When the
option is set, the middleware handler will respond only to `GET /`
and pass any other request down the middleware chain.

See strongloop/generator-loopback#80

/to @ritch or @raymondfeng please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "loopback-datasource-juggler": "^2.8.0",
     "loopback-testing": "~0.2.0",
     "mocha": "~1.21.4",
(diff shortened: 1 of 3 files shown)


ritch:
The change looks good. @crandmck or @bajtos - we'll need to add `scopeToGetRoot` to the config.json wiki page.

bajtos:
Closing this PR as rejected after our discussion in strongloop/loopback-workspace#182

PR #998, opened 2015-01-13
Optimize the creation of handlers for rest

raymondfeng:
/to @bajtos or @ritch

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -23,27 +23,27 @@ module.exports = rest;
  */
 
 function rest() {
+  var handlers; // Cached handlers
+
   return function restApiHandler(req, res, next) {
     var app = req.app;
-    var restHandler = app.handler('rest');
 
     if (req.url === '/routes') {
-      return res.send(restHandler.adapter.allRoutes());
+      return res.send(app.handler('rest').adapter.allRoutes());
     } else if (req.url === '/models') {
       return res.send(app.remotes().toJSON());
     }
 
-    var preHandlers;
-
-    if (!preHandlers) {
-      preHandlers = [];
+    if (!handlers) {
+      handlers = [];
       var remotingOptions = app.get('remoting') || {};
 
       var contextOptions = remotingOptions.context;
       if (contextOptions !== false) {
-        if (typeof contextOptions !== 'object')
+        if (typeof contextOptions !== 'object') {
           contextOptions = {};
-        preHandlers.push(loopback.context(contextOptions));
+        }
+        handlers.push(loopback.context(contextOptions));
       }
 
       if (app.isAuthEnabled) {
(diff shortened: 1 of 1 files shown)


server/middleware/rest.js
```diff
@@ -23,9 +23,15 @@ module.exports = rest;
...
+  var router; // Cache
+
   return function restApiHandler(req, res, next) {
     var app = req.app;
-    var restHandler = app.handler('rest');
+    var restHandler;
+
+    if (!router) {
+      restHandler = app.handler('rest');
+    }
```
bajtos: This is a breaking change. Note that `app.handler('rest')` is already caching the handler fn. However, when a new model is added, this cache is cleared, so that the next call of `app.handler('rest')` builds a new handler fn that includes the new model too.

bajtos: See the CI test results for proof.

raymondfeng:
@bajtos Regression fixed.

server/middleware/rest.js
```diff
@@ -57,9 +59,16 @@ function rest() {
...
-      handler(req, res, done);
-    }, next);
+    if (preHandlers.length > 0) {
+      // Build a router to chain all handlers
+      var router = loopback.Router();
+      for (var i = 0, n = preHandlers.length; i < n; i++) {
+        router.use(preHandlers[i]);
+      }
+      router.use(restHandler);
+      return router(req, res, next);
```
bajtos: I am surprised that building a new Router for each request is faster than running `async.each`. Did the benchmark confirm your assumption?

FWIW, since the list of pre-handlers is static, it should be possible to cache the Router instance and just wrap `restHandler` in a function, so that it's dynamically fetched for each request.

```js
var router = loopback.Router();
for (var i = 0, n = preHandlers.length; i < n; i++) {
  router.use(preHandlers[i]);
}
router.use(function(req, res, next) { return app.handler('rest')(res, req, next); });
```

Of course, you need to run the benchmark to see which version is faster.

raymondfeng:
Now we cache the router instance.

bajtos:
@slnode test please

bajtos:
LGTM. For posterity, what is the performance improvement you have measured?

raymondfeng:
@bajtos Benchmarking shows some overhead in express Router. I reverted it to async.eachSeries() and optimize it for the single handler case. 

Here are some throughput numbers (16 threads * 5000 transactions per thread):

- Without this PR: 2835/s, 2875/s, 2711/s
- With this PR: 2911/s, 2907/s, 2824/s

bajtos:
>  I reverted it to async.eachSeries() and optimize it for the single handler case.

FWIW, that's IMO not the typical use case. I'd expect that most people run their server with authentication enabled.

PR #999, opened 2015-01-13
Remove unnecessary allocations from rest handler

bajtos:
Correctly cache the list of `preHandlers` so that they are built only once in application lifetime.

Remove array concatenation, use a new callback function instead.

/to @raymondfeng please review

Note: the problem addressed by this patch was pointed out in strongloop/loopback-benchmarks#1 as a possible improvement. I run the benchmark again with the patched version of loopback, but I did not observe any performance improvement greater that the variations in the measured data.

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -23,6 +23,8 @@ module.exports = rest;
  */
 
 function rest() {
+  var preHandlers;
+
   return function restApiHandler(req, res, next) {
     var app = req.app;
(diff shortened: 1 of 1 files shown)


bajtos:
LOL, apparently we were both thinking about the same problem.

@raymondfeng feel free to close this in favour of your #998.

raymondfeng:
See https://github.com/strongloop/loopback/pull/998

PR #1008, opened 2015-01-14
Fix/removes use of deprecated req.param method

aars:
Removes use of deprecated req.param() method.

req.param() would look in req.params, req.query and req.body. I tried to make logical choices on where to look for certain params (e.g. don't lookup redirect parameter in request body). Personally I would not allow an access-token in the request body and only as a query parameter or header, but people might depend on it so I kept it.

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -165,13 +165,15 @@ module.exports = function(AccessToken) {
     var i = 0;
     var length;
     var id;
+    var param;
 
     params = params.concat(['access_token']);
     headers = headers.concat(['X-Access-Token', 'authorization']);
     cookies = cookies.concat(['access_token', 'authorization']);
 
     for (length = params.length; i < length; i++) {
-      id = req.param(params[i]);
+      param = params[i];
+      id = req.params[param] || req.query[param];
 
       if (typeof id === 'string') {
         return id;
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -564,7 +564,11 @@ module.exports = function(User) {
     UserModel.on('attached', function() {
       UserModel.afterRemote('confirm', function(ctx, inst, next) {
         if (ctx.req) {
-          ctx.res.redirect(ctx.req.param('redirect'));
+          // Find redirect parameter as param or query. Not in body.
+          var redirectTo = (ctx.req.params && ctx.req.params.redirect) ||
+            (ctx.req.query && ctx.req.query.redirect);
+
+          ctx.res.redirect(redirectTo);
         } else {
           next(new Error('transport unsupported'));
         }
--- a/lib/application.js
+++ b/lib/application.js
@@ -295,13 +295,20 @@ app.enableAuth = function() {
   remotes.before('**', function(ctx, next, method) {
     var req = ctx.req;
     var Model = method.ctor;
-    var modelInstance = ctx.instance;
-    var modelId = modelInstance && modelInstance.id || req.param('id');
+    var modelInstance = ctx.instance || {};
+    // req.params and req.query default to {}. req.body can be undefined.
+    var modelId = modelInstance.id ||
+                  req.params.id    ||
+                  req.query.id     ||
+                  (req.body && req.body.id);
 
     var modelSettings = Model.settings || {};
-    var errStatusCode = modelSettings.aclErrorStatus || app.get('aclErrorStatus') || 401;
-    if (!req.accessToken) {
-      errStatusCode = 401;
+
+    var errStatusCode = 401;
+    if (req.accessToken) {
+      errStatusCode = modelSettings.aclErrorStatus ||
+                      app.get('aclErrorStatus')    ||
+                      errStatusCode;
     }
 
     if (Model.checkAccess) {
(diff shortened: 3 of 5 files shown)


slnode:

aars:
I'm not sure how to handle the appVeyor error.

common/models/access-token.js
```diff
@@ -171,7 +171,8 @@ module.exports = function(AccessToken) {
     cookies = cookies.concat(['access_token', 'authorization']);
 
     for (length = params.length; i < length; i++) {
-      id = req.param(params[i]);
+      // Only look up params in query. There is no body here.
+      id = req.query && req.query[params[i]];
```
bajtos: IMO you should check `req.params` too.

lib/application.js
```diff
@@ -296,7 +296,9 @@ app.enableAuth = function() {
     var req = ctx.req;
     var Model = method.ctor;
     var modelInstance = ctx.instance;
-    var modelId = modelInstance && modelInstance.id || req.param('id');
+    var modelId = (modelInstance && modelInstance.id) ||
+      (req.params && req.params.id) ||
+      (req.body && req.body.id);
```
bajtos: Note that `req.param` checked `req.body` before `req.params` as can be seen [here](https://github.com/strongloop/express/blob/4db51fbeba0cd543949e243672c0e6c3c3c70639/lib/request.js#L213-L215).

Please add a unit-test to prevent such regressions from resurfacing in the future.

bajtos: Correction: `req.body` before `req.query`

common/models/access-token.js
dougwilson: And, really, unless this is used outside of Express, `req.query` is guaranteed to be an object.

lib/application.js
dougwilson: Hopefully `req.params` should be guaranteed to be an object at this location.

aars: I put it at the bottom because `req.param()` does indeed check the body first, but the id should not be in the request body. Should both `req.params.id` and `req.body.id` I believe the body should be ignored.

Why not remove `req.body.id`?

common/models/access-token.js
aars: Ok, I'll remove the check on `req.query`.

lib/application.js
aars: And since `req.params` defaults to `{}` I'd suggest simply:
``` js
var modelId = (modelInstance) ? modelInstance.id : req.params.id;
```

common/models/access-token.js
aars: I'm refactoring this, but there is one test-case that mocks a request that does _not_ contain `req.params` or `req.query`. I've contacted @bajtos since it's his test.

lib/application.js
bajtos: > but the id should not be in the request body

That's AFAIK not true, e.g. `updateOrCreate` takes the id from the request body.

More importantly, we must preserve backwards compatibility with previous loopback versions. So even if the `id` was not supposed to be read from the body, I would argue that we should still check the body for the sake of staying backwards compatible.

aars: `updateOrCreate` is a good point! Though I wonder if a requestor should ever be able to determine a `modelId`, but that is another discussion. (Backwards compatibility is not always a good thing) Let's keep the `req.body` lookup.

@bajtos Do you agree that `req.params.id` should take precedence over `req.body.id`?

bajtos: >  Do you agree that `req.params.id` should take precedence over `req.body.id`?

Yes, we must preserve the behaviour of `req.param()`.

common/models/user.js
```diff
@@ -564,7 +564,11 @@ module.exports = function(User) {
     UserModel.on('attached', function() {
       UserModel.afterRemote('confirm', function(ctx, inst, next) {
         if (ctx.req) {
-          ctx.res.redirect(ctx.req.param('redirect'));
+          // Find redirect parameter as param or query. Not in body.
+          var redirectTo = (ctx.req.params && ctx.req.params.redirect) ||
+            (ctx.req.query && ctx.req.query.redirect);
```
bajtos: Why the redirect parameter cannot be in the body? Unless you can prove that such request would not be valid, or that such usage is explicitly forbidden by the docs, then this patch must check `req.body` too in the same way as `req.param()` does.

lib/application.js
```diff
@@ -295,13 +295,20 @@ app.enableAuth = function() {
...
     var modelSettings = Model.settings || {};
-    var errStatusCode = modelSettings.aclErrorStatus || app.get('aclErrorStatus') || 401;
-    if (!req.accessToken) {
-      errStatusCode = 401;
+
+    var errStatusCode = 401;
+    if (req.accessToken) {
+      errStatusCode = modelSettings.aclErrorStatus ||
+                      app.get('aclErrorStatus')    ||
+                      errStatusCode;
```
bajtos: What is the purpose of this change? It looks unrelated to `req.param()` fixes.

bajtos:
> I'm not sure how to handle the appVeyor error.

Please ignore that, @cgole is working on disabling that integration.

bajtos:
Closing in favour of #1038 to get this fixed ASAP.

Issue #1011, opened 2015-01-14
ACL permission checking


jrhone:
Reproduction:
Create a remote method on a model that is called via GET, do not set the access type (it will default to READ since the verb is GET)

    Poll.remoteMethod(
        'inbox',
        {
            accepts: [{ arg: 'limit', type: 'number' }],
            returns: { arg: 'results', type: 'object' },
            http: { verb: 'get' },
            description: 'Get inbox'
        }
    );

Set the ACL perms on the model json (poll.json) to be 
    {
      "accessType": "EXECUTE",
      "principalType": "ROLE",
      "principalId": "$authenticated",
      "permission": "ALLOW",
      "property": "inbox"
    }

You should receive an Authorization Required message.  EXECUTE does not seem to be superset'ing READ

ritch:
@jrhone does https://github.com/strongloop/loopback/pull/1014 fix your issue?

raymondfeng:
Fixed by https://github.com/strongloop/loopback/issues/1014

fabien:
@raymondfeng I'm having issues after updating an older app. If I gather correctly, it means that allowing EXECUTE will also allow WRITE. Is this correct? To me EXECUTE should be useful to distinguish between any other custom operation and explicit READ or WRITE.

raymondfeng:
EXECUTE allows you to READ or WRITE.

PR #1014, opened 2015-01-14
Make sure EXECUTE access type matches READ or WRITE

raymondfeng:
/to @ritch 

The PR fixes a regression reported by https://github.com/strongloop/loopback/issues/1011.


ritch:
LGTM

PR #1015, opened 2015-01-14
Extend AccessToken to parse Basic auth headers

rmg:
Allow convenient URLs for curl and browsers such as:
 - http://some-long-token@localhost:3000/
 - http://token:some-long-token@localhost:3000/

Basic Auth specifies a 'Basic' scheme for the Authorization header
similar to how OAuth specifies 'Bearer' as an auth scheme.

Following a similar convention, extract the access token from the
Authorization header when it specifies the 'Basic' scheme, assuming
it is the larger of the \<user\>:\<pass\> segments.

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -189,6 +189,20 @@ module.exports = function(AccessToken) {
           // Decode from base64
           var buf = new Buffer(id, 'base64');
           id = buf.toString('utf8');
+        } else if (/^Basic /i.test(id)) {
+          id = id.substring(6);
+          id = (new Buffer(id, 'base64')).toString('utf8');
+          // The spec says the string is user:pass, so if we see both parts
+          // we will assume the longer of the two is the token, so we will
+          // extract "a2b2c3" from:
+          //   "a2b2c3"
+          //   "a2b2c3:"   (curl http://a2b2c3@localhost:3000/)
+          //   "token:a2b2c3" (curl http://token:a2b2c3@localhost:3000/)
+          //   ":a2b2c3"
+          var parts = /^([^:]*):(.*)$/.exec(id);
+          if (parts) {
+            id = parts[2].length > parts[1].length ? parts[2] : parts[1];
+          }
         }
         return id;
       }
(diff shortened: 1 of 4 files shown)


common/models/access-token.js
```diff
@@ -189,6 +189,25 @@ module.exports = function(AccessToken) {
...
+          //   ":a2b2c3"
+          if (id.indexOf(':')) {
+            // get the longest part in <part1>:<part2>
+            id = [
+              id.slice(0, id.indexof(':')),
+              id.slice(id.indexOf(':'))
+            ].reduce(function (a, b) {
+              return a.length > b.length ? a : b;
+            });
+          }
```
raymondfeng: If the `id` doesn't contain `:`, we'll get wrong token. For example,

> '1234'.slice(-1) ==> '4'
> '1234'.slice(0, -1) ==> '123'

rmg: Ya, that was just shamefully sloppy of me. I'm changing it to a style I'm more comfortable with.

raymondfeng:
LGTM. @ritch What do you think?

ritch:
:+1: 

Wanted to add this a while ago!

rmg:
I'll rebase and merge shortly. I want to reread the code with fresh eyes and double check before merging that this won't interfere with any auth middleware people might be using.

rmg:
@raymondfeng @ritch I think this is ready to merge if someone wants to sign off on it.

rmg:
Our own CI is failing on iojs only, which is the same as the master branch. Travis seems to be intermittently failing on master as well, so I don't think this PR is breaking anything.

bajtos:
> Our own CI is failing on iojs only, which is the same as the master branch.

Yeah, `npm install phantomjs` is failing on io.js, see my patch here: https://github.com/AvianFlu/ncp/pull/78

raymondfeng:
LGTM

bajtos:
> > Our own CI is failing on iojs only, which is the same as the master branch.

> Yeah, `npm install phantomjs` is failing on io.js, see my patch here: https://github.com/AvianFlu/ncp/pull/78

Strangely enough, it seems that our build is not choking on phantomjs. Is it using a globally installed instance?

bajtos:
Anyways, let's not worry about io.js, I am slowly working to get that fixed.

rmg:
Yes, the phantomjs binary is pre-installed in our CI environment. The npm module detects it and uses it.

PR #1022, opened 2015-01-19
Update use of request.param(name) to request.params[name]

bradplank:
Express deprecated the use of request.param(name).

Signed-off-by: bradplank <brad@bradplank.com>


slnode:

bradplank:
I agree with bajtos, request.param(name) is not the same as request.params[name].

bajtos:
FWIW, here is another pull request fixing this problem: #1008.

@bradplank I would still like to thank you for the effort you made to contribute this change. Keep going :+1:

bradplank:
Next time I will check for open pull requests first :stuck_out_tongue_winking_eye:

PR #1038, opened 2015-01-21
Remove usages of deprecated `req.param()`

bajtos:
Express has recently deprecated `req.param()` to force developers to be explicit about the source of the value. To avoid deprecation warnings, this commit replaces all calls of `req.param()` with a simplified inline version.

This supersedes a community-contributed PR https://github.com/strongloop/loopback/pull/1008 that is suboptimal.

/to @raymondfeng or @ritch please review

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -172,7 +172,12 @@ module.exports = function(AccessToken) {
     cookies = cookies.concat(['access_token', 'authorization']);
 
     for (length = params.length; i < length; i++) {
-      id = req.param(params[i]);
+      var param = params[i];
+      // replacement for deprecated req.param()
+      id = req.params && req.params[param] !== undefined ? req.params[param] :
+        req.body && req.body[param] !== undefined ? req.body[param] :
+        req.query && req.query[param] !== undefined ? req.query[param] :
+        undefined;
 
       if (typeof id === 'string') {
         return id;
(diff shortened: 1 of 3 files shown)


common/models/access-token.js
```diff
@@ -172,7 +172,12 @@ module.exports = function(AccessToken) {
     cookies = cookies.concat(['access_token', 'authorization']);
 
     for (length = params.length; i < length; i++) {
-      id = req.param(params[i]);
+      var param = params[i];
+      // replacement for deprecated req.param()
+      id = req.params && req.params[param] !== undefined ? req.params[param] :
+        req.body && req.body[param] !== undefined ? req.body[param] :
+        req.query && req.query[param] !== undefined ? req.query[param] :
+        undefined;
```
raymondfeng: Do we want to refactor the compat logic to be utility function?

bajtos: I'd rather not. The utility function already exists, it's called `req.param()`, right? Ideally, we should revisit this code at the time of 3.x release and keep only the sources that make sense. For example, I am not sure if it makes sense to pass the access token in the URL path (`req.params`) . Or perhaps we should force the users to be more specific about where to get the token from (body, query, path param).

bajtos: If you think the utility function is worth it, where should it be implemented, so that we can share it between strong-remoting and loopback?

raymondfeng:
LGTM

Issue #698, opened 2014-10-24
Per-model prefix for the REST path (e.g. version or domain)


wascou:
Hi,

I haven't found the answer in the issues stack neither in the documentation, so here's my question.
Is it possible to prefix all routes to API exposed resources by a string which can be used either to scope the API by domain either for versioning purpose ?

Cheers and good job guys.

Max

bajtos:
#1039 added a test to verify and document this contract. I am closing this issue a resolved.

PR #1039, opened 2015-01-22
Add test case for loopback issue #698

PradnyaBaviskar:
@bajtos , I've Added new test case to verify custom http path - as mentioned in "loopback issue 698". Can you please verify?

### DIFF
--- a/test/rest.middleware.test.js
+++ b/test/rest.middleware.test.js
@@ -99,6 +99,18 @@ describe('loopback.rest', function() {
       .expect(200, done);
   });
 
+  it('allows models to provide a custom HTTP path', function(done) {
+    var ds = app.dataSource('db', { connector: loopback.Memory });
+    var CustomModel = ds.createModel('CustomModel',
+      {name: String},
+      {'http': { 'path': 'domain1/CustomModelPath' }
+    });
+
+    app.model(CustomModel);
+    app.use(loopback.rest());
+    request(app).get('/domain1/CustomModelPath').expect(200).end(done);
+  });
+
   it('includes loopback.token when necessary', function(done) {
     givenUserModelWithAuth();
     app.enableAuth();


test/rest.middleware.test.js
```diff
@@ -99,6 +99,15 @@ describe('loopback.rest', function() {
       .expect(200, done);
   });
 
+  it("allows models to provide a custom HTTP path", function(done) {
+    var ds = app.dataSource('db', { connector: loopback.Memory });
+    var CustomModel = ds.createModel('CustomModel', {name: String}, {"http": { "path": "domain1/CustomModelPath" } });
```
bajtos: We try to keep the lines at most 80 characters long, that way the diffs can be reviewed without horizontal scrolling.

Please split this line.

test/rest.middleware.test.js
```diff
@@ -99,6 +99,15 @@ describe('loopback.rest', function() {
...
+
+    app.model(CustomModel);
+    app.use(loopback.rest());
+    request(app).get(CustomModel.http.path).expect(200).end(done);
```
bajtos: Please use an explicit string instead of `CustomModel.http.path`. It is more robust, as it is testing exactly the URL that we expected. Your version is relying on LoopBack to build the path. If LoopBack ignored the http.path setting passed to `ds.createModel`, your version of this test will probably still pass.

bajtos:
> Added test case to verify custom http path

We use imperative tense in commit messages, see [A note about git commit  messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). Please reword the message.

bajtos:
Here is what I would recommend to you: when you fixed the code per my comments above, run `git commit --amend` instead of `git commit`. This way your changes will be added to the last commit and you will have a chance to reword the commit message. Then you need to force-push the changes using `git push -f`.

bajtos:
One more thing: your code does not pass code style check, please run `npm test` and fix all errors reported. You can also see the errors in the Travis CI build output - see the links at the bottom of this page.

```
Running "jscs:test" (jscs) task
Invalid quote mark found at test/rest.middleware.test.js :
   100 |  });
   101 |
   102 |  it("allows models to provide a custom HTTP path", function(done) {
-------------^
   103 |    var ds = app.dataSource('db', { connector: loopback.Memory });
   104 |    var CustomModel = ds.createModel('CustomModel', {name: String}, {"http": { "path": "domain1/CustomModelPath" } });
Invalid quote mark found at test/rest.middleware.test.js :
   102 |  it("allows models to provide a custom HTTP path", function(done) {
   103 |    var ds = app.dataSource('db', { connector: loopback.Memory });
   104 |    var CustomModel = ds.createModel('CustomModel', {name: String}, {"http": { "path": "domain1/CustomModelPath" } });
-----------------------------------------------------------------------------^
   105 |
   106 |    app.model(CustomModel);
Invalid quote mark found at test/rest.middleware.test.js :
   102 |  it("allows models to provide a custom HTTP path", function(done) {
   103 |    var ds = app.dataSource('db', { connector: loopback.Memory });
   104 |    var CustomModel = ds.createModel('CustomModel', {name: String}, {"http": { "path": "domain1/CustomModelPath" } });
---------------------------------------------------------------------------------------^
   105 |
   106 |    app.model(CustomModel);
Invalid quote mark found at test/rest.middleware.test.js :
   102 |  it("allows models to provide a custom HTTP path", function(done) {
   103 |    var ds = app.dataSource('db', { connector: loopback.Memory });
   104 |    var CustomModel = ds.createModel('CustomModel', {name: String}, {"http": { "path": "domain1/CustomModelPath" } });
-----------------------------------------------------------------------------------------------^
   105 |
   106 |    app.model(CustomModel);
>> 4 code style errors found!
```

The javascript convention is to use single quotes (`'`), not double quotes (`"`).

PradnyaBaviskar:
@bajtos,
I've made following changes. Can you please review?

 - Verified that the code lines do not exceed, 80 characters

 - Used explicit string "/domain1/CustomModelPath", instead of CustomModel.http.path. One thing I noticed - when creating model, I used http path as "domain1/CustomModelPath", but when asserting the test, I had to prefix "/" to the get path - only then the test would succeed. Is that desired?

- Replaced double quotes with single quote.

test/rest.middleware.test.js
```diff
@@ -99,6 +99,18 @@ describe('loopback.rest', function() {
       .expect(200, done);
   });
 
+  it('allows models to provide a custom HTTP path', function(done) {
+    var ds = app.dataSource('db', { connector: loopback.Memory });
+    var CustomModel = ds.createModel('CustomModel',
+      {name: String},
+      {'http': { 'path': 'domain1/CustomModelPath' }
```
bajtos: Coding style: we put spaces inside square brackets, e.g. `{ name: String }`. Also property names (keys) don't need quotes, i.e. `{ http: {...} }`.

bajtos:
> Add test case for loopback issue #698 …
> New test case to verify per-model prefix for the REST path
> @bajtos, can you please review?

The commit message should describe only the change made by the commit. Comments related to the review process belong to the Pull Request. That way when the commit is merged, the git history contains only the relevant information (what was changed).

bajtos:
I made the fixes myself in 7dcc2fc and landed the patch. Thank you for the contribution!

bajtos:
> Used explicit string "/domain1/CustomModelPath", instead of CustomModel.http.path. One thing I noticed - when creating model, I used http path as "domain1/CustomModelPath", but when asserting the test, I had to prefix "/" to the get path - only then the test would succeed. Is that desired?

I suppose it would be more correct to specify the http.path as "/domain1/CustomModelPath". AFAIK the strong-remoting layer "normalizes" the path and prepends the missing slash, so it's ok to keep it this way.

Issue #1065, opened 2015-02-04
User.confirm has 'redirect' as required argument


umbrella-ivan-zhukov:
User.confirm remote method shouldn't have `redirect` as required argument, because you can request this endpoint via ajax.

Also, the code allows to do this (from common/models/user.js)

```
var redirectUrl =
            params && params.redirect !== undefined ? params.redirect :
            body && body.redirect !== undefined ? body.redirect :
            query && query.redirect !== undefined ? query.redirect :
            undefined;

          // redirect user if path defined
          if (ctx.req.param('redirect'))
            ctx.res.redirect(redirectUrl);
          else
            ctx.res.sendStatus(200)
        } else {
          next(new Error('transport unsupported'));
```

umbrella-ivan-zhukov:
Me bad, above I posted code from my patch, but now it's in PR.
Now trying to pass tests...

ritch:
Agree for supporting `confirm()` without `redirect` set.

bajtos:
This issue is a duplicate of #416. I'll keep it open though since there is already a pull request associated with this issue.

umbrella-ivan-zhukov:
Setting Location header and calling `next()` (like @ritch says) is now triggering afterRemote hook (for duplicate issue).
But I have to make some change in strong-remoting repo. Please, take a look on PR (linked above).

pthieu:
Closing because issue has been resolved: https://github.com/strongloop/strong-remoting/pull/174 points to https://github.com/strongloop/strong-remoting/issues/175

PR #1066, opened 2015-02-04
Setted 'redirect' arg as optional for User.confirm remote

umbrella-ivan-zhukov:
issue #1065

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -582,6 +582,7 @@ module.exports = function(User) {
 
           if (redirectUrl !== undefined) {
             ctx.res.location(redirectUrl);
+            ctx.res.status(302);
           }
           next()
         } else {
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -581,10 +581,9 @@ module.exports = function(User) {
             undefined;
 
           if (redirectUrl !== undefined) {
-            ctx.res.redirect(redirectUrl);
-          } else {
-            ctx.res.sendStatus(204);
+            ctx.res.location(redirectUrl);
           }
+          next()
         } else {
           next(new Error('transport unsupported'));
         }
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -583,7 +583,7 @@ module.exports = function(User) {
           if (redirectUrl !== undefined) {
             ctx.res.redirect(redirectUrl);
           } else {
-            ctx.res.sendStatus(204)
+            ctx.res.sendStatus(204);
           }
         } else {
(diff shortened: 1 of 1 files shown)


slnode:

common/models/user.js
```diff
@@ -580,7 +580,11 @@ module.exports = function(User) {
             query && query.redirect !== undefined ? query.redirect :
             undefined;
 
-          ctx.res.redirect(redirectUrl);
+          if (redirectUrl !== undefined) {
+            ctx.res.redirect(redirectUrl);
+          } else {
+            ctx.res.sendStatus(204);
```
ritch: This should just call `next()`

ritch:
ok to test

bajtos:
Please see the discussion in #416 too.

umbrella-ivan-zhukov:
I performed an update.

common/models/user.js
```diff
@@ -580,7 +580,10 @@ module.exports = function(User) {
             query && query.redirect !== undefined ? query.redirect :
             undefined;
 
-          ctx.res.redirect(redirectUrl);
+          if (redirectUrl !== undefined) {
+            ctx.res.location(redirectUrl);
```
bajtos: Per my comment in https://github.com/strongloop/strong-remoting/pull/174, this code should set both response location and status code.

common/models/user.js
```diff
@@ -580,7 +580,10 @@ module.exports = function(User) {
...
+          }
+          next()
         } else {
           next(new Error('transport unsupported'));
```
bajtos: The error should be reported only when there is `redirectUrl`. When the url is not provided then there is no reason to throw the error.

umbrella-ivan-zhukov: Can you, please, tell, why should I rise `transport unsupported` error if `redirectUrl` is passed?

bajtos: If there is a `redirectUrl` parameter, then the caller of the API is expecting to receive a redirect. Since non-REST transports (JSON-RPC, WebSocket) don't support redirects, an error should be reported in such case, which is the current behaviour.

bajtos:
Please add a unit-test verifying the implementation and make sure `npm test` passes on your machine. At the moment, you are missing a semicolon on line 587.

ritch:
@umbrella-ivan-zhukov I'm closing this PR for now. If you make the requested changes we can merge it.

Also if you don't have time to make the update please just open an issue and ping me (@ritch) and I'll finish the PR.

PR #1072, opened 2015-02-04
Document user settings

ritch:
See https://github.com/strongloop/loopback/issues/1070

/to @raymondfeng 
/cc @bajtos @dashby3000 @superkhau


bajtos:
FWIW, LGTM.

PR #1083, opened 2015-02-10
Fix #1080 - domain memory leak.

### LINKED ISSUE #1080: Large memory leak caused by context.js's domains
This regression was introduced in https://github.com/strongloop/loopback/commit/ca0208ddd936a8573f1af6eb18edaff329fd4571.

I am documenting as I go - this was causing large memory usage in my application that would eventually cause it to crash. 

An example of its memory usage (I updated from loopback `2.8.4` to latest on Feb 1):

Imgur

To replicate this, clone [loopback-example-app](https://github.com/strongloop/loopback-example-app) and run:

`ab -n 1000 -c 10 "http://127.0.0.1:3000/api/cars"`

Simple as that. It appears that the domain is not properly being torn down after the request has ended.

Here's a few snapshots of some heapdumps:

Imgur

Here's the retainers (coming back to `currentDomain`). Note that this is a comparison between Snapshot 5 and Snapshot 6, so the listed items are those retained between the snapshots. All I did was run a thousand hits via `ab`.

Imgur

STRML:
Domains are created per request rather than creating a single
domain for all requests. This kills the memory leak, as a single
domain would keep a reference to every single req and res, causing
rapid heap growth.

### DIFF
--- a/server/middleware/context.js
+++ b/server/middleware/context.js
@@ -46,12 +46,6 @@ function context(options) {
   var enableHttpContext = options.enableHttpContext || false;
   var ns = createContext(scope);
 
-  var currentDomain = process.domain = domain.create();
-  currentDomain.oldBind = currentDomain.bind;
-  currentDomain.bind = function(callback, context) {
-    return currentDomain.oldBind(ns.bind(callback, context), context);
-  };
-
   // Return the middleware
   return function contextHandler(req, res, next) {
     if (req.loopbackContext) {
@@ -61,6 +55,12 @@ function context(options) {
     // Bind req/res event emitters to the given namespace
     ns.bindEmitter(req);
     ns.bindEmitter(res);
+
+    var currentDomain = domain.create();
+    currentDomain.oldBind = currentDomain.bind;
+    currentDomain.bind = function(callback, context) {
+      return currentDomain.oldBind(ns.bind(callback, context), context);
+    };
 
     currentDomain.add(req);
     currentDomain.add(res);


STRML:
See #1080.

After applying this fix, `node-memwatch` reports no leaks after 10000 hits to `/api/cars`.

The heap sits at about 90MB after the stress test, as opposed to just over 300MB before this patch.

`ab` stats:

```
~/g/f/l/server (master|✚1) $ ab -n 10000 -c 10 "http://127.0.0.1:3000/api/cars"
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:
Server Hostname:        127.0.0.1
Server Port:            3000

Document Path:          /api/cars
Document Length:        204962 bytes

Concurrency Level:      10
Time taken for tests:   295.533 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      2052380000 bytes
HTML transferred:       2049620000 bytes
Requests per second:    33.84 [#/sec] (mean)
Time per request:       295.533 [ms] (mean)
Time per request:       29.553 [ms] (mean, across all concurrent requests)
Transfer rate:          6781.91 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       2
Processing:    89  295  27.5    286     448
Waiting:       88  295  27.5    286     448
Total:         90  295  27.5    286     448

Percentage of the requests served within a certain time (ms)
  50%    286
  66%    303
  75%    312
  80%    317
  90%    335
  95%    348
  98%    366
  99%    377
 100%    448 (longest request)
```

We should consider how we might write a test for `loopback` that would catch leaks.

STRML:
Not sure what Travis' problem is, the `0.10` build is fine but `0.11` fails.

raymondfeng:
LGTM

STRML:
Can we talk about how we might write a test to prevent this in the future? This went unnoticed in my test environments due to lower API traffic and caused several crashes in production.

raymondfeng:
Sure, do you have any ideas in mind to measure/verify the memory footprint? For example, http://nodejs.org/api/process.html#process_process_memoryusage? Maybe we should have a set of `benchmark` tests, for performance and memory.

STRML:
`memwatch` does a pretty good job as well - it will emit `leak` events if ten consecutive GCs result in a growing heap.

I think we should do some basic client/server stress test in a `benchmarks` folder, and fail if a `leak` event is thrown.

STRML:
When can we expect a release? This is a critical leak.

wpjunior:
This release is very important for us, Many our servers are consuming a expensive CPU and memory. 
Thanks.

PR #1112, opened 2015-02-20
Fix test broken by recent juggler changes

bajtos:
The patch strongloop/loopback-datasource-juggler#436 changed the way how `Model.extend` works, which broke a loopback test relying on the old behaviour.

This commit fixes the failing test. The test is checking now that the model base was not changed, instead of checking that the base is undefined.

/cc @raymondfeng @PradnyaBaviskar

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "underscore.string": "~2.3.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

PR #1116, opened 2015-02-20
Fix change detection & tracking

### LINKED ISSUE #1048: Refactor change tracking to use the new hook infrastructure
The new hook API: https://github.com/strongloop/loopback-datasource-juggler/pull/403

~~Once change tracking uses the new hooks, the following issues should become resolved/fixed:
 https://github.com/strongloop/loopback/issues/617~~

bajtos:
Add unit-tests to verify that all DAO methods correctly create change records.

Rework the change detection to use the new operation hooks, this fixes the bugs where operations like "updateOrCreate" did not update change records.

Close #1048 

/to @ritch please review
/cc @raymondfeng @BerkeleyTrue

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -755,10 +755,10 @@ PersistedModel.changes = function(since, filter, callback) {
   filter.fields[idName] = true;
 
   // TODO(ritch) this whole thing could be optimized a bit more
-  Change.find({
+  Change.find({ where: {
     checkpoint: {gt: since},
     modelName: this.modelName
-  }, function(err, changes) {
+  }}, function(err, changes) {
     if (err) return callback(err);
     if (!Array.isArray(changes) || changes.length === 0) return callback(null, []);
     var ids = changes.map(function(change) {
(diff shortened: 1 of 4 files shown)


lib/persisted-model.js
```diff
@@ -1072,6 +1066,54 @@ PersistedModel.enableChangeTracking = function() {
...
+function rectifyOnSave(ctx, next) {
+  if (ctx.instance) {
+    ctx.Model.rectifyChange(ctx.instance.getId(), reportErrorAndNext);
+  } else {
+    ctx.Model.rectifyAllChanges(reportErrorAndNext);
+  }
+
+  function reportErrorAndNext(err) {
+    if (err) {
+      console.error(
```
ritch: -1 for hardcoded `console.error()`. I'd prefer just about anything else. At least allow the user to provide a function to call when an error occurs.

ritch:
Only one major comment. I'm fine if you merge as is and create an issue to implement it or convince me it isn't required.

lib/persisted-model.js
bajtos: The current implementation uses `console.error` too, see [lib/persisted-model.js#L1066-L1071](https://github.com/strongloop/loopback/blob/a82b33ec5c3cd043c53fed222ece2db67fc77c3c/lib/persisted-model.js#L1066-L1071). There are other places where PersistedModel prints to the console on error, I'd rather fix this properly in a standalone pull request - see #1126

PR #1120, opened 2015-02-23
Add fix to override afterRemote comfirm method

PradnyaBaviskar:
to / @bajtos -
In reference to https://github.com/strongloop/loopback/issues/416 , 

I've made changes so that redirect URL is an optional param. And in case redirect URL is specified then the response status is 302.

For this to work correctly, the strong remoting fix - https://github.com/strongloop/strong-remoting/pull/183, will have to be taken up first.

Can you please review?

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -561,7 +561,7 @@ module.exports = function(User) {
         accepts: [
           {arg: 'uid', type: 'string', required: true},
           {arg: 'token', type: 'string', required: true},
-          {arg: 'redirect', type: 'string', required: true}
+          {arg: 'redirect', type: 'string'}
         ],
         http: {verb: 'get', path: '/confirm'}
       }
@@ -580,21 +580,14 @@ module.exports = function(User) {
 
     UserModel.on('attached', function() {
       UserModel.afterRemote('confirm', function(ctx, inst, next) {
-        if (ctx.req) {
-          // replacement for deprecated req.param()
-          var params = ctx.req.params;
-          var body = ctx.req.body;
-          var query = ctx.req.query;
-          var redirectUrl =
-            params && params.redirect !== undefined ? params.redirect :
-            body && body.redirect !== undefined ? body.redirect :
-            query && query.redirect !== undefined ? query.redirect :
-            undefined;
-
-          ctx.res.redirect(redirectUrl);
-        } else {
-          next(new Error('transport unsupported'));
+        if (ctx.args.redirect !== undefined) {
+          if (!ctx.res) {
+            return next(new Error('The transport does not support HTTP redirects.'));
+          }
+          ctx.res.location(ctx.args.redirect);
+          ctx.res.status(302);
         }
+        next();
       });
     });
 
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -858,6 +858,28 @@ describe('User', function() {
         }, done);
       });
 
+      it('Should report 302 when redirect url is set', function(done) {
+        testConfirm(function(result, done) {
+          request(app)
+            .get('/users/confirm?uid=' + (result.uid) +
+              '&token=' + encodeURIComponent(result.token) +
+              '&redirect=http://foo.com/bar')
+            .expect(302)
+            .expect('Location', 'http://foo.com/bar')
+            .end(done);
+        }, done);
+      });
+
+      it('Should report 204 when redirect url is not set', function(done) {
+        testConfirm(function(result, done) {
+          request(app)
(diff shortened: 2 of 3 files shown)


common/models/user.js
```diff
@@ -591,7 +591,11 @@ module.exports = function(User) {
...
 
-          ctx.res.redirect(redirectUrl);
+          if (redirectUrl !== undefined) {
+            ctx.res.location(redirectUrl);
+            ctx.res.status(302);
+          }
+          next();
         } else {
           next(new Error('transport unsupported'));
         }
```
bajtos: While you are changing this part, please rework the implementation as follows:

 1. Use `ctx.args.redirectUrl` to obtain the parameter value, don't use `ctx.req` directly.
 2. Don't throw `Error("transport unsupported")` when `redirectUrl` is not set

bajtos:
@slnode test please

test/user.test.js
```diff
@@ -858,6 +858,31 @@ describe('User', function() {
...
 
+      it('Should report 302 when redirect url is set', function(done) {
+        testConfirm(function(result, done) {
+          request(app)
+            .get('/users/confirm?uid=' + (result.uid) +
+              '&token=' + encodeURIComponent(result.token) +
+              '&redirect=' + encodeURIComponent(options.redirect))
+            .expect(302);
+
+          done();
```
bajtos: You have to pass `done` as a callback of `expect` and/or `end`.

The current implementation marks the test as done before the HTTP request is finished.

The same applies to the next test below.

```js
.expect(302)
.end(done);
```

PradnyaBaviskar:
@bajtos,

I've modified the code, so as to use ```ctx.args.redirectUrl```, rather than using ```ctx.req```.
Also modified the test cases as suggested above.

Please review.

common/models/user.js
```diff
@@ -580,21 +580,11 @@ module.exports = function(User) {
...
-            undefined;
-
-          ctx.res.redirect(redirectUrl);
-        } else {
-          next(new Error('transport unsupported'));
+        if (ctx.args.redirect !== undefined) {
+          ctx.res.location(ctx.args.redirect);
+          ctx.res.status(302);
         }
+        next();
```
bajtos: `ctx.res` may be undefined when non-REST transport like WebSockets is used.

Here is a mockup of what I'd like to see here: 

```js
if (ctx.args.redirect !== undefined) {
  if (!ctx.res) {
    return next(new Error('The transport does not support HTTP redirects.'));
  }
  // redirect via ctx.res.location & status
}
next();
```

PradnyaBaviskar:
Made changes to handle undefined ```ctx.res``` for non-REST transports.

@bajtos, can you review?

common/models/user.js
```diff
@@ -580,21 +580,14 @@ module.exports = function(User) {
...
-            params && params.redirect !== undefined ? params.redirect :
-            body && body.redirect !== undefined ? body.redirect :
-            query && query.redirect !== undefined ? query.redirect :
-            undefined;
-
-          ctx.res.redirect(redirectUrl);
-        } else {
-          next(new Error('transport unsupported'));
+        if (ctx.args.redirect !== undefined) {
+          if(!ctx.res) {
```
bajtos: style: missing space after "if" keyword (see `npm test`)

bajtos:
Please update the strong-remoting version in package.json to `^2.13.2`.

test/user.test.js
```diff
@@ -858,6 +858,38 @@ describe('User', function() {
...
+          request(app)
+            .get('/users/confirm?uid=' + (result.uid) +
+              '&token=' + encodeURIComponent(result.token))
+            .expect(204)
+            .end(function(err, res) {
+              if (err) {
+                return done(err);
+              }
+              done();
+            });
```
bajtos: Since you are not making any extra checks here, the code can be simplified to

```js
  .expect(204)
  .end(done);
```

test/user.test.js
```diff
@@ -858,6 +858,38 @@ describe('User', function() {
...
+              '&token=' + encodeURIComponent(result.token) +
+              '&redirect=http://foo.com/bar')
+            .expect(302)
+            .expect('Location', 'http://foo.com/bar')
+            .end(function(err, res) {
+              if (err) {
+                return done(err);
+              }
+              done();
+            });
```
bajtos: Same comment here, please simplify to 

```js
  .expect(Location, ...)
  .end(done);
```

bajtos:
I have improved the commit message and landed the patch via 3c43ecc. Thank you for the contribution!

PR #1127, opened 2015-02-24
travis.yml: drop 0.11, add 0.12 and iojs

bajtos:
This patch fixes the configuration of Travis CI to run the tests on stable major versions.

@rmg @ritch @raymondfeng I'd like to use this opportunity to consider whether we want to continue using Travis CI for loopback. The browser (Karma) tests are failing there due to a timing issue. The only benefit of Travis CI compared to our Jenkins CI is the openness, meaning that the result of a CI build in is visible to anonymous users in Travis CI, but they are protected by (StrongLoop) login in Jenkins CI.

Thoughts?

@rmg How much work does is required to make all PR jobs for public GH repos readable to anonymous users?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
 node_js:
-  - "0.11"
   - "0.10"
+  - "0.12"
+  - "iojs"
+


rmg:
@bajtos I haven't seen a simple way to do that. An additional complication is it could expose credentials that are injected into the build for access to test DB and other shared resources.

bajtos:
>  I haven't seen a simple way to do that.

That's really bad news for community contributors - they don't have any way how to check why the CI build of their pull request failed. Is it perhaps possible to configure Jenkins to post a comment to the pull request with the relevant part of the build log describing the failure?

If not, then we should probably enable Jenkins for all projects where we accept community contributions.

> An additional complication is it could expose credentials that are injected into the build for access to test DB and other shared resources.

The Android jobs use two Jenkins plugins for keeping the credentials secret. IIRC they set the values via ENV variable and also mask the values with `***` when printed to the console output. Would that help?

Or do you think a malicious PR can expose the credentials in a way that is not covered by those plugins?

--- 

The Jenkins server running Node.js pull requests allows anonymous access to build logs, thus there must be a way how to get this feature. Is the Node.js CI that much different from ours?

PR #1131, opened 2015-02-24
Add a workaround to avoid conflicts with NewRelic

raymondfeng:
/to @bajtos 

NewRelic intercepts express `app.use` and injects its middleware. It also wraps middleware handlers. This monkey-patching trick breaks the assumption in LoopBack that the last item in the router stack is the one being added. As a result, the phase name is not set to the correct handler and middleware ordering is broken.

See https://github.com/bstar/newrelic_test to reproduce the problem.

### DIFF
--- a/lib/server-app.js
+++ b/lib/server-app.js
@@ -161,12 +161,49 @@ proto.middleware = function(name, paths, handler) {
 
   this._skipLayerSorting = true;
   this.use(paths, handler);
-  this._router.stack[this._router.stack.length - 1].phase = fullPhaseName;
+
+  var layer = this._findLayerByHandler(handler);
+  if (layer) {
+    // Set the phase name for sorting
+    layer.phase = fullPhaseName;
+  } else {
+    debug('No matching layer is found for %s %s', fullPhaseName, handlerName);
+  }
+
   this._skipLayerSorting = false;
 
   this._sortLayersByPhase();
 
   return this;
+};
+
+/*
+ * Find the corresponding express layer by handler
+ *
+ * This is needed because monitoring agents such as NewRelic can add handlers
+ * to the stack. For example, NewRelic adds sentinel handler. We need to search
+ * the stackto find the correct layer.
+ */
+proto._findLayerByHandler = function(handler) {
+  // Other handlers can be added to the stack, for example,
+  // NewRelic adds sentinel handler. We need to search the stack
+  for (var k = this._router.stack.length - 1; k >= 0; k--) {
+    if (this._router.stack[k].handle === handler ||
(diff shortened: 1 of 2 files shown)


lib/server-app.js
```diff
@@ -161,7 +161,28 @@ proto.middleware = function(name, paths, handler) {
...
+      // Agressively check if the original handler has been wrapped
+      // into a new function with a property pointing to the original handler
+      for (var p in this._router.stack[k].handle) {
+        if (this._router.stack[k].handle[p] === handler) {
+          this._router.stack[k].phase = fullPhaseName;
+          found = true;
+        }
+      }
+    }
+  }
```
bajtos: Please extract this new block into a standalone function. Perhaps a function that accepts the `handler` and returns the appropriate stack layer:

```
// usage
var layer = findHandlerLayer(this._router.stack, handler);
if (layer) {
  layer.phase = fullPhaseName;
} else {
  /* print a warning? */ 
}
```

Add unit tests for the two new scenarios:

 - the handler function is stored in `handle[__NR_original]`
 - the handler function is stored in a different property

lib/server-app.js
```diff
@@ -161,7 +161,28 @@ proto.middleware = function(name, paths, handler) {
...
   this.use(paths, handler);
-  this._router.stack[this._router.stack.length - 1].phase = fullPhaseName;
+
+  // Other handlers can be added to the stack, for example,
+  // NewRelic adds sentinel handler. We need to search the stack
+  var found = false;
+  for (var k = this._router.stack.length - 1; k >= 0 && !found; k--) {
+    if (this._router.stack[k].handle === handler ||
+      // NewRelic replaces the handle and keeps it as __NR_original
+      this._router.stack[k].handle['__NR_original'] === handler
```
bajtos: Is this optimisation worth the extra code? We will be checking most of the middleware entries using the full property scan anyways.

raymondfeng:
Comments addressed with tests.

lib/server-app.js
```diff
@@ -161,7 +161,15 @@ proto.middleware = function(name, paths, handler) {
...
+  if (layer) {
+    // Set the phase name for sorting
+    layer.phase = fullPhaseName;
+  } else {
+    debug('No matching layer is found for %s %s', fullPhaseName, handlerName);
+  }
+
   this._skipLayerSorting = false;
 
   this._sortLayersByPhase();
```
bajtos: Nitpick: when the layer was not found, there is no need to sort the layers. Feel free to ignore this comment.

bajtos:
IMO it would be nicer to write the tests in an integration style that would verify that the handlers were correctly sorted, but I can live with the current version too.

LGTM.

PR #1132, opened 2015-02-25
Skip hashing password if it's already hashed

raymondfeng:
/to @ritch 

`User.setter.password` is evil. It's triggered when a user record is loaded from the DB too. The `bcrypt` is very expensive.

See https://github.com/strongloop/loopback-datasource-juggler/issues/471


ritch:
LGTM

Issue #1134, opened 2015-02-25
Move /models /routes to explorer middleware


ritch:
[This issue](https://github.com/strongloop/loopback/issues/1124#issuecomment-76008965) raises the concern for these routes always being available. They should not be available from a vanilla use of `loopback.rest()`. Instead whatever is using them (I'm assuming explorer) should add them or turn them on.

bajtos:
Note: loopback-explorer does not use this routes, it has it's own routes for ages. If there was not a backwards-compatibility concern, I would say the routes can be immediately deleted.

I am proposing to add a flag  that will control whether these routes are added or not. This new flag should be:
 - Enabled by default to preserve backwards compatibility.
 - Deprecated from the beginning, so that users are warned about this problem. Use [depd](https://www.npmjs.com/package/depd) for deprecation, that's what express uses too.
 - Disabled in the configuration scaffolded by `slc loopback` via loopback-workspace. That way new applications come without these routes.

greaterweb:
Thanks guys for looking into this. 

I can take a stab at the first two items @bajtos has suggested. Depending on time, I might be able to look into the loopback-workspace one as well.

I'll be sure to include a few tests as good measure.

greaterweb:
> Enabled by default to preserve backwards compatibility.

Trying to think through a config option that is meaningful and concise for this one, two options
* `disableLegacyExplorer` - (`boolean`, `false` by default) seems appropriate but is a bit lengthy - `true` disables the routes, `false` or omitting the option keeps them enabled
* `legacyExplorer` - (`boolean`, `true` by default) this shorter option seems better - `false` disables the routes, `true` or omitting the option keeps them enabled

Thoughts?

greaterweb:
Work completed to resolve this in PR. Though I wonder if we should improve the middleware to limit only `GET` requests.

bajtos:
+1 for `legacyExplorer`, `disableLegacyExplorer=false` is a double negative which is better to avoid.

> Depending on time, I might be able to look into the loopback-workspace one as well.

This part is easy, just add a new entry to [templates/components/api-server/component.js](https://github.com/strongloop/loopback-workspace/blob/5275efe737215ce7a4d338407c2ab173fd91ab6d/templates/components/api-server/component.js#L50) and a new unit-test to [test/end-to-end.js](https://github.com/strongloop/loopback-workspace/blob/5275efe737215ce7a4d338407c2ab173fd91ab6d/test/end-to-end.js#L26).

[...]

PR #1135, opened 2015-02-25
ability to disalbe /models /routes routes

greaterweb:
PR for #1134 

WIP though feedback ok for existing work.

[Suggested Functionality](https://github.com/strongloop/loopback/issues/1134#issuecomment-76024882)
* [x] Enabled by default to preserve backwards compatibility.
* [x] Deprecated from the beginning, so that users are warned about this problem. Use depd for deprecation, that's what express uses too.

Also:
* [x] Tests to ensure `/models` `/routes` can be disabled

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -4,6 +4,7 @@
 
 var loopback = require('../../lib/loopback');
 var async = require('async');
+var deprecate = require('depd')('loopback');
 
 /*!
  * Export the middleware.
@@ -28,10 +29,17 @@ function rest() {
   return function restApiHandler(req, res, next) {
     var app = req.app;
 
-    if (req.url === '/routes') {
-      return res.send(app.handler('rest').adapter.allRoutes());
-    } else if (req.url === '/models') {
-      return res.send(app.remotes().toJSON());
+    // added for https://github.com/strongloop/loopback/issues/1134
+    if (app.get('legacyExplorer') !== false) {
+      deprecate(
+        'Routes "/methods" and "/models" are considered dangerous and should not be used.\n' +
+        'Disable them by setting "legacyExplorer=false" in "server/config.json" or via "app.set()".'
+      );
+      if (req.url === '/routes') {
+        return res.send(app.handler('rest').adapter.allRoutes());
+      } else if (req.url === '/models') {
+        return res.send(app.remotes().toJSON());
+      }
     }
 
     if (!handlers) {
--- a/test/rest.middleware.test.js
+++ b/test/rest.middleware.test.js
@@ -164,6 +164,44 @@ describe('loopback.rest', function() {
     }, done);
   });
 
+  it('should report 200 for legacy explorer route /routes', function(done) {
+    app.use(loopback.rest());
+    request(app).get('/routes')
+      .expect(200)
+      .end(function(err, res) {
+        if (err) return done(err);
+        expect(res.body).to.eql([]);
+        done();
+      });
+  });
+
+  it('should report 200 for legacy explorer route /models', function(done) {
+    app.use(loopback.rest());
+    request(app).get('/models')
+      .expect(200)
+      .end(function(err, res) {
+        if (err) return done(err);
+        expect(res.body).to.eql({});
+        done();
+      });
+  });
+
+  it('should report 404 for disabled legacy explorer route /routes', function(done) {
+    app.set('legacyExplorer', false);
+    app.use(loopback.rest());
+    request(app).get('/routes')
+      .expect(404)
+      .end(done);
+  });
+
+  it('should report 404 for disabled legacy explorer route /models', function(done) {
+    app.set('legacyExplorer', false);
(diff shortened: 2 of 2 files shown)


greaterweb:
Condensed into a single commit and added the functionality to limit the `/models` and `/routes` availability to just `GET` requests. @bajtos, this PR is ready for review whenever anyone has time.

package.json
```diff
@@ -49,7 +49,8 @@
     "stable": "^0.1.5",
     "strong-remoting": "^2.13.2",
     "uid2": "0.0.3",
-    "underscore.string": "~2.3.3"
+    "underscore.string": "~2.3.3",
+    "depd": "~1.0.0"
```
bajtos: Please use `^1.0.0` so that we are automatically picking up new compatible versions in the future.

server/middleware/rest.js
```diff
@@ -28,10 +29,16 @@ function rest() {
   return function restApiHandler(req, res, next) {
     var app = req.app;
 
-    if (req.url === '/routes') {
-      return res.send(app.handler('rest').adapter.allRoutes());
-    } else if (req.url === '/models') {
-      return res.send(app.remotes().toJSON());
+    // added for https://github.com/strongloop/loopback/issues/1134
+    var legacyExplorer = (undefined === app.get('legacyExplorer') || app.get('legacyExplorer')) ? true : false;
```
bajtos: Please simplify to `app.get('legacyExplorer') !== false`.

server/middleware/rest.js
```diff
@@ -28,10 +29,16 @@ function rest() {
...
+
+    if (legacyExplorer && 'get' === req.method.toLowerCase()) {
```
bajtos: Please keep support for all methods in order to preserve backwards compatibility, even if it's silly.

server/middleware/rest.js
```diff
@@ -28,10 +29,16 @@ function rest() {
...
+      deprecate('legacyExplorer: Update config.json to disable legacyExplorer routes');
```
bajtos: I'd like to improve this message to give the user more context. How about this?

```
Routes "/methods" and "/models" are considered dangerous and should not be used.\n
Disable them by setting "legacyExplorer=false" in "server/config.json" or via "app.set()".
```

test/rest.middleware.test.js
```diff
@@ -164,6 +164,40 @@ describe('loopback.rest', function() {
     }, done);
   });
 
+  it('should report 200 for legacy explorer route /routes', function(done) {
+    app.model(MyModel);
```
bajtos: Is it necessary to add `MyModel` to the app? It's better to keep the test as small as possible and remove anything not needed. The same comment applies to the other new tests.

bajtos:
@greaterweb Good job, thank you for the pull request. Please address my comments above.

greaterweb:
Thank you for the feedback. I will correct these items and update the PR.

test/rest.middleware.test.js
greaterweb: It wasn't necessary, I just wanted to ensure there would be values returned inside the responses. By removing it though I can improve the test a bit and look for `[]` and `{}` for the `/routes` and `/models` response bodies respectively.

greaterweb:
Updates implemented as suggested by @bajtos, rewrote the history to keep it clean as one small commit.

I can't see why the Merged build is failing unfortunately. I'm getting prompted me to authorize StrongLoop Jenkins with permissions to read/write for all of my public and private repos, not sure I want to do that :-)

 Should any additional updates be required, please feel free to let me know.

bajtos:
The code LGTM, thank you.

The tests are failing because `depd` does not work in the browser. My pull request #1140 will fix that, so we need to wait until it lands.

bajtos:
@greaterweb #1140 has been landed, could you please rebase your patch on top of the current master? 

In case you haven't done this before, here is one of the many ways: `git fetch && git rebase -i upstream/master && git push -f` (assuming that "upstream" remote points to strongloop/loopback).

bajtos:
@greaterweb #1140 has been landed, could you please rebase your patch on top of the current master? 

In case you haven't done this before, here is one of the many ways: `git fetch upstream && git rebase -i upstream/master && git push -f` (assuming that "upstream" remote points to strongloop/loopback).

greaterweb:
Sounds good, I'll get it ready.

greaterweb:
Ok, should be good to go. Test are working ok locally but seem to be failing for CI.

Made a small update to remove my addition of `depd` from package.json and fixed the commit message as it mentioned limiting the supporting methods to just `GET`.

If you can see anything on my end that needs to be updated to fix the CI tests, please let me know.

bajtos:
For posterity, Travis CI is failing due to PhantomJS timeouts. Our Jenkins CI passed on 4 slaves out of 6, the remaining two slaves ran into race condition when checking out the code from git. It's a know problem of Jenkins which we did not managed to fix yet.

bajtos:

PR #1140, opened 2015-02-26
Guids for replicated models

### LINKED ISSUE #981: Enforce guid for replication
Replication in its current form does not support serial `Number` -ish ids. When a model is created on an offline client with a conflicting id created on the server loopback assumes that the models are the same.

This is easily prevented by using a guid id for your replicated models. Requires https://github.com/strongloop/loopback-datasource-juggler/issues/388.

/cc @raymondfeng @bajtos

bajtos:
- Print a deprecation warning when a persisted model is tracking changes but does not have a client-generated unique string id property (GUID).

- Fix all test models that are tracking changes so that they have a generated unique string id.

- Make model names unique where possible to prevent tests reusing the same model which causes unintented side effects.

Closes #981 

/to @ritch please review. I decided to use soft deprecation warning instead of throwing an exception. This way we are not breaking backwards compatibility and make the upgrade process easier (IMO).

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -7,6 +7,7 @@ var registry = require('./registry');
 var runtime = require('./runtime');
 var assert = require('assert');
 var async = require('async');
+var deprecated = require('depd')('loopback');
 
 /**
  * Extends Model with basic query and CRUD support.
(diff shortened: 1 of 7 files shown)


lib/persisted-model.js
```diff
@@ -1041,6 +1042,13 @@ PersistedModel.enableChangeTracking = function() {
...
     ' is not attached to a dataSource');
 
+  var idName = this.getIdName();
+  var idProp = this.definition.properties[idName];
+  if (!idProp || idProp.type !== String || !/^[ug]uid$/.test(idProp.defaultFn)) {
+    deprecated('The model ' + this.modelName + ' is tracking changes, ' +
+      'which requries a string id with GUID/UUID default value.');
+  }
+
   Change.attachTo(this.dataSource);
```
ritch: That regex is a bit silly.

bajtos: Shall I rewrite it as `!(id.Prop.defaultFn === 'uuid' || id.Prop.defaultFn === 'guid')`?

bajtos:
Ouch, `depd` does not work in the browser :( I'll need to fix juggler as well.

bajtos:
Fixed, the patch depends on https://github.com/strongloop/loopback-datasource-juggler/pull/473 now.

@ritch LGTY now?

ritch:
LGTM

Issue #838, opened 2014-11-19
Add ctx. isNewInstance to "before save" and "after save" hooks


pungoyal:
Extend `before save` and `after save` hooks and add a new property `ctx.isNewInstance` (or similar).

This new property will have one of the following values for `Model.updateOrCreate`:

- `before save`:
  - true when the data has not id and the method falls back to a regular create
  - true/false when a non-atomic implementation is used
  - undefined when an atomic implementation is used

- `after save`:
 - true when the connector reported that a record was created
 - false when the connector reported that a record was updated
 - undefined when the connector did not report anythin

A thing to figure out: how to handle atomic `findOrCreate`, as it may fire the `before save` hook even if the model already exists?

---

**The original issue description**

Does it make sense to have a `created` event on `model`? Currently we have `changed`, but that fires for any changes to the model.

bajtos:
It is probably better to defer this enhancement until https://github.com/strongloop/loopback-datasource-juggler/issues/367 is implemented.

bajtos:
The new [Operation hooks](http://docs.strongloop.com/display/LB/Operation+hooks) implemented by  strongloop/loopback-datasource-juggler#367 provide `after save` hook which does not distinguish between `created` and `updated`. One of the reasons is that this information may not be available from `updateOrCreate` (a.k.a `upsert` and HTTP `PUT`) implemented by connectors as an atomic operation.

However, if you are able to come up with a way how to distinguish between `CREATE` and `UPDATE` that works in all connectors, then I am ok with adding a new property to the `ctx` object passed to `after save` hook and use this property to decide which event to fire.

Another alternative is to let `updateOrCreate()` trigger always `updated` event, even in the case where a new model instance was created.

Thoughts?

clarkorz:
@bajtos *loopback-connector-mongodb* implements `updateOrCreate` by using **findAndModify**.  The  result that **findAndModify** command responses could be use to distinguish `FIND`, `CREATE`, and `UPDATE`.  

For *loopback-connector-mysql*, I am not familiar with mysql, but I do find a hint in [mysql document](http://dev.mysql.com/doc/refman/5.5/en/insert-on-duplicate.html): 
>With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values.

[...]

PR #1148, opened 2015-03-02
Replace deprecated hooks with Operation hooks

bajtos:
```
AccessToken.beforeCreate -> AccessToken.observe('before save')
Application.beforeCreate -> Application.observe('before save')
Checkpoint.beforeSave -> Checkpoint.observe('before save')
```

/to @ritch please review
/cc @raymondfeng 

Loosely related: https://github.com/strongloop/loopback/issues/838 that will implement `ctx.isNewInstance`. Although I don't think it will be very helpful in these three cases.

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -65,20 +65,18 @@ module.exports = function(AccessToken) {
   /*!
    * Hook to create accessToken id.
    */
-
-  AccessToken.beforeCreate = function(next, data) {
-    data = data || {};
+  AccessToken.observe('before save', function(ctx, next) {
+    if (!ctx.instance || ctx.instance.id) {
+      // We are running a partial update or the instance already has an id
+      return next();
+    }
 
     AccessToken.createAccessTokenId(function(err, id) {
-      if (err) {
-        next(err);
-      } else {
-        data.id = id;
-
-        next();
-      }
+      if (err) return next(err);
+      ctx.instance.id = id;
+      next();
     });
-  };
+  });
 
   /**
    * Find a token for the given `ServerRequest`.
--- a/common/models/application.js
+++ b/common/models/application.js
@@ -81,8 +81,15 @@ module.exports = function(Application) {
    * A hook to generate keys before creation
    * @param next
    */
-  Application.beforeCreate = function(next) {
-    var app = this;
+  Application.observe('before save', function(ctx, next) {
+    if (!ctx.instance) {
+      // Partial update - don't generate new keys
+      // NOTE(bajtos) This also means that an atomic updateOrCreate
+      // will not generate keys when a new record is creatd
+      return next();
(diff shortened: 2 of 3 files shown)


common/models/application.js
```diff
@@ -81,8 +81,14 @@ module.exports = function(Application) {
    * A hook to generate keys before creation
    * @param next
    */
-  Application.beforeCreate = function(next) {
-    var app = this;
+  Application.observe('before save', function(ctx, next) {
+    if (!ctx.instance) {
+      // Example: Application.updateAll() and Application.updateOrCreate()
+      var msg = 'Application object does not support data-based updates';
+      return next(new Error(msg));
```
raymondfeng: I believe we do have the use case to reset api keys for a given application.

bajtos: Can you please post a short snippet of code that you need to keep working with my change in place?

common/models/access-token.js
```diff
@@ -65,20 +65,23 @@ module.exports = function(AccessToken) {
...
    * Hook to create accessToken id.
    */
+  AccessToken.observe('before save', function(ctx, next) {
+    if (!ctx.instance) {
+      if (ctx.data.id) return next();
 
-  AccessToken.beforeCreate = function(next, data) {
-    data = data || {};
+      // Example: AccessToken.updateAll()
+      var msg = 'AccessToken does not support data-based updates.';
```
ritch: This message is confusing... I'm not even sure what it is trying to say. It must be updated with `inst.save()` and counterparts? And not `updateAttributes(data)`?

ritch:
Message is pretty vague... otherwise it LGTM.

common/models/application.js
raymondfeng:
```js
  Application.prototype.resetKeys = function(cb) {
    this.clientKey = generateKey('client');
    this.javaScriptKey = generateKey('javaScript');
    this.restApiKey = generateKey('restApi');
    this.windowsKey = generateKey('windows');
    this.masterKey = generateKey('master');
    this.modified = new Date();
    this.save(cb);
  };
```

bajtos: There is already a unit-test for this and it's passing: https://github.com/strongloop/loopback/blob/baf372b8d7f1bd578759e11e596ec7d633b70cb5/test/model.application.test.js#L97-L120

bajtos:
I have simplified the implementation and make it closer to the original version. Only `Checkpoint` reports an error now, I have reworded the message to make it hopefully more clear. I cannot make the message specific (e.g. list method names), as the specifics may change in the future, depending on the implementation details of DAO methods.

@raymondfeng @ritch PTAL, LGTY now?

ritch:
:+1:

raymondfeng:
LGTM

PR #1152, opened 2015-03-02
Pass options from User.login to createAccessToken

raymondfeng:
/to @ritch 

It will allow subclass of User to create access token based on additional properties such as 'scope'.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -69,11 +69,23 @@ module.exports = function(User) {
    * customize how access tokens are generated
    *
    * @param {Number} ttl The requested ttl
-   * @callack {Function} cb The callback function
+   * @param {Object} [options] The options for access token, such as scope, appId
+   * @callback {Function} cb The callback function
    * @param {String|Error} err The error string or object
    * @param {AccessToken} token The generated access token object
    */
-  User.prototype.createAccessToken = function(ttl, cb) {
+  User.prototype.createAccessToken = function(ttl, options, cb) {
+    if (cb === undefined && typeof options === 'function') {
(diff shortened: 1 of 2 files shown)


common/models/user.js
```diff
@@ -69,11 +69,23 @@ module.exports = function(User) {
    * customize how access tokens are generated
    *
    * @param {Number} ttl The requested ttl
+   * @param {Object} options The options for access token
```
ritch: This could be a bit clearer.

ritch: You should also update https://github.com/strongloop/loopback/pull/1152/files#diff-5d81af95449574544e7317f3eda2badeL144 with some docs for this.

raymondfeng: I assume you mean to update docs of User.login at docs.strongloop.com.

common/models/user.js
```diff
@@ -210,19 +236,11 @@ module.exports = function(User) {
...
-                  //  1. token.user already exists, it's a function injected by
-                  //     "AccessToken belongsTo User" relation
-                  //  2. ModelBaseClass.toJSON() ignores own properties, thus
-                  //     the value won't be included in the HTTP response
-                  // See also loopback#161 and loopback#162
-                  token.__data.user = user;
-                }
-                fn(err, token);
-              });
+              if (user.createAccessToken.length === 2) {
```
seriousben: @raymondfeng Not sure what this means? How can I pass credentials to to createAccessToken?

seriousben: NVM, just learned something new.

PR #1153, opened 2015-03-03
karma conf: prevent timeouts on Travis CI

bajtos:
Increase the timeout values to prevent DISCONNECTED errors when
running the tests on slow machines (e.g. Travis CI).

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -61,6 +61,11 @@ module.exports = function(config) {
     // If browser does not capture in given timeout [ms], kill it
     captureTimeout: 60000,
 
+    // to avoid DISCONNECTED messages
+    browserDisconnectTimeout : 10000, // default 2000
+    browserDisconnectTolerance : 1, // default 0
+    browserNoActivityTimeout : 60000, //default 10000
+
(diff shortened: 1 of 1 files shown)


bajtos:
Here we go, Travis builds are green again :dart:

bajtos:
For posterity, here is some info that helped me to solve this issue.

https://github.com/karma-runner/karma/issues/598
> I've noticed that using pre-processors can cause this to increase as well.
> Since karma launches the browser then processes all the files. If your
> preprocessors take a long time to run then I'll bet you see a direct
> correlation.

https://github.com/webpack/karma-webpack/issues/8
> I've solved the issue by increasing the browserNoActivityTimeout in karma
> config. It seems it took too long to compile a bundle so the browser was
> killed before the bundle was ready.

    // Timeout for capturing a browser (in ms).
    captureTimeout: 60 * 1e3,

    // to avoid DISCONNECTED messages
    browserDisconnectTimeout : 10000, // default 2000
    browserDisconnectTolerance : 1, // default 0
    browserNoActivityTimeout : 60 * 1e3, //default 10000

Issue #985, opened 2015-01-09
Fix checkpoint implementation


bajtos:
The REST API uses one checkpoint id. In practice, the client checkpoint number will be different than the server-side number. (Think about two clients connected to the same server).

#1157

 - [x]  `Model.replicate` will accept two `since` values - one for the source model, another for the target model. The single `since` value will remain supported for backwards compatibility with a deprecation warning.

 - [x] Either associate changes with CP+1, or bump the CP as the first step in the `replicate` method. I prefer the latter as otherwise the method `Checkpoint.current()` does not make sense anymore.

 - [x] Modify `replicate` to return the current CP in the callback, probably for both source and target models. (I need to give this some more thoughts.)

 - [x] Write integration tests to verify that the scenarios I have identified while working on this issue are fixed by the changes above.

#1176

 - [x] Improve change detection to reduce the amount of false conflicts
 - [x] Write integration tests for common replication scenarios for single server with multiple clients.

#1187

 - [x] Write integration tests for conflict scenarios for single server with multiple clients.

Out of scope:

 - Rework the `Change` model to keep a history of changes instead of the last record only. Possibly rework the `PersistedModel.diff` and/or `PersistedModel.changes` to return a history of changes too.

BerkeleyTrue:
This is something I am running into now, but haven't thought of a way to fix. What about adding local/remote checkpoint values to the Change model?

bajtos:
> What about adding local/remote checkpoint values to the Change model?

That would not work as you can have multiple clients for the same server/model/change.

Here is what I'd like to implement instead: when a replication is finished, the new checkpoint should be created as `max(localCheckpoint, remoteCheckpoint)+1`.

bajtos:
Uff, this issue is much more complex than I initially thought.

Problems we need to address:

 1. Ensure that no changes are added to the checkpoint that is being
   synchronized while the replication is in progress.

    *Example:*

     ```
     Code1: start replication of CP 1
     Code1: get the list of changes (diff): A, B, C
     Code2: make a new change D
     Code2: D is associated with the current CP 1
     Code1: finish the replication, create CP2
     ```

    *The problem:*

    The next run will consider change D as already synchronized

    *Solution:*

    Increment the checkpoint before querying for the list of changes,
    i.e. as the first step of the `diff` method.

 2. When the second step of replication (conflict resolution + batch update)
   is aborted, e.g. because there are conflicts or because update fails,
   the subsequent replication should process all changes that were not
[...]

Issue #1155, opened 2015-03-03
Fix the way how Checkpoint model is defined


bajtos:
At the moment, each `Change`-based model creates a new `Checkpoint` model, but with the same model name "checkpoint". AFAIK, this is not a valid supported usage of LoopBack model building.

We should rework the way how `Checkpoint` models are created and give more control to LoopBack users.

PR #1157, opened 2015-03-03
Replication improvements: Checkpoint

bajtos:
A first round of `PersistedModel.replicate()` improvements described in #985 (connect to #985):

 - Support different "since" for source and target
 - Fix checkpoint-related race condition, create a remote checkpoint during replication too
 - Return new checkpoints in callback of replicate()

/to @ritch Please review, preferably today.

My next work will most likely depend on this PR, so if I don't manage to get this merged today, I'll have to add new stuff to this PR instead of opening a new one.

/cc @raymondfeng

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -428,6 +428,8 @@ module.exports = function(Change) {
   Change.getCheckpointModel = function() {
     var checkpointModel = this.Checkpoint;
     if (checkpointModel) return checkpointModel;
+    // FIXME(bajtos) This code creates multiple different models with the same
+    // model name, which is not a valid supported usage of juggler's API.
     this.Checkpoint = checkpointModel = loopback.Checkpoint.extend('checkpoint');
     assert(this.dataSource, 'Cannot getCheckpointModel(): ' + this.modelName +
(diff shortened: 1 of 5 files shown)


common/models/change.js
```diff
@@ -428,6 +428,8 @@ module.exports = function(Change) {
   Change.getCheckpointModel = function() {
     var checkpointModel = this.Checkpoint;
     if (checkpointModel) return checkpointModel;
+    // FIXME(bajtos) This code creates multiple different models with the same
+    // model name, which is not a valid supported usage of juggler's API.
```
bajtos: I think it this cannot be fixed while preserving backwards compatibility. I have opened a follow-up issue to make the fix in LoopBack 3.0 - see https://github.com/strongloop/loopback/issues/1155.

bajtos:
Added one extra commit fixing the behaviour I always found surprising - checkpoint.seq should start from 1, not from zero.

ritch:
LGTM :+1: good job... I think the tests alone in this PR show that breaking this out into its own module (perhaps with a more test-able API) would be a good idea.

Issue #209, opened 2014-03-02
Remote methods mapping in JSON


nikita-leonov:
Would be nice to have an option to define remote methods mapping to an actual classes in JSON. So we will have complete REST interface defined in one place with a logic and complex cases in JS. I do not see why some of such constructions could not be a part of JSON mapping:
ApplicationModel.authenticate,
    {
      accepts: [
        {arg: 'appId', type: 'string', required: true},
        {arg: 'key', type: 'string', required: true}          
      ],
      returns: {arg: 'accessToken', type: 'object', root: true},
      http: {verb: 'post'}
    }

mrfelton:
fwiw I'm doing something like this in my own setup by using the unimplemented `methods` key in the model json schema to define the remote methods in the same way as you would normally in the `remoteMethod()` method, and then using that to setup the remote methods on the fly.

eg

common/models/my-model.json
```
  ...
  "relations": [],
  "acls": [],
  "methods": {
    "syncToRemote": {
      "doc": "Syncronise a person to a supported provider account.",
      "accepts": [{
        "arg": "id",
        "type": "number",
        "required": true,
        "description": "PersistedModel id"
      }, {
        "arg": "provider",
        "type": "string",
        "required": true,
        "http": {
          "source": "form"
        },
        "description": "The external account provider to update."
      }],
      "returns": {
        "arg": "person",
        "type": "object",
        "root": true,
        "description": "The updated person object."
      },
      "http": {
        "path": "/:id/syncToRemote",
        "verb": "post"
      }
    }
  }
```

common/models/my-model.js:
```
  // Set up remote methods from model config schema json.
  // TODO: Remove / Refactor this once loopback supports this natively.
  MyModel.on('attached', function () {
    var app = MyModel.app;
    _.each(app.models.MyModel.settings.methods, function (value, key) {
      MyModel.remoteMethod(key, value);
    });
  });
```

Seems to be working ok so far. My only gripe with it is that the remoteMethod call expects documentation for the properties to be in a 'description' key, which is inconsistent with all the rest of the documentation in common/models/my-model.json which uses the `doc` key instead.

Would be great to have some support like this in loopback itself.

bajtos:
@mrfelton can you submit a pull request with your implementation? Remote methods should be defined by [registry.createModel](https://github.com/strongloop/loopback/blob/d5b7447750c4ca881744f5f9ce8be559a993d2ab/lib/registry.js#L87).

Few things to change in the second snippet:
 
[...]

Issue #597, opened 2014-09-30
Inconsistent API for remote methods and hooks


bajtos:
As of LoopBack 2.0, there is a new method for defining remote method metadata `Model.remoteMethod`. To configure a prototype method, one has to supply a flag `isStatic`:

```js
Model.remoteMethod('foo', { isStatic: false, ... });
```

A) This is inconsistent with the approach we have already in place for hooks ([docs](http://docs.strongloop.com/display/LB/Defining+remote+hooks)), where we prefix the method name with `prototype.`:

```js
Model.beforeRemote('prototype.foo', ...);
```

B) The usage of this API - e.g. `('find', true)` - is very ambiguous. What does `true` mean? One has to be very familiar with the API to understand what is the intent of such code snippet.

---

We should modify `Model.remoteMethod` to support the `prototype.foo` notation too.

Since both `Model.remoteMethod` and `Model.beforeRemote` are thin wrappers around strong-remoting API, we should fix this problem in strong-remoting too.

bajtos:
Actually, I find `sharedClass.find` name rather ambiguous, a better name would be `sharedClass.findMethod`. That reveals a backwards-compatible solution: create a new method `sharedClass.findMethod` accepting a single string (`prototype.foo` or `foo`) and deprecate the `find` method.

Unfortunately fixing `disableRemoteMethod` is still a breaking change.

bajtos:
@0candy in reference to https://github.com/strongloop-internal/scrum-loopback/issues/617, here is what I would like you to implement:

1. Create a new function `SharedClass.prototype.findMethod` that could be invoked like `findMethod('create')` and `findMethod('prototype.updateAttributes')` - the first call should find a static method, the second call a prototype method. Implement this function in both 2.x and 3.0 branches.
2. Deprecate `SharedClass.prototype.find` in 3.0 only
3. Modify `SharedClass.prototype.disableRemoteMethod` to accept a single string argument (to follow the signature of `findMethod`), in 3.0 only - this is a breaking change that needs an entry in 3.0 release notes.
4. Compile a list of other places that use `isStatic` when searching for a method, so that we can discuss whether we should change them too (3.0 only).
5. Take a look at places in LoopBack (Model, PersistedModel) that are affected by these API changes, rework the LoopBack API to follow the new strong-remoting API (3.0 only).
[...]

PR #1163, opened 2015-03-04
Define remote methods using json schema definitions. Prototype for #209

mrfelton:
Enables remote methods to be defined in the json schema and applied automatically to the generated models. #209


slnode:

raymondfeng:
@mrfelton It's a great addition. Can you add some test cases?

lib/registry.js
```diff
@@ -201,12 +201,24 @@ registry.configureModel = function(ModelCtor, config) {
       'must be an array of objects', modelName);
   }
 
+  // Remote methods
+  if (typeof settings.methods === 'object' && settings.methods !== null) {
+    var methods = settings.methods = settings.methods || {};
+    Object.keys(settings.methods).forEach(function(key) {
+      ModelCtor.remoteMethod(key, settings.methods[key]);
```
bajtos: Please rework this part according to https://github.com/strongloop/loopback/issues/209#issuecomment-77103970, so that:

 - key `prototype.methodName` is mapped to `remoteMethod('methodName', { isStatic: false, ... })`
 - key `methodName` is mapped to `remoteMethod('methodName', { isStatic: true, ... })`

bajtos:
Nice to have: rework built-in models in `common/models/*` to use this new declarative approach too.

bajtos:
@slnode ok to test

bajtos:
Besides adding tests, we also need you @mrfelton to sign our CLA: https://cla.strongloop.com/agreements/strongloop/loopback

ritch:
I've been wanting to add this for quite some time... :+1:

lib/registry.js
ritch: @bajtos please consider the workspace implementation. Lets not make it impractical to implement the generator via the workspace. I'm not against `prototype.methodName` itself... but we might be making more work for ourselves in the workspace without much benefit.

I'd really like to see remote method creation support added to workspace alongside this addition... These JSON files are specifically designed to serve the workspace well so this new feature is no exception.

bajtos: > Please consider the workspace implementation. Lets not make it impractical to implement the generator via the workspace.

Can you please elaborate more why my proposal is impractical for loopback-workspace to implement? Are you afraid of parsing `prototype.{name}` in the workspace?

We can modify the implementation as follows if you think it will make the workspace implementation simpler:

 - `methodName` is mapped to `isStatic:true` unless the config object provides a different value
 - `prototype.methodName` is always mapped to `isStatic:false`, regardless of the value in the config object

```js
// static methods
methods: { "work": { isStatic: true, /*...*/ } }
methods: { "work": { /*...*/ } }

// prototype methods
methods: { "work": { isStatic: false, /*...*/ } }
methods: { "prototype.work": { /*...*/ } }
```

What I really want to avoid is adding more stuff that we will have to rework when fixing #209 and thus causing #209 to introduce even more breaking changes.

mrfelton:
Have added a basic test case to verify that remote methods defined via the json spec are properly added to the model. As for the handling of static vs prototype methods, I feel like that should be handled separately in #597 as it's a much wider change that would affect far more than what I was trying to do in #209 (making it possible to define remote methods in json).

bajtos:
> As for the handling of static vs prototype methods, I feel like that should be handled separately in #597 as it's a much wider change that would affect far more than what I was trying to do in #209 (making it possible to define remote methods in json).

I really don't want people to start adding JSON metadata in the format that is already known to get changed in the near future. All I am asking is to modify the piece of code calling `ModelCtor.remoteMethod` to pre-process the `key`, no other changes are needed.

@ritch what's your opinion on this? If you agree with me, then I can add the necessary code on top of  @mrfelton's patch to get the desired API.

ritch:
> I really don't want people to start adding JSON metadata in the format that is already known to get changed in the near future.

Agreed. Lets not be hasty in adding this to the format... there's a lot of downstream work when adding anything to the `$model.json` file. I'm **-1** for landing this without **at least** an initial PR that adds support for it in the workspace.

Regarding `methods: { "prototype.work": { /*...*/ } }`. Why not leave this off as it is entirely additive (in your last proposal @bajtos)? Although I am fine having this style supported (if it is also supported in the workspace PR mentioned above).

lib/registry.js
ritch: > Can you please elaborate more why my proposal is impractical for loopback-workspace to implement

The workspace uses keys as identifiers for entities (in the case a method would be an entity).

> `prototype.methodName` is always mapped to isStatic:false, regardless of the value in the config object

I like where you are going with this... but consider what the workspace does. It loads properties, and other `Entities` and also writes them back to files. We would have to store a bit that says "this method was defined using `prototype.methodName`" and recall that during serialization. Of course this is all possible... but probably not worth it to get rid of the annoying `isStatic` property.

bajtos:
> Regarding methods: { "prototype.work": { /*...*/ } }. Why not leave this off as it is entirely additive (in your last proposal @bajtos)?

Because it's not additive, it's a breaking change as explained in #597. ATM, if you don't specify `isStatic`, it defaults to `false`, i.e. `{ methods: { "work":  { /*...*/ }  } }` is interpreted as a prototype method in the implementation proposed by this pull request.

What I would like to see is that  `{ methods: { "work":  { /*...*/ }  } }` is interpreted as a static method.

> Lets not be hasty in adding this to the format... there's a lot of downstream work when adding anything to the $model.json file. I'm -1 for landing this without at least an initial PR that adds support for it in the workspace.

I understand your concerns, we are building technical debt in the workspace API. On the other hand, I feel it's not fair to ask people contributing useful loopback core features to implement support for them in loopback-workspace too.

bajtos:
@ritch what's your opinion on my comment above? It would be nice to come to an agreement to unblock progress of this pull request.

If you are still not convinced about my proposal, then how these two alternatives?

 1. Make `isStatic` required, don't provide (broken) default value.
 2. Make `isStatic` default to `true`, this will allow us to add support for `prototype.name` later and in a way that does not break compatibility.

bajtos:
@ritch ping, this pull request is waiting for your comment, can you please read https://github.com/strongloop/loopback/pull/1163#issuecomment-78608932 and https://github.com/strongloop/loopback/pull/1163#issuecomment-87608282 and let me know what solution do you consider as the best (or at least acceptable to you)?

ritch:
> Make isStatic required, don't provide (broken) default value.

Once we add the nicer syntax we can make this not required if you specify the sugary syntax (prototype.foo). So I am in favor of this option to move this forward.

ritch:
> I feel it's not fair to ask people contributing useful loopback core features to implement support for them in loopback-workspace too

I understand that its not fair to say "Thanks for the contribution, but we won't accept it unless you add support in x, y and z". But you also have to understand (as someone contributing to core) that their are upstream / downstream projects that were purposefully developed concurrently with loopback core to ensure that once a capability was designed it would function throughout the entire platform.

In this case it is not a matter of technical debt. Its a matter of compatability. The reason we have these JSON files in the first place is because of the workspace. We've been bitten in the past by hastily designing syntax (cough.. ACLs cough...) without thinking about its use in JSON files. What we are talking about here is the other way around sure (we've got some good JSON syntax ideas) but we cannot gloss over the fact that this will be parsed by the workspace API, and eventually implemented in the composer UI.

bajtos:
Closing in favour of #1366.

PR #1169, opened 2015-03-04
Allow 'me' literal to represent the current user in url

raymondfeng:
/to @ritch @bajtos 

This PR adds the ability to use `me` or a custom literal to represent the current user in req url. For example,

GET /api/users/me

The token middleware will replace `me` with the current user id for downstream handlers.

It adds an option to loopback.token middleware as `currentUserLiteral`.

### DIFF
--- a/server/middleware/token.js
+++ b/server/middleware/token.js
@@ -4,12 +4,34 @@
 
 var loopback = require('../../lib/loopback');
 var assert = require('assert');
+var debug = require('debug')('loopback:middleware:token');
 
 /*!
  * Export the middleware.
  */
 
 module.exports = token;
+
+/*
+ * Rewrite the url to replace current user literal with the logged in user id
+ */
+function rewriteUserLiteral(req, currentUserLiteral) {
+  if (req.accessToken && req.accessToken.userId && currentUserLiteral) {
+    // Replace /me/ with /current-user-id/
+    var urlBeforeRewrite = req.url;
+    req.url = req.url.replace(
+      new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g'),
+        '/' + req.accessToken.userId + '$1');
+    if (req.url !== urlBeforeRewrite) {
+      debug('req.url has been rewritten from %s to %s', urlBeforeRewrite,
+        req.url);
+    }
+  }
+}
+
+function escapeRegExp(str) {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
 
 /**
  * Check for an access token in cookies, headers, and query string parameters.
@@ -39,19 +61,37 @@ module.exports = token;
  * @property {Array} [cookies] Array of cookie names.
  * @property {Array} [headers] Array of header names.
  * @property {Array} [params] Array of param names.
- * @property {Array} [model] An AccessToken object to use.
+ * @property {Function|String} [model] AccessToken model name or class to use.
+ * @property {String} [currentUserLiteral] String literal for the current user.
  * @header loopback.token([options])
  */
 
 function token(options) {
   options = options || {};
   var TokenModel = options.model || loopback.AccessToken;
-  assert(TokenModel, 'loopback.token() middleware requires a AccessToken model');
+  if (typeof TokenModel === 'string') {
+    // Make it possible to configure the model in middleware.json
+    TokenModel = loopback.getModel(TokenModel);
+  }
+  var currentUserLiteral = options.currentUserLiteral;
+  if (currentUserLiteral && (typeof currentUserLiteral !== 'string')) {
+    debug('Set currentUserLiteral to \'me\' as the value is not a string.');
+    currentUserLiteral = 'me';
+  }
+  if (typeof currentUserLiteral === 'string') {
+    currentUserLiteral = escapeRegExp(currentUserLiteral);
+  }
+  assert(typeof TokenModel === 'function',
+    'loopback.token() middleware requires a AccessToken model');
 
   return function(req, res, next) {
-    if (req.accessToken !== undefined) return next();
+    if (req.accessToken !== undefined) {
+      rewriteUserLiteral(req, currentUserLiteral);
+      return next();
+    }
     TokenModel.findForRequest(req, options, function(err, token) {
       req.accessToken = token || null;
+      rewriteUserLiteral(req, currentUserLiteral);
       var ctx = loopback.getCurrentContext();
       if (ctx) ctx.set('accessToken', token);
       next(err);
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -108,6 +108,51 @@ describe('loopback.token(options)', function() {
       });
   });
 
+  it('should rewrite url for the current user literal at the end without query',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var id = this.token.id;
+      var userId = this.token.userId;
+      request(app)
+        .get('/users/me')
+        .set('authorization', id)
+        .end(function(err, res) {
+          assert(!err);
+          assert.deepEqual(res.body, {userId: userId});
+          done();
+        });
+    });
+
+  it('should rewrite url for the current user literal at the end with query',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var id = this.token.id;
+      var userId = this.token.userId;
+      request(app)
+        .get('/users/me?state=1')
+        .set('authorization', id)
+        .end(function(err, res) {
+          assert(!err);
+          assert.deepEqual(res.body, {userId: userId, state: 1});
+          done();
+        });
+    });
+
+  it('should rewrite url for the current user literal in the middle',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var id = this.token.id;
+      var userId = this.token.userId;
+      request(app)
+        .get('/users/me/1')
+        .set('authorization', id)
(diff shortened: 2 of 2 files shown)


ritch:
We need to make it clear that you can use `req.originalUrl` to differentiate between `/users/me` and `/users/7`.

ritch:
We need to add some examples using this to the docs.

ritch:
We also need to update the `loopback.token()` reference to include this option.

raymondfeng:
@bajtos Any comments before I merge?

server/middleware/token.js
```diff
@@ -11,6 +11,18 @@ var assert = require('assert');
...
 module.exports = token;
 
+/*
+ * Rewrite the url to replace current user literal with the logged in user id
+ */
+function rewriteUserLiteral(req, currentUserLiteral) {
+  if (req.accessToken && req.accessToken.userId && currentUserLiteral) {
+    // Replace /me/ with /current-user-id/
+    req.url = req.url.replace(
+      new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g'),
```
bajtos: This assumes "currentUserLiteral" does not contain any special regexp characters like ".". This requirement should be captured in the docs and API docs.

bajtos: I am also wondering why are you replacing all occurrences of "/me" via the "g" flag?

server/middleware/token.js
```diff
@@ -39,19 +51,33 @@ module.exports = token;
...
  */
 
 function token(options) {
   options = options || {};
   var TokenModel = options.model || loopback.AccessToken;
-  assert(TokenModel, 'loopback.token() middleware requires a AccessToken model');
+  if (typeof TokenModel === 'string') {
+    // Make it possible to configure the model in middleware.json
+    TokenModel = loopback.getModel(TokenModel);
+  }
```
bajtos: Unrelated change, please move it to a standalone commit (or even a PR).

server/middleware/token.js
```diff
@@ -39,19 +51,33 @@ module.exports = token;
...
+  var currentUserLiteral = options.currentUserLiteral;
+  if (currentUserLiteral && (typeof currentUserLiteral !== 'string')) {
```
bajtos: You should report a warning (at least via `debug`) if currentLiteralString is defined and not a string. Otherwise users can be confused why their setting was ignored.

Should we support `currentLiteralString: false` to disable this feature?

test/access-token.test.js
```diff
@@ -108,6 +108,48 @@ describe('loopback.token(options)', function() {
...
+        .get('/users/me')
+        .set('authorization', id)
+        .end(function(err, res) {
+          assert(!err);
+          assert.deepEqual(res.body, {userId: '123'});
+          done();
+        });
+    });
+
+  it('should rewrite url for the current user literal at the end',
```
bajtos: This is the same name as on L111, please make it unique and descriptive.

test/access-token.test.js
```diff
@@ -108,6 +108,48 @@ describe('loopback.token(options)', function() {
...
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var id = this.token.id;
+      request(app)
+        .get('/users/me')
+        .set('authorization', id)
+        .end(function(err, res) {
+          assert(!err);
+          assert.deepEqual(res.body, {userId: '123'});
```
bajtos: You should not use magic values like `123` that are specified deep in test helpers. You should be able to read the user id from `token.userId`, right?

Please fix this issue in all other tests too.

test/access-token.test.js
```diff
@@ -284,7 +326,7 @@ describe('app.enableAuth()', function() {
 
 function createTestingToken(done) {
   var test = this;
-  Token.create({}, function(err, token) {
+  Token.create({userId: '123'}, function(err, token) {
```
bajtos: See my previous comment, this should not be needed.

bajtos:
A couple more things to consider.

1) Is the middleware the right place for this feature? What if the app is using JSON-RPC or WebSockets? Shouldn't this be implemented by `User.findById` instead?

2) Is it a good idea to shoehorn this into token middleware? Shouldn't this be a standalone middleware, e.g. `loopback.currentUserUrl`?

server/middleware/token.js
raymondfeng: By default, this feature is NOT enabled as `currentUserLiteral === undefined`. Setting `currentUserLiteral` to a `falsey` value will disable it too.

bajtos:
> We need to make it clear that you can use req.originalUrl to differentiate between /users/me and /users/7.

IIRC, `originalUrl` is a full path including any prefixes stripped by outside apps, i.e. if you mount REST at "/api", the rest adapter will see `url: /users/1234` and `originalUrl: /api/users/1234`. In which case it may be worth to store the "me" url in a new property. I may be wrong though, feel free to ignore this comment.

server/middleware/token.js
bajtos: > By default, this feature is NOT enabled as currentUserLiteral === undefined. Setting currentUserLiteral to a falsey value will disable it too.

I see, makes sense.

I still think we should let the user know when we decide to ignore his config option, e.g. because it's an object or an array.

raymondfeng:
The `originalUrl` is the full path and it won't be rewritten.

bajtos:
> The originalUrl is the full path and it won't be rewritten.

This is my point: If my assumptions are correct, then there is no way for the rest adapter to get the unmodified local url, i.e. "/users/me". That's why I am proposing to save this URL in a new property.

test/access-token.test.js
raymondfeng: This is needed to make sure there is a userId property in the token instance.

bajtos: Fair enough, just don't use the magic constant `123` in test assertions please.

raymondfeng:
> 1) Is the middleware the right place for this feature? What if the app is using JSON-RPC or WebSockets? Shouldn't this be implemented by User.findById instead?

The currentUserLiteral rewrite is only supported for the token-based authentication. It doesn't make sense to implement it in `User.findById` because the `me` context is not a concern of `User.findById`.

> 2) Is it a good idea to shoehorn this into token middleware? Shouldn't this be a standalone middleware, e.g. loopback.currentUserUrl?

In the longer term, I'm open to refactor it into a `loopback.urlrewrite` middleware which will support more rules beyond `currentUserLiteral`. Any rules that reference the `currentUserLiteral` can only be activated in a phase after `auth` though. For example, 
```json
"intial": {
    "loopback#urlrewrite": {
      "params": {
        "rules": [...]
      }
    }
  },
"auth:after": {
    "loopback#urlrewrite": {
      "params": {
        "currentUserLiteral": "me",
        "rules": [...]
      }
    }
  }
```

bajtos:
> The currentUserLiteral rewrite is only supported for the token-based authentication. It doesn't make sense to implement it in User.findById because the me context is not a concern of User.findById.

How about a "beforeRemote" hook then? It has access to the parsed value of "id" parameter and also access to the request context (to get the current user id).

The implementation proposed in this pull request does not discriminate between different places where the "me" string can occur in the URL, thus it's possible it will modify non-user paths too.

I don't like the implementation proposed here from the reasons above. However, if you believe this is the best solution possible, then I can live with it and you can land this patch as it is now.

bajtos:
> you can land this patch as it is now.

After you fix the failing build please.

raymondfeng:
The build error should be fixed. It was due to a style issue.

raymondfeng:
> How about a "beforeRemote" hook then? It has access to the parsed value of "id" parameter and also access to the request context (to get the current user id).
The implementation proposed in this pull request does not discriminate between different places where the "me" string can occur in the URL, thus it's possible it will modify non-user paths too.

It depends on the scope of url rewrite. If we just want to deal with user model (only /api/users/:id/...), it can be implemented as a beforeRemote hook. I see cases that `me` can be used beyond the user model apis, including the api gateway with oAuth 2. We also need to understand if the url rewrite should happen before other middleware that rely on the userId.

server/middleware/token.js
bajtos: @raymondfeng how about this part? ^^^

bajtos:
> It depends on the scope of url rewrite. If we just want to deal with user model (only /api/users/:id/...), it can be implemented as a beforeRemote hook. I see cases that me can be used beyond the user model apis, including the api gateway with oAuth 2. We also need to understand if the url rewrite should happen before other middleware that rely on the userId.

Fair enough. My primary goal was to ensure that we have considered all options and picked the best one.

I don't like the unrelated change squashed into this patch, but that's a detail I can live with.

LGTM.

james-relyea:
Currently, if you use an expired/invalid access token when accessing relations of the user model, using the 'me' literal causes a 404 error instead of 401. Is this intentional? To reproduce:

GET /api/users/<user-id>/<some-relation>?access_token=<valid_access_token>, returns 200
GET /api/users/me/<some-relation>?access_token=<valid_access_token>, returns 200

GET /api/users/<user-id>/<some-relation>?access_token=<invalid_access_token>, returns 401
GET /api/users/me/<some-relation>?access_token=<invalid_access_token>, returns 404

Should this be a 404 if the path exists for authenticated users? I would think that if using the 'me' literal requires an authenticated user, it should return a 401 if the authentication fails.

james-relyea:
@raymondfeng ^^

PR #1172, opened 2015-03-05
Upgrade deps

raymondfeng:
/to @ritch or @bajtos 

I'm tired of the deprecation warnings :-).


test/replication.test.js
```diff
@@ -75,7 +75,8 @@ describe('Replication / Change APIs', function() {
         if (err) return done(err);
         SourceModel.changes(FUTURE_CHECKPOINT, {}, function(err, changes) {
           if (err) return done(err);
-          expect(changes).to.be.empty();
+          /*jshint -W030 */
+          expect(changes).to.be.empty;
```
bajtos: This is weird, why did you have to replace the function call `empty()` with a property access `empty`? Are you sure this test is not passing just because `empty` is not defined at all?

The same comment applies to other test changes.

bajtos: This is weird, why did you have to replace the function call `empty()` with a property access `empty`? Are you sure this test is not passing just because `empty` is not defined at all?

The same comment applies to other similar changes in the test files.

bajtos: Oh, I found the explanation in chai's release notes:

> Unfortunately with 1.10.0 - compatibility broke with older versions because of
the addChainableNoop. This change has been reverted.
> Any plugins using addChainableNoop should cease to do so.
> Any developers wishing for this behaviour can use dirty-chai
by @joshperry

https://www.npmjs.com/package/dirty-chai

In which case we should really start using dirty-chai.

PR #1174, opened 2015-03-05
Static ACL support array of properties now

ulion:
to make the acl rule reusable for multiple properties. now the static acl in model json support array of string to setup acl in batch for group of properties.

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -256,10 +256,14 @@ module.exports = function(ACL) {
     var staticACLs = [];
     if (modelClass && modelClass.settings.acls) {
       modelClass.settings.acls.forEach(function(acl) {
-        if (!acl.property || acl.property === ACL.ALL || property === acl.property) {
+        var prop = acl.property;
+        // We support static ACL property with array of string values.
+        if (Array.isArray(prop) && prop.indexOf(property) >= 0)
+          prop = property;
+        if (!prop || prop === ACL.ALL || property === prop) {
           staticACLs.push(new ACL({
             model: model,
-            property: acl.property || ACL.ALL,
+            property: prop || ACL.ALL,
             principalType: acl.principalType,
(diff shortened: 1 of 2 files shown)


slnode:

common/models/acl.js
```diff
@@ -256,10 +256,10 @@ module.exports = function(ACL) {
     var staticACLs = [];
     if (modelClass && modelClass.settings.acls) {
       modelClass.settings.acls.forEach(function(acl) {
-        if (!acl.property || acl.property === ACL.ALL || property === acl.property) {
+        if (!acl.property || acl.property === ACL.ALL || property === acl.property || Array.isArray(acl.property) && acl.property.indexOf(property) >= 0) {
```
raymondfeng: Please reformat the code so that it doesn't go beyond 80 chars per line.

raymondfeng:
Thanks for the patch. Please add a test case too.

ulion:
adjusted and added test code for it.

2015-03-06 0:33 GMT+08:00 Raymond Feng <notifications@github.com>:

> Thanks for the patch. Please add a test case too.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/strongloop/loopback/pull/1174#issuecomment-77397350>.
>



-- 
Ulion

ulion:
Thank you

raymondfeng:
@crandmck Can we add it to our docs? For the same ACL rule, now we support an array of values for the `property` attribute to match more than one methods. For example,
```json
"acls": [ {
  "principalType": "ROLE",
  "principalId": "$everyone",
  "property": ["create", "update"],
  "permission": "DENY"
  }
]
```

crandmck:
Added a mention to http://docs.strongloop.com/display/LB/Controlling+data+access and http://docs.strongloop.com/display/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-ACLs

PR #1176, opened 2015-03-05
Prevent more kinds of false replication conflicts

bajtos:
- Rework the Change model to merge changes made within the same Checkpoint.

 - Rework `replicate()` to run multiple iteration until there were no changes replicated. This ensures that the target model is left in a clean state with no pending changes associated with the latest (current) checkpoint.

Here is the test that was failing before this patch:

```
it('replicates multiple updates within the same CP', function(done) {
   async.series([
     replicateExpectingSuccess(),
     verifyModelsAreEqual(),

     updateSourceInstanceNameTo('updated'),
     updateSourceInstanceNameTo('again'),
     replicateExpectingSuccess(),
     verifyModelsAreEqual()
   ], done);
});
```

Connect to #985

/to @ritch please review

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -840,18 +840,6 @@ PersistedModel.replicate = function(since, targetModel, options, callback) {
   options = options || {};
 
   var sourceModel = this;
-  var diff;
-  var updates;
-  var Change = this.getChangeModel();
-  var TargetChange = targetModel.getChangeModel();
-  var changeTrackingEnabled = Change && TargetChange;
-  var newSourceCp, newTargetCp;
-
-  assert(
-    changeTrackingEnabled,
-    'You must enable change tracking before replicating'
-  );
-
   callback = callback || function defaultReplicationCallback(err) {
     if (err) throw err;
   };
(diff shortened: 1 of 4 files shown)


bajtos:
@ritch I have added one more commit with integration tests. It turns out that the checkpoint-based filtering is pretty powerful and prevents false conflicts in the typical replication scenarios.

I suspect we may still report a false conflict after a conflict was merged with a manual edit, because each change has only one parent. I think that's an acceptable limitation for v1.0. I did not write a test for that thought, I'll try to write one later this week.

lib/persisted-model.js
```diff
@@ -839,21 +840,64 @@ PersistedModel.replicate = function(since, targetModel, options, callback) {
   options = options || {};
 
   var sourceModel = this;
-  var diff;
-  var updates;
-  var Change = this.getChangeModel();
+  callback = callback || function defaultReplicationCallback(err) {
+    if (err) throw err;
```
ritch: Is this new behavior? I am wary of ever throwing in a callback...

lib/persisted-model.js
```diff
@@ -839,21 +840,64 @@ PersistedModel.replicate = function(since, targetModel, options, callback) {
...
+  var MAX_ATTEMPTS = 3;
+
+  run(1, since);
+
+  function run(attempt, since) {
+    debug('\titeration #%s', attempt);
+    tryReplicate(sourceModel, targetModel, since, options, next);
+
+    function next(err, conflicts, cps, updates) {
+      var finished = err || conflicts.length ||
```
ritch: Hmm... thats a lot of `or`s. -1

lib/persisted-model.js
```diff
@@ -866,17 +910,38 @@ PersistedModel.replicate = function(since, targetModel, options, callback) {
...
+      cb(err, result);
+    }
   }
 
   function getDiffFromTarget(sourceChanges, cb) {
-    targetModel.diff(since.target, sourceChanges, cb);
+    targetModel.diff(since.target, sourceChanges, debug.enabled ? log : cb);
+
+    function log(err, result) {
+      if (err) return cb(err);
```
ritch: should probably start this with

`if(!debug.enabled) return;`

lib/persisted-model.js
bajtos: Nope, this is just moving existing code around.

lib/persisted-model.js
bajtos: What do you propose?

lib/persisted-model.js
bajtos: See L924 - `debug.enabled ? log : cb`

lib/persisted-model.js
ritch: assign some of the sub results to descriptive variables... but I can live with this

ritch:
> I suspect we may still report a false conflict after a conflict was merged with a manual edit

Is this in the case of a race? Once you `resolve()` the change should point to the correct `prev` and no longer be a conflict.

ritch:
> Rework replicate() to run multiple iteration until there were no changes replicated. 

I'm not sure why this is required... but I'll take your word for it. I'm a bit worried about how costly a single `replicate()` call is becoming... As in 4-5+ (just an estimate) requests.

Although with the forthcoming pubsub module... we should be able to reduce the need to call `replicate()` many times.

ritch:
Anyways I'm fine merging this now as is. :+1:

Issue #1181, opened 2015-03-06
Cannot create model without id property


komarik0:
Model has id property by default so it's not possible to create model without id.

prashantbaid:
+1 Same problem for me. @komarik0 your solution works. Wonder when this will be fixed.

loay:
Hi @komarik0 @prashantbaid 
in order to create a model without a default id, you need to do the following:

in the model json file, please add the following
1)  "idInjection": false,
2) in your property that is not the defaulted id, add  "id": true

Example: 
```
  "name": "myModelName",
  "base": "PersistedModel",
  "idInjection": false,
  "properties": {
    "myModelFirstProperty": {
      "type": "string",
      "required": true,
            "id": true
    }
  },
```
_

loay:
I am closing the issue. If you are still having a problem with that fix, please feel free to leave a comment and I will re-open. Thanks.

PR #1182, opened 2015-03-06
Cannot create model without id property #1181

komarik0:


slnode:

komarik0:
I think I have found a better solution to fix this. It will affect loopback-datasource-juggler.

PR #1187, opened 2015-03-06
Add replication tests for conflict resolution; run replication tests in the browser too

bajtos:
Connect to #985 

/to @ritch please review

### DIFF
--- a/test/checkpoint.test.js
+++ b/test/checkpoint.test.js
@@ -3,7 +3,11 @@ var loopback = require('../');
 
 // create a unique Checkpoint model
 var Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');
-Checkpoint.attachTo(loopback.memory());
+
+var memory = loopback.createDataSource({
+  connector: loopback.Memory
+});
+Checkpoint.attachTo(memory);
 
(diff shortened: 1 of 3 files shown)


bajtos:
Few comments:

 - I did not include tests for handling of "deleted" conflicts. If you think it's important, then I can add them next week, possibly in a new pull request

 - It would be nice if `Conflict` provided API for the three ways how conflicts are usually resolved. With the current implementation, I had to look at loopback-example-full-stack to learn what's the correct solution.

 - DOC note: When doing bidirectional sync, it's important to replicate from Client to Server first, so that any conflicts are resolved at the client side.

ritch:
> It would be nice if Conflict provided API for the three ways how conflicts are usually resolved. With the current implementation, I had to look at loopback-example-full-stack to learn what's the correct solution.

Agreed 

LGTM

ritch:
I'm not worried about the delete tests. Also the ordering issue needs to be pretty clear in the docs otherwise it sounds like a bug.

bajtos:
I filled a new issue for the better `Conflict` API - see https://github.com/strongloop/loopback/issues/1188

Issue #1166, opened 2015-03-04
Replication and ACLs


bajtos:
The replication algorithm should check that the user invoking the action has permissions to modify the model instances replicated.

**Tasks**

 - [x] replication #1270
 - [x] conflict resolution #1294
 - [x] documentation update

chandadharap:
link PR

bajtos:
@chandadharap @ritch I'd like to check with you what's the status of this story? It's assigned to Sync v1.0 milestone, but not in the current sprint 67.

bajtos:
Related: https://github.com/strongloop/loopback-connector-remote/issues/3

bajtos:
Done.

PR #1202, opened 2015-03-11
Replication + ACLs

ritch:
This is a fix for: https://github.com/strongloop/loopback/issues/1166

Connect to #1166

@bajtos [This test](https://github.com/strongloop/loopback/pull/1202/files#diff-54ee614ae54f9f63bad0b4bf00121622R70-R79) already passes. I must be missing something. Are you suggesting that we check before we start replicating?

Also can you comment on the tests I've added. I'd like to ensure I've got those right before I move much further.

### DIFF
--- a/test/remote-replication.test.js
+++ b/test/remote-replication.test.js
@@ -1,0 +1,90 @@
+describe('Remote Replication', function() {
+  beforeEach(function(done) {
+    var test = this;
+
+    test.serverApp = loopback();
+    test.serverApp.use(function(req, res, next) {
+      console.log(req.method, req.url);
+      next();
+    });
+    test.serverApp.use(loopback.rest());
+    test.serverApp.enableAuth();
+    test.clientApp = loopback();
+    this.serverApp.set('legacyExplorer', false);
+    var settings = {
+      base: 'PersistedModel',
+      trackChanges: true,
+      dataSource: null,
+      http: {
+        path: 'my-model'
+      }
+    };
+    test.ServerModel = test.serverApp.model('ServerModel', settings);
+    test.ClientModel = test.clientApp.model('ClientModel', settings);
+    var server = test.server = this.serverApp.listen(function(err) {
+      if(err) return done(err);
+      test.port = server.address().port;
+      done();
+    });
+  });
+
+  afterEach(function(done) {
+    this.server.close(done);
+  });
+
+  beforeEach(function(done) {
+    var test = this;
+
+    test.remotes = loopback.createDataSource({
+      connector: 'remote',
+      url: 'http://localhost:' + test.port
+    });
+
+    this.ClientModel.attachTo(test.remotes);
+    this.ServerModel.attachTo(loopback.createDataSource({
+      connector: 'memory'
+    }));
+
+    this.ClientModel.create({foo: 'bar'}, done);
+  });
+
+  describe('client to server replication', function() {
+    it('should replicate a simple change from client to server', function(done) {
+      simpleReplication(this, done);
+    });
+
+    describe('with access controls in place', function() {
+      beforeEach(function() {
+        this.ServerModel.settings.acls = [{
+          "accessType": "*",
+          "principalType": "ROLE",
+          "principalId": "$everyone",
+          "permission": "DENY"
+        },{
+          "accessType": "WRITE",
+          "principalType": "ROLE",
+          "principalId": "$authenticated",
+          "permission": "ALLOW",
+        }];
+      });
+      it('should fail without a logged in user', function(done) {
+        simpleReplication(this, function(err) {
+          expect(err).to.exist;
+          expect(err.message).to.equal('Authorization Required');
+          done();
+        });
+      });
+    });
+  });
+});
+
+
+function simpleReplication(test, done) {
+  test.ClientModel.replicate(test.ServerModel, function(err) {
+    if(err) return done(err);
+    test.ServerModel.findOne(function(err, model) {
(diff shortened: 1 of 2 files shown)


test/remote-replication.test.js
```diff
@@ -0,0 +1,90 @@
+describe('Remote Replication', function() {
+  beforeEach(function(done) {
+    var test = this;
+
+    test.serverApp = loopback();
+    test.serverApp.use(function(req, res, next) {
+      console.log(req.method, req.url);
```
bajtos: Perhaps use debug logs here? `require('debug')('test')('%s %s', req.method, req.url)`

test/remote-replication.test.js
```diff
@@ -0,0 +1,90 @@
...
+    test.serverApp.enableAuth();
+    test.clientApp = loopback();
+    this.serverApp.set('legacyExplorer', false);
+    var settings = {
+      base: 'PersistedModel',
+      trackChanges: true,
+      dataSource: null,
+      http: {
+        path: 'my-model'
+      }
```
bajtos: IIRC this setting is not honoured by the remoting connector, see https://github.com/strongloop/loopback-connector-remote/issues/16. Perhaps I am wrong?

test/remote-replication.test.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+  });
+});
+
+
+function simpleReplication(test, done) {
+  test.ClientModel.replicate(test.ServerModel, function(err) {
+    if(err) return done(err);
+    test.ServerModel.findOne(function(err, model) {
+      expect(model.foo).to.equal('bar');
```
bajtos: Magic string :( Please save the created model in the beforeeach hook and rework this check to `expect(model.toObject()).to.eql(test.model.toObject())`

test/remote-replication.test.js
```diff
@@ -0,0 +1,90 @@
...
+    var settings = {
+      base: 'PersistedModel',
+      trackChanges: true,
+      dataSource: null,
+      http: {
+        path: 'my-model'
+      }
+    };
+    test.ServerModel = test.serverApp.model('ServerModel', settings);
+    test.ClientModel = test.clientApp.model('ClientModel', settings);
```
bajtos: You should use unique model names. Registering multiple models with the same name causes all sorts of weird behaviour.

bajtos:
> This test already passes. I must be missing something. 

I don't know enough about our auth&auth implementation to be able to help you. Few things that stand out in the current code:

 - You are calling `app.enableAuth()` for all tests, even for the test that verifies anonymous mode
 - The restrictive ACLs are enabled after the application and the model was already set up. Is that valid approach, I mean will the auth code pick up these changes?

I don't know how to help you here without debugging your new tests, which is something you can do yourself too.

> Are you suggesting that we check before we start replicating?

I am not following you here, what should we check before we start?

> Also can you comment on the tests I've added. I'd like to ensure I've got those right before I move much further.

Yes, it looks good so far. Few more scenarios to consider:

 - a happy path where replication succeeds for a logged-in user
 - replication is rejected for a logged-in user due to permissions

It may be a good idea to test all three operations in all three scenarios: create, update, delete.

On more idea to consider: bulkUpdate should be as atomic as possible. If any of the updates is rejected by auth, then no updates should be applied at all.

bajtos:
> This test already passes. I must be missing something.

I think the test is passing because the ACL implementation reject anonymous requests when there are ACL configured for the operation, thus I guess this is mostly expected.

A more interesting scenario is described in my comment above, where the user is authenticated, but they don't have permissions to create/update/delete the specific model instance.

The built-in `User` model is a good example:
 - regular users cannot create new `User` instances (accounts)
 - a regular user can edit only his own account
 - a regular user can either delete only his own account or cannot delete any account at all

I am not sure how exactly the current implementation works, please take the above items with a grain of salt.

test/remote-replication.test.js
ritch: What? They are unique...

ritch:
> I don't know how to help you here without debugging your new tests, which is something you can do yourself too.

I am more concerned with what exactly led you to create https://github.com/strongloop/loopback/issues/1166.

bajtos:
As discussed in chat, I didn't have any specific case where it should not work, just that we didn't have any tests and I had a feeling that a the current implementation of authentication may not work, because:

 - auth is implemented as a `beforeRemote`, thus it verifies permissions at the granularity of a single method (`bulkUpdate` in this case)
 - bulkUpdate makes possibly multiple create/update/delete calls, however these calls don't go through the authentication layer

What I would like to see is a suite of tests that will show that bulkUpdate respects permissions like I described in my comment above, i.e.g that when a user "Joe" can read all Products but cannot modify any of them, then a bulk update will reject a request containing updates of Product instances. When "Employee" record is viewable by all users but only owner can edit it, then a bulk update containing update of different employee than the logged in user will fail.

test/remote-replication.test.js
bajtos: No. This code is called before each test, and every time you call this code, a new model class is created and registered in the global static registry under the same name "ServerModel". It's safer to use a name that is unique *per test*, see replication tests for an example: https://github.com/strongloop/loopback/blob/131633f50d9d37f8f081bfe1f19005db92e44e99/test/replication.test.js#L21-L31

This is not strictly necessary, in many cases it's ok to override model with the same name. It's just that the case where it's not ok are difficult to identify, because the errors are very subtle.

bajtos:
Closing in favour of #1270

PR #1208, opened 2015-03-12
fix implementation of Role methods: users,roles, and applications

### LINKED ISSUE #1151: role.users doesn't invoke callback
The role.users function doesn't invoke the callback if the query succeeds. Check comments in the code below to see where...

```javascript
Role.prototype.users = function(callback) {
  roleMappingModel.find({where: {roleId: this.id,
    principalType: RoleMapping.USER}}, function(err, mappings) {
    if (err) {
      if (callback) callback(err); // callback is invoked on error, good
      return;
    }
    return mappings.map(function(m) {
      return m.principalId;
    });
   // callback is never invoked...hangs
  });
};
```
Should this just be returning the array via callback? 

Another thing I noticed is that its checking for the existence of the callback
```javascript
if (err) {
  if (callback) callback(err);
  return;
}
```
Since this is happening within a an anonymous function after a call to `roleMappingModel.find` which is not being returned, shouldn't a callback be required?

https://github.com/strongloop/loopback/blob/d24da8c37d9b5ba26c7572f5cc950698c0831e3d/common/models/role.js#L39

The `Role.prototype.applications` and `Role.prototype.roles` functions have the same issue.

esco:
Fixes #1151

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -29,42 +29,55 @@ module.exports = function(Role) {
   // Set up the connection to users/applications/roles once the model
   Role.once('dataSourceAttached', function() {
     var roleMappingModel = this.RoleMapping || loopback.getModelByType(RoleMapping);
+
+    /**
+     * Fetch the ids of all users assigned to this role
+     * @param  {Function} callback
+     */
     Role.prototype.users = function(callback) {
-      roleMappingModel.find({where: {roleId: this.id,
-        principalType: RoleMapping.USER}}, function(err, mappings) {
+      roleMappingModel.find({
+        where: {roleId: this.id, principalType: RoleMapping.USER}
+      }, function(err, mappings) {
         if (err) {
-          if (callback) callback(err);
-          return;
+          return callback(err);
         }
-        return mappings.map(function(m) {
+        callback(null, mappings.map(function(m) {
           return m.principalId;
-        });
+        }));
       });
     };
 
+    /**
+     * Fetch the ids of all applications assigned to this role
+     * @param  {Function} callback
+     */
     Role.prototype.applications = function(callback) {
-      roleMappingModel.find({where: {roleId: this.id,
-        principalType: RoleMapping.APPLICATION}}, function(err, mappings) {
+      roleMappingModel.find({
+          where: {roleId: this.id, principalType: RoleMapping.APPLICATION}
+      }, function(err, mappings) {
         if (err) {
-          if (callback) callback(err);
-          return;
+          return callback(err);
         }
-        return mappings.map(function(m) {
+        callback(null, mappings.map(function(m) {
           return m.principalId;
-        });
+        }));
       });
     };
 
+    /**
+     * Fetch the ids of all roles assigned to this role
+     * @param  {Function} callback
+     */
     Role.prototype.roles = function(callback) {
-      roleMappingModel.find({where: {roleId: this.id,
-        principalType: RoleMapping.ROLE}}, function(err, mappings) {
+      roleMappingModel.find({
+        where: {roleId: this.id, principalType: RoleMapping.ROLE}
+      }, function(err, mappings) {
         if (err) {
-          if (callback) callback(err);
-          return;
+          return callback(err);
         }
-        return mappings.map(function(m) {
+        callback(null, mappings.map(function(m) {
           return m.principalId;
-        });
+        }));
       });
     };
 
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -63,7 +63,7 @@ module.exports = function(Role) {
        */
       Role.prototype[pluralName] = function(query, callback) {
         listByPrincipalType(model, principalType, query, callback);
-      }
+      };
     });
 
     /**
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -72,7 +72,7 @@ module.exports = function(Role) {
      * @param {*} model model type to fetch
      * @param {String} [principalType] principalType used in the rolemapping for model
      * @param {object} [query] query object passed to model find call
-     * @param  {Function} [callback]
+     * @param  {Function} [callback] callback function called with `(err, models)` arguments.
      */
     function listByPrincipalType(model, principalType, query, callback) {
       if (callback === undefined) {
@@ -92,7 +92,9 @@ module.exports = function(Role) {
         });
         query.where = query.where || {};
         query.where.id = {inq: ids};
-        model.find(query, callback);
+        model.find(query, function(err, models) {
+          callback(err, models);
+        });
       });
     }
 
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -68,6 +68,7 @@ module.exports = function(Role) {
 
     /**
      * Fetch all models assigned to this role
+     * @private
      * @param {*} model model type to fetch
      * @param {String} [principalType] principalType used in the rolemapping for model
      * @param {object} [query] query object passed to model find call
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -40,27 +40,27 @@ module.exports = function(Role) {
     principalTypesToModels[RoleMapping.APPLICATION] = Application;
     principalTypesToModels[RoleMapping.ROLE] = Role;
 
-    Object.keys(principalTypesToModels).forEach(function(principalType){
+    Object.keys(principalTypesToModels).forEach(function(principalType) {
       var model = principalTypesToModels[principalType];
       var pluralName = model.pluralModelName.toLowerCase();
-      /** 
+      /**
        * Fetch all users assigned to this role
        * @function Role.prototype#users
        * @param {object} [query] query object passed to model find call
        * @param  {Function} [callback]
-       */ 
-      /** 
+       */
+      /**
        * Fetch all applications assigned to this role
        * @function Role.prototype#applications
        * @param {object} [query] query object passed to model find call
        * @param  {Function} [callback]
        */
-      /** 
+      /**
        * Fetch all roles assigned to this role
        * @function Role.prototype#roles
        * @param {object} [query] query object passed to model find call
        * @param {Function} [callback]
-       */ 
+       */
       Role.prototype[pluralName] = function(query, callback) {
         listByPrincipalType(model, principalType, query, callback);
       }
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -6,6 +6,10 @@ var async = require('async');
 var AccessContext = require('../../lib/access-context').AccessContext;
 
 var RoleMapping = loopback.RoleMapping;
+var Role = loopback.Role;
+var User = loopback.User;
+var Application = loopback.Application;
+
 assert(RoleMapping, 'RoleMapping model must be defined before Role model');
 
 /**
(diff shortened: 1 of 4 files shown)


slnode:

common/models/role.js
```diff
@@ -29,42 +29,55 @@ module.exports = function(Role) {
...
+        where: {roleId: this.id, principalType: RoleMapping.USER}
+      }, function(err, mappings) {
         if (err) {
-          if (callback) callback(err);
-          return;
+          return callback(err);
         }
-        return mappings.map(function(m) {
+        callback(null, mappings.map(function(m) {
           return m.principalId;
```
raymondfeng: I wonder if we should return the principalId or resolve it to the user object.

esco: I considered that, but there were concerns around the potential size of the document so I decided to follow the original idea of returning the id. What do you think about accepting a fields param as the second argument and passing it to the underlying `find` so that the user isn't forced to load the entire document every time they use this function?

```
Role.prototype.users = function(fields, callback) {
  if (callback === undefined) {
    callback = fields;
    fields = {};
  }

  roleMappingModel.find({
    where: {roleId: this.id, principalType: RoleMapping.USER}
  }, function(err, mappings) {
    var ids;

    if (err) {
      return callback(err);
    }
    
    ids = mappings.map(function(m) {
      return m.principalId;
    });

    User.find({
      where: {id: {inq: ids}, fields: fields}
    }, callback);
  });
};
```

esco: Now that I think about it more, another concern is that the user is being forced to load every single document associated with the Role. Imagine there were 10,000 User documents associated with a role, a reasonable scenario. This function would be unusable unless there was a way to implement pagination. Maybe we should accept a query param instead of fields so that the user could use limit, skip, and order.

lib/builtin-models.js
```diff
@@ -13,6 +13,10 @@ module.exports = function(loopback) {
     require('../common/models/access-token.json'),
     require('../common/models/access-token.js'));
 
+  loopback.User = createModel(
+    require('../common/models/user.json'),
+    require('../common/models/user.js'));
+
   loopback.RoleMapping = createModel(
```
esco: moved user up so that the User model is accessible when Roles is required

package.json
```diff
@@ -83,6 +83,7 @@
     "loopback-datasource-juggler": "^2.19.1",
     "loopback-testing": "^1.1.0",
     "mocha": "^2.1.0",
+    "sinon": "^1.13.0",
```
esco: added dev dependency sinon to use spies in the roles test

common/models/role.js
```diff
@@ -29,44 +33,66 @@ module.exports = function(Role) {
...
+       */
+      /** 
+       * Fetch all roles assigned to this role
+       * @function Role.prototype#roles
+       * @param {object} [query] query object passed to model find call
+       * @param {Function} [callback]
+       */ 
+      Role.prototype[pluralName] = function(query, callback) {
+        listByPrincipalType(model, principalType, query, callback);
+      }
```
esco: consolidate most of the logic from users, applications, and roles into one function

common/models/role.js
```diff
@@ -29,44 +33,66 @@ module.exports = function(Role) {
...
-          return;
+          return callback(err);
         }
-        return mappings.map(function(m) {
+        ids = mappings.map(function(m) {
           return m.principalId;
         });
+        query.where = query.where || {};
+        query.where.id = {inq: ids};
+        model.find(query, callback);
```
esco: return the model instead of just the id

raymondfeng:
@esco Have you explored http://docs.strongloop.com/display/LB/Polymorphic+relations? I think the role mapping is a good candidate for polymorphic relations. It would be ideal to use that.

esco:
@raymondfeng I checked out the docs for polymorphic relations. How does this sound: Role hasMany enrollable through RoleMapping, where "enrollable" is a model that can be assigned to a role such as User,Application and Role.

raymondfeng:
What about `principal`? The foreign key is principalId` and discriminator is `principalType'.

esco:
@raymondfeng I believe that would only get us the RoleMapping model, but we want the actual model that the role is assigned to (User, Application, or Role).

raymondfeng:
I'm thinking about the following:
```json
"RoleMapping": {
  "relations": {
    "principal": {
      "type": "belongsTo",
      "polymorphic": {
        "as": "principal",
        "foreignKey": "principalId",
    }
  }
```

raymondfeng:
BTW, if you have immediate need to land the PR, we can merge it 1st before working out the ideal fix.

esco:
It would be great if you could merge the PR. I'll be looking into polymorphic relations towards the end of the week when I get more time.

Thanks!

common/models/role.js
```diff
@@ -30,44 +34,66 @@ module.exports = function(Role) {
...
-      });
-    };
+    /**
+     * Fetch all models assigned to this role
+     * @param {*} model model type to fetch
+     * @param {String} [principalType] principalType used in the rolemapping for model
+     * @param {object} [query] query object passed to model find call
+     * @param  {Function} [callback]
+     */
+    function listByPrincipalType(model, principalType, query, callback) {
```
ritch: Include `@private` to ensure this isn't picked up by the class doc.

common/models/role.js
ritch: Is this breaking backwards compatability?

test/role.test.js
```diff
@@ -208,4 +210,64 @@ describe('role model', function() {
...
+        });
+      });
+    });
+
+    it('should apply query', function(done) {
+      User.create({name: 'Raymond', email: 'x@y.com', password: 'foobar'}, function(err, user) {
+        Role.create({name: 'userRole'}, function(err, role) {
+          role.principals.create({principalType: RoleMapping.USER, principalId: user.id}, function(err, p) {
+            var query = {fields:['id', 'name']};
+            sandbox.spy(User, 'find');
```
ritch: I'm not familiar with `sinon`... what is this `spy()` even used for?

I'd just get rid of it... I like the idea of spies / mocks / etc... but I think its something that should be added from the beginning not in a minor PR a year after the tests were written.

On the other hand... if you want to create a PR that adds sinon to all the tests that would be pretty cool.

ritch:
> It would be great if you could merge the PR. I'll be looking into polymorphic relations towards the end of the week when I get more time.

@esco this mostly looks good. Can you address the minor comments I made? If not we can take on the PR but it will most likely not get to the top of the queue for a while.

common/models/role.js
esco: The previous version did not work, the callback was never called. I don't believe anyone could have been using it because of this issue.

test/role.test.js
esco: A spy is used to examine the calls of a function and the arguments passed to it. In unit tests, its important to test with laser focus. Instead of relying on the result of the .find function, we only need to test that the correct input was passed to it. I would suggest an iterative approach to adding sinon so that some tests can benefit while others are being converted. I could set it up in a way that makes it easy to use in other tests, and implement in some of the other tests when I have time. I saw a few other places that could benefit from it.

ritch: > A spy is used to examine the calls of a function and the arguments passed to it.

Haha yes I know this part. :)

> In unit tests, its important to test with laser focus.

Agreed... And we aren't doing a good job of this really. But I prefer the same kind of focus you are talking about with PRs so we can avoid having to go off track in the PR comments like this :(.

> I would suggest an iterative approach to adding sinon so that some tests can benefit while others are being converted.

Sounds good... just not alongside another PR. Please move it to a separate one. That way we can get some more peoples (eg. @superkhau @bajtos) opinions and chat about it.

>  I could set it up in a way that makes it easy to use in other tests, and implement in some of the other tests when I have time. I saw a few other places that could benefit from it.

Since you might not have time, even just starting a conversation about the specifics could be helpdul. Just keep in mind that this isn't really a priority.

common/models/role.js
bajtos: Please be specific about the arguments you are passing to `callback`. Your current implementation creates a contract saying `listByPrincipalType` passes the same callback arguments as `PersistedModel.find`, which may become difficult to preserve in future refactorings.

Also the jsdoc entry should include description of callback arguments.

test/role.test.js
esco: > Haha yes I know this part. :)

Ah ok ha, I thought you were asking what a spy is, but now realize you were talking about the spy function. The function accepts two args (Object, methodString) and creates a spy out of the method associated with the string.

> Agreed... And we aren't doing a good job of this really. But I prefer the same kind of focus you are talking about with PRs so we can avoid having to go off track in the PR comments like this :(.

Sure that makes sense. I didn't want to submit tests with my PR that I weren't comfortable with but I get your point. Since theres no agreed upon module to use for stubs, spies, and mocks...etc, then basically I'm introducing a module thats expected to be used across all tests without any consensus :blush: 

> Since you might not have time, even just starting a conversation about the specifics could be helpful. Just keep in mind that this isn't really a priority.

I like this approach better. I'll take sinon out of this PR and make an issue to discuss choosing a testing module for spies, and mocks...etc

Thanks for the advice.

OwenBrotherwood:
The npm test fails with indentation, then fails with:

```

  1) Replication / Change APIs Model.replicate(since, targetModel, options, call
back) with 3rd-party changes detects UPDATE made during UPDATE:
     Uncaught TypeError: undefined is not a function


  2) Replication / Change APIs Model.replicate(since, targetModel, options, call
back) with 3rd-party changes detects CREATE made during CREATE:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being call
ed in this test.


  3) Replication / Change APIs Model.replicate(since, targetModel, options, call
back) with 3rd-party changes detects UPDATE made during DELETE:
     Uncaught TypeError: undefined is not a function


  4) Replication / Change APIs Model.replicate(since, targetModel, options, call
back) with 3rd-party changes handles DELETE made during DELETE:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being call
ed in this test.


  5) Replication / Change APIs change detection detects "findOrCreate":
     Uncaught TypeError: undefined is not a function


  6) User User.login should handle multiple `include`:
     Uncaught AssertionError: body.user: expected undefined to not equal undefin
ed
      at Test.<anonymous> (c:\Users\ob_000\esco\loopback\test\user.test.js:398:5
0)
      at net.js:1392:10
``´

Patch for indentation
```
diff --git a/common/models/role-mapping.js b/common/models/role-mapping.js
index 3e87e56..6c7b435 100644
--- a/common/models/role-mapping.js
+++ b/common/models/role-mapping.js
@@ -12,7 +12,7 @@ var loopback = require('../../lib/loopback');
  */

 module.exports = function(RoleMapping) {
-// Principal types
+  // Principal types
   RoleMapping.USER = 'USER';
   RoleMapping.APP = RoleMapping.APPLICATION = 'APP';
   RoleMapping.ROLE = 'ROLE';
diff --git a/common/models/role.js b/common/models/role.js
index 340874f..83b4ba3 100644
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -100,7 +100,7 @@ module.exports = function(Role) {

   });

-// Special roles
+  // Special roles
   Role.OWNER = '$owner'; // owner of the object
   Role.RELATED = '$related'; // any User with a relationship to the object
   Role.AUTHENTICATED = '$authenticated'; // authenticated user
diff --git a/test/app.test.js b/test/app.test.js
index 995ac24..9ca1ac2 100644
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -838,7 +838,7 @@ describe('app', function() {
   });

   describe('app.connector', function() {
-     // any connector will do
+    // any connector will do
     it('adds the connector to the registry', function() {
       app.connector('foo-bar', loopback.Memory);
       expect(app.connectors['foo-bar']).to.equal(loopback.Memory);
```

PR #1212, opened 2015-03-14
Per-app Models

ritch:
This PR adds the following behavior:

**`loopback.registry` is now a true global registry**

If you have multiple `loopback` modules in your `require.cache` you will still get the same registry from `loopback.registry`. All previous `Registry` methods (eg. `loopback.createModel()` and `loopback.getModel()`) will use `loopback.registry`.

This has a side benefit of being able to clear all `loopback` state by resetting this registry (this PR does not add an API for this yet, but it would be trivial). 

**`app.registry`** is unique per `app` object. This allows you to define models with the same name in two separate applications. See the new `registry.test.js` for an example of this.

**`Model.registry`** is set when a `Model` is created using any registry method, or when a `Model` is `extend()`ed (and does not yet have a `registry`).

**`loopback.localRegistry`** and **`loopback({localRegistry: true})`** when set to `true` this will create a `Registry` per `Application`. It defaults to `false`.

**TODO**

 - [x] Fix `autoAttach` errors (like the one below)

```
{ [AssertionError: cannot autoAttach model "PersistedModel". No dataSource found of type db]
  name: 'AssertionError',
  actual: false,
  expected: true,
  operator: '==',
  message: 'cannot autoAttach model "PersistedModel". No dataSource found of type db',
  generatedMessage: false }
```

 - [x] Add a flag that sets `app.registry` as `loopback.registry` for true backwards compatability

/to @bajtos @raymondfeng 
/cc @kraman

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -375,6 +375,8 @@ module.exports = function(ACL) {
    */
 
   ACL.checkAccessForContext = function(context, callback) {
+    var registry = this.registry;
+
     if (!(context instanceof AccessContext)) {
       context = new AccessContext(context);
     }
@@ -394,7 +396,7 @@ module.exports = function(ACL) {
     var staticACLs = this.getStaticACLs(model.modelName, property);
 
     var self = this;
-    var roleModel = loopback.getModelByType(Role);
+    var roleModel = registry.getModelByType(Role);
     this.find({where: {model: model.modelName, property: propertyQuery,
       accessType: accessTypeQuery}}, function(err, acls) {
       if (err) {
--- a/common/models/role-mapping.js
+++ b/common/models/role-mapping.js
@@ -24,9 +24,11 @@ module.exports = function(RoleMapping) {
    * @param {Application} application
    */
   RoleMapping.prototype.application = function(callback) {
+    var registry = this.constructor.registry;
+
     if (this.principalType === RoleMapping.APPLICATION) {
       var applicationModel = this.constructor.Application ||
-        loopback.getModelByType(loopback.Application);
+        registry.getModelByType('Application');
       applicationModel.findById(this.principalId, callback);
     } else {
       process.nextTick(function() {
@@ -42,9 +44,10 @@ module.exports = function(RoleMapping) {
    * @param {User} user
    */
   RoleMapping.prototype.user = function(callback) {
+    var RoleMapping = this.constructor;
     if (this.principalType === RoleMapping.USER) {
-      var userModel = this.constructor.User ||
-        loopback.getModelByType(loopback.User);
+      var userModel = RoleMapping.User ||
+        RoleMapping.registry.getModelByType('User');
       userModel.findById(this.principalId, callback);
     } else {
       process.nextTick(function() {
@@ -60,9 +63,11 @@ module.exports = function(RoleMapping) {
    * @param {User} childUser
    */
   RoleMapping.prototype.childRole = function(callback) {
+    var registry = this.constructor.registry;
+
     if (this.principalType === RoleMapping.ROLE) {
       var roleModel = this.constructor.Role ||
-        loopback.getModelByType(loopback.Role);
+        registry.getModelByType(loopback.Role);
       roleModel.findById(this.principalId, callback);
     } else {
       process.nextTick(function() {
--- a/lib/application.js
+++ b/lib/application.js
@@ -3,7 +3,7 @@
  */
 
 var DataSource = require('loopback-datasource-juggler').DataSource;
-var registry = require('./registry');
+var Registry = require('./registry');
 var assert = require('assert');
 var fs = require('fs');
 var extend = require('util')._extend;
@@ -104,6 +104,8 @@ app.disuse = function(route) {
 
 app.model = function(Model, config) {
   var isPublic = true;
+  var registry = this.registry;
+
   if (arguments.length > 1) {
     config = config || {};
     if (typeof Model === 'string') {
@@ -130,7 +132,7 @@ app.model = function(Model, config) {
     configureModel(Model, config, this);
     isPublic = config.public !== false;
   } else {
-    assert(Model.prototype instanceof registry.Model,
+    assert(Model.prototype instanceof Model.registry.getModel('Model'),
       Model.modelName + ' must be a descendant of loopback.Model');
   }
 
@@ -216,7 +218,7 @@ app.models = function() {
  * @param {Object} config The data source config
  */
 app.dataSource = function(name, config) {
-  var ds = dataSourcesFromConfig(config, this.connectors);
+  var ds = dataSourcesFromConfig(name, config, this.connectors, this.registry);
   this.dataSources[name] =
   this.dataSources[classify(name)] =
   this.dataSources[camelize(name)] = ds;
@@ -362,14 +364,14 @@ app.boot = function(options) {
     '`app.boot` was removed, use the new module loopback-boot instead');
 };
 
-function dataSourcesFromConfig(config, connectorRegistry) {
+function dataSourcesFromConfig(name, config, connectorRegistry, registry) {
   var connectorPath;
 
   assert(typeof config === 'object',
     'cannont create data source without config object');
 
   if (typeof config.connector === 'string') {
-    var name = config.connector;
+    name = config.connector;
     if (connectorRegistry[name]) {
       config.connector = connectorRegistry[name];
     } else {
@@ -385,7 +387,7 @@ function dataSourcesFromConfig(config, connectorRegistry) {
 }
 
 function configureModel(ModelCtor, config, app) {
-  assert(ModelCtor.prototype instanceof registry.Model,
+  assert(ModelCtor.prototype instanceof ModelCtor.registry.getModel('Model'),
     ModelCtor.modelName + ' must be a descendant of loopback.Model');
 
   var dataSource = config.dataSource;
@@ -405,7 +407,7 @@ function configureModel(ModelCtor, config, app) {
   config = extend({}, config);
   config.dataSource = dataSource;
 
-  registry.configureModel(ModelCtor, config);
+  app.registry.configureModel(ModelCtor, config);
 }
 
 function clearHandlerCache(app) {
--- a/lib/loopback.js
+++ b/lib/loopback.js
@@ -10,6 +10,9 @@ var ejs = require('ejs');
 var path = require('path');
 var merge = require('util')._extend;
 var assert = require('assert');
+var Registry = require('./registry');
+var getGlobalRegistry = require('./global-registry');
+var juggler = require('loopback-datasource-juggler');
 
 /**
  * LoopBack core module. It provides static properties and
@@ -25,6 +28,7 @@ var assert = require('assert');
  * @property {String} mime
  * @property {Boolean} isBrowser True if running in a browser environment; false otherwise.  Static read-only property.
  * @property {Boolean} isServer True if running in a server environment; false otherwise.  Static read-only property.
+ * @property {Registry} registry The global `Registry` object.
  * @property {String} faviconFile Path to a default favicon shipped with LoopBack.
  * Use as follows: `app.use(require('serve-favicon')(loopback.faviconFile));`
  * @class loopback
@@ -45,6 +49,22 @@ loopback.version = require('../package.json').version;
 
 loopback.mime = express.mime;
 
+Object.defineProperty(loopback, 'registry', {
+  get: getGlobalRegistry
+});
+
+Object.defineProperty(loopback, 'Model', {
+  get: function() {
+    return this.registry.getModel('Model');
+  }
+});
+
+Object.defineProperty(loopback, 'PersistedModel', {
+  get: function() {
+    return this.registry.getModel('PersistedModel');
+  }
+});
+
 /*!
  * Create an loopback application.
  *
@@ -52,7 +72,7 @@ loopback.mime = express.mime;
  * @api public
  */
 
-function createApplication() {
+function createApplication(options) {
   var app = loopbackExpress();
 
   merge(app, proto);
@@ -76,6 +96,13 @@ function createApplication() {
   app.connector('memory', loopback.Memory);
   app.connector('remote', loopback.Remote);
 
+  if (loopback.localRegistry || options && options.localRegistry === true) {
+    // setup the app registry
+    var registry = app.registry = new Registry();
+  } else {
+    app.registry = loopback.registry;
+  }
+
   return app;
 }
 
@@ -91,7 +118,6 @@ function mixin(source) {
 }
 
 mixin(require('./runtime'));
-mixin(require('./registry'));
 
 /*!
  * Expose static express methods like `express.errorHandler`.
@@ -191,8 +217,198 @@ loopback.template = function(file) {
 
 require('../server/current-context')(loopback);
 
+/**
+ * Create a named vanilla JavaScript class constructor with an attached
+ * set of properties and options.
+ *
+ * This function comes with two variants:
+ *  * `loopback.createModel(name, properties, options)`
+ *  * `loopback.createModel(config)`
+ *
+ * In the second variant, the parameters `name`, `properties` and `options`
+ * are provided in the config object. Any additional config entries are
+ * interpreted as `options`, i.e. the following two configs are identical:
+ *
+ * ```js
+ * { name: 'Customer', base: 'User' }
+ * { name: 'Customer', options: { base: 'User' } }
+ * ```
+ *
+ * **Example**
+ *
+ * Create an `Author` model using the three-parameter variant:
+ *
+ * ```js
+ * loopback.createModel(
+ *   'Author',
+ *   {
+ *     firstName: 'string',
+ *     lastName: 'string'
+ *   },
+ *   {
+ *     relations: {
+ *       books: {
+ *         model: 'Book',
+ *         type: 'hasAndBelongsToMany'
+ *       }
+ *     }
+ *   }
+ * );
+ * ```
+ *
+ * Create the same model using a config object:
+ *
+ * ```js
+ * loopback.createModel({
+ *   name: 'Author',
+ *   properties: {
+ *     firstName: 'string',
+ *     lastName: 'string'
+ *   },
+ *   relations: {
+ *     books: {
+ *       model: 'Book',
+ *       type: 'hasAndBelongsToMany'
+ *     }
+ *   }
+ * });
+ * ```
+ *
+ * @param {String} name Unique name.
+ * @param {Object} properties
+ * @param {Object} options (optional)
+ *
+ * @header loopback.createModel
+ */
+
+loopback.createModel = function(name, properties, options) {
+  return this.registry.createModel.apply(this.registry, arguments);
+};
+
+/**
+ * Alter an existing Model class.
+ * @param {Model} ModelCtor The model constructor to alter.
+ * @options {Object} config Additional configuration to apply
+ * @property {DataSource} dataSource Attach the model to a dataSource.
+ * @property {Object} [relations] Model relations to add/update.
+ *
+ * @header loopback.configureModel(ModelCtor, config)
+ */
+
+loopback.configureModel = function(ModelCtor, config) {
+  return this.registry.configureModel.apply(this.registry, arguments);
+};
+
+/**
+ * Look up a model class by name from all models created by
+ * `loopback.createModel()`
+ * @param {String} modelName The model name
+ * @returns {Model} The model class
+ *
+ * @header loopback.findModel(modelName)
+ */
+loopback.findModel = function(modelName) {
+  return this.registry.findModel.apply(this.registry, arguments);
+};
+
+/**
+ * Look up a model class by name from all models created by
+ * `loopback.createModel()`. Throw an error when no such model exists.
+ *
+ * @param {String} modelName The model name
+ * @returns {Model} The model class
+ *
+ * @header loopback.getModel(modelName)
+ */
+loopback.getModel = function(modelName) {
+  return this.registry.getModel.apply(this.registry, arguments);
+};
+
+/**
+ * Look up a model class by the base model class.
+ * The method can be used by LoopBack
+ * to find configured models in models.json over the base model.
+ * @param {Model} modelType The base model class
+ * @returns {Model} The subclass if found or the base class
+ *
+ * @header loopback.getModelByType(modelType)
+ */
+loopback.getModelByType = function(modelType) {
+  return this.registry.getModelByType.apply(this.registry, arguments);
+};
+
+/**
+ * Create a data source with passing the provided options to the connector.
+ *
+ * @param {String} name Optional name.
+ * @options {Object} options Data Source options
+ * @property {Object} connector LoopBack connector.
+ * @property {*} [*] Other&nbsp;connector properties.
+ *   See the relevant connector documentation.
+ */
+
+loopback.createDataSource = function(name, options) {
+  return this.registry.createDataSource.apply(this.registry, arguments);
+};
+
+/**
+ * Get an in-memory data source. Use one if it already exists.
+ *
+ * @param {String} [name] The name of the data source.
+ * If not provided, the `'default'` is used.
+ */
+
+loopback.memory = function(name) {
+  return this.registry.memory.apply(this.registry, arguments);
+};
+
+/**
+ * Set the default `dataSource` for a given `type`.
+ * @param {String} type The datasource type.
+ * @param {Object|DataSource} dataSource The data source settings or instance
+ * @returns {DataSource} The data source instance.
+ *
+ * @header loopback.setDefaultDataSourceForType(type, dataSource)
+ */
+
+loopback.setDefaultDataSourceForType = function(type, dataSource) {
+  return this.registry.setDefaultDataSourceForType.apply(this.registry, arguments);
+};
+
+/**
+ * Get the default `dataSource` for a given `type`.
+ * @param {String} type The datasource type.
+ * @returns {DataSource} The data source instance
+ */
+
+loopback.getDefaultDataSourceForType = function(type) {
+  return this.registry.getDefaultDataSourceForType.apply(this.registry, arguments);
+};
+
+/**
+ * Attach any model that does not have a dataSource to
+ * the default dataSource for the type the Model requests
+ */
+
+loopback.autoAttach = function() {
+  return this.registry.autoAttach.apply(this.registry, arguments);
+};
+
+loopback.autoAttachModel = function(ModelCtor) {
+  return this.registry.autoAttachModel.apply(this.registry, arguments);
+};
+
+// temporary alias to simplify migration of code based on <=2.0.0-beta3
+Object.defineProperty(loopback, 'DataModel', {
+  get: function() {
+    return this.registry.DataModel;
+  }
+});
+
 /*!
  * Built in models / services
  */
 
 require('./builtin-models')(loopback);
+
+loopback.DataSource = juggler.DataSource;
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -1,12 +1,3 @@
-/*
- * This file exports methods and objects for manipulating
- * Models and DataSources.
- *
- * It is an internal file that should not be used outside of loopback.
- * All exported entities can be accessed via the `loopback` object.
- * @private
- */
-
 var assert = require('assert');
 var extend = require('util')._extend;
 var juggler = require('loopback-datasource-juggler');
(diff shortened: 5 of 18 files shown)


common/models/role-mapping.js
```diff
@@ -24,9 +24,11 @@ module.exports = function(RoleMapping) {
    * @param {Application} application
    */
   RoleMapping.prototype.application = function(callback) {
+    var registry = this.constructor.registry || loopback.registry;
```
bajtos: Most of the other code relies on the fact that `ModelCtor.registry` is always set. What is the reason for using `|| loopback.registry` in this method? If `this.constructor.registry` is always set, then the second branch is redundant IMO.

lib/application.js
```diff
@@ -369,7 +371,7 @@ function camelize(str) {
   return stringUtils.camelize(str);
 }
 
-function dataSourcesFromConfig(config, connectorRegistry) {
+function dataSourcesFromConfig(name, config, connectorRegistry, registry) {
```
bajtos: What is the purpose of the new `name` parameter? AFAICT, it's not used anywhere.

lib/loopback.js
```diff
@@ -194,8 +216,197 @@ loopback.getCurrentContext = function() {
...
+loopback.autoAttachModel = function(ModelCtor) {
+  return this.registry.autoAttachModel.apply(this.registry, arguments);
+};
+
+// temporary alias to simplify migration of code based on <=2.0.0-beta3
+Object.defineProperty(loopback, 'DataModel', {
+  get: function() {
+    return this.registry.DataModel;
+  }
+});
```
bajtos: Is there any particular reason for writing these methods by hand, instead of performing an automatic mixin?

```js
Object.keys(loopback.registry).forEach(function(k) {
  var target = loopback.registry[key];
  if (typeof target !== 'function') return;
  if (loopback[key]) return;
  loopback[key] = function() { this.registry[key].apply(this.registry, arguments); };
});
```

lib/registry.js
```diff
@@ -445,6 +432,3 @@ Object.defineProperty(registry, 'DataModel', {
     return this.PersistedModel;
   }
 });
-
-// Set the default model base class. This is done after the Model class is defined.
-registry.modelBuilder.defaultModelBaseClass = registry.Model;
```
bajtos: This code was removed with not replacement. Are you sure it's not a breaking change?

server/middleware/rest.js
```diff
@@ -55,14 +56,8 @@ function rest() {
...
       if (app.isAuthEnabled) {
-        // NOTE(bajtos) It would be better to search app.models for a model
-        // of type AccessToken instead of searching all loopback models.
-        // Unfortunately that's not supported now.
-        // Related discussions:
-        // https://github.com/strongloop/loopback/pull/167
-        // https://github.com/strongloop/loopback/commit/f07446a
-        var AccessToken = loopback.getModelByType(loopback.AccessToken);
-        handlers.push(loopback.token({ model: AccessToken }));
+        var AccessToken = registry.getModelByType(loopback.AccessToken);
```
bajtos: You can't use `loopback.AccessToken` here, as it refers to a model registered in the global registry, while `registry.getModelByType` is searching the local per-app registry.

Also are you sure the comment is no longer relevant?

server/middleware/token.js
```diff
@@ -68,10 +68,16 @@ function escapeRegExp(str) {
 
 function token(options) {
   options = options || {};
+
+  var registry = options.registry;
+  if(!registry && options.app && options.app.registry) {
+    registry = options.app.registry;
+  }
+  
   var TokenModel = options.model || loopback.AccessToken;
```
bajtos: This is wrong, you can't use `loopback.AccessToken` as the default, you need to use the app-local model instead.

The solution is to move the code resolving TokenModel to the middleware, because then you have access to the loopback app object.

```js
function token(options) {
  var TokenModel;
  // etc.
  return function(req, res, next) {
    if (TokenModel === undefined) {
      var app = req.app;
      var registry = app.registry;
      TokenModel = options.model || registry.AccessToken;
      if (typeof TokenModel === 'string') {
        // Make it possible to configure the model in middleware.json
        TokenModel = registry.getModel(TokenModel);
      }
    }
    // etc.
  };
}
```

test/data-source.test.js
```diff
@@ -60,8 +60,8 @@ describe('DataSource', function() {
       });
 
       var Color = ds.createModel('color', {name: String});
-      assert(Color.prototype instanceof loopback.Model);
-      assert.equal(Color.base, loopback.Model);
+      assert(Color.prototype instanceof Color.registry.getModel('Model'));
+      assert.equal(Color.base.modelName, 'PersistedModel');
```
bajtos: I'd prefer to keep constructor-equality checks here instead of switching to name-based checks. Is there any particular reason why it's not enough to replace `loopback.` with `Color.registry.` and keep the rest unchanged?

test/registries.test.js
```diff
@@ -0,0 +1,48 @@
+describe('Registry', function () {
+  describe('one per app', function () {
+    it('should allow two apps to reuse the same model name', function (done) {
+      var appFoo = loopback();
+      var appBar = loopback();
+      var modelName = 'MyModel';
+      var subModelName = 'Sub' + modelName;
```
bajtos: I was expecting that `subModel` is a subclass of `Model`. Can you use a different prefix please, e.g. `relatedModelName`? Or perhaps be more specific and use `Product` and `Category` instead of `Model` and `subModel`.

test/app.test.js
```diff
@@ -708,6 +702,7 @@ describe('app', function() {
     it('looks up the connector in `app.connectors`', function() {
       app.connector('custom', loopback.Memory);
       app.dataSource('custom', { connector: 'custom' });
+      console.log(app.dataSources.custom.name);
```
bajtos: Forgotten console.log?

test/app.test.js
```diff
@@ -674,13 +674,7 @@ describe('app', function() {
...
     it('attaches the model to a datasource', function() {
       app.dataSource('db', { connector: 'memory' });
-      var TestModel = loopback.Model.extend('TestModel');
-      // TestModel was most likely already defined in a different test,
-      // thus TestModel.dataSource may be already set
-      delete TestModel.dataSource;
-
-      app.model(TestModel, { dataSource: 'db' });
-
+      app.model('TestModel', { dataSource: 'db' });
```
bajtos: Unless this test is running with a per-app registry, you need to keep the code deleting `TestModel.dataSource` around. Please add an assertion that `TestModel.dataSource` is not defined, that will show you whether you need to delete it first or not.

Note that if you remove the code and thus the dataSource will be already set, this test will become a tautology, i.e. it will always pass, regardless of the implementation.

test/app.test.js
```diff
@@ -628,7 +628,7 @@ describe('app', function() {
       var Foo = app.models.foo;
       var f = new Foo();
 
-      assert(f instanceof loopback.Model);
+      assert(f instanceof app.registry.getModel('Model'));
```
bajtos: Will this work too? `f instanceof app.registry.Model`

server/middleware/rest.js
```diff
@@ -28,6 +28,7 @@ function rest() {
 
   return function restApiHandler(req, res, next) {
     var app = req.app;
+    var registry = app.registry || loopback.registry;
```
bajtos: Please don't ask users of `app.registry` to handle the case when the per-app registries are not enabled. `app.registry` should be always defined and point to `loopback.registry` when a global registry is used.

common/models/role-mapping.js
```diff
@@ -24,9 +24,11 @@ module.exports = function(RoleMapping) {
...
+
     if (this.principalType === RoleMapping.APPLICATION) {
       var applicationModel = this.constructor.Application ||
-        loopback.getModelByType(loopback.Application);
+        registry.getModelByType(loopback.Application);
```
bajtos: As I explained elsewhere, you must not use `loopback.Application` here, as it is potentially defined in a different registry.

test/registries.test.js
```diff
@@ -0,0 +1,48 @@
...
+      var settings = {base: 'PersistedModel'};
+      appFoo.set('perAppRegistries', true);
+      appBar.set('perAppRegistries', true);
```
bajtos: This flag does not seem to be implemented at all. I'd also prefer a different name. Since the flag is per app object, using plural name does not make sense to me.

I am proposing `app.set('localModelRegistry', true)` or `app.set('isolateModels', true)`.

lib/loopback.js
```diff
@@ -76,6 +95,10 @@ function createApplication() {
   app.connector('memory', loopback.Memory);
   app.connector('remote', loopback.Remote);
 
+  // setup the app registry
+  var registry = app.registry = new Registry();
+  require('./builtin-models')(registry);
+
```
bajtos: This should be hidden by a flag, right? And if the flag is not set, then `app.registry = loopback.registry`.

In the future (LB 3.x?) I'd like to stop auto-defining built-in models. This is good for reducing the size of the browser bundle and it will also make it easier to move built-in models to a standalone npm module.

I am proposing to use the opportunity created by implementing local per-app registries and don't include builtin-models in the local registries.

bajtos: Also the flag implementation has to be clever, because the `app.registry` object is initialised before `app.set` can be called.

However, I still think it is a good idea to let the application decide whether it wants local or global model registry.

A possible solution if we want to keep the flag configurable via `config.json`:
 - `var app = loopback()` initialises `app.registry = loopback.registry`
 - `app.set('flag', true)` is intercepted and calls `app.registry = new Registry()`

I don't like the fact that if you happen to register models between the line where you create the app and the line where you switch to per-app registry, these models will be "lost".

A better approach is to keep the local/global registry switch in the code.

```js
// Option A - add "options" object, though it's rather un-express-like :(

// Option B - add another factory method
var app = loopback.isolated();

loopback.isolated = function() {
  var app = loopback();
  app.registry = new Registry();
  return app;
}
```

bajtos: Ha, I noticed now that the flag is a known TODO item :)

> Add a flag that sets app.registry as loopback.registry for true backwards compatability

The part "don't include builtin-models in the local registries" is still relevant.

test/registries.test.js
bajtos: I'm deleting this comment, the PR description says that the flag is not implemented yet.

test/data-source.test.js
bajtos: I think the answer is that `Color.registry.Model` is not defined, therefore we have to use `getModel()`. I am wondering if it makes sense to expose at least `Model` and `BaseModel` on the `registry` object?

bajtos:
Great stuff, @ritch! 

I left a bunch of comments, the most important problem I noticed is the usage of `loopback.{modelName}` in model lookups.

bajtos:
>  Fix autoAttach errors (like the one below)

My plan is to drop `autoAttach` in LB 3.x, as it should not be used anymore and it is around only for backwards compatibility.

I am proposing to disable `autoAttach` when using a local per-app registry.

Together with not including built-in models in the local per-app registry, this will give us a great feedback whether we can promote these two changes to a default behaviour in LB 3.x.

bajtos:
@ritch I have renamed the PR from "App Model Namespacing" to "Per-app Models". When I see "namespacing", I am thinking about namespaces in model names preventing clashes between loopback components, e.g. `core.User` vs. `oauth2.User`.

ritch:
> My plan is to drop autoAttach in LB 3.x, as it should not be used anymore and it is around only for backwards compatibility.

Sure that is my plan as well...

> I don't like the fact that if you happen to register models between the line where you create the app and the line where you switch to per-app registry, these models will be "lost".

Ok... so why don't we just make this change (and this PR) the basis for `3.0` and forget about the flag?

ritch:
> In the future (LB 3.x?) I'd like to stop auto-defining built-in models. This is good for reducing the size of the browser bundle and it will also make it easier to move built-in models to a standalone npm module.

More signs that perhaps this should just be `3.0`.

ritch:
> I left a bunch of comments, the most important problem I noticed is the usage of loopback.{modelName} in model lookups.

If we ensure that `app.registry` is always set we should be able to just always reference models using it. Some tests use `loopback.Model.extend()` and similar ways to get at models though.

bajtos:
> Ok... so why don't we just make this change (and this PR) the basis for 3.0 and forget about the flag?

Two reasons:

 - To have an opportunity to test the new approach "in production" before we make a 3.0 release.
 - A simpler upgrade path - people can switch back from the per-all registry to a global registry without having to downgrade the loopback version.

However, if we can come up with a plan how to make all breaking changes planned for 3.0 in a near future (2-6 weeks), then I am ok with starting a 3.0 branch and making preview releases like `3.0.0-alpha.1` which can be used to test the changes in user projects.

bajtos:
Let me elaborate a bit more: refactoring the code relying on a global loopback registry to use a per-app registry only will be probably a very time-consuming task. By shipping a loopback version that exposes the new API like `MyModel.registry` and `app.registry` but preserves the old behaviour of a single global registry, we allow LoopBack users to upgrade their code gradually over longer period of time. I.e. they can locally switch the flag on, rework one or two models to use the new API, switch the flag off; then commit the partial changes they have made and leave the rest of old code to be upgraded later, when there is time available again.

Here is my vision of a roadmap to get rid of global registry:

 1. Provide the new API, implement it using the global registry. Provide a flag to enable per-app registries. (This pull request.)
 2. Test the new mode in as many real-world applications as possible, fix all issues discovered along the way. (We should allow at least several weeks for this).
 3. Deprecate the global registry, start reporting deprecation warnings in application using the old way. This will nudge even more people to upgrade, we should expect new issues to be discovered and we will need to fix them before proceeding to the next step.
 4. Drop the global registry completely, provide only per-app registries.

raymondfeng:
The registry is basically a JavaScript class loader (i.e., Model builder). I see the following opportunities:

1. We can allow registries to form a class loading hierarchy. The root is the global registry. Each app will have its own registry and app.registry.parent === global registry.

2. registry.createModel() will always add the model to the local registry. registry.getModel() will look up in its own scope first before delegate to the parent.

3. Each LoopBack model (or mixin) has two perspectives - definitions and configurations. The definitions should be immutable after loading and they are safe to be defined in the global registry. The configurations are subject to the owning applications. For example, a model can be attached to different data sources for different apps.

4. We can still define (not configure) the built-in models into the global registry and they should be referenceable from apps for config, inheritance or mixin. The global registry should be a place to have the shareable model/mixin definitions.

bajtos:
> We can allow registries to form a class loading hierarchy. The root is the global registry. Each app will have its own registry and app.registry.parent === global registry.

This does not work. Each model can be attached to a single datasource only.

Imagine the situation when "app1" and "app2" are sharing a same model from the global registry, e.g. "AccessToken". The model cannot be attached to both app1 and app2 datasource at the same time.

raymondfeng:
> This does not work. Each model can be attached to a single datasource only.
Imagine the situation when "app1" and "app2" are sharing a same model from the global registry, e.g. "AccessToken". The model cannot be attached to both app1 and app2 datasource at the same time.

In this case, app1 and app2 cannot share the AccessToken as they have different configuration requirements.

bajtos:
> > This does not work. Each model can be attached to a single datasource only.
Imagine the situation when "app1" and "app2" are sharing a same model from the global registry, e.g. "AccessToken". The model cannot be attached to both app1 and app2 datasource at the same time.

> In this case, app1 and app2 cannot share the AccessToken as they have different configuration requirement.

Initially, I had the same idea as you had:

> We can allow registries to form a class loading hierarchy. The root is the global registry. Each app will have its own registry and app.registry.parent === global registry.

But then I realised this approach does not work, because a model definition is not immutable. When you attach a model to a datasource, several aspects of the model definition are modified - an id field is injected if needed, connector-specific methods are added, relation foreign keys and methods are added.

Also since we don't have a hierarchical registry of datasources that would allow applications to share the same datasource, every model attached to a non-null datasource has a different configuration requirement and thus cannot be shared as you said. 

In other words, a typical LoopBack project can share only `Model` and `PersistedModel` and that's really not worth the extra complexity of maintaining hierarchical registries IMO.

lib/loopback.js
bajtos: Also please include `loopback.DataSource`, loopback-boot is using this API in a [unit-test](https://github.com/strongloop/loopback-boot/blob/61798455f8014d8066783e5329125a2cbdd7a245/test/executor.test.js#L205.

ritch:
@raymondfeng -1 for hierarchical registries. LoopBack Model graphs are complex enough :(

lib/loopback.js
ritch: > However, I still think it is a good idea to let the application decide whether it wants local or global model registry.

I don't think we need to support `app.set()` and `config.localRegsitry`... Thats overkill IMO. Also its impractical. See more on comments below RE: specific implementation for the flag.

server/middleware/token.js
ritch: This worked ...thanks! heh

test/app.test.js
ritch: Removed

test/app.test.js
ritch: I'd like to avoid any `regsitry` "special" properties like this. Model will always be available via `getModel('Model')`.

ritch:
> Also the flag implementation has to be clever, because the app.registry object is initialised before app.set can be called.

I'm leaning toward `loopback({localRegistry: true})`... The problem is once I've implemented this all the tests just pass (since it defaults to false). Of course I can duplicate tests... but I'd like to avoid that if possible (even duplicating at runtime).

ritch:
Added a global `loopback.localRegistry` setting... I don't really like this, but I couldn't come up with a better way to set the flag in certain tests and not in others.

lib/loopback.js
ritch: > Is there any particular reason for writing these methods by hand, instead of performing an automatic mixin?

Mostly for nice documentation support (sdocs). I could be convinced to avoid this (I already hate the idea of having duplicate methods at all) if we had great support for `@mixin` (not just a link to a `Mixin` class, but the actual methods defined inline).

ritch:
> I left a bunch of comments, the most important problem I noticed is the usage of loopback.{modelName} in model lookups.

@bajtos I've removed this in favor of

`registry.getModelByType(registry.getModel('BaseModel'))` // eg. BaseModel => AccessToken

lib/loopback.js
bajtos: > Mostly for nice documentation support (sdocs)

I see. The problem with this approach is that the documentation is in two places and will get out of sync eventually.

Perhaps we can deprecate these methods in 2.x and hide them from API docs in 3.x? That way to duplicated documentation will be only in the 2.x branch which is unlikely to change much going forward.

This is not a big deal, the PR can be landed even without addressing my concerns expressed in this thread.

bajtos:
> @bajtos I've removed this in favor of
> ```js
registry.getModelByType(registry.getModel('BaseModel')) // eg. BaseModel => AccessToken
```

I am proposing to hide this complexity in `getModelByType` and modify the function to support string argument.

```js
getModelByType(model) {
  if (typeof model === 'string') 
    model = this.getModel(model);
  // etc. as before
}
```

Thoughts?

bajtos:
> I'm leaning toward loopback({localRegistry: true})... The problem is once I've implemented this all the tests just pass (since it defaults to false). Of course I can duplicate tests... but I'd like to avoid that if possible (even duplicating at runtime).

How about using an ENV variable in the tests only to control whether the tests are using a local or a global registry? Then we can modify Grungfile to run the tests with both values.

```js
// test/support.js
loopback.localRegistry = !process.env.LB_GLOBAL_REG;
// etc.
```

```
$ LB_GLOBAL_REG=1 mocha
 ... using a global registry ...

$ mocha
 ... using a local registry ...
```

ritch:
> How about using an ENV variable in the tests only to control whether the tests are using a local or a global registry? Then we can modify Grungfile to run the tests with both values.

But some tests will not pass since the use the global API (eg. `loopback.Model.extend()`). I guess we could refactor these tests... but if I'm doing that I might as well just make all the tests pass.

I'd like to move this patch forward in a way where we are not doing more than we really need (do we really need to run all tests under both flags?) and instead put that time into polishing this feature (and others) for 3.0.

ritch:
test please

bajtos:
> I'd like to move this patch forward in a way where we are not doing more than we really need (do we really need to run all tests under both flags?) and instead put that time into polishing this feature (and others) for 3.0.

Sure, I am just trying to help you finding a solution that will be easiest/fastest to implement. Perhaps you can run only a subset of test files when using a local registry?

```
$ mocha
... run all tests using a global registry ...
$ LB_LOCAL_REG=1 mocha test/support.js test/app.test.js (etc.)
... run a subset of tests using local registries ...
```

IMO, this patch should include any changes necessary to test both global and per-app registry as part of `npm test`. I agree that the solution does not have to be perfect, but we should really have at least some test coverage for this new feature/mode.

ritch:
@bajtos and @raymondfeng I'd like to merge this as is and rework the tests in https://github.com/strongloop/loopback/issues/1268

server/middleware/token.js
```diff
@@ -81,10 +77,34 @@ function token(options) {
...
+    var TokenModel;
+
+    switch (typeof options.model) {
+      case 'string':
+        TokenModel = registry.getModel(options.model);
+      break;
+      case 'function':
+        TokenModel = options.model;
+      break;
+    }
```
bajtos: Resolving a "string or constructor" into a model constructor is IMO a common use case that should be implemented in registry, e.g. as `registry.resolveModel(nameOrCtor)`. Would you mind making this improvement as part of this PR?

server/middleware/token.js
```diff
@@ -81,10 +77,34 @@ function token(options) {
...
+
+    TokenModel = TokenModel || registry.getModelByType(BaseAcessToken) || BaseAcessToken;
+
+    assert(typeof TokenModel === 'function',
+      'loopback.token() middleware requires a AccessToken model');
+
+    TokenModel = options.model || loopback.AccessToken;
```
bajtos: This is discarding the lookup result from lines 87-93, is that intentional?

server/middleware/token.js
```diff
@@ -81,10 +77,34 @@ function token(options) {
...
+
+    if (typeof TokenModel === 'string') {
+      // Make it possible to configure the model in middleware.json
+      TokenModel = registry.getModel(TokenModel);
+    }
```
bajtos: This is more or less duplicating lines 87-93, is it intentional?

test/support.js
```diff
@@ -12,6 +12,9 @@ TaskEmitter = require('strong-task-emitter');
 request = require('supertest');
 var RemoteObjects = require('strong-remoting');
 
+// Use local registries
+loopback.localRegistry = true;
```
bajtos: If this PR is not going to change Gruntfile to run the tests twice, then this change must be reverted together with other places where you manipulate `loopback.localRegistry` in before/after hooks.

The idea is to run existing unit-tests with global registry to ensure we are not introducing any regression in the existing applications.

server/middleware/rest.js
```diff
@@ -55,14 +56,8 @@ function rest() {
...
       if (app.isAuthEnabled) {
-        // NOTE(bajtos) It would be better to search app.models for a model
-        // of type AccessToken instead of searching all loopback models.
-        // Unfortunately that's not supported now.
-        // Related discussions:
-        // https://github.com/strongloop/loopback/pull/167
-        // https://github.com/strongloop/loopback/commit/f07446a
-        var AccessToken = loopback.getModelByType(loopback.AccessToken);
-        handlers.push(loopback.token({ model: AccessToken }));
+        var AccessToken = registry.getModelByType(registry.getModel('AccessToken'));
```
bajtos: As I mentioned in past, this is IMO a common use case. Please improve `registry.getModelByType` to accept both model constructor and model name.

bajtos:
> @bajtos and @raymondfeng I'd like to merge this as is and rework the tests in #1268

I am fine with that as long as you remove all `loopback.localRegistry` changes in the tests, see https://github.com/strongloop/loopback/pull/1212#discussion_r27636119. I don't mind adding a new feature that is not yet fully covered by tests, but I do mind removing test coverage for scenarios that every loopback user is using now.

I did a partial review of the changes, see my comments above. I would prefer if you addressed them as part of this PR, but I suppose most of them can be resolved later too.

You are still incorrectly resolving models using the global registry in few places, e.g. `registry.getModelByType(loopback.Application)`. Please consider fixing this problem before landing the PR. (Search for all occurences `getModelByType(loopback` and fix them as appropriate.)

bajtos:
Last but not least, the CI build is failing on both Jenkins and Travis-CI, please fix it before landing the patch.

test/support.js
ritch: OK I've removed this for now.

ritch:
> I am fine with that as long as you remove all loopback.localRegistry 

done

lib/loopback.js
ritch: > Perhaps we can deprecate these methods in 2.x and hide them from API docs in 3.x? That way to duplicated documentation will be only in the 2.x branch which is unlikely to change much going forward.

:+1:

ritch:
> I am proposing to hide this complexity in getModelByType and modify the function to support string argument.

OK

ritch:
test please

common/models/acl.js
```diff
@@ -394,7 +396,7 @@ module.exports = function(ACL) {
     var staticACLs = this.getStaticACLs(model.modelName, property);
 
     var self = this;
-    var roleModel = loopback.getModelByType(Role);
+    var roleModel = registry.getModelByType(Role);
```
bajtos: `Role` is the global `loopback.Role`, this should be IMO changed to `getModelByType('Role')`.

common/models/role-mapping.js
```diff
@@ -42,9 +44,10 @@ module.exports = function(RoleMapping) {
    * @param {User} user
    */
   RoleMapping.prototype.user = function(callback) {
+    var RoleMapping = this.constructor;
     if (this.principalType === RoleMapping.USER) {
-      var userModel = this.constructor.User ||
-        loopback.getModelByType(loopback.User);
+      var userModel = RoleMapping.User ||
+        RoleMapping.registry.getModelByType(loopback.User);
```
bajtos: `loopback.User` is a global model.

bajtos: Seems that the usage of global models was not fixed/removed yet, I am not going to post any more comments about this.

lib/loopback.js
```diff
@@ -76,6 +96,14 @@ function createApplication() {
   app.connector('memory', loopback.Memory);
   app.connector('remote', loopback.Remote);
 
+  if (loopback.localRegistry || options && options.localRegistry === true) {
+    // setup the app registry
+    var registry = app.registry = new Registry();
+    require('./builtin-models')(registry);
```
bajtos: It would be nice if `localRegistry` was initialized only with `Model` and `PersistedModel`. That will make it easier to exclude unused built-in models from the browser bundle. I want to make this change in 3.0 anyways, having it in 2.x under a flag would make migration from 2.x to 3.0 easier.

lib/registry.js
```diff
@@ -424,18 +428,25 @@ registry.autoAttachModel = function(ModelCtor) {
...
+ * @param {String|Function} name
+ * @returns {Function} The `Model` constructor
  */
 
-registry.Model = require('./model');
-registry.PersistedModel = require('./persisted-model');
+Registry.prototype.resolveModel = function(name) {
+  switch (typeof name) {
+    case 'string':
+      return registry.findModel(name);
```
bajtos: I think we should have two methods: `tryResolveModel` that returns null/undefined when the string name is not found, and `resolveModel` that throws when the string name is not found. Thoughts?

bajtos: Alternatively we can fix `getModel` and `findModel` to support a function type argument too, which may be actually the best solution.

lib/registry.js
bajtos: > This code was removed with no replacement. Are you sure it's not a breaking change?

@ritch ^^^

server/middleware/token.js
```diff
@@ -81,10 +77,36 @@ function token(options) {
...
+          break;
+        }
+      }
+    }
+
+    if(options.model) {
+      TokenModel = registry.resolveModel(options.model);
+    } else {
+      TokenModel = BaseAcessToken;
+    }
```
bajtos: I don't understand why there are two sections resolving TokenModel, one on L88-L98, the second on L101-L104. Can you please explain?

Also please cache the result of TokenModel lookup so that you resolve it only once. At the moment, you are resolving the model on every incoming request, which is IMHO premature pessimization.

bajtos:
@ritch I have rebased this patch on top of #1271 to make the changeset smaller and prevent merge conflicts. Please run `git fetch && git reset --hard origin/feature/app-registries` before making any new changes/commits.

common/models/acl.js
ritch: We'll address this in another PR.

lib/registry.js
ritch: Moved that code to the `Registry` constructor

ritch:
test please

PR #1214, opened 2015-03-16
Detect 3rd-party changes made during replication

### LINKED ISSUE #986: Race condition in the replication algorithm
When two clients are preforming replication at the same time and the same object was changed in both clients, some of the changes may get lost. The replication algorithm should check the data before the update to verify it was not changed by another party during the replication process. ~~This will most likely require LoopBack to support transactions for SQL datasources and [Update if Current](http://docs.mongodb.org/manual/tutorial/isolate-sequence-of-operations/#update-if-current] for MongoDB.~~

Another example are external changes. At the moment, there is a timer-based hook to detect any changes made by non-loopback applications and update change tracking records. If the sync algorithm is started before this timed check was run, changes made by the external writer may get discarded.

bajtos:
- Modify `Change.diff()` to include current data revision in each delta reported back. The current data revision is stored in `delta.prev`.

- Modify `PersistedModel.bulkUpdate()` to check that the current data revision matches `delta.prev` and report a conflict if a third party has modified the database under our hands.

- Fix `Change` implementation and tests so that they are no longer attempting to create instances with duplicate ids. (This used to work because the memory connector was silently converting such requests to updateOrCreate/findOrCreate.)

Closes #986
Requires https://github.com/strongloop/loopback-datasource-juggler/pull/502
Requires https://github.com/strongloop/loopback-datasource-juggler/pull/511
Requires strongloop/loopback-datasource-juggler#510

/to @ritch please review
/cc @BerkeleyTrue 

Note: This change may be viewed as a breaking change, because a `bulkUpdate` called with a hand-crafted list of updates will probably start reporting conflict errors. If you think this is a valid concern and we should preserve BC for such users, then I am proposing to add a flag to `bulkUpdate` that will control this behaviour.

Thoughts?

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -952,7 +952,20 @@ function tryReplicate(sourceModel, targetModel, since, options, callback) {
   function bulkUpdate(_updates, cb) {
     debug('\tstarting bulk update');
     updates = _updates;
-    targetModel.bulkUpdate(updates, cb);
+    targetModel.bulkUpdate(updates, function(err) {
+      var conflicts = err && err.details && err.details.conflicts;
+      if (conflicts && err.statusCode == 409) {
+        diff.conflicts = conflicts;
+        // filter out updates that were not applied
+        updates = updates.filter(function(u) {
+          return conflicts
+            .filter(function(d) { return d.modelId === u.change.modelId; })
+            .length === 0;
+        });
+        return cb();
+      }
+      cb(err);
+    });
   }
 
   function checkpoints() {
@@ -974,7 +987,7 @@ function tryReplicate(sourceModel, targetModel, since, options, callback) {
 
     debug('\treplication finished');
     debug('\t\t%s conflict(s) detected', diff.conflicts.length);
-    debug('\t\t%s change(s) applied', updates && updates.length);
+    debug('\t\t%s change(s) applied', updates ? updates.length : 0);
     debug('\t\tnew checkpoints: { source: %j, target: %j }',
       newSourceCp, newTargetCp);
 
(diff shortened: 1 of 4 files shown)


lib/persisted-model.js
```diff
@@ -1060,19 +1073,25 @@ PersistedModel.bulkUpdate = function(updates, callback) {
...
-        // tasks.push(model.save.bind(model));
         tasks.push(function(cb) {
-          var model = new Model(update.data);
-          model.save(cb);
+          applyCreate(Model, id, update.data, update.change, conflicts, cb);
         });
         break;
       case Change.DELETE:
+        // NOTE(bajtos) Delete does not care whether the model has
+        // the same revision as expected, we are going to delete it anyway
```
ritch: > Delete does not care whether the model has the same revision as expected

The delete should be a conflict if the current target revision doesn't match the previous revision of the source. If someone has changed the target, we shouldn't blindly delete it (although I'm not certain that is what you are doing).

lib/persisted-model.js
```diff
@@ -1081,9 +1100,88 @@ PersistedModel.bulkUpdate = function(updates, callback) {
...
+        debug('UpdateAll detected non-rectified change of %s %j',
+          Model.modelName, id);
+        conflicts.push(change);
+        // NOTE(bajtos) updateAll triggers change rectification
+        // for all model instances, even when no records were updated,
+        // thus we don't need to rectify explicitly ourselves
+      } else if (count !== 1) {
+        debug('%s.updateAll modified unexpected number of instances: %j',
+          Model.modelName, count);
+        return cb(new Error(
```
ritch: `count !== 1` is really a good enough check for this? If `count` is `2` we should report an error? What about `typeof count` or `arguments.length`...

lib/persisted-model.js
```diff
@@ -1081,9 +1100,88 @@ PersistedModel.bulkUpdate = function(updates, callback) {
...
+}
+
+function applyCreate(Model, id, data, change, conflicts, cb) {
+  Model.create(data, function(createErr) {
+    if (!createErr) return cb();
+
+    // We don't have a reliable way how to detect the situation
+    // where he model was not create because of a duplicate id
+    // The workaround is to query the DB to check if the model already exists
+    Model.findById(id, function(findErr, inst) {
```
ritch: This isn't so great. But I actually like it better than trying to parse the error for some meaningful info. The juggler should probably take this on though (an optmization or reporting a consistent error).

test/replication.test.js
```diff
@@ -291,6 +289,96 @@ describe('Replication / Change APIs', function() {
...
+              conflicts[0].resolve(next);
+            });
+        },
+
+        replicateExpectingSuccess(),
+        verifyInstanceWasReplicated(SourceModel, TargetModel, '1')
+      ], done);
+    });
+
+    it('detects 3rd-party CREATE made during replication', function(done) {
```
ritch: Why aren't we testing for `DELETE`?

lib/persisted-model.js
bajtos: OK, I'll add a check for that too.

lib/persisted-model.js
bajtos: If `count` is `2` then the connector and/or the database does not correctly implement unique constraint on model keys.

I'll rework the code to distinguish between the case "connector did not provide the count value" and "the count values is unexpected".

lib/persisted-model.js
bajtos: > The juggler should probably take this on though (an optmization or reporting a consistent error).

I have already looked into this, databases error messages/code don't distinguish between a duplicate primary key (e.g. id) and a duplicate value in a unique column, at least not in MySQL and MongoDB.

See e.g. https://github.com/felixge/node-mysql/issues/372:

> Detecting which fields are duplicated is not part of the mysql protocol (afaik), and we're not going to start human readable error strings from the server.

test/replication.test.js
bajtos: I'll add the test as part of modifying the DELETE implementation as mentioned above.

lib/persisted-model.js
bajtos: The fix for this requires https://github.com/strongloop/loopback-datasource-juggler/pull/511 and possibly implementation in all connectors.

bajtos:
@ritch I have addressed all your comments, PTAL again.

ritch:
LGTM

bajtos:
@slnode test please

bajtos:
The Jenkins build is passing on v0.10 and v0.12 and failing on io.js 1.6.0 due to https://github.com/iojs/io.js/issues/1215 that has been fixed in 1.6.1. @rmg could you please upgrade our build slaves?

rmg:
Yep, I'll upgrade in the morning.

bajtos:
> Note: This change may be viewed as a breaking change, because a `bulkUpdate` called with a hand-crafted list of updates will probably start reporting conflict errors. If you think this is a valid concern and we should preserve BC for such users, then I am proposing to add a flag to `bulkUpdate` that will control this behaviour.

@chandadharap @dashby3000 @sumitha Can you check with our customers if they are using `bulkUdate` on it's own? We don't have exact day when this will be released, I'd say it will be a week or two at most.

PR #1221, opened 2015-03-18
Set up the correct remoting type for id/fk

raymondfeng:
/to @ritch 

Connect to #1221

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -97,6 +97,34 @@ var extend = require('util')._extend;
 
 var Model = module.exports = registry.modelBuilder.define('Model');
 
+function getRemotingType(prop) {
+  if (prop.type === Number) {
+    return 'number';
+  }
+  if (prop.type === Boolean) {
+    return 'number';
+  }
+  return 'string';
+}
+
+Model._getRemotingIdType = function getIdType(ModelCtor) {
+  var idName = ModelCtor.definition.idName();
+  var idProp = ModelCtor.definition.properties[idName];
+  return getRemotingType(idProp);
+};
+
+Model._getRemotingFkType = function getForeignKeyType(ModelCtor, relationName) {
+  var relation = ModelCtor.relations[relationName];
+  var prop;
+  if (relation.type === 'belongsTo') {
+    prop = relation.modelFrom.definition.properties[relation.keyFrom];
+  } else if (relation.modelThrough) {
+    prop = relation.modelThrough.definition.properties[relation.keyTo];
+  } else {
+    prop = relation.modelTo.definition.properties[relation.keyTo];
+  }
+  return getRemotingType(prop);
+};
 /*!
  * Called when a model is extended.
  */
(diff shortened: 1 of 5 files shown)


test/util/model-tests.js
```diff
@@ -169,11 +170,11 @@ module.exports = function defineModelTestsWithDataSource(options) {
 
     describe('Model.upsert(data, callback)', function() {
       it('Update when record with id=data.id found, insert otherwise', function(done) {
-        User.upsert({first: 'joe', id: 7}, function(err, user) {
+        User.upsert({first: 'joe', id: '7'}, function(err, user) {
```
ritch: ...? why does this have to change

ritch:
@raymondfeng 1 minor change... not comfortable merging this when you are changing the tests.

raymondfeng:
I found some issues related to how loopback-connector-remote sets up the models. As the PR defers the remoting setup after a model is attached to the data source, the loopback-connector-remote fails as it has some assumptions about the timing. Let's chat more.

ritch:
@raymondfeng we can chat offline if you want. Just ping me today.

lib/model.js
```diff
@@ -98,6 +98,34 @@ var stringUtils = require('underscore.string');
 
 var Model = module.exports = registry.modelBuilder.define('Model');
 
+function getRemotingType(prop) {
+  if (prop.type === Number) {
+    return 'number';
+  }
+  if (prop.type === Boolean) {
+    return 'number';
```
ritch: This should be boolean @raymondfeng

bajtos:
@raymondfeng @ritch what's the status of this patch?

ritch:
Not sure... @raymondfeng ?

fabien:
This is a critical issue for many users - why hasn't this been merged yet? I read about failing tests in `loopback-connector-remote`, but why has this been preventing anyone to prioritize and fix this?

/cc @raymondfeng

raymondfeng:
The patch fixes the issue but it breaks the assumption of loopback-connector-remote which expects the remoting metadata to be available earlier. We need to rebase the fix and have @ritch look into the loopback-connector-remote issue.

fabien:
@raymondfeng @ritch @bajtos - any updates?

lib/model.js
bajtos: @raymondfeng please add a unit-test for this case, to make sure we don't introduce a similar bug in the future.

lib/model.js
```diff
@@ -164,10 +190,41 @@ Model.setup = function() {
     }
   };
 
+  // resolve relation functions
+  sharedClass.resolve(function resolver(define) {
+
```
superkhau: extra space here

test/util/model-tests.js
```diff
@@ -169,11 +170,11 @@ module.exports = function defineModelTestsWithDataSource(options) {
...
           assert(!err);
           assert.equal(user.first, 'joe');
 
-          User.upsert({first: 'bob', id: 7}, function(err, updatedUser) {
+          User.upsert({first: 'bob', id: '7'}, function(err, updatedUser) {
```
superkhau: same here why string now

bajtos:
I am closing this pull request as abandoned. If there is anybody willing to take over this work in progress and make the changes requested in the comments above, then please open a new pull request.

PR #1225, opened 2015-03-18
Typecast debug info to number in lib/persisted-model

BerkeleyTrue:


bajtos:
This will be fixed by #1214, see https://github.com/strongloop/loopback/pull/1214/files#diff-3bf07e630a94e80eef9557e874f96f56L977

PR #1227, opened 2015-03-19
Update user.js

qazs:
Check for empty email string.


slnode:

PR #1228, opened 2015-03-19
Password reset fixes

qazs:
Check for empty email string.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -456,7 +456,7 @@ module.exports = function(User) {
     var ttl = UserModel.settings.resetPasswordTokenTTL || DEFAULT_RESET_PW_TTL;
 
     options = options || {};
-    if (typeof options.email === 'string') {
+    if (typeof options.email === 'string' && options.email.trim() != '') {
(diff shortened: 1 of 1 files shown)


slnode:

common/models/user.js
```diff
@@ -476,7 +476,10 @@ module.exports = function(User) {
             }
           });
         } else {
-          cb();
+          err = new Error("User '" + options.email + "' does not exist");
```
ritch: The caller of this function should already have the email... Why put it in the error message?

ritch:
I'd be for merging this if we removed the `email` from the error message. Otherwise looks good.

ritch:
Actually on second thought I don't think we should make any of these changes. Its adding behavior that I'm not comfortable with being the default.

Eg. you could use the error message to determine whether or not an account exists. If you'd really like this behavior just implement your own remote method. You could create your own version of the reset password method. You can also emit your own reset password event.

Closing for now (if you want it re-opened just chime in here and we can discuss).

qazs:
The problem is there's no error message when an account doesn't exist, the response is the same as when a password is successfully being reset.

PR #1233, opened 2015-03-20
Add conflict resolution API

### LINKED ISSUE #1188: Conflict API for common resolution actions
The `Conflict` model should provided API for the three usual ways how conflicts are usually resolved: "use mine", "use theirs", "edit & resolve manually".

See https://github.com/strongloop/loopback/pull/1187#issuecomment-77584907 for background info.

bajtos:
New methods:
```
  conflict.resolveUsingSource(cb)
  conflict.resolveUsingTarget(cb)
  conflict.resolveManually(data, cb)
```

Close #1188

/to @ritch please review
/cc @BerkeleyTrue

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -616,6 +616,13 @@ module.exports = function(Change) {
   /**
    * Resolve the conflict.
    *
+   * Set the source change's previous revision to the current revision of the
+   * (conflicting) target change. Since the changes are no longer conflicting
+   * and appear as if the source change was based on the target, they will be
+   * replicated normally as part of the next replicate() call.
+   *
+   * This is effectively resolving the conflict using the source version.
+   *
    * @callback {Function} callback
    * @param {Error} err
    */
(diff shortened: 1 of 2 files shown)


BerkeleyTrue:
:+1:

bajtos:
> I understand the use of target/source but I wish there was a more descriptive way to define these, like theirs and mine, or local/remote. Debugging is always an issue as I have to check model name to determine which model is that target and which is the source.

Yeah, there is definitely space for improvement. Especially with two-way replication.

On the other hand, you must be careful to resolve the conflicts always on the source side, otherwise subsequent replication may not pick up the resolution correctly.

ritch:
Target and source is confusing... I wish I used something more concrete but I could not think of anything honestly. As a user of the API you do need a distinction. Keep in mind that you could be resolving conflicts where the "mine" describes the author of both the target and source (eg. replicating from one local model to another local model).

ritch:
Big +1 for these APIs... this might be what really makes using replication practical now! I'd love to see used in the full stack example.

common/models/change.js
```diff
@@ -630,6 +637,74 @@ module.exports = function(Change) {
...
+   */
+  Conflict.prototype.resolveUsingTarget = function(cb) {
+    var conflict = this;
+
+    conflict.models(function(err, source, target) {
+      if (err) return done(err);
+      if (target === null) {
+        return conflict.SourceModel.deleteById(conflict.modelId, done);
+      }
+      var inst = new conflict.SourceModel(target);
```
ritch: I cringe when I see this (creating an instance just to call `save()`). I'm guessing you are depending on something like auto-setting the ID properly. Otherwise cant you use `SourceModel.update()`?

ritch: I'm OK with leaving this... but `new Foo().save()` is starting to bother me.

ritch:
Code LGTM.... Merge it!

common/models/change.js
bajtos: I just copied your code from the full-stack example, I am ok to use something else.

AFAICT, the difference between `save` and `updateAttributes` is in the way how undefined properties are handled. `updateAttributes` is ignoring undefined properties and preserving data in the database, while `save` will apply undefined properties in the database record too.

PR #1238, opened 2015-03-23
Use `ctx.instance` provided by "after delete" hook

bajtos:
Use the recently added context property `ctx.instance` (see #488) to improve the accuracy of the algorithm detecting whether a single or multiple models were deleted.

/to @ritch


ritch:
LGTM

Issue #989, opened 2015-01-09
Reduce the size of loopback client bundle


bajtos:
At the moment, a browserified loopback module has about 1.4MB uncompressed, 650KB compressed. This size is unacceptable, especially for mobile browsers on a slow internet connection. We need to get at least below 100kb.

This GH issue is a time-scoped story. The goal is to analyse the dependencies, fix as many low-hanging fruits as possible and then create follow-up stories for the unsolved items.

Subtasks:
- [x] strongloop/loopback-datasource-juggler#538 - remove lodash
- [x] strongloop/strong-remoting#187 - remove js2xmlparser and cors
- [x] #1246 - optimise underscode.string usage
- [ ] #1247 - rework places using crypto module
- [ ] #1249 - don't use bcryptjs in browser
- [x] strongloop/strong-remoting#177 - use a smaller HTTP client in the browser
- [ ] https://github.com/strongloop/loopback/issues/1211 - pluggable HTTP client
- [ ] Investigate if we can get rid of "buffer" usage in the browser

BerkeleyTrue:
@bajtos any advice or strategies on this?

bajtos:
My plan is to produce profiling data using file (bundle) size instead of execution time, so that we can identify the files and modules that contribute most to the overall size. There may be an existing browserify module for that, I did not check yet. (Please let me know if you figure out how to do this.)

Once we have that information, it should be easy to pick the biggest dependencies, check if they are used by the browser client, and exclude them using `browser` field in the relevant package.json field if they are not needed.

BerkeleyTrue:
So the first step is to get the size of each file/script in the bunde?

bajtos:
For each file/script, you need two values: the size of the file itself, and the cumulated size of all dependencies including the file itself.

For example, a single-line file `var loopback = require('loopback')` is small itself (about 20 bytes), but huge when dependencies are included (1.4MB).

bajtos:
I did some investigation in the past two days while I was travelling.

The biggest contributor to the bundle size is the "crypto" module, it adds ~650kb unminified. See https://github.com/strongloop/loopback/issues/1247 for a follow-up discussion.

Another big dependency is lodash with 367kb unminified. https://github.com/strongloop/loopback-datasource-juggler/pull/538 removes this dependency.

Smaller modules that were easy to remove:

 - Use only the relevant subset of "underscore.string" - https://github.com/strongloop/loopback/pull/1246 saves about 25kb.
[...]

PR #1246, opened 2015-03-26
Import subset of underscore.string scripts only

bajtos:
Require individual methods like `classify` instead of the whole module. This reduces the size of underscore.string browser bundle from ~27kb down to ~2kb.

See https://github.com/strongloop/loopback/issues/989

/to @ritch please review


ritch:
Cool! 

LGTM

Issue #397, opened 2014-07-24
A better way for customizing built-in models and models from components


bajtos:
Consider an application with a "User" model that has few extra properties or methods compared to the built-in User model.

At the moment, the recommended solution is to create a `user` model extending the built-in `User` model, which is rather silly.

**What we would like to have instead**

Support `redefine:true` as an alternative to `base: "User"`. A redefined "User" model is defined in `common/models/user.json` and scaffolded by `yo loopback`. This will allow redefinition of common models in the server facet.

In the long term (loopback 3.x), we were discussing a mixin-based approach, where User would be just a thin wrapper mixing in login/logout and other features provided by the current User implementation.

---

**Original issue description**

Possible solutions:
1. Implement namespaces, move all built-in models to `loopback` namespace. Application models can live in the global namespace, i.e. `User` extends `loopback.User`.
2. Implement means allowing applications to modify the built-in models, so that they can use them directly.

See also https://github.com/strongloop/loopback-datasource-juggler/issues/119

A probably related problem: the canonical project layout has three facets - "common", "server", "client". In many cases it is needed to provide different implementation of model methods on the server. The current approach is to create a new "server" model extending the shared "common" model and make any changes in this subclass. While this keeps the design clean, it's also rather cumbersome.

bajtos:
Related: how to customize the `Change` model used for change-tracking. https://github.com/strongloop/loopback/issues/926

uNki23:
hi there! 

i'm facing the same problems: pain to customize the built in User model. Extending it by using my own 'user' model isn't working good with angular (injecting 'user' into the function doesn't work, 'User' does, but refers to the base model - see: https://stackoverflow.com/questions/29458355/loopback-angularjs-extending-user-model?answertab=active#tab-top) is there any best practice solution to customize the base User model? i'm doing in the boot-script way with code.

pulkitsinghal:
@uNki23 - I switched to the nomenclature where I create a file named `user-model.json` which extends the base `User` and named it `UserModel` ... it prettifies everything for me in core and angular code ... a bit wordy though.

alFReD-NSH:
I want to add a property to the built-in Role model. Any idea how to do this currently?

uNki23:
correct me if i'm wrong, but it seems to work fine if you just "use" the property.

here is a sample function within an AngularJS UserController:

vm.addUser = function () {
            User.create({
                firstName: vm.firstName,
                lastName: vm.lastName,
                division: vm.division,
[...]

PR #1251, opened 2015-03-26
Improve error handling in replication

### LINKED ISSUE #1126: Better error handling in the replication algorithm
At the moment, the change tracking & replication algorithm reports errors via `console.error`. We should  allow the user to provide a function to call when an error occurs.

bajtos:
Deprecate `Change.handleError`, it was used inconsistenly for a subset of possible errors only. Rework all `Change` methods to always report all errors to the caller via the callback.

Rework `PersistedModel` to report change-tracking errors via the existing method `PersistedModel.handleChangeError`. This method can be customized on a per-model basis to provide different error handling.

Close #1126

/to @ritch please review
/cc @BerkeleyTrue

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -9,6 +9,7 @@ var CJSON = {stringify: require('canonical-json')};
 var async = require('async');
 var assert = require('assert');
 var debug = require('debug')('loopback:change');
+var deprecate = require('depd')('loopback');
 
 /**
  * Change list entry.
@@ -73,18 +74,43 @@ module.exports = function(Change) {
    */
 
   Change.rectifyModelChanges = function(modelName, modelIds, callback) {
-    var tasks = [];
     var Change = this;
+    var errors = [];
 
-    modelIds.forEach(function(id) {
-      tasks.push(function(cb) {
+    var tasks = modelIds.map(function(id) {
+      return function(cb) {
         Change.findOrCreateChange(modelName, id, function(err, change) {
-          if (err) return Change.handleError(err, cb);
-          change.rectify(cb);
+          if (err) return next(err);
+          change.rectify(next);
         });
-      });
+
+        function next(err) {
+          if (err) {
+            err.modelName = modelName;
+            err.modelId = id;
+            errors.push(err);
+          }
+          cb();
+        }
+      };
     });
-    async.parallel(tasks, callback);
+
+    async.parallel(tasks, function(err) {
+      if (err) return callback(err);
+      if (errors.length) {
+        var desc = errors
+          .map(function(e) {
+            return '#' + e.modelId + ' - ' + e.toString();
+          })
+          .join('\n');
+
+        var msg = 'Cannot rectify ' + modelName + ' changes:\n' + desc;
+        err = new Error(msg);
+        err.details = { errors: errors };
+        return callback(err);
+      }
+      callback();
+    });
   };
 
   /**
@@ -217,7 +243,7 @@ module.exports = function(Change) {
     var model = this.getModelCtor();
     var id = this.getModelId();
     model.findById(id, function(err, inst) {
-      if (err) return Change.handleError(err, cb);
+      if (err) return cb(err);
       if (inst) {
         cb(null, Change.revisionForInst(inst));
       } else {
@@ -459,6 +485,9 @@ module.exports = function(Change) {
   };
 
   Change.handleError = function(err) {
+    deprecate('Change.handleError is deprecated, ' +
+      'you should pass errors to your callback instead.');
+
     if (!this.settings.ignoreErrors) {
       throw err;
     }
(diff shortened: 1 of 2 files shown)


common/models/change.js
```diff
@@ -464,6 +490,9 @@ module.exports = function(Change) {
     }
   };
 
+  Change.handleError = deprecate(Change.handleError,
+    'Change.handleError is deprecated, report your errors via the callback.');
```
ritch: Do you mean to say "via the `replicate()` callback?"... This is vague.

lib/persisted-model.js
```diff
@@ -1421,11 +1417,13 @@ PersistedModel.rectifyAllChanges = function(callback) {
...
 
-PersistedModel.handleChangeError = function(err) {
-  if (err) {
-    console.error(Model.modelName + ' Change Tracking Error:');
-    console.error(err);
-  }
+PersistedModel.handleChangeError = function(err, operation) {
+  if (!err) return;
+
+  console.error('%s Change Tracking Error%s:',
```
ritch: -1 for `console` usage... 

Can't we do something else? Like an `error` event?

ritch:
I have always regretted using `console.error()` in a library. We should provide some sort of hook and the user should chose to `console.log` or `debug()` etc.

bajtos:
> I have always regretted using console.error() in a library. We should provide some sort of hook and the user should chose to console.log or debug() etc. I think emit('error', err, ...) is probably the idiomatic node way.

I see. So there are two questions to answer:

1) What should be the default behaviour? In this patch, we are using `console.error`. Should I `debug` instead? My concern is that if we don't print the errors by default, users won't notice the problem and won't know that he should install his own error handler.

2) How to change/customize the default behaviour? In this patch, the user should overwrite `PersistedModel.handleChangeError` or `MyModel.handleChangeError` with his own error logging implementation. Optionally, they can call the original handler if they want to preserve the default output too.

Users can also create a base class for all replicated models and override the method there, because `PersistedModel` cannot be modified via loopback-boot until we implement `redefine:true` in #397.

  ```js
  // replicated-model.json
  {
    "name": "ReplicatedModel",
    "base": "PersistedModel",
    "strict": true
  }
  ```

  ```js
  // replicated-model.js
  module.exports = function(ReplicatedModel) {
    ReplicatedModel.handleChangeError = function(err) {
      // optionally:
      ReplicatedModel.base.handleChangeError(err);
      // user-provided handler
      bugsnag.notify(err);
    };
  };
  ```

> I think `emit('error', err, ...)` is probably the idiomatic node way.

I am ok to consider this option too, though I see few problems:

 1. When you don't install any listener, Node will crash your process due to an unhandled error. That seems pretty harsh to me.

 2. If LoopBack adds any default handler (event listener), users cannot easily remove it (disable the default logger).

 3. Event listeners are not inherited. You have to install the error handler/listener in each model.

1 & 3 can be solved by using hooks (`observe` and `notifyObserversOf`) instead of events (`on` and `emit`). 

2 can be partially addressed via `removeAllListeners` or https://github.com/strongloop/loopback-datasource-juggler/pull/500 for operation hooks, but this solution feels little bit hacky to me. What if there are more listeners besides the default one? The first custom listener must call `removeAllListeners`, the second listener must not call it.

@ritch Thoughts?

common/models/change.js
bajtos: Well, this message is printed if your code calls `Change.handleError`. I want to tell the user to send the error to the callback of the method which called `handleError`.

```js
Change.custom = function(cb) {
  var self = this;
  someAsyncWork(function(err) {
    // deprecated
    if (err) return handleError(err);
    // the correct way
    if (err) return cb(err);
   
    // do more stuff on success
    cb();
  });
};
```

I don't think anybody is calling `handleError` directly. Of course, I cannot be sure, thus I added this deprecation warning instead of removing the method completely.

How about this message?

```
Change.handleError is deprecated, you should pass errors to your callback instead.
```

ritch:
> 1) What should be the default behaviour? In this patch, we are using console.error. Should I debug instead? My concern is that if we don't print the errors by default, users won't notice the problem and won't know that he should install his own error handler.

I think the node idiom here is to `throw`... Not so great for production... but pretty useful for dev. If you `emit('error')` and it isn't handled it will throw so there's that.

ritch:
> When you don't install any listener, Node will crash your process due to an unhandled error. That seems pretty harsh to me.

This is the case for other things... why is our case special?

ritch:
> Event listeners are not inherited. You have to install the error handler/listener in each model.

This is on of the reasons I added `Model.setup()`. Just put it there.

lib/persisted-model.js
bajtos: As we discussed elsewhere, I'll replace `console.error` with `this.getChangeModel().emit('error', err)`.

bajtos: Or with `this.emit('error', err).` We should pass the "operation" value to the event too.

bajtos:
@ritch reworked, LGTY now?

ritch:
LGTM

PR #1270, opened 2015-04-02
Basic access control for change replication

bajtos:
1) Add integration tests running change replication over REST to verify
that access control at model level is correctly enforced.

2) Implement a new access type "REPLICATE" that allows principals
to create new checkpoints, even though they don't have full WRITE
access to the model. Together with the "READ" permission, these
two types allow principals to replicate (pull) changes from the server.

Note that anybody having "WRITE" access type is automatically
granted "REPLICATE" type too.

3) Add a new model option "enableRemoteReplication" that exposes
replication methods via strong remoting, but does not configure
change rectification. This option should be used the clients
when setting up Remote models attached to the server via the remoting
connector.

---

The scope of this pull request is to implement access control checks for replication at model-level, i.e. when a user can either access (read, write) all mode instances or cannot access any model instance. Instance-level access control checks (where e.g. a user can read & write only his own user profile) are out of scope of this pull request.

Connect to #1166

### DIFF
--- a/test/replication.rest.test.js
+++ b/test/replication.rest.test.js
@@ -1,0 +1,484 @@
+var async = require('async');
+var debug = require('debug')('test');
+var extend = require('util')._extend;
+var loopback = require('../');
+var expect = require('chai').expect;
+var supertest = require('supertest');
+
+describe('Replication over REST', function() {
+  var ALICE = { id: 'a', username: 'alice', email: 'a@t.io', password: 'p' };
+  var PETER = { id: 'p', username: 'peter', email: 'p@t.io', password: 'p' };
+  var EMERY = { id: 'e', username: 'emery', email: 'e@t.io', password: 'p' };
+
+  var serverApp, serverUrl, ServerUser, ServerCar, serverCars;
+  var aliceId, peterId, aliceToken, peterToken, emeryToken, request;
+  var clientApp, LocalUser, LocalCar, RemoteUser, RemoteCar, clientCars;
+
+  before(setupServer);
+  before(setupClient);
+  beforeEach(seedServerData);
+  beforeEach(seedClientData);
+
+  describe('the replication scenario scaffolded for the tests', function() {
+    describe('Car model', function() {
+      it('rejects anonymous READ', function(done) {
+        listCars().expect(401, done);
+      });
+
+      it('rejects anonymous WRITE', function(done) {
+        createCar().expect(401, done);
+      });
+
+      it('allows EMERY to READ', function(done) {
+        listCars()
+          .set('Authorization', emeryToken)
+          .expect(200, done);
+      });
+
+      it('denies EMERY to WRITE', function(done) {
+        createCar()
+          .set('Authorization', emeryToken)
+          .expect(401, done);
+      });
+
+      it('allows ALICE to READ', function(done) {
+        listCars()
+          .set('Authorization', aliceToken)
+          .expect(200, done);
+      });
+
+      it('denies ALICE to WRITE', function(done) {
+        createCar()
+          .set('Authorization', aliceToken)
+          .expect(401, done);
+      });
+
+      it('allows PETER to READ', function(done) {
+        listCars()
+          .set('Authorization', peterToken)
+          .expect(200, done);
+      });
+
+      it('allows PETER to WRITE', function(done) {
+        createCar()
+        .set('Authorization', peterToken)
+        .expect(200, done);
+      });
+
+      function listCars() {
+        return request.get('/Cars');
+      }
+
+      function createCar() {
+        return request.post('/Cars').send({ model: 'a-model' });
+      }
+    });
+  });
+
+  describe('sync with model-level permissions', function() {
+    describe('as anonymous user', function() {
+      it('rejects pull from server', function(done) {
+        RemoteCar.replicate(LocalCar, expectHttpError(401, done));
+      });
+
+      it('rejects push to the server', function(done) {
+        LocalCar.replicate(RemoteCar, expectHttpError(401, done));
+      });
+    });
+
+    describe('as user with READ-only permissions', function() {
+      beforeEach(function() {
+        setAccessToken(emeryToken);
+      });
+
+      it('rejects pull from server', function(done) {
+        RemoteCar.replicate(LocalCar, expectHttpError(401, done));
+      });
+
+      it('rejects push to the server', function(done) {
+        LocalCar.replicate(RemoteCar, expectHttpError(401, done));
+      });
+    });
+
+    describe('as user with REPLICATE-only permissions', function() {
+      beforeEach(function() {
+        setAccessToken(aliceToken);
+      });
+
+      it('allows pull from server', function(done) {
+        RemoteCar.replicate(LocalCar, function(err, conflicts, cps) {
+          if (err) return done(err);
+          if (conflicts.length) return done(conflictError(conflicts));
+
+          LocalCar.find(function(err, list) {
+            if (err) return done(err);
+            expect(list.map(carToString)).to.include.members(serverCars);
+            done();
+          });
+        });
+      });
+
+      it('rejects push to the server', function(done) {
+        LocalCar.replicate(RemoteCar, expectHttpError(401, done));
+      });
+    });
+
+    describe('as user with READ and WRITE permissions', function() {
+      beforeEach(function() {
+        setAccessToken(peterToken);
+      });
+
+      it('allows pull from server', function(done) {
+        RemoteCar.replicate(LocalCar, function(err, conflicts, cps) {
+          if (err) return done(err);
+          if (conflicts.length) return done(conflictError(conflicts));
+
+          LocalCar.find(function(err, list) {
+            if (err) return done(err);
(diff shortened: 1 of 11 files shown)


bajtos:
As I am thinking about this issue more, I see two "feature" levels we can implement:

Level1 - model-level access control. A user can either read/write all model instances, or no model instance at all. This should be rather easy to implement, as the ACL check can stay at method-level.

Level2 - instance-level access control. A user can read/write only a subset of model instances. This is more difficult to implement, partially because we don't support this fine-grained access control in existing code either (e.g. `Model.find` does not filter instances to those that the current user can access). I want to leave this out of Sync v1.0 release.

bajtos:
@ritch could you please comment with a code snippet showing how to configure ACLs so that user Alice can read & write all Car instances, user Peter can only read Car instances, and user Jane cannot read nor write any Car instances.

bajtos:
@ritch please review. I have implemented the first part of what we have discussed on Friday: the new setting "enableRemoteReplication" and a new access type "REPLICATE". 

If we manage to land this PR soon, then I'll send the second part (conflict resolution fixes) in a new pull request. Otherwise I'll add more commits to this one.

Let me know what works best for you.

Two sets of tests that I left out, let me know if you think they are important and I should add them:
 - Tests covering ACLs per each individual PersistedModel method. 
 - Tests checking the new "REPLICATE" access type on a lower level, i.e. directly via AccessContext and/or ACL methods.

/cc @raymondfeng you may want to review this PR too, as it is adding a new feature to access control.

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+  var serverApp, serverUrl, ServerUser, ServerCar, serverCars;
+  var aliceId, peterId, aliceToken, peterToken, emeryToken, request;
+  var clientApp, LocalUser, LocalCar, RemoteUser, RemoteCar, clientCars;
+
+  before(setupServer);
+  before(setupClient);
+  beforeEach(seedServerData);
+  beforeEach(seedClientData);
+
+  describe('scenario under test', function() {
```
ritch: Missing describe?

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+          LocalCar.find(function(err, list) {
+            if (err) return done(err);
+            expect(list.map(carToString)).to.include.members(serverCars);
+            done();
+          });
+        });
+      });
+
+      it('rejects push to the server', function(done) {
+        LocalCar.replicate(RemoteCar, expectHttpError(401, done));
```
ritch: I assumed this would be possible with `REPLICATE` permission. We should make this very clear in the docs.

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+          ServerCar.find(function(err, list) {
+            if (err) return done(err);
+            expect(list.map(carToString)).to.include.members(clientCars);
+            done();
+          });
+        });
+      });
+    });
+
+    // TODO conflict resolution
```
ritch: Are these big enough that we want to create an issue to track?

Also if this stays in the code... change to `TODO(bajtos)`

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+            if (err) return next(err);
+            expect(found.toObject())
+              .to.not.have.property('fullname');
+            next();
+          });
+        }
+      ], done);
+    });
+
+    // TODO verify conflict resolution
```
ritch: Same comment as above

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+    serverApp.model(ServerCar, {  dataSource: 'db', public: true });
+
+    serverApp.use(function(req, res, next) {
+      debug(req.method + ' ' + req.path);
+      next();
+    });
+    serverApp.use(loopback.token({ model: ServerToken }));
+    serverApp.use(loopback.rest());
+
+    serverApp.set('legacyExplorer', false);
```
ritch: Outside the scope of this PR... but global defaults might be a nice feature to add. Something like `loopback.defaults.legacyExplorer = false`.

test/replication.rest.test.js
```diff
@@ -0,0 +1,484 @@
...
+  function conflictError(conflicts) {
+    var err = new Error('Unexpected conflicts\n' +
+      conflicts.map(JSON.stringify).join('\n'));
+    err.name = 'ConflictError';
+  }
+
+  function carToString(c) {
+    return c.maker ? c.maker + ' ' + c.model : c.model;
+  }
+});
```
ritch: Wow this is an impressive amount of testing. :+1: 

This is making it blatantly obvious that we need to move all of this into another module.

test/rest.middleware.test.js
```diff
@@ -3,6 +3,7 @@ describe('loopback.rest', function() {
   beforeEach(function() {
     var ds = app.dataSource('db', { connector: loopback.Memory });
     MyModel = ds.createModel('MyModel', {name: String});
+    loopback.autoAttach();
```
ritch: ???

This is concerning... why do we need this?

test/user.test.js
```diff
@@ -23,6 +23,7 @@ describe('User', function() {
 
     // Update the AccessToken relation to use the subclass of User
     AccessToken.belongsTo(User);
+    User.hasMany(AccessToken);
```
ritch: This is also concerning...? What prompted this change?

test/support.js
```diff
@@ -30,9 +30,6 @@ beforeEach(function() {
       {type: 'STUB'}
     ]
   });
-
-  // auto attach data sources to models
-  loopback.autoAttach();
```
ritch: ....? why?

test/replication.rest.test.js
bajtos: This is intentional. The test suite is checking that I the before/beforeEach hooks have correctly set up all models and permissions. You would not believe how much are the tests intertwined through the global model registry and how easy it is to mess up the setup.

Can you come up with a better name that will make my intent more clear?

test/replication.rest.test.js
bajtos: At the moment, replicate = pull changes. I am fine with renaming it if it makes the intent more clear. It's much better than relying on docs, as most people don't read them. Should I use `PULLCHANGES` instead of `REPLICATE`?

test/replication.rest.test.js
bajtos: I'll start to work on this tomorrow, as part of the sprint story.

test/replication.rest.test.js
bajtos: Sounds good :+1:

test/support.js
bajtos: All these three comments are related to the first commit https://github.com/strongloop/loopback/commit/699bc7aa974c4a46cd8b6d2da71d5088ad532fb6 which is changing the test setup, removing `autoAttach` from a global beforeEach hook and enabling it only in the test suites that depend on that.

The reason why I need the change is that it's the fastest way how to reduce the amount of crazy cross-test side effect caused the global shared registry. I spend about one or two hours investigating the problem described below and I didn't want to waste more time on that.

Here is the quiz for you. Look at the following code and tell me, why the type of the property `ServerToken.userId` was changed from `String` to `Number` between the line 2 and line 3?

```js
before(setupServer);
before(setupClient);
beforeEach(seedServerData);
beforeEach(seedClientData);
```

.
.
.
.
.
.
.
.
.

Hint: the global beforeEach hook calling `autoAttach` is executed between the line 2 and 3. 

.
.
.
.

Do you know the answer yet? ;-)

.
.
.
.
.
.


`autoAttach` attached `User` and `AccessToken` models to a in-memory datasource. When these models are attached, model relations are resolved and `AccessToken.userId` is defined as a `Number` property, because `User.id` is a Number property injected by the in-memory datasource.

Unfortunately, it seem like the action has also a side effect of modifying the type of `userId` property in the descendant model `ServerToken`, which is related to `ServerUser` (and `ServerUser.id` has the id as a `String`).

Again, I really didn't want to spend any more time investigating it. Considering that we want to remove both autoAttach in the future, this small change is IMO going in the right direction, as it's reducing the number of tests depending on autoAttach.

However, if you feel I should debug the underlying issue and fix the probable bug in the juggler instead, then I can do that too.

/cc @raymondfeng

test/replication.rest.test.js
bajtos: perhaps `describe('the replication scenario scaffolded for the tests")`?

test/support.js
ritch: > However, if you feel I should debug the underlying issue and fix the probable bug in the juggler instead, then I can do that too.

I was more concerned that this was a side effect of #1212.

ritch: > Do you know the answer yet? ;-)

Yes I got it :). Its very odd that attaching to a `dataSource` modifies your property definitions :(.

test/replication.rest.test.js
ritch: ya that would be better

ritch:
I'm happy with the changes. A bit concerned about the 3-4 TODO's... But if those are captured by other stories than this LGTM. :+1:

PR #1284, opened 2015-04-07
user: throw error if email is not found in reset

yorkie:
Sometimes the user needs to get if someone who wants to reset his password has typed the correct email.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -500,7 +500,10 @@ module.exports = function(User) {
             }
           });
         } else {
-          cb();
+          err = new Error('User not found by the email: ' + options.email);
+          err.statusCode = 404;
+          err.code = 'USER_NOT_FOUND';
+          cb(err);
         }
       });
     } else {


slnode:

yorkie:
can someone take a look at this?

raymondfeng:
I wonder if this feature should be controlled by an option. Telling people that an email doesn't match the users could be a security concern. @ritch WDYT?

OwenBrotherwood:
@raymondfeng @ritch 
Use Case: 
User friendly, maybe.
Hacker friendly, likely.
If it is a guess the username, the attacker obtains immediate feedback that she should try another combination.
It is also a candidate for rate limiting if this is not already done.

yorkie:
How is this pr going?

slnode:

raymondfeng:
@ritch Can you review the PR?

ritch:
This cannot be the default behavior. I'm ok with it being behind a setting.

ritch:
Will re-open once the flag is added @yorkie.

PR #1294, opened 2015-04-09
Conflict resolution and Access control

bajtos:
**Fix PersistedModel._defineChangeModel** (the first commit)

 - Correctly handle the case when the model is attached multiple times during the lifecycle, this happens because `loopback.createModel` always makes an attempt to auto-attach.

**Conflict resolution and Access control** (the second commit)

- Add end-to-end unit-tests verifying enforcement of access control during conflict resolution.

- Implement two facade methods providing REST API for Change methods used
by conflict resolution:

        PersistedModel.findLastChange
        GET /api/{model.pluralName}/{id}/changes/last

        PersistedModel.updateLastChange
        PUT /api/{model.pluralName}/{id}/changes/last

  By providing these two methods on PersistedModel, replication users don't have to expose the Change model via the REST API. What's even more important, these two methods use the same set of ACL rules as other (regular) PersistedModel methods.

- Rework `Conflict.prototype.changes()` and `Conflict.prototype.resolve()` to use these new facade methods.

- Implement a new method `Conflict.prototype.swapParties()` that provides better API for the situation when a conflict detected in Remote->Local replication should be resolved locally (i.e. in the replication target).

Connect to strongloop/loopback#1166

/to @ritch please review

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -617,9 +617,8 @@ module.exports = function(Change) {
     ], done);
 
     function getSourceChange(cb) {
-      conflict.SourceChange.findOne({where: {
-        modelId: conflict.modelId
-      }}, function(err, change) {
+      var SourceModel = conflict.SourceModel;
+      SourceModel.findLastChange(conflict.modelId, function(err, change) {
         if (err) return cb(err);
         sourceChange = change;
         cb();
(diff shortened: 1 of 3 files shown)


common/models/change.js
```diff
@@ -703,6 +707,23 @@ module.exports = function(Change) {
   };
 
   /**
+   * Return a new Conflict instance with swapped Source and Target models.
+   *
+   * This is useful for example when resolving a conflict in one-way
```
ritch: Consider:

> This is useful for example when resolving a conflict in one-way

changing to

> This is useful when resolving a conflict in one-way

ritch:
LGTM

common/models/change.js
bajtos: :+1:

Issue #386, opened 2014-07-20
Inclusion filter present security issue


offlinehacker:
So let's say i have following relation `user <-groupMembers-> group`, now if i allow group owner to query users, he can also include all user tokens, and login as user, nice.

grosem:
@fabien & @raymondfeng 

I just tested it with the latest version and it is still possible. For example like this (manager is extending the User model):

http://localhost:3000/api/events?filter={"where":{"nr":1},"include":{"relation":"manager","scope":{"include":[{"relation":"accessTokens"}]}}}

https://groups.google.com/forum/#!topic/loopbackjs/8XmsoaHpNP4 seems to be the same problem.

fabien:
ping @raymondfeng @bajtos @ritch

bajtos:
@raymondfeng mentioned in the mailing list that one can disable includes in model/relation settings using the option `disableInclude: true`. Does that fix the problem? 

Is it enough to change configuration of the built-in relation "User hasMany accessTokens" to prevent inclusion? Or is that just a workaround and we need to implement @raymondfeng proposal "the ACL will have to check the permissions for the included relations"?

grosem:
From my point of view this should definitely covered by ACL.

raymondfeng:
@bajtos User.hasMany accessTokens probably should have `disableInclude` set to `true` by default.

bajtos:
Closing the issue as fixed by #1318.

erikverheij:
The ability to access related objects via include is a large security hole. Essentially it makes the whole ACL useless as long as users can use the include filter. 

I suggest to add a big warning on the include documentation page. In fact I think filter.include should be disabled by default until ACL deals with it.

fabien:
@erikverheij agreed, this should be handled with caution. 

@raymondfeng @bajtos what do you think?

A workaround for now (from the top of my head) would be:

```
Model.observe('access', function(ctx, next) {
  ctx.query.include = {}; // clear
  next();
});
```

Shyri:
I totally agree with @erikverheij . Include feature is a nice way to avoid multiple requests from the frontend but it is a must to be handled by ACL in order to have decent security. At the moment the only solution I can see is to use "disableInclude" for all models that needs to be blocked to at least one role (other than $everyone or $unauthorized) and access them with specific requests only allowed for specific roles

ben-girardet:
Any chance that the ACL would handle this security issue in the future ?

dmisdm:
[...]

PR #1301, opened 2015-04-14
Upgrade test fixtures to 2.x project layouts

raymondfeng:
/to @bajtos 
/cc @ritch 

See https://github.com/strongloop/loopback/issues/386


common/models/user.json
```diff
@@ -90,7 +90,10 @@
     "accessTokens": {
       "type": "hasMany",
       "model": "AccessToken",
-      "foreignKey": "userId"
+      "foreignKey": "userId",
+      "options": {
+        "disableInclude": true
+      }
```
bajtos: FWIW, this option is not covered by your new tests, because the tests are explicitly overriding this option in `models.json`. Please consider writing few more test to verify that a) the built-in user disables include of access tokens b) user submodels inherit this setting by default too.

bajtos:
The model change is not covered by tests. Other than that, the patch LGTM.

raymondfeng:
@bajtos PTAL again. I had to convert the test apps into LB 2.x layout, which is a great side effect.

bajtos:
>  I had to convert the test apps into LB 2.x layout, which is a great side effect.

For posterity, can you please explain why was this needed?

 - LB 2.x project layout does not use `defaultForType` option, please remove it from all `datasources.json` files

test/fixtures/user-integration-app/server/server.js
```diff
@@ -0,0 +1,14 @@
...
+app.enableAuth();
+boot(app, __dirname);
+app.use(loopback.favicon());
+app.use(loopback.cookieParser({secret: app.get('cookieSecret')}));
+app.use(loopback.token({model: app.models.AccessToken}));
+var apiPath = '/api';
+app.use(apiPath, loopback.rest());
+app.use(loopback.static(path.join(__dirname, 'public')));
+app.use(loopback.urlNotFound());
+app.use(loopback.errorHandler());
```
bajtos: Is there any particular reason why you are not using `middleware.json` and middleware phases in this app?

Are your tests relying on `cookieParser` at all? And why do they need static files from the `public` directory?

test/user.integration.js
```diff
@@ -0,0 +1,166 @@
...
+var SIMPLE_APP = path.join(__dirname, 'fixtures', 'user-integration-app');
+var app = require(path.join(SIMPLE_APP, 'server/server.js'));
+var expect = require('chai').expect;
+
+describe('users - integration', function() {
+
+  lt.beforeEach.withApp(app);
+
+  before(function(done) {
+    // HACK: [rfeng] We have to reset the relations are they are polluted by
```
bajtos: Typo: `... relations are they are polluted...`

test/user.integration.js
```diff
@@ -0,0 +1,166 @@
...
+    // other tests
+    app.models.User.hasMany(app.models.post);
+    app.models.User.hasMany(app.models.AccessToken,
+      {options: {disableInclude: true}});
```
bajtos: You are explicitly setting `disableInclude: true` here, thus the tests are again not testing the default behaviour AFAICT.

test/user.integration.js
```diff
@@ -0,0 +1,166 @@
...
+    app.models.User.destroyAll(function(err) {
+      if (err) return done(err);
+      app.models.post.destroyAll(function(err) {
+        if (err) return done(err);
+        app.models.blog.destroyAll(function(err) {
+          if (err) return done(err);
+          done();
+        });
+      });
+    });
```
bajtos: Perhaps you can run `app.dataSources.db.automigrate()` instead of this cascade of `destroyAll`?

test/user.integration.js
```diff
@@ -0,0 +1,166 @@
...
+          }
+          expect(res.body.title).to.be.eql('T1');
+          expect(res.body.content).to.be.eql('C1');
+          expect(res.body.userId).to.be.eql(userId);
+          done();
+        });
+    });
+
+    // FIXME: [rfeng] The test is passing if run alone. But it fails with
+    // `npm test` as the loopback models are polluted by other tests
```
bajtos: Did you try to use the new feature "localRegistry" (see https://github.com/strongloop/loopback/pull/1212) in these tests? I think it should fix this problem.

```js
```

We should eventually migrate all unit-tests to use local (per-app) registry, but that's a task for 3.0.

bajtos:
> I had to convert the test apps into LB 2.x layout, which is a great side effect.

Actually I think this is a wrong thing to do. LoopBack tests should not depend on loopback-boot, they should setup the application via code only.

 - Loading an external app from `fixtures` makes tests difficult to understand, one has to scan many files to understand what's going on.
 - loopback-boot is an optional component providing opinionated convention for project layout. LoopBack must work without loopback-boot too, and thus the unit-tests should not depend on loopback-boot.

IIRC, the main reason why the existing LoopBack tests are depending on loopback-boot is to ensure that LoopBack 1.x apps can be upgraded to LoopBack 2.x + loopback-boot 1.x with no breaking changes.

I think this issue deserves some discussion and it will be best to decouple it from your work on "Set disableInclude for User->AccessToken relation". Is there any way how to finish your fix without upgrading loopback-boot?

test/user.integration.js
raymondfeng: Yes, I tried and it didn't work. The issue is that we try to test the base User class which can be attached to different apps. Unless the per-app registry clones the User, we still deal with a shared User class.

test/user.integration.js
raymondfeng: I'll try.

test/user.integration.js
raymondfeng: That was unfortunately due to the hack I had to use to reset the relations when this test is run as part of `npm test`. Running this test alone doesn't require the `options`.

test/user.integration.js
raymondfeng: I'll fix it.

test/fixtures/user-integration-app/server/server.js
raymondfeng: No, I just copied them from the existing fixture. I can clean them up.

raymondfeng:
> Actually I think this is a wrong thing to do. LoopBack tests should not depend on loopback-boot, they should setup the application via code only.
Loading an external app from fixtures makes tests difficult to understand, one has to scan many files to understand what's going on.
loopback-boot is an optional component providing opinionated convention for project layout. LoopBack must work without loopback-boot too, and thus the unit-tests should not depend on loopback-boot.
IIRC, the main reason why the existing LoopBack tests are depending on loopback-boot is to ensure that LoopBack 1.x apps can be upgraded to LoopBack 2.x + loopback-boot 1.x with no breaking changes.
I think this issue deserves some discussion and it will be best to decouple it from your work on "Set disableInclude for User->AccessToken relation". Is there any way how to finish your fix without upgrading loopback-boot?

The project already has dependency on `loopback-boot` before the PR. I simply upgrade the fixtures to 2.x layout. 

Migrating from 1.x to 2.x is not a main concern any more. I think it's better to use 2.x layout as developers sometimes try to follow what we do. If you feel strongly about removing the test fixtures, we can discuss that as a separate issue.

With 1.x layout, there is no way to use the base class such as `User` for REST. The test is meant to cover our main stream - LB 2.x.

test/user.integration.js
bajtos: The per-app registry creates a new unique `User` instance in a unique `modelBuilder` instance. Just not that you have to access the User via `app.registry.getModel('User')`, not `loopback.User`. It's also possible that your app did not have the relations set up correctly, or the token middleware was using the global models instead of per-app local models. Last but not least, the local registry does not come with built-in models, you have to define them yourself (and there isn't an easy API for that yet).

bajtos:
> The project already has dependency on loopback-boot before the PR. I simply upgrade the fixtures to 2.x layout.

> Migrating from 1.x to 2.x is not a main concern any more. I think it's better to use 2.x layout as developers sometimes try to follow what we do. If you feel strongly about removing the test fixtures, we can discuss that as a separate issue.

> With 1.x layout, there is no way to use the base class such as User for REST. The test is meant to cover our main stream - LB 2.x.

I dig up the commit that added the dependency on loopback-boot - see c896c78d. It seems the change was made purely in order to support test fixtures booting the app via `loopback.boot()`, in which case I agree with you that it's ok to upgrade loopback-boot to 2.x.

However, I still believe that we should keep the number of fixtures as small as possible, create loopback apps inside the tests purely code-wise and keep these apps as small as possibly.

raymondfeng:
I decoupled the LB 2.x upgrade from `disableInclude User.accessTokens`. This PR is for the upgrade.

test/user.test.js
```diff
@@ -16,14 +18,21 @@ describe('User', function() {
...
 
+  var defaultApp;
+
   beforeEach(function() {
-    User = loopback.User.extend('user');
+    // FIXME: [rfeng] Remove loopback.User.app so that remote hooks don't go
+    // to the wrong app instance
+    defaultApp = loopback.User.app;
+    loopback.User.app = null;
+    User = loopback.User.extend('TestUser', {}, {http: {path: 'test-users'}});
```
bajtos: Is there any particular reason why your are not using HTTP path `/users`? That way you will not have to change all the test cased below.

test/user.test.js
```diff
@@ -78,7 +88,7 @@ describe('User', function() {
         assert(err);
         assert.equal(err.name, 'ValidationError');
         assert.equal(err.statusCode, 422);
-        assert.equal(err.details.context, 'user');
+        assert.equal(err.details.context, 'TestUser');
```
bajtos: The following is better as it should be resistant against future changes:

```js
assert.equal(err.details.context, TestUser.modelName);
```

test/user.test.js
raymondfeng: Yes, because the model in use is a sub-model of User. If we don't explicitly specify `http.path`, it will default to /testusers.

test/user.test.js
raymondfeng: Sure.

test/user.test.js
bajtos: So why can't you specify `http.path: 'users'`, or even `plural: 'Users'`, since you are already overriding the default?

bajtos: Anyways. If you believe that it's better to have `/test-users` URL even at the cost of making future code archaeology more difficult, then I can live with that.

bajtos:
LGTM.

PR #1306, opened 2015-04-15
Expose the `filter` argument for findById

raymondfeng:
/to @bajtos or @ritch 

This PR depends on https://github.com/strongloop/loopback-datasource-juggler/pull/564

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -142,15 +142,20 @@ module.exports = function(registry) {
   };
 
   /**
-   * Find object by ID.
+   * Find object by ID with an optional filter for include/fields.
    *
    * @param {*} id Primary key value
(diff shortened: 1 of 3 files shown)


bajtos:
Please add a unit-test to test that the new `filter` argument works as expected. You can add it for example to [test/model.test.js](https://github.com/strongloop/loopback/blob/bd973def6308a771bdd7212d9a1d7021f3772491/test/model.test.js#L139-L166).

Consider also writing a test capturing the use case of @pulkitsinghal, where `findById` + `filter.include` is used as a workaround for missing role-based filtering in `find`.

raymondfeng:
@bajtos Tests added.

raymondfeng:
@ritch Docs updated.

ritch:
LGTM

bajtos:
:+1:

PR #1311, opened 2015-04-17
Rework global registry to be per-module-instance

bajtos:
Fix a regression introduced by b917075 where two loopback projects,
each one using a different instance of loopback module, were
sharing the global registry and thus not working correctly.

The issue was discovered by unit-tests in loopback-workspace.

/to @ritch please review


ritch:
:+1:

ritch:
We can just get rid of that file now.

bajtos:
> We can just get rid of that file now.

Done, PTAL again. (The original commit for posterity: 7442c12)

ritch:
Looks good to me! :ship: it

PR #1318, opened 2015-04-20
Disable inclusion of User.accessTokens

raymondfeng:
/to @bajtos

### DIFF
--- a/test/fixtures/user-integration-app/server/server.js
+++ b/test/fixtures/user-integration-app/server/server.js
@@ -1,0 +1,10 @@
+var loopback = require('../../../../index');
+var boot = require('loopback-boot');
+var app = module.exports = loopback();
+app.enableAuth();
+boot(app, __dirname);
+app.use(loopback.token({model: app.models.AccessToken}));
+var apiPath = '/api';
+app.use(apiPath, loopback.rest());
+app.use(loopback.urlNotFound());
+app.use(loopback.errorHandler());
--- a/test/user.integration.js
+++ b/test/user.integration.js
@@ -1,0 +1,166 @@
+/*jshint -W030 */
+var loopback = require('../');
+var lt = require('loopback-testing');
+var path = require('path');
+var SIMPLE_APP = path.join(__dirname, 'fixtures', 'user-integration-app');
+var app = require(path.join(SIMPLE_APP, 'server/server.js'));
+var expect = require('chai').expect;
+
+describe('users - integration', function() {
+
+  lt.beforeEach.withApp(app);
+
+  before(function(done) {
+    // HACK: [rfeng] We have to reset the relations as they are polluted by
+    // other tests
+    app.models.User.hasMany(app.models.post);
+    app.models.User.hasMany(app.models.AccessToken,
+      {options: {disableInclude: true}});
+    app.models.AccessToken.belongsTo(app.models.User);
+    app.models.User.destroyAll(function(err) {
+      if (err) return done(err);
+      app.models.post.destroyAll(function(err) {
+        if (err) return done(err);
+        app.models.blog.destroyAll(function(err) {
+          if (err) return done(err);
+          done();
+        });
+      });
+    });
+  });
+
+  describe('base-user', function() {
+    var userId;
+    var accessToken;
+
+    it('should create a new user', function(done) {
+      var url = '/api/users';
+
+      this.post(url)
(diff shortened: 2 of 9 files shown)


test/fixtures/user-integration-app/server/server.js
```diff
@@ -0,0 +1,10 @@
+var loopback = require('../../../../index');
+var boot = require('loopback-boot');
+var app = module.exports = loopback();
+app.enableAuth();
+boot(app, __dirname);
+app.use(loopback.token({model: app.models.AccessToken}));
+var apiPath = '/api';
+app.use(apiPath, loopback.rest());
+app.use(loopback.urlNotFound());
+app.use(loopback.errorHandler());
```
bajtos: Can you please build the app in `test/user.integration.js` and don't use loopback-boot?

test/user.integration.js
```diff
@@ -0,0 +1,166 @@
...
+          expect(res.body.title).to.be.eql('T1');
+          expect(res.body.content).to.be.eql('C1');
+          expect(res.body.userId).to.be.eql(userId);
+          done();
+        });
+    });
+
+    // FIXME: [rfeng] The test is passing if run alone. But it fails with
+    // `npm test` as the loopback models are polluted by other tests
+    it('should prevent access tokens from being included', function(done) {
```
bajtos: This test is useless as it does not verify anything. If you per-app registry does not solve the problem of cross-test interference, then it's better to remove this test completely and use a different way to ensure that access tokens are not include-able by default.

test/fixtures/user-integration-app/server/server.js
raymondfeng: Why? The purpose of the user.integration.js is to test a LB app's built-in user model.

test/user.integration.js
raymondfeng: It does verify with the User/AccessToken relation setup (disableInclude = true), the accessTokens cannot be included via a user, even though the pollution from other tests forces me to use the hack to restore the relations to mimic the original configuration from the test fixture.

I'm open to better ways to test the feature but it's better not to let the security hole drags too long.

test/fixtures/user-integration-app/server/server.js
bajtos: > The purpose of the user.integration.js is to test a LB app's built-in user model.

As you wrote yourself, there is nothing related to loopback-boot in the scenario you are testing.

It is a good practice to write the test code in such way that it's easy to quickly understand what is the scenario being tested. The test scenario should be also as minimal as possible and include only things that are absolutely necessary for the test.

By using a fixture defined in several files living outside of the test file, you are making it difficult to understand what is exactly being tested. By including lots of unrelated stuff in your fixture, you make it difficult to see what is the important (test-specific) part and what is just boilerplate. Last but not least, all these extra unneeded bits are making the maintenance more expensive.

This is already visible in this pull request, as the `server/server.js` file is not following the project structure as scaffolded by recent versions of loopback-workspace.

To test the built-in user model, you need just few lines of code:

```js
var app = loopback();
app.model(loopback.User, { dataSource: 'db' });
app.model(loopback.AccessToken, { dataSource: 'db' });
app.model(loopback.ACL, { dataSource: 'db' });
app.model(loopback.Role, { dataSource: 'db' });
app.model(loopback.RoleMapping, { dataSource: 'db' }); // this may be not necessary

app.enableAuth();

app.use(loopback.rest());
request = supertest(app);
```

I admit this setup is not as short as it should be, let's discuss the best API in https://github.com/strongloop/loopback/pull/1323 and keep the code above for the purposes of this PR.

test/user.integration.js
bajtos: My point is: if you remove `disableInclude:true` from `common/models/user.json`, all your tests will keep passing.

> it's better not to let the security hole drags too long.

That's true. Let's land the fix with no unit tests then (just the small change in `common/models/users.json`) and implement tests in a follow-up pull request.

test/fixtures/user-integration-app/server/server.js
raymondfeng: There are more models such as `post` and `blog` are needed. I'll merge the PR as is and defer the cleanup of tests.

bajtos: I am rather unhappy that you have added tests that always pass, regardless of whether the underlying implementation works or not. Such tests are giving false sense of security :-1:

PR #1323, opened 2015-04-21
Auto-configure models required by `app.enableAuth`

bajtos:
Modify `app.enableAuth` to automaticaly setup all required models that are not attached to the app nor a datasource.

Users wishing to use this option must provide the name of the data-source to use for these models.

Example usage:

    var app = loopback();
    app.enableAuth({ dataSource: 'db' });

    app.use(loopback.rest());
    app.listen(3000);

The purpose of this new API is to make it easier to setup authentication from code that does not use "slc loopback" project scaffolding, typically from unit-tests.

/to @ritch @raymondfeng please review

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -291,9 +291,36 @@ app.dataSources = app.datasources = {};
  * Enable app wide authentication.
  */
 
-app.enableAuth = function() {
+app.enableAuth = function(options) {
+  var AUTH_MODELS = ['User', 'AccessToken', 'ACL', 'Role', 'RoleMapping'];
+
   var remotes = this.remotes();
   var app = this;
+
+  if (options && options.dataSource) {
+    var appModels = app.registry.modelBuilder.models;
+    AUTH_MODELS.forEach(function(m) {
+      var Model = app.registry.findModel(m);
+      if (!Model) {
+        throw new Error(
+          'Authentication requires model ' + m + ' to be defined.');
+      }
+
+      if (m.dataSource || m.app) return;
+
+      for (var name in appModels) {
+        var candidate = appModels[name];
+        var isSubclass = candidate.prototype instanceof Model;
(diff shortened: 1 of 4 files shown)


lib/application.js
```diff
@@ -291,10 +291,27 @@ app.dataSources = app.datasources = {};
  * Enable app wide authentication.
  */
 
-app.enableAuth = function() {
+app.enableAuth = function(options) {
+  var AUTH_MODELS = ['User', 'AccessToken', 'ACL', 'Role', 'RoleMapping'];
```
raymondfeng: What's going to happen if these models are sub-classed?

bajtos: That's a good question. This shortcut is intended only for the simple situations where you typically don't create subclasses. To keep the implementation simple, I can modify the condition determining whether a base model should be attached or not, and do not attach a base model if there is a subclassed model already attached?

From user perspective, if you create let's say Customer extending User:

 - If Customer is attached to a data source or the app, then the builtin User model is left unchanged.
 - If Customer is not attached to the app nor the datasource, then we attach the builtin User model for you.

Thoughts?

bajtos:
@ritch @raymondfeng ping

lib/application.js
raymondfeng: Sounds good.

ritch:
:+1:

bajtos:
Reworked.

@raymondfeng PTAL again.

@ritch FYI, I have added a new `loopback` option `loadBuiltinModels` that can be used together with `localRegistry`, see 0ccc1e2.

ritch:
> I have added a new loopback option loadBuiltinModels

Cool. We should start marking code like this somehow... so we can easily review it.

bajtos:
@slnode test please

ritch:
LGTM

raymondfeng:
LGTM

PR #1329, opened 2015-04-23
Allow dataSource === false

raymondfeng:
/to @bajtos 
/cc @kraman @sam-github

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -236,7 +236,7 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
     ModelCtor.attachTo(config.dataSource);
     debug('Attached model `%s` to dataSource `%s`',
       modelName, config.dataSource.name);
-  } else if (config.dataSource === null) {
+  } else if (config.dataSource === null || config.dataSource === false) {
     debug('Model `%s` is not attached to any DataSource by configuration.',
       modelName);
   } else {
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -604,6 +604,10 @@ describe('app', function() {
       app.model('MyTestModel', { dataSource: null });
     });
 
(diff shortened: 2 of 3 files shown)


lib/registry.js
```diff
@@ -236,7 +236,7 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
     ModelCtor.attachTo(config.dataSource);
     debug('Attached model `%s` to dataSource `%s`',
       modelName, config.dataSource.name);
-  } else if (config.dataSource === null) {
+  } else if (config.dataSource === null || config.dataSource === false) {
```
sam-github: Above doesn't handle `undefined`, is that intentional?

Why not `if (!config.dataSource) {`, so that all falsy values are treated equally?

raymondfeng: It's intentional as `undefined` means the property is not set in json and we want to warn the user.

bajtos:
Please add a unit test.

What is the reasoning behind the change? What's wrong with `null`?

bajtos:
Oh I see, the error message says that both `null` and `false` should work.

Well, if you think `false` should be supported, then I won't argue against that too much.

raymondfeng:
@bajtos Can you point me to the original unit test for `dataSource === null`? I can make sure it allows `false` too?

bajtos:
See #483 and [this line](https://github.com/strongloop/loopback/pull/483/files#diff-d118d017711e90fc081f7548515c5436R73) in particular.

Since we are generally not testing the logs, I think it's ok to simply copy "accepts null dataSource" and create a new test "accepts false dataSource".

raymondfeng:
@bajtos Tests added. PTAL.

bajtos:
LGTM. Please wait for the CI to finish & pass before landing.

PR #1332, opened 2015-04-24
Fix remoting metadata for User.login#include

bajtos:
Change the type of the "include" argument to "string array".

The type used to be "string" before and thus requests sending multiple
include items were technically incorrect.

See https://github.com/strongloop/strong-remoting/pull/207

Connect to strongloop-internal/scrum-loopback#243

/to @ritch please review


ritch:
LGTM

PR #1346, opened 2015-04-29
Access token : option noConcat 

OwenBrotherwood:
Full control over which methods are used to finde Access token.
Requires full specification of which methods and values are to be used with loopback.token


slnode:

Issue #1326, opened 2015-04-22
Use Case: ability to disable that Access Tokens are accepted when in an URL


OwenBrotherwood:
Use Case: 
- ability to disable that Access Tokens are accepted when in an URL
- alternativly, possibility to enforce elsewhere as an alternative

It may be such that it is not possible to disable the use of Access Tokens when sent in an URL.
As to if one can enforce in a model is unknown.

Addresses a case if the use of URL (or cookie) based Access Tokens are not allowed.

---

### Implementation for options.searchDefaultTokenKeys

---

```
$ mocha -g 'api.*token'

  api:loopback:middleware:token(options)
    loopback.token({})
      V header (125ms)
      Breaking change
        V None
    options
      searchDefaultTokenKeys:[true|false]
        Normal usage: false for use case, restrict location of token
          Example, token in header
            V options.headers is empty, true (46ms)
            V options.headers has header, false
        Unormal usage: testing or unknown usage
          Example, token in header
searchDefaultTokenKeys is false and no headers: no token keys will be found
            V options.headers is empty, false
        Breaking change
          V None

  6 passing (297ms)
```

---

### Compare master

https://github.com/strongloop/loopback/compare/master...sloopback:loopback-issue-1326?expand=1

OwenBrotherwood:
The Use Case is possible input to https://github.com/strongloop/loopback/issues/1091

bajtos:
This should be relatively easy to implement as a new configuration option for `loopback#token` middleware. The options are processed in [common/models/access-token.js](https://github.com/strongloop/loopback/blob/f93590de1bab0b19c00f2c0a5956243e2776d58e/common/models/access-token.js#L163-L173).

OwenBrotherwood:
@bajtos 
I will probably try an option.noConcat around the concats.
This means specific enablement of type/value with enableAuthentication (also for previous default values)

Exposing the parameter that are used in enableAuthentication could also allow a better audit of a (future) loopback AAA model that contains the type and text used for tokens in the Authentication /Authorization

Effected doc
- http://docs.strongloop.com/display/public/LB/Introduction+to+User+model+authentication
  -- investigate how set token works ie cookie/url/header ...
- http://docs.strongloop.com/display/public/LB/Making+authenticated+requests

OwenBrotherwood:
@bajtos 
On the pull train @ https://github.com/strongloop/loopback/pull/1347
@crandmck 
with a small doc example in the pull :)

[...]

PR #1347, opened 2015-04-29
loopback.token option.defaultTokenKeys

OwenBrotherwood:
Allow full control of which methods and values are used to obtain Access Token.
Requires specific methods and values as option 

Example: 
Only allow headers as source of Access Token from X-Access-Token and authorization 
```
loopback.token({headers: ['X-Access-Token', 'authorization'], searchDefaultTokenKeys: false })
```
https://github.com/strongloop/loopback/issues/1326 

Signed-off-by: Owen Brotherwood <owen@brotherwood.biz>

Edit: 
https://github.com/strongloop/loopback/pull/1347#discussion_r29793818 option name change from options.defaultTokenKeys to options.searchDefaultTokenKeys


slnode:

bajtos:
@slnode ok to test

bajtos:
@OwenBrotherwood Thank you for the pull request.
1) We use two spaces for indentation, not tab characters.
2) Can you please add unit-tests to verify that your implementation works correctly and to prevent regressions in the future?

OwenBrotherwood:
@bajtos 
Will do: I am not a full time programmer: and at the time I was using vi as it was a small patch but forgot  space/tab

I wanted to do unit-tests: I just wanted an excuse to find out how to.
- Perhaps a link to a unit-test I can use as the base (there must be one for the original routines somewhere) would save me a little time.

bajtos:
> Any comments about the option name noConcat 

Perhaps `includeDefaults: false` or `excludeDefaults: true` would be a better name?

Regarding the unit-tests, [test/access-token.test.js](https://github.com/strongloop/loopback/blob/242bcec539abe3fadac475d7fa60a67f7c7b34a3/test/access-token.test.js) is the best place where to add the new test. Take a look at "should populate req.token from the query string" for an example. You will need to modify `createTestAppAndRequest` to accept options for the token middleware.

Let me know how far you get and if there is anything else I can do to help you along the way.

OwenBrotherwood:
@bajtos 
- Add option.defaultTokenKeys
  - the doc uses wording "additional keys" @ http://apidocs.strongloop.com/loopback/#loopback-token
    - https://github.com/OwenBrotherwood/loopback/blob/master/common/models/access-token.js#L171-176
- Prepare for some form of test by altering createTestApp to support option.defaultTokenKeys
  - https://github.com/OwenBrotherwood/loopback/blob/master/test/access-token.test.js#L354-375
- Next steps
  - find what is to be tested
    - make a test ...

OwenBrotherwood:
@bajtos 
Take a look 
Comments are in https://github.com/strongloop/loopback/pull/1347#issuecomment-97952341
1 test made and support for others

OwenBrotherwood:
https://player.vimeo.com/video/77870960
Use mocha to run a specific test
- mine seems to fail just now: WIP

OwenBrotherwood:
@bajtos 
On a clone of strongloop/loopback (from today), mocha  test/access-token.js gives a lot of fails.

bajtos:
@OwenBrotherwood loopback tests have this ugly quirk that you cannot run a single test file, you must include `test/support.js` too:

```
$ mocha test/support.js test/access-token.js
```

The easiest way to run a subset of unit-tests is to use the `-g` option:
 
```
$ mocha -g AccessToken
```

> I think when everything is ok, it might be nicer to reject this push and I can create a "nicer" push: to many commits for my taste

You can clean up the commit history by running `git rebase -i master` and then `git push -f`. But don't worry about that, I can clean it up myself once the patch is ready to be merged.

test/access-token.test.js
```diff
@@ -41,7 +41,17 @@ describe('loopback.token(options)', function() {
       .end(done);
   });
 
-  describe('populating req.toen from HTTP Basic Auth formatted authorization header', function() {
+  it('should populate req.token from an authorization header with bearer token using option.defaultTokenKeys', function(done) {
```
bajtos: Please keep the line length at 80 characters max.

test/access-token.test.js
```diff
@@ -351,10 +361,33 @@ function createTestApp(testToken, settings, done) {
...
+  } else {
+    tokenOptions = {
+      model: Token,
+      currentUserLiteral: 'me',
+      defaultTokenKeys: defaultTokenKeys,
+      headers: headers,
+      cookies: cookies,
+      params: params
+    };
+  }
```
bajtos: This looks overcomplicated to me. How about the following solution?

```js
var tokenOptions = settings.token || {};
tokenOptions.model = Token;
tokenOptions.currentUserLiteral = 'me';
```

Usage:

```js
createTestAppAndRequest(this.token,
  { token: { headers: ['authorization'], defaultTokenKeys: false } },
  done)
```

bajtos:
@OwenBrotherwood  Please add another test to verify that the default token keys are not added. I.e. send a request that would be authorized with `defaultTokenKeys: true`, and verify that it returns 401 when `defaultTokenKeys: false`.

@ritch I feel the option name `defaultTokenKeys` is not the best one, can you come up with a better name?

OwenBrotherwood:
Understood : 
Work In Progress, I really want to understand how to make effective tests for node at this stage of my learning curve

OwenBrotherwood:
I did a clone of loopback "just now" and did a mocha test/support.js test/access-token.test.js on the master

Are the errors shown in this link the same as you have for the current master? http://pastebin.com/hM95D2PQ

OwenBrotherwood:
@bajtos I deceided to make my own test program
I have dificulty in ensurring a defaultTokenKeys=false, and no other options should fail.
I google a bit :)

https://github.com/OwenBrotherwood/loopback/blob/master/test/access-token-ob.test.js#L123-L129

OwenBrotherwood:
@bajtos Sorry about the spam :)
Let's see if this get's me somewhere:
https://github.com/OwenBrotherwood/loopback/blob/784db50300fec892ca94aba01c5d486d221d1ecb/test/access-token-defaultTokenKeys.test.js#L3-L4

https://github.com/OwenBrotherwood/loopback/blob/784db50300fec892ca94aba01c5d486d221d1ecb/test/app-access-token.js

OwenBrotherwood:
@bajtos 
Doc the  API and the tests needed, implement a TEST environment then add the code to implement the change in the code.

The description of API and the tests used:
https://github.com/OwenBrotherwood/loopback/blob/master/test/API-loopback-token-defaultTokenKeys.test.js
The implementation of TEST and the components needed to create the test environment:
https://github.com/OwenBrotherwood/loopback/blob/master/test/TEST-loopback-token-defaultTokenKeys.test.js

common/models/access-token.js
```diff
@@ -168,9 +168,12 @@ module.exports = function(AccessToken) {
     var length;
     var id;
 
-    params = params.concat(['access_token']);
-    headers = headers.concat(['X-Access-Token', 'authorization']);
-    cookies = cookies.concat(['access_token', 'authorization']);
+    // https://github.com/strongloop/loopback/issues/1326
+    if (options.defaultTokenKeys === undefined || options.defaultTokenKeys === true) {
```
ritch: I would suggest a different option name:

```js
searchDefaultTokenKeys: false
```

OwenBrotherwood: I agree, I was getting tired of the other name

OwenBrotherwood:
@bajtos 
As far as I can see, there is a possible problem with the main test:
https://github.com/strongloop/loopback/blob/master/test/access-token.test.js?#L364-L365
- it only tests the contents of the req and not the validity of the req with referance to the model/user role/ACL/Authentication/Authorization

I now have the same test:
https://github.com/OwenBrotherwood/loopback/blob/eb44d7744dfb855a43d1630b86d7ea4663f42623/test/TEST-loopback-token-searchDefaultTokenKeys.js?#L75-L91

I will now try and make sure that the user's use of a model is allowed by allowing the code to generate 200 or 401 instead of the assert test.

bajtos:
> it only tests the contents of the req and not the validity of the req with referance to the model/user role/ACL/Authentication/Authorization

Just call a different URL path instead of `/`. Here is an existing test [test/access-token.test.js#L284-L298](https://github.com/strongloop/loopback/blob/694ab814f984119f380c398124647f958a25439a/test/access-token.test.js#L284-L298) that expects a 401 response.

common/models/access-token.js
```diff
@@ -168,9 +168,12 @@ module.exports = function(AccessToken) {
     var length;
     var id;
 
-    params = params.concat(['access_token']);
-    headers = headers.concat(['X-Access-Token', 'authorization']);
-    cookies = cookies.concat(['access_token', 'authorization']);
+    // https://github.com/strongloop/loopback/issues/1326
+    if (options.searchDefaultTokenKeys === undefined || options.searchDefaultTokenKeys === true) {
```
bajtos: We usually use a simpler check `options.searchDefaultTokenKeys !== false`. It works well for `true`, `false` and `undefined`, even though it behaves differently for edge cases like `null` or non-boolean value.

bajtos:
@OwenBrotherwood To be honest, I am quite confused about the direction you are taking and don't know what's the best way how I can help you to move forward. Please let me know if there is anything else to answer or explain in addition to my comments above.

common/models/access-token.js
OwenBrotherwood: I will use !== false so that it the same as is currently used for this type of test

OwenBrotherwood:
@bajtos 
I understand the comment :)
I close this PR will continue to work on my fork.
Why? I use https://gitter.im/strongloop/loopback and all my work in my fork keeps creating "noise" in the right hand panel as I push to my fork.
I use https://gitter.im/strongloop/loopback to watch and learn: maybe others are doing the same and all the OwenBrotherwood push's are basically "noise".
We have tracking on https://github.com/strongloop/loopback/issues/1326 ie when I am ready to make the final PR, I will @bajtos there (or if I have a problem)

The current status is, there is a reasonably alteration to support options.searchDefaultTokenKeys https://github.com/OwenBrotherwood/loopback/blob/2bef4d861bd70ac823c62cfe4580af598d956d83/common/models/access-token.js?#L171-176 
BUT
If someone uses the api wrongly, ie loopback.token({searchDefaultTokenKeys:false),the call to the API will suceed even with the call not making any sense.
The wrong usage should throw an error, and the user is warned: code alteration in the call to the api to support the new option to be friendly to the user.

And of course a tests to ensure that the api loopback.token(options) is tested all the way from the call to the api to what is expected as the result.

bajtos:
> We have tracking on #1326 ie when I am ready to make the final PR, I will @bajtos there (or if I have a problem)

OK, thanks for letting me know.

> If someone uses the api wrongly, ie loopback.token({searchDefaultTokenKeys:false),the call to the API will suceed even with the call not making any sense.
> The wrong usage should throw an error, and the user is warned: code alteration in the call to the api to support the new option to be friendly to the user.

That's a good idea! :+1:

PR #1366, opened 2015-05-06
Define remote methods via model settings/config

### LINKED ISSUE #209: Remote methods mapping in JSON

bajtos:
Process `settings.methods` and `config.methods` as a key-value map
where the key is the method name and the value is an object describing
the method in the format expected by strong-remoting.

Example: a static method `Model.create`

    "methods": {
      "create": {
        "isStatic": true,
        "accepts": {
           "arg": "data", "type": "Car",
           "http": { "source": "body" }
        },
        "returns": { "arg": "data", "type": "Car", "root": true }
      }
    }

This patch is a reworked version of #1163 and fixes #209 .

/to @ritch please review
/cc @mrfelton

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -122,6 +122,8 @@ Registry.prototype.createModel = function(name, properties, options) {
   var model = BaseModel.extend(name, properties, options);
   model.registry = this;
 
+  this._defineRemoteMethods(model, options.methods);
+
   // try to attach
   try {
     this.autoAttachModel(model);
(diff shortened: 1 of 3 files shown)


lib/registry.js
```diff
@@ -247,6 +249,30 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
...
+  }
+
+  Object.keys(methods).forEach(function(key) {
+    var meta = methods[key];
+    if (typeof meta.isStatic !== 'boolean') {
+      console.warn('Remoting metadata for "%s.%s" is missing "isStatic" ' +
+        'flag, the method is registered as an instance method.',
+        ModelCtor.modelName,
+        key);
+      console.warn('This behaviour may change in the next major version.');
```
bajtos: I decided to report a warning instead of throwing an exception, because what if somebody was already using the "methods" option and implemented a solution similar to this patch? Throwing an exception would break backwards-compatibility for that case.

However, if you @ritch think I am too cautious and it's ok to throw an error, then I'll rework this part.

ritch: I think what you have is fine...

ritch:
@mrfelton thanks for the original contribution. We appreciate you moving this forward :+1: :+1: :+1:

mrfelton:
no worries... Will be great to see this finally committed in some form. I'm for the backwards compatibility option of providing a warning rather than throwing an exception, since I am using my earlier path on several systems :)

bajtos:
@ritch LGTY now?

ritch:
LGTM :+1:

gausie:
This should be reflected in the docs!

bajtos:
> This should be reflected in the docs!

Good catch, /cc @crandmck

PR #1106, opened 2015-02-18
Add attachments option to user verification email

reichert621:
This PR allows an array of `attachments` to be sent in the `sendEmail` function of `User#verify`.

Reference: https://github.com/andris9/Nodemailer#attachments

PR #1368, opened 2015-05-06
Pass the full options object to the email send method in user verification process

### LINKED ISSUE #1107: Can't add attachments to user verification email
When using the `User.prototype.verify` function (http://docs.strongloop.com/display/public/LB/User#user-prototype-verify), I'd like to include `attachments` in the options. The verification email we're sending includes an image, and this image can't be displayed with adding it as an attachment. See https://github.com/andris9/Nodemailer#attachments

I've created a pull request for this issue here: https://github.com/strongloop/loopback/pull/1106

alexandrusavin:
I think that there should be full flexibility for the creation of any the emails that are sent to the users. In order to achieve that the options object shouldn't be recreated. Consider this pull request that also fixes #1107.

@ritch I know that there is already PR #1106 but it only adds the attachment field to the options object. What if one needs to add alternative content or change the SMTP envelope? All these options are supported by the nodemailer library (the one that is used by loopback to send emails) so why ditch them?

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -391,15 +391,16 @@ module.exports = function(User) {
 
       options.text = options.text.replace('{href}', options.verifyHref);
 
+      options.to = options.to || user.email;
+
+      options.subject = options.subject || 'Thanks for Registering';
+
+      options.headers = options.headers || {};
+
       var template = loopback.template(options.template);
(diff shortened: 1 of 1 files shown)


slnode:

Issue #1363, opened 2015-05-05
Add "authorize" hook


ritch:
In order to fix https://github.com/strongloop/loopback/issues/1191 we need support for a special hook that is executed as the first "before" hook. Something like this should work:

```js

remotes.authorize = function(ctx, next) {
   // called first, preceding any other before hook
   next();
}

```

`checkAccess()` would be called using this hook ensuring ACLs are checked before any other hooks.

/cc @fabien @bajtos @raymondfeng 

**Tasks**

 - [x] create [PR](https://github.com/strongloop/strong-remoting/pull/212) adding authorization hook to strong-remoting
 - [x] create [PR](https://github.com/strongloop/loopback/pull/1370) using the new authorization hook in `enableAuth()` instead of `beforeRemote`

seriousben:
Really last, not as the first remote hook?

ritch:
> Really last, not as the first remote hook?

edited :)

seriousben:
:)

bajtos:
As long as there is only one hook of this type, then it makes sense. Can we please use a different method name please, to make it more obvious what is the purpose of this hook? For example:

```js
remotes.authorize = function(ctx, next) { next(); };
```

ritch:
`authorize` sounds good to me.

ritch:
Landed this new feature. Released as **v2.17.3**

Issue #1191, opened 2015-03-09
Remote hooks called before ACL rules are applied


toreader:
Right now ACLs are called after before remote hook. That means this stack of calls:
1. Before remote Hook
2. ACL
3. Remote method
4. After remote hook
I'd expect ACLs to be checked prior calling before remote hooks and not the other way round.

Example:
```javascript
Event.beforeRemote("prototype.__unlink__guests", function(ctx, modelInstance, next){
    console.log("before remote hook");
    next();
});
```
with this ACL
```
{
  "accessType": "*",
  "principalType": "ROLE",
  "principalId": "$everyone",
  "permission": "DENY",
  "property": "__unlink__guests"
}
```
Even though the Loopback explorer displays a 401 error, the console displays "before remote hook". In my opinion, this should not be the case.

raymondfeng:
@ritch ACL check is done from one of the remote hooks. Can we ensure it's called before any other remote hooks?

ritch:
If you create the hook after boot it should run in the expected order.

fabien:
@raymondfeng @ritch I think the expected behavior would be to have the ACL check be the first call in the stack. Or at least it should be configurable to work that way.

fabien:
@raymondfeng @ritch I strongly believe this should be fixed ASAP. Any before remote hook defined at the `models/model.js` or mixin level will just execute the before hook without any regard to ACL checks. This is a security disaster waiting to happen.

ritch:
@fabien agreed. I've implemented this as both a new feature (a new authorization hook) and a fix (ensuring we use this hook instead of adding blindly as a `remotes.before()` hook).

fabien:
@ritch thanks!

ritch:
@toreader @fabien fixed @ **2.17.3**

PR #1370, opened 2015-05-06
Use the new remoting.authorization hook for check access

ritch:
/to @raymondfeng 
/cc @bajtos @fabien 

This PR uses the authorization hook added with [this strong-remoting PR](https://github.com/strongloop/strong-remoting/pull/212).

Connect strongloop/loopback#1363
Connect strongloop/loopback#1191

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -295,7 +295,8 @@ app.enableAuth = function() {
   var remotes = this.remotes();
   var app = this;
 
-  remotes.before('**', function(ctx, next, method) {
+  remotes.authorization = function(ctx, next) {
+    var method = ctx.method;
     var req = ctx.req;
     var Model = method.ctor;
     var modelInstance = ctx.instance;
@@ -354,7 +355,7 @@ app.enableAuth = function() {
     } else {
       next();
     }
-  });
+  };
 
   this.isAuthEnabled = true;
 };
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -356,3 +356,42 @@ describe('security ACLs', function() {
     });
   });
 });
+
+describe('access check', function() {
+  var app;
+  before(function() {
+    app = loopback();
+    app.use(loopback.rest());
+    app.enableAuth();
+    app.dataSource('test', {connector: 'memory'});
+  });
+
+  it('should occur before other remote hooks', function(done) {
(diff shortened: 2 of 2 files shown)


lib/application.js
```diff
@@ -295,7 +295,8 @@ app.enableAuth = function() {
   var remotes = this.remotes();
   var app = this;
 
-  remotes.before('**', function(ctx, next, method) {
+  remotes.authorization = function(ctx, next) {
```
bajtos: Should we save the original `authorization` implementation and call after loopback auth passed? I think that's the canonical way of overriding methods like this.

```js
var baseAuthorization = remotes.authorization;
remotes.authorization = function(ctx, next) {
  if (authorized) {
    baseAuthorization(ctx, next); // or perhaps .call(this, ctx, next)
  } else {
    next(new Error('Not authorized.'));
  }
};
```

lib/application.js
```diff
@@ -354,7 +355,7 @@ app.enableAuth = function() {
     } else {
       next();
     }
-  });
+  }
```
bajtos: Missing semicolon.

test/acl.test.js
```diff
@@ -356,3 +356,42 @@ describe('security ACLs', function() {
...
+    app = loopback();
+    app.use(loopback.rest());
+    app.enableAuth();
+    app.dataSource('test', {connector: 'memory'});
+  });
+
+  it('should occur before other remote hooks', function(done) {
+    var MyTestModel = app.model('MyTestModel', {base: 'PersistedModel', dataSource: 'test'});
+    var checkAccessCalled = false;
+    var beforeHookCalled = false;
```
bajtos: Please consider using an array of strings instead of boolean properties. Each hook pushes a string name, and then the assertion can verify the order of execution.

```js
var called = [];

MyTestModel.checkAccess = function() {
  called.push('checkAccess');
  // ...
}

// ...

expect(called).to.eql(['checkAccess', 'beforeHook']);
```

bajtos:
The CI build is red, please fix it.

test/acl.test.js
ritch: Cool idea...

ritch:
> The CI build is red, please fix it.

Most likely due to the missing `authorization` hook (since it is not merged yet).

ritch:
test please

raymondfeng:
LGTM.

Issue #1059, opened 2015-02-02
Testing best practices / guide / docs


ritch:
We should provide some guidance on how to test loopback applications. `loopback-testing` is really not well designed or documented for external usage.

/cc @bajtos @crandmck

jolafrite:
Hi
There is no +1 button on Github but I definitely would be interested by this feature
Is there a place where I can start to have a look ?

OwenBrotherwood:
@crandmck I think this is important: Google Docs with use of r/w access and comments

OwenBrotherwood:
@crandmck What about some short mention of tests in https://github.com/strongloop/loopback/blob/8d851e8b5f574a16ee480251b1d5b5fc2d989913/CONTRIBUTING.md

crandmck:
@rmg what do you think?

OwenBrotherwood:
(Yep: I noted it was @rmg that was last on the CONTRIBUTING.md: so I agree with anything he says .... almost)

rmg:
As far as I can tell, this issue is about guidelines for testing loopback _applications_, not writing tests for loopback _contributions_.

@crandmck @OwenBrotherwood I'm not sure what you are asking for my opinion on.

OwenBrotherwood:
@rmg : yep, I know
@crandmck fills in the details as to "why are we 'here' now"

crandmck:
OK, I think the two issues of testing applications and testing contributions have gotten somewhat conflated here. 

@rmg you are right that this issue was for the former, and the above query seems to be about the latter.  Since I had nothing to do with `CONTRIBUTING.md`, I just punted it to you.  Perhaps it makes sense to suggest that any contributions to LoopBack have corresponding tests--I don't know.  That's the gist, and perhaps it does deserve a separate issue, but I will leave it to others if so.

bajtos:
> We should provide some guidance on how to test loopback applications. loopback-testing is really not well designed or documented for external usage.

My two cents: developers should use `supertest` together with whatever testing framework they prefer (mocha, tap, etc.) If they find data-builders helpful, then they can use that part of loopback-testing too. For everything else, they should write their own helpers specific to the application they are developing.

crandmck:
The `loopback-testing` repo has been removed.  If there is still a desire to cover testing best practices in docs, lets open a separate issue.

PR #1388, opened 2015-05-13
Add test case to highlight fatal error when trying to include a scope

mrfelton:
As noted in strongloop/loopback-datasource-juggler#589 when trying to include a relationship through a polymorphic relationship as part of an include filter, the server crashes with a fatal error.

This patch highlights the issue.

The test titled 'includes related models scoped to the related parent model' fails.

### DIFF
--- a/test/relations.integration.js
+++ b/test/relations.integration.js
@@ -67,9 +67,9 @@ describe('relations - integration', function() {
       var test = this;
       app.models.Group.create({ name: 'Group 1' },
         function(err, group) {
-        if (err) return done(err);
-        test.group = group;
-        app.models.Reader.create({ name: 'Reader 1' },
+          if (err) return done(err);
+          test.group = group;
+          app.models.Reader.create({ name: 'Reader 1' },
           function(err, reader) {
             if (err) return done(err);
             test.reader = reader;
@@ -79,7 +79,7 @@ describe('relations - integration', function() {
             reader.save(done);
           });
         }
-      )
+      );
     });
 
     after(function(done) {
--- a/test/relations.integration.js
+++ b/test/relations.integration.js
@@ -27,6 +27,109 @@ describe('relations - integration', function() {
     this.app.models.widget.destroyAll(done);
   });
 
+  describe.only('polymorphicHasMany', function() {
+
+    before(function defineProductAndCategoryModels() {
+      var Group = app.model(
+        'Group',
+        { properties: { name: 'string' },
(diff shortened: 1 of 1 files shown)


slnode:

OwenBrotherwood:
The tests are failining due to wrong indention,
Then the tests fail with:
```
 2 passing (110ms)
  1 failing

  1) relations - integration polymorphicHasMany includes related models scoped to the related parent model:
     Uncaught AssertionError: expected undefined to deeply equal { name: 'Group 1', id: 1 }
      at Test.<anonymous> (c:\Users\ob_000\mrfelton\loopback\test\relations.integration.js:126:53)
      at net.js:1392:10
```

Patch for indention:
```
diff --git a/test/relations.integration.js b/test/relations.integration.js
index a0062c8..6ffcd69 100644
--- a/test/relations.integration.js
+++ b/test/relations.integration.js
@@ -67,9 +67,9 @@ describe('relations - integration', function() {
       var test = this;
       app.models.Group.create({ name: 'Group 1' },
         function(err, group) {
-        if (err) return done(err);
-        test.group = group;
-        app.models.Reader.create({ name: 'Reader 1' },
+          test.group = group;
+          app.models.Reader.create({ name: 'Reader 1' },
           function(err, reader) {
             if (err) return done(err);
             test.reader = reader;
@@ -79,7 +79,7 @@ describe('relations - integration', function() {
             reader.save(done);
           });
         }
-      )
+      );
     });
 
     after(function(done) {

```

mrfelton:
Thanks @OwenBrotherwood - I was just running the `mocha` command to run the tests. Thats fixed up now, so the tests are properly failing due to the actual bug in loopback when running with `npm test`.

OwenBrotherwood:
@mrfelton just @ me if anything else is needed

----
@crandmck  I think there may be a doc problem to point out that npm test should be used before making a pull request.

crandmck:
@OwenBrotherwood I would love to add some information on testing, but I need the raw material.  
Would you be interested in contributing something?
BTW, there is a related issue https://github.com/strongloop/loopback/issues/1059 .

bajtos:
@raymondfeng I am assigning this PR for you to review since I think you are most familiar with this part of juggler. Feel free to reassign if there is a more suitable person to review the change.

slnode:

raymondfeng:
@mrfelton I fixed the issue in your test case. Thanks!

PR #1390, opened 2015-05-13
Initial relation auth implementation

ritch:
/to @bajtos 
/cc @raymondfeng 

#### WIP: DO NOT MERGE

This could be considered a controversial change. It changes the behavior for access control of the dynamically defined relation remote methods.

Here is a basic example:

 - `/customers/1/reviews?include=foo` => `Customers.prototype.__get__reviews()`
 - `checkAccess()` for customer with id 1
 - `checkAccess()` for `Reviews.find()`
 - `checkAccess()` for `Foo.find()`

This same concept is applied to all other relation methods.

Where I am a bit concerned about this change... what if you don't want to check access in the way mentioned above? 

....well technically you can modify the auth hook (or just remove it) like this

```js
var sharedMethod = Customer.sharedClass.find('__get__reviews');
sharedMethod.authorization = function(ctx, done) {
  // same signature as the `Model.checkAccess()` callback
  done(null, true); // allow all
};

// or extend the built in logic
var defaultCheck = sharedMethod.authorization;

sharedMethod.authorization = function myAuthorizationExtension(ctx, done) {
  if(ctx.args.myArg == 'someValue') {
    return done(null, true);
  }

  defaultCheck(ctx, done);
}

// or just disable the hook
delete sharedMethod.authorization;
```

@bajtos do you share the concern above? I don't want to break existing apps unless it is in a helpful way (eg. patching a security gap). Also can you look at the implementation and suggest how to test it or any major changes? I am going to actually start over since I am not entirely happy with how it is implemented (very duplicative).

**Alternative Impl**

I'm thinking of passing all the hooks through `Model.checkRelationAccess` which would allow you to change the way the check is performed just by overriding that function.

Connect strongloop/loopback#1362

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -301,18 +301,16 @@ app.enableAuth = function() {
     var Model = method.ctor;
     var modelInstance = ctx.instance;
 
-    var modelId = modelInstance && modelInstance.id ||
-      // replacement for deprecated req.param()
-      (req.params && req.params.id !== undefined ? req.params.id :
-       req.body && req.body.id !== undefined ? req.body.id :
-       req.query && req.query.id !== undefined ? req.query.id :
-       undefined);
+    var modelId = modelInstance && modelInstance.id || ctx.args.id;
+    ctx.modelId = modelId;
 
     var modelName = Model.modelName;
 
     var modelSettings = Model.settings || {};
     var errStatusCode = modelSettings.aclErrorStatus || app.get('aclErrorStatus') || 401;
-    if (!req.accessToken) {
+    if (req.accessToken) {
+      ctx.accessToken = req.accessToken;
+    } else {
       errStatusCode = 401;
     }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -5,6 +5,8 @@ var assert = require('assert');
 var RemoteObjects = require('strong-remoting');
 var SharedClass = require('strong-remoting').SharedClass;
 var extend = require('util')._extend;
+var normalizeInclude = require('loopback-datasource-juggler/lib/include').normalizeInclude;
+var async = require('async');
 
 module.exports = function(registry) {
 
@@ -289,9 +291,10 @@ module.exports = function(registry) {
    */
 
   Model.checkAccess = function(token, modelId, sharedMethod, ctx, callback) {
+    var Model = this;
+    var aclModel = Model._ACL();
     var ANONYMOUS = registry.getModel('AccessToken').ANONYMOUS;
     token = token || ANONYMOUS;
-    var aclModel = Model._ACL();
 
     ctx = ctx || {};
     if (typeof ctx === 'function' && callback === undefined) {
@@ -310,7 +313,11 @@ module.exports = function(registry) {
       remotingContext: ctx
     }, function(err, accessRequest) {
       if (err) return callback(err);
-      callback(null, accessRequest.isAllowed());
+      if(typeof sharedMethod.authorization === 'function') {
+        sharedMethod.authorization(ctx, accessRequest, callback);
+      } else {
+        callback(null, accessRequest.isAllowed());
+      }
     });
   };
 
@@ -436,6 +443,7 @@ module.exports = function(registry) {
     var modelName = relation.modelTo && relation.modelTo.modelName;
     modelName = modelName || 'PersistedModel';
     var fn = this.prototype[relationName];
+    var registry = this.registry;
     var pathName = (relation.options.http && relation.options.http.path) || relationName;
     define('__get__' + relationName, {
       isStatic: false,
@@ -443,7 +451,11 @@ module.exports = function(registry) {
       accepts: {arg: 'refresh', type: 'boolean', http: {source: 'query'}},
       accessType: 'READ',
       description: 'Fetches belongsTo relation ' + relationName + '.',
-      returns: {arg: relationName, type: modelName, root: true}
+      returns: {arg: relationName, type: modelName, root: true},
+      authorization: function(ctx, done) {
+        var targetSharedMethod = relation.modelTo.sharedClass.find('findOne', true);
+        relation.modelTo.checkAccess(ctx.accessToken, ctx.modelId, targetSharedMethod, ctx, done);
+      }
     }, fn);
   };
 
@@ -461,6 +473,9 @@ module.exports = function(registry) {
   Model.hasOneRemoting = function(relationName, relation, define) {
     var pathName = (relation.options.http && relation.options.http.path) || relationName;
     var toModelName = relation.modelTo.modelName;
+    var Model = this;
+    var registry = Model.registry;
+    var TargetModel = registry.getModel(toModelName);
 
     define('__get__' + relationName, {
       isStatic: false,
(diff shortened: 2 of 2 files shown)


lib/application.js
```diff
@@ -301,18 +301,16 @@ app.enableAuth = function() {
...
+    var modelId = modelInstance && modelInstance.id || ctx.args.id;
+    ctx.modelId = modelId;
 
     var modelName = Model.modelName;
 
     var modelSettings = Model.settings || {};
     var errStatusCode = modelSettings.aclErrorStatus || app.get('aclErrorStatus') || 401;
-    if (!req.accessToken) {
+    if (req.accessToken) {
+      ctx.accessToken = req.accessToken;
```
ritch: This doesn't seem like the right place for this, but it is required to be able to just pass the `ctx` to the auth hook.

lib/model.js
```diff
@@ -629,7 +687,24 @@ module.exports = function(registry) {
...
-      returns: {arg: scopeName, type: [toModelName], root: true}
+      returns: {arg: scopeName, type: [toModelName], root: true},
+      authorization: function(ctx, done) {
+        var modelsToCheck = [toModelName];
+        var include = ctx.args.filter && ctx.args.filter.include;
+        if (include) {
+          modelsToCheck = modelsToCheck.concat(normalizeInclude(include));
+        }
+
+        async.map(modelsToCheck, function(modelName, cb) {
```
ritch: @bajtos @raymondfeng do you see any issues with running the checks in parallel?

bajtos:
>  do you share the concern above?

Yes, I agree we must be very careful here. I think it's very likely that there are projects which actually depend on the current behaviour. Can we hide this new behaviour behind a "compat" flag, which would be disabled by default, but enabled in the project template in loopback-workspace?

lib/application.js
```diff
@@ -301,18 +301,16 @@ app.enableAuth = function() {
     var Model = method.ctor;
     var modelInstance = ctx.instance;
 
-    var modelId = modelInstance && modelInstance.id ||
-      // replacement for deprecated req.param()
-      (req.params && req.params.id !== undefined ? req.params.id :
-       req.body && req.body.id !== undefined ? req.body.id :
-       req.query && req.query.id !== undefined ? req.query.id :
-       undefined);
+    var modelId = modelInstance && modelInstance.id || ctx.args.id;
```
bajtos: I don't remember how exactly does strong-remoting handle arguments of prototype method. I think `ctx.args.id` is set for the shared constructor only, the method itself does not necessarily have to have an `id` argument.

bajtos: ... In which case this may be a breaking change.

lib/application.js
bajtos: Well, is it necessary for the auth hook to share the `ctx` object with strong-remoting? I mean you don't have to pass strong-remoting context to the auth hook, in my opinion you are free to create your own auth-hook-specific context.

lib/model.js
```diff
@@ -5,6 +5,8 @@ var assert = require('assert');
 var RemoteObjects = require('strong-remoting');
 var SharedClass = require('strong-remoting').SharedClass;
 var extend = require('util')._extend;
+var normalizeInclude = require('loopback-datasource-juggler/lib/include').normalizeInclude;
```
bajtos: I think it's a bad practice to load deep internals of a dependency. Can you modify juggler to export `normalizeInclude` in such way, that you can load it here via `require('loopback-datasource-juggler').normalizeInclude` or something similar?

lib/model.js
```diff
@@ -443,7 +451,11 @@ module.exports = function(registry) {
       accepts: {arg: 'refresh', type: 'boolean', http: {source: 'query'}},
       accessType: 'READ',
       description: 'Fetches belongsTo relation ' + relationName + '.',
-      returns: {arg: relationName, type: modelName, root: true}
+      returns: {arg: relationName, type: modelName, root: true},
+      authorization: function(ctx, done) {
+        var targetSharedMethod = relation.modelTo.sharedClass.find('findOne', true);
+        relation.modelTo.checkAccess(ctx.accessToken, ctx.modelId, targetSharedMethod, ctx, done);
```
bajtos: I think you need to call `accessRequest.isAllowed()` too, otherwise the statement from PR description is not accurate, in particular "checkAccess() for customer with id 1" is not called. Unless I am mistaken?

lib/model.js
bajtos: I think it's ok to run the checks in parallel.

bajtos:
> Also can you look at the implementation and suggest how to test it or any major changes?

In general, it would be great to write several tests for each relation method and check that

 1. an authorized user can access the method
 2. an unauthorized user cannot access the method

Preferably one test for each cause of "unauthorized user". 

Using your basic example outlined in the PR description, there are three permissions being checked:
 - (1) checkAccess() for customer with id 1
 - (2) checkAccess() for Reviews.find()
 - (3) checkAccess() for Foo.find()

So you need four tests for this particular relation method:
 1. user is allowed to do (1), (2), (3) -> requests is allowed
 2. user is allowed (1), (2), but not (3) -> 403
 3. user is allowed (1),(3) but not (2) -> 403
 4. user is allowed (2),(3) but not (1) -> 403

I think the important part is to come up with a good DSL that will allow you to write this tests with minimum amount of duplicated code. Perhaps something similar to  what we have in [test/access-control.integration.js](https://github.com/strongloop/loopback/blob/master/test/access-control.integration.js). Or you can invent a new language, for example:

```js
describe('hasMany relation methods', function() {
  describe('find', function() {
    it('allows users with all permissions', function(done) {
      requestWithPermissions({ Customer: 'READ', Review: 'READ', FOO: 'READ' })
        .get('/customers/1/reviews?include=foo')
        .expect(200).end(done);
    });
    it('rejects user without permission to read related model', function(done) {
      requestWithPermissions({ Customer: 'READ', Review: null, FOO: 'READ' })
         .get('/customers/1/reviews?include=foo')
         .expect(403).end(done);
    });
   // etc.
  });
});

function requestWithPermissions(map) {
  // create a new user
  // setup ACLs to give the user permissions from "map"
  // login the new user
  // setup supertest object with Authorization header set to user's access token
}
```

Or you can even push this further:

```js
describe('hasMany relation methods', function() {
  describe('find', function() {
    describe('/customers/1/reviews?include=foo', function() {
      verifyAccessCheck({ Customer: 'READ', Review: 'READ', FOO: 'READ' }, 200);
      verifyAccessCheck({ Customer: 'READ', Review: null, FOO: 'READ' }, 403);
    });
  });
});
```

lib/application.js
ritch: Hmmm... thats right! I can add in the sharedCtor args to the remote context but I'm not sure that is useful. But I'm not happy with the current mess of param checking (which is http specific for no reason).

lib/model.js
ritch: Sure, its just annoying to create such a small PR.

lib/model.js
ritch: Ah this is right...

ritch:
The testing comment is really useful thanks @bajtos

fabien:
@ritch looking forward to this - are there any remaining issues with this?

ritch:
Yes... We need to re-work this once https://github.com/strongloop/loopback/issues/1495 is implemented as a core feature.

zanemcca:
Is there an update on this?

PR #1416, opened 2015-05-29
Config option: (do not) search default token keys

bajtos:
A follow-up for abandoned #1347.

Add a new option `searchDefaultTokenKeys` to the token middleware. Users can set this option to false to prevent AccessToken from checking default places like "access_token" in the request query string.

Connect to #1326

/to @ritch please review
/cc @OwenBrotherwood

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -168,9 +168,12 @@ module.exports = function(AccessToken) {
     var length;
     var id;
 
-    params = params.concat(['access_token']);
-    headers = headers.concat(['X-Access-Token', 'authorization']);
-    cookies = cookies.concat(['access_token', 'authorization']);
+    // https://github.com/strongloop/loopback/issues/1326
+    if (options.searchDefaultTokenKeys !== false) {
+      params = params.concat(['access_token']);
+      headers = headers.concat(['X-Access-Token', 'authorization']);
+      cookies = cookies.concat(['access_token', 'authorization']);
+    }
 
     for (length = params.length; i < length; i++) {
       var param = params[i];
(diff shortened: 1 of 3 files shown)


OwenBrotherwood:
@bajtos Some Javadoc may be needed
https://github.com/strongloop/loopback/blob/9cd1b96c31c5868f085f2b6eb1b73540c06e253e/server/middleware/token.js?#L64

bajtos:
>  Some Javadoc may be needed
https://github.com/strongloop/loopback/blob/9cd1b96c31c5868f085f2b6eb1b73540c06e253e/server/middleware/token.js?#L64

@OwenBrotherwood good catch, it should be fixed now in 86ed472

bajtos:
@ritch I forgot to assign this to you in GH, could you PTAL?

ritch:
Can you make sure the ci is passing? Otherwise the code LGTM.

bajtos:
@slnode test please

bajtos:
> Can you make sure the ci is passing? Otherwise the code LGTM.

This looks like a problem with downstream dependencies. LoopBack's master fails on our CI build and passes on Travis. This patch passes on Travis and fails on Jenkins. The error messages are the same AFAICT:

```
// beforeEach hook - access control - integration /users when called by logged in user
TypeError: Cannot read property 'constructor' of undefined
// afterEach hook - access control - integration /users when called by logged in user
Cannot call method 'destroy' of undefined
TypeError: Cannot call method 'destroy' of undefined
```

Was there any breaking change in loopback-testing that would explain this problem?

The other test error does not seem to be related to this patch either:

```
// Change Change.diff(modelName, since, remoteChanges, callback) 
//    should return delta and conflict lists
AssertionError: 4 == 1
    at test/change.test.js:354:16
    at common/models/change.js:445:7
```

I am proposing to land this PR now and investigate the failing build later.

@ritch Thoughts?

ritch:
> Was there any breaking change in loopback-testing that would explain this problem?

Yes. LoopBack should use `loopback-testing@~1.1.0`.

ritch:
> I am proposing to land this PR now and investigate the failing build later.

Sounds good.

Issue #1437, opened 2015-06-08
server break down if password is in numeric form on reset password


dhirender-tyagi:
server get break down if i give password as numeric in reset password case because it is handled as number so number.indexOf() get break

BerkeleyTrue:
This method is also causing us huge issues when used in combination with loopback-component-passport. We had to revert our production servers back to our old setup after a surge of (crashing) errors in this method from our live users using oauth.

Seems like it tries to set password when oauth is being used. Since no password is given and no password is required in this situation why isn't there a check here for existence and type of string?

@ritch Would you be open to a PR. This is blocking for us from moving forward with loopback.

something like this:

```js
UserModel.setter.password = function(plain) {
      if (!plain) {
         return;
      }

      // for the issue above
      plain = '' + plain;
      if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
        // The password is already hashed. It can be the case
        // when the instance is loaded from DB
        this.$password = plain;
      } else {
        this.$password = this.constructor.hashPassword(plain);
      }
};
```


cc @quincylarson @terakilobyte

![keymetrics___app](https://cloud.githubusercontent.com/assets/6775919/8185917/e8281cb4-13fb-11e5-9a76-b5761a4ac98f.png)

PR #1438, opened 2015-06-08
change password to string if it is detected as number

dhirender-tyagi:
server get break down if i give password as numeric in reset password case because it is handled as number so number.indexOf() get break

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -544,6 +544,10 @@ module.exports = function(User) {
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
     UserModel.setter.password = function(plain) {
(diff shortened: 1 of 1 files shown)


slnode:

ritch:
@dhirender-tyagi can you sign the CLA? Otherwise this LGTM and can be merged.

dhirender-tyagi:
@ritch I have signed the CLA.

dhirenderbp:
Is there is any feedback that does't received

dhirender-tyagi:
Is there any feedback that is doesn't get

0candy:
Already fixed by https://github.com/strongloop/loopback/issues/1437

Issue #1444, opened 2015-06-10
New Method: PersistedModel.subscribe


ritch:
This new method should callback with a readable (object mode) stream flowing with `Change` objects. A `Change` object should be added to the stream when a change is made to the `PersistedModel` that matches the provided `filter`.

It should be made remote-able using this: https://github.com/strongloop/strong-remoting/issues/214

```js
// Score is a PersistedModel
Score.subscribe({
  where: {
    views: {gt: 100}
  }
}, function(err, changes) {
  changes.on('data', function(change) {
    console.log('score changed', change.data);
  });
});
```

The source of the `Change` stream is configurable.

 - local
   - only uses hooks to observe changes on the local model
 - pubsub
   - subscribes to a default topic (model name) or a configurable topic
   - must provide a data source name for connecting to a broker

ritch:
This approach has been re-worked: see the new issue https://github.com/strongloop/loopback/issues/1419

BerkeleyTrue:
Wow cool! @ritch what are your thoughts on following the observable pattern established by RxJS instead? It could be made extendable to include composable methods similar in fashion to RxJS and it's already a well established pattern.

ritch:
> what are your thoughts on following the observable pattern established by RxJS instead?

This is actually a great idea. The reason I am using the `Stream` interface is purely because its compose-ability in node land. Eg. if we support `Stream`s in strong-remoting you can do things like upload and download files like this:

```js
// streaming upload
function handleUpload(function(path) {
  cb(null, fs.createWriteStream(path));
}

// streaming download
function download(path) {
  cb(null, fs.createReadStream(path));
}
```

I think the important point is that `Stream`s are not for reactive programming. They are for moving data and in a way that doesn't overload consumers (and handling other tricky cases).

It probably makes sense to only support streams at the `strong-remoting` layer. In your application code it would be nice to have an rxjs compatible api. Implementing this is fairly similar to supporting both callbacks and promises.

I think the advantage of using rxjs is the compatibility / compose-ability, so it doesn't make sense to force users to use that style API if they don't intend to use it in the rest of their application.   

What if we have a low-level api in loopback core (like what I've outlined here) and a wrapper library that provides an rxjs api on top of the low level stream+callback api?

[...]

PR #1445, opened 2015-06-11
Add PersistedModel.subscribe()

ritch:
Connect to strongloop/loopback#1444

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -7,6 +7,7 @@ var assert = require('assert');
 var async = require('async');
 var deprecated = require('depd')('loopback');
 var debug = require('debug')('loopback:persisted-model');
+var PassThrough = require('stream').PassThrough;
 
 module.exports = function(registry) {
   var Model = registry.getModel('Model');
@@ -761,6 +762,24 @@ module.exports = function(registry) {
         http: {verb: 'post', path: '/:id/rectify-change'}
       });
     }
+
+    setRemoting(PersistedModel, 'createChangeStream', {
+      description: 'Create a change stream.',
+      accessType: 'READ',
+      http: [
+        {verb: 'post', path: '/change-stream'},
+        {verb: 'get', path: '/change-stream'}
+      ],
+      accepts: {
+        arg: 'options',
+        type: 'object'
+      },
+      returns: {
+        arg: 'changes',
+        type: 'ReadableStream',
+        json: true
+      }
+    });
   };
 
   /**
@@ -1525,6 +1544,101 @@ module.exports = function(registry) {
 
       inst.updateAttributes(data, cb);
     });
+  };
+
+  /**
+   * Create a change stream.
+   *
+   * @param {Object} options
+   * @param {Object} options.where Only changes to models matching this where filter will be included in the `ChangeStream`.
+   * @callback {Function} callback
+   * @param {Error} err
+   * @param {ChangeStream} changes
+   */
+
+  PersistedModel.createChangeStream = function(options, cb) {
+    if (typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+
+    var idName = this.getIdName();
+    var Model = this;
+    var changes = new PassThrough({objectMode: true});
+    var writeable = true;
+
+    changes.destroy = function() {
+      changes.removeAllListeners('error');
+      changes.removeAllListeners('end');
+      writeable = false;
+      changes = null;
+    };
+
+    changes.on('error', function() {
+      writeable = false;
+    });
+    changes.on('end', function() {
+      writeable = false;
+    });
+
+    process.nextTick(function() {
+      cb(null, changes);
+    });
+
+    Model.observe('after save', createChangeHandler('save'));
+    Model.observe('after delete', createChangeHandler('delete'));
+
+    function createChangeHandler(type) {
+      return function(ctx, next) {
+        // since it might have set to null via destroy
+        if (!changes) {
+          return next();
+        }
+
+        var where = ctx.where;
+        var data = ctx.instance || ctx.data;
+        var whereId = where && where[idName];
+
+        // the data includes the id
+        // or the where includes the id
+        var target;
+
+        if (data && (data[idName] || data[idName] === 0)) {
+          target = data[idName];
+        } else if (where && (where[idName] || where[idName] === 0)) {
+          target = where[idName];
+        }
+
+        var hasTarget = target === 0 || !!target;
+
+        var change = {
+          target: target,
+          where: where,
+          data: data
+        };
+
+        switch (type) {
+          case 'save':
+            if (ctx.isNewInstance === undefined) {
+              change.type = hasTarget ? 'update' : 'create';
+            } else {
+              change.type = ctx.isNewInstance ? 'create' : 'update';
+            }
+
+            break;
+          case 'delete':
+            change.type = 'remove';
+            break;
+        }
+
+        // TODO(ritch) this is ugly... maybe a ReadableStream would be better
+        if (writeable) {
+          changes.write(change);
+        }
+
+        next();
+      };
+    }
   };
 
   PersistedModel.setup();
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "karma-script-launcher": "^0.1.0",
     "loopback-boot": "^2.7.0",
     "loopback-datasource-juggler": "^2.19.1",
-    "loopback-testing": "^1.1.0",
+    "loopback-testing": "~1.1.0",
     "mocha": "^2.1.0",
     "sinon": "^1.13.0",
     "strong-task-emitter": "^0.0.6",
--- a/test/change-stream.test.js
+++ b/test/change-stream.test.js
@@ -1,0 +1,87 @@
+describe('PersistedModel.createChangeStream()', function() {
+  describe('configured to source changes locally', function() {
+    before(function() {
+      var test = this;
+      var app = loopback({localRegistry: true});
+      var ds = app.dataSource('ds', {connector: 'memory'});
+      this.Score = app.model('Score', {
+        dataSource: 'ds',
+        changeDataSource: false // use only local observers
+      });
+    });
+
+    it('should detect create', function(done) {
+      var Score = this.Score;
+
+      Score.createChangeStream(function(err, changes) {
+        changes.on('data', function(change) {
+          expect(change.type).to.equal('create');
+          changes.destroy();
+          done();
+        });
+
+        Score.create({team: 'foo'});
+      });
+    });
+
+    it('should detect update', function(done) {
+      var Score = this.Score;
+      Score.create({team: 'foo'}, function(err, newScore) {
+        Score.createChangeStream(function(err, changes) {
+          changes.on('data', function(change) {
+            expect(change.type).to.equal('update');
+            changes.destroy();
+            done();
+          });
+          newScore.updateAttributes({
+            bat: 'baz'
+          });
+        });
+      });
+    });
+
+    it('should detect delete', function(done) {
+      var Score = this.Score;
+      Score.create({team: 'foo'}, function(err, newScore) {
+        Score.createChangeStream(function(err, changes) {
+          changes.on('data', function(change) {
+            expect(change.type).to.equal('remove');
+            changes.destroy();
+            done();
+          });
+
+          newScore.remove();
+        });
+      });
+    });
+  });
+
+  // TODO(ritch) implement multi-server support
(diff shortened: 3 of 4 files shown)


lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+          return next();
+        }
+
+        // we don't support optimistic create yet
+        if(ctx.isNewInstance && optimistic) {
+          return next();
+        }
+
+        var where = ctx.where;
+        var data = ctx.instance || ctx.data;
```
bajtos: `ctx.instance.toJSON() || ctx.data`?

lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+        var change = {
+          target: target,
+          where: where,
+          optimistic: optimistic,
+          data: data
+        };
+
+        switch(type) {
+          case 'save':
+            change.type = ctx.isNewInstance
```
bajtos: Note that `ctx.isNewInstance` has three possible values: `true`, `false` and `undefined`. The last one is used when we cannot tell whether the operation is update or create. Examples: "before save" hook for updateOrCreate; "after save" hook after updateOrCreate when the connector does not report the operation it performed.

bajtos:
As mentioned in the slack chat, I can't review this thoroughly right now. I did a quick review and didn't find any major issues, I think it's ok to merge this as it is and refine the implementation later as needed.

BerkeleyTrue:
This method is being renamed, correct?

lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+    changes.on('end', function() {
+      writeable = false;
+    });
+
+    process.nextTick(function() {
+      cb(null, changes);
+    });
+
+    Model.observe('before save', createChangeHandler('save', true));
+    Model.observe('before delete', createChangeHandler('delete', true));
```
bajtos: In my opinion it does not make much sense to send optimistic changes from "before" hooks.
 - When running on the server, the latency of database calls is usually negligible, thus an "optimistic" update does provide much value IMO.
 - It would make sense to emit optimistic changes when running in the browser, but IIRC, the isomorphic client does not invoke operation hooks because the remote connector completely bypasses DAO methods.

I think we need to find a different solution for emitting optimistic updates.

lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+    Model.observe('after save', createChangeHandler('save'));
+    Model.observe('after delete', createChangeHandler('delete'));
+
+    function createChangeHandler(type, optimistic) {
+      return function(ctx, next) {
+        // note: this leaks the ref and writeable variables
+        // but not the entire change stream (with its huge buffer)
```
bajtos: Are you sure about this? I believe that references are transitive, thus if A holds a reference to REF and REF holds reference to CHANGES, then it's the same as if A held a reference to CHANGES.

Perhaps @bnoordhuis or @piscisaureus can provide more insights here?

lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+
+        var where = ctx.where;
+        var data = ctx.instance || ctx.data;
+        var whereId = where && where[idName];
+
+        // the data includes the id
+        // or the where includes the id
+        var target =
+          (data && data[idName]) ||
+          (where && where[idName]);
```
bajtos: This does not work correctly for `data.id=0`.

test/subscribe.test.js
```diff
@@ -0,0 +1,89 @@
+describe('Subscribe', function() {
+  describe('PersistedModel', function() {
+    describe('configured to source changes locally', function() {
+      before(function() {
+        var test = this;
+        var app = loopback({localRegistry: true});
+        var ds = app.dataSource('ds', {connector: 'memory'});
+        this.Score = app.model('Score', {
+          dataSource: 'ds',
+          changeDataSource: false // use only local observers
```
bajtos: Could you please point me to the code which handles `changeDataSource`? I'd like to understand better the consequences of setting this flag to `false`.

Perhaps you should add unit-tests for `changeDataSource:true` too?

test/subscribe.test.js
```diff
@@ -0,0 +1,89 @@
...
+        });
+      });
+
+      it('should detect create', function(done) {
+        var Score = this.Score;
+
+        Score.subscribe(function(err, changes) {
+          changes.on('data', function(change) {
+            expect(change.type).to.equal('create');
+            changes.destroy();
```
bajtos: `changes.destroy` is not called in case the assertion fails (throws an error). I think you should call the destroy method first, before making any assertions. Please fix this in the other unit-tests too.

lib/persisted-model.js
bnoordhuis: `ref.changes` is the stream with the big buffer?  This closure keeps it alive, yes.

ritch: @bnoordhuis even with `delete ref.changes` (see the `destroy()` method)? How...? Sure the `ref` object will never be garbage collected since its being referenced by the hook function but `ref.changes` should be... No?

bajtos: I think the stream should be de-referenced after `delete ref.changes` and eventually GC-ed too. But then isn't `delete ref.changes` equivalent to setting a local variable to null `changes = null`?

```diff
var changes = new PassThrough({objectMode: true});
- var ref = { changes: changes }
- // work with ref.changes
- delete ref.changes
+ // work with changes directly
+ changes = null
```

ritch: > thus if A holds a reference to REF and REF holds reference to CHANGES

How is it holding reference after `delete ref.changes`? I am aware that A will hold a reference via a closure variable to REF since the hook code is still referencing / able to invoke A (the hook function).

bajtos: I think the stream should be de-referenced after `delete ref.changes` and eventually GC-ed too, so your thinking should be correct @ritch.

But then isn't `delete ref.changes` equivalent to setting a local variable to null `changes = null`?

```diff
var changes = new PassThrough({objectMode: true});
- var ref = { changes: changes }
- // work with ref.changes
- delete ref.changes
+ // work with changes directly
+ changes = null
```

test/subscribe.test.js
ritch: That is placeholder for allowing you to specify a remote broker (eg. redis or mosquitto) that allows you to subscribe to changes originating from other servers.

We also might use it to subscribe to changes from a db op log.

lib/persisted-model.js
bajtos: Sorry @ritch if my comment was misleading. The point I wanted to make is that I think you don't need the `ref` variable and can use `changes` directly. I believe the scope references the variable, not the object it points to. Thus when you set `changes = null`, the stream won't be referenced any more.

ritch: > But then isn't delete ref.changes equivalent to setting a local variable to null changes = null?

Yeah this is true... There really isn't a need for a `ref` object at all.

test/subscribe.test.js
bajtos: Can we use a more specific name? Consider the existing change tracking & synchronization algorithm. I can imagine that we may want to attach model's Change object to a different datasource than the model (e.g. Car is in Oracle, Car-Changes is in MongoDB). In which case I would expect that the setting `changeDataSource` refers to the datasource used by the Change model.

Perhaps we can omit this placeholder altogether if there is no real implementation yet?

lib/persisted-model.js
```diff
@@ -1527,6 +1546,107 @@ module.exports = function(registry) {
...
+
+    var idName = this.getIdName();
+    var Model = this;
+    var changes = new PassThrough({objectMode: true});
+    var writeable = true;
+    var ref = {changes: changes};
+
+    changes.destroy = function() {
+      changes.removeAllListeners('error');
+      changes.removeAllListeners('end');
```
bajtos: It would be good to remove hook observers too.

I think that otherwise the list of registered hook observers will grow without limits, each DAO operation will have to invoke a lot of inactive hooks and the server will gradually slow down.

lib/persisted-model.js
ritch: @bajtos 

> When running on the server, the latency of database calls is usually negligible, thus an "optimistic" update does provide much value IMO.

The purpose of an optimistic update is to avoid latency entirely. We don't have to wait for the operation to be acknowledged by the database (or rest api, other datasource) since we are only telling the client about a likely change.

I think we are disagreeing on what latency is acceptable. IMO any IO is going to add significant latency. Ideally we would emit the optimistic change right after it is sanitized. Ideally it would be emitted before timely ACL checks, but this introduces obvious vulnerabilities.

In order to move this forward, I am OK with removing this code, as it can be added later as an optimization.

ritch:
> This method is being renamed, correct?

@BerkeleyTrue Yep... renamed to `createChangeStream()`

ritch:
@bajtos I'm seeing some errors in the phantomjs tests that appear to be unrelated to this change. I'm getting them on master as well. I'd like to fix those test in another PR. Any objections?

package.json
```diff
@@ -81,7 +81,7 @@
     "karma-script-launcher": "^0.1.0",
     "loopback-boot": "^2.7.0",
     "loopback-datasource-juggler": "^2.19.1",
-    "loopback-testing": "^1.1.0",
+    "loopback-testing": "~1.1.0",
```
bajtos: Is there any particular reason for locking down loopback-testing version to 1.1.x?

test/change-stream.test.js
```diff
@@ -0,0 +1,87 @@
+describe('PersistedModel.createChangeStream()', function() {
+  describe('configured to source changes locally', function() {
+    before(function() {
+      var test = this;
+      var app = loopback({localRegistry: true});
+      var ds = app.dataSource('ds', {connector: 'memory'});
+      this.Score = app.model('Score', {
+        dataSource: 'ds',
+        changeDataSource: false // use only local observers
```
bajtos: Can we use a more specific name instead of `changeDataSource`? See https://github.com/strongloop/loopback/pull/1445#discussion_r33587585

test/change-stream.test.js
```diff
@@ -0,0 +1,87 @@
...
+      });
+    });
+
+    it('should detect update', function(done) {
+      var Score = this.Score;
+      Score.create({team: 'foo'}, function(err, newScore) {
+        Score.createChangeStream(function(err, changes) {
+          changes.on('data', function(change) {
+            expect(change.type).to.equal('update');
+            changes.destroy();
```
bajtos: `changes.destroy` is not called in case the assertion fails (throws an error). I think you should call the destroy method first, before making any assertions. Please fix this in the other unit-tests too.

lib/persisted-model.js
```diff
@@ -1527,6 +1546,101 @@ module.exports = function(registry) {
...
+    }
+
+    var idName = this.getIdName();
+    var Model = this;
+    var changes = new PassThrough({objectMode: true});
+    var writeable = true;
+
+    changes.destroy = function() {
+      changes.removeAllListeners('error');
+      changes.removeAllListeners('end');
```
bajtos: It would be good to remove hook observers too.

I think that otherwise the list of registered hook observers will grow without limits, each DAO operation will have to invoke a lot of inactive hooks and the server will gradually slow down.

bajtos:
> I'm seeing some errors in the phantomjs tests that appear to be unrelated to this change. I'm getting them on master as well. I'd like to fix those test in another PR.

Sure, that's fine with me.

lib/persisted-model.js
bajtos: However, I am ok to implement this later (outside of this PR) as an additional optimization.

bajtos:
Some of my older comments were not addressed, I re-posted them on the updated diff. None of them is a blocker though, feel free to land this PR without addressing them.

For the sake of our future selves, it would be good to explain why do you need to lock the version of loopback-testing to `1.1.x` (see https://github.com/strongloop/loopback/pull/1445#discussion_r34245142).

ritch:
> For the sake of our future selves, it would be good to explain why do you need to lock the version of loopback-testing to 1.1.x (see #1445 (comment)).

1.2 introduces a bug.  https://github.com/strongloop/loopback-testing/issues/62

lib/persisted-model.js
ritch: Ok i'll create a follow up issue.

ritch:
test please

lib/persisted-model.js
fabien: @ritch any update on removal of these hook observers?

PR #1465, opened 2015-06-17
fix: exit early when password is non-string

### LINKED ISSUE #1437: server break down if password is in numeric form on reset password

BerkeleyTrue:
prevent errors when plain is non-string by using call on indexOf method
closes #1437

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -544,6 +544,9 @@ module.exports = function(User) {
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
     UserModel.setter.password = function(plain) {
+      if (typeof plain !== 'string') {
+        return;
+      }
       if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
         // The password is already hashed. It can be the case
         // when the instance is loaded from DB


BerkeleyTrue:
Any feedback on this?

common/models/user.js
```diff
@@ -544,7 +544,11 @@ module.exports = function(User) {
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
     UserModel.setter.password = function(plain) {
-      if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
+      var indexOf = String.prototype.indexOf;
+      if (!plain) {
+        return;
+      }
+      if (indexOf.call(plain, '$2a$') === 0 && plain.length === 60) {
```
raymondfeng: Why do you have to the reflection api instead of `plain.indexOf`?

BerkeleyTrue: #1437 states that if plain is type number indexOf breaks as it a method on string instances. Using indexOf.call keeps this from breaking on numbers as it will cast it to string first.

BerkeleyTrue:
Any way we can move this along? I really don't like the idea of having to support our own fork of loopback for lack of this PR.

BerkeleyTrue:
@raymondfeng I hate to ping you directly but it is important that this fix gets in. Our development servers using loopback have shown lots of promise, but the fact that there has been no traction on this issue scares me.

If there is anything I can do to help move this along please let me know.

common/models/user.js
BerkeleyTrue: I can remove the indexOf.call if it will help move things along.

The main thing I am worried about is users who have no passwords throwing and crashing.

The indexOf call was added as a bonus as address #1437

BerkeleyTrue:
ping @raymondfeng

common/models/user.js
raymondfeng: I would rather test if typeof plain === 'string'.

raymondfeng:
Sorry for the delay. It skipped out of my eyes. I added a comment.

common/models/user.js
BerkeleyTrue: like this?
```js
if (typeof plain !== 'string') {
  return;
}
```

BerkeleyTrue: @raymondfeng I don't think this really covers our use case. Most of our users coming in through github/twitter auth, they don't set passwords. We need to make sure users who do not have passwords do not crash the site just by logging in. 

I can get ride of the indexOf reflection and just set it to this:
```js
      if (!plain) {
        return;
      }
      if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
        // The password is already hashed. It can be the case
        // when the instance is loaded from DB
        this.$password = plain;
      } else {
      }
    };
```

This prevents crashes when plain is undefined, which is the use case I am trying to solve, but won't do anything for the case when plain is a number, which I don't really need right now. 

@raymondfeng How does that sound?

raymondfeng: The following test returns for null, undefined, and non-strings.
```
if (typeof plain !== 'string') {
  return;
}
```

BerkeleyTrue: Got it. Will do.

BerkeleyTrue:
@raymondfeng done

Issue #418, opened 2014-07-29
Support promises


coodoo:
LoopBack should provide Promise-enabled API out of the box. The plan is to use the approach outlined in https://github.com/iojs/io.js/pull/173.

**Tasks**
- [x] strongloop/strong-remoting#179 support promises in strong-remotin
- [x] strongloop/loopback-datasource-juggler#449 loopback-datasource-juggler - support promises in "operation hooks" 
- [x]  strongloop/loopback-datasource-juggler#451 and  strongloop/loopback-datasource-juggler#452 loopback-datasource-juggler - convert all DAO & relation methods to dual callback/promise API 
- [x] promisify juggler's DataSource methods like `automigrate`, `autoupdate`, `discover*`, etc.
  - https://github.com/strongloop/loopback-datasource-juggler/pull/630
- [ ] promisify juggler's relation and scope methods
  - https://github.com/strongloop/loopback/issues/2709
- [ ] promisify custom validators in juggler
 - strongloop/loopback-datasource-juggler#1229
- [ ] convert loopback builtin models to support dual callback/promise API
  - [x] User - https://github.com/strongloop/loopback/pull/1493
  - [x] Application - https://github.com/strongloop/loopback/pull/1533
  - [x] PersistedModel (replication, change tracking) - https://github.com/strongloop/loopback/pull/1536
  - [ ] Change
  - [ ] Checkpoint
  - [ ] Scope
  - [ ] Model.checkAccess()
  - [ ] AccessToken
  - [x] Role (and role resolvers)
  - [x] RoleMapping
  - [x] ACL
- [ ] Promisify `app.listen`
- [ ] Modify connectors and convert all mixin methods to dual callback/promise API
  - [ ] loopback-connector-remote
  - [ ] Email model & connector
- (?) Allow middleware handlers to return a promise
- [ ] Support promises in loopback-phase, both for API and for handler functions
- [ ] Update examples and docs
  - [x] Enhance strong-docs to support `@promise` annotation https://github.com/strongloop-internal/strong-docs/issues/63
  - [ ] Edit API docs / Mark all functions supporting promises with `@promise` attribute
    - [ ] [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler/issues/833)
    - [x] [loopback-core](https://github.com/strongloop/loopback/issues/1998)
    - [ ] [strong-remoting](https://github.com/strongloop/strong-remoting/issues/276)
[...]

PR #1493, opened 2015-07-01
Promisify User model

PradnyaBaviskar:
Connect https://github.com/strongloop/loopback/issues/418

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -3,6 +3,7 @@
  */
 
 var loopback = require('../../lib/loopback');
+var utils = require('../../lib/utils');
 var path = require('path');
 var SALT_WORK_FACTOR = 10;
 var crypto = require('crypto');
@@ -80,6 +81,9 @@ module.exports = function(User) {
       cb = options;
       options = undefined;
     }
+
+    cb = cb || utils.createPromiseCallback();
+
     if (typeof ttl === 'object' && !options) {
       // createAccessToken(options, cb)
       options = ttl;
@@ -91,6 +95,7 @@ module.exports = function(User) {
     this.accessTokens.create({
       ttl: ttl
     }, cb);
+    return cb.promise;
   };
 
   function splitPrincipal(name, realmDelimiter) {
@@ -168,6 +173,8 @@ module.exports = function(User) {
       include = undefined;
     }
 
+    fn = fn || utils.createPromiseCallback();
+
     include = (include || '');
     if (Array.isArray(include)) {
       include = include.map(function(val) {
@@ -191,13 +198,15 @@ module.exports = function(User) {
       var err1 = new Error('realm is required');
       err1.statusCode = 400;
       err1.code = 'REALM_REQUIRED';
-      return fn(err1);
+      fn(err1);
+      return fn.promise;
     }
     if (!query.email && !query.username) {
       var err2 = new Error('username or email is required');
       err2.statusCode = 400;
       err2.code = 'USERNAME_EMAIL_REQUIRED';
-      return fn(err2);
+      fn(err2);
+      return fn.promise;
     }
 
     self.findOne({where: query}, function(err, user) {
@@ -234,7 +243,7 @@ module.exports = function(User) {
               err = new Error('login failed as the email has not been verified');
               err.statusCode = 401;
               err.code = 'LOGIN_FAILED_EMAIL_NOT_VERIFIED';
-              return fn(err);
+              fn(err);
             } else {
               if (user.createAccessToken.length === 2) {
                 user.createAccessToken(credentials.ttl, tokenHandler);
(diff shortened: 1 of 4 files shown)


test/user.test.js
```diff
@@ -300,6 +338,20 @@ describe('User', function() {
...
+          assert(!accessToken);
+          done();
+        })
+        .catch(function(err) {
+          assert(err);
+          assert.equal(err.code, 'LOGIN_FAILED');
+          done();
+        });
+    });
+
```
PradnyaBaviskar: Rather than falling into `catch`, this `User.login` throws "Unhandled rejection Error: login failed". And so, when executing the test it says "Cannot call method 'then' of undefined". Whereas, https://github.com/strongloop/loopback/pull/1493/files#diff-5d81af95449574544e7317f3eda2badeR259, correctly returns `fn.promise`

test/user.test.js
```diff
@@ -441,6 +506,21 @@ describe('User', function() {
...
+      })
+      .catch(function(err) {
+        // strongloop/loopback#931
+        // error message should be "login failed" and not "login failed as the email has not been verified"
+        assert(err && !/verified/.test(err.message), ('expecting "login failed" error message, received: "' + err.message + '"'));
+        assert.equal(err.code, 'LOGIN_FAILED');
+        done();
+      });
+    });
+
```
PradnyaBaviskar: Same thing happens here. Hence skipped.

pulkitsinghal:
@PradnyaBaviskar It would seem that the loopback team intentionally limited the promisification to CRUD methods

> > @pulkitsinghal wrote: So the CRUD methods for models and relatedModels now use promises ... but something like User.login() still does not ... correct?

> @bajtos wrote: Yes, that's a correct description of the current status.

But that doesn't mean we shouldn't get the other ones also promisified, thanks for opening this, +1

common/models/user.js
```diff
@@ -80,6 +80,9 @@ module.exports = function(User) {
       cb = options;
       options = undefined;
     }
+
+    cb = cb || app.createPromiseCallback();
```
bajtos: Where is the `app` variable defined? Perhaps this is throwing an error that is later converted to the error message you are seeing?

bajtos: Oh, are you relying on the fact that `test/support.js` is setting up a global `app` variable?

common/models/user.js
```diff
@@ -222,11 +230,13 @@ module.exports = function(User) {
       if (err) {
         debug('An error is reported from User.findOne: %j', err);
         fn(defaultError);
+        return fn.promise;
```
bajtos: This is returning from `self.findOne` callback, not from `User.login`.

common/models/user.js
```diff
@@ -222,11 +230,13 @@ module.exports = function(User) {
...
       } else if (user) {
         user.hasPassword(credentials.password, function(err, isMatch) {
           if (err) {
             debug('An error is reported from User.hasPassword: %j', err);
             fn(defaultError);
+            return fn.promise;
```
bajtos: ditto

common/models/user.js
```diff
@@ -234,7 +244,8 @@ module.exports = function(User) {
               err = new Error('login failed as the email has not been verified');
               err.statusCode = 401;
               err.code = 'LOGIN_FAILED_EMAIL_NOT_VERIFIED';
-              return fn(err);
+              fn(err);
+              return fn.promise;
```
bajtos: ditto

common/models/user.js
```diff
@@ -245,11 +256,13 @@ module.exports = function(User) {
           } else {
             debug('The password is invalid for user %s', query.email || query.username);
             fn(defaultError);
+            return fn.promise;
```
bajtos: ditto

common/models/user.js
```diff
@@ -245,11 +256,13 @@ module.exports = function(User) {
...
           }
         });
       } else {
         debug('No matching record is found for user %s', query.email || query.username);
         fn(defaultError);
+        return fn.promise;
```
bajtos: ditto

common/models/user.js
```diff
@@ -245,11 +256,13 @@ module.exports = function(User) {
...
       }
     });
```
bajtos: Here is the place where to add `return fn.promise`.

```js
    });
    return fn.promise;
  };

  /**
   * Logout...
```

PradnyaBaviskar:
Done with promisifying User model. 

@raymondfeng, can you please review?
/cc @bajtos

bajtos:
@slnode test please

bajtos:

bajtos:
FWIW, the Jenkins build is failing due to an issue that was fixed on master by https://github.com/strongloop/loopback/commit/34eb0e1a4577739d5ec8221238dcb2b0c92a8f87

Issue #1489, opened 2015-06-26
Non-authenticated POSTing to /User/logout produces an error 500


doublemarked:
The User.logout code appears to do this intentionally. It feels like the wrong error response code, but I don't know what the correct code is. Here is the User.logout code, for reference,

```
  User.logout = function(tokenId, fn) {
    this.relations.accessTokens.modelTo.findById(tokenId, function(err, accessToken) {
      if (err) {
        fn(err);
      } else if (accessToken) {
        accessToken.destroy(fn);
      } else {
        fn(new Error('could not find accessToken'));
      }
    });
  };
```

Please advise if that is the intended functionality or not.

raymondfeng:
Maybe 400 is better. Do you want to submit a patch?

rus0000:
401. User.logout requires logged-in user. And as any other protected method it has to return 401.

doublemarked:
Perhaps the problem here is just that the ACL is wrong?

```
    {
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "ALLOW",
      "property": "logout"
    },
```

rus0000:
principalId has to be $authenticated

doublemarked:
@raymondfeng can you confirm that the ACL is off?  If so, I'm happy to put together a patch.

superkhau:
@raymondfeng Assigning to you since you are already working on this.

kevintechie:
A missing or invalid access token should return a 401 per HTTP 1.1 RFC. See my response to related issue: https://github.com/strongloop/loopback/issues/1496

benawarez:
I'm also having the 500 error.  I'm essentially using the same AuthService and controllers that came with the loopback-getting-started-intermediate.

The only difference is the following, any ideas?  Thanks in advance.

    $scope.login = function(e) {
      console.log('logging...');
      AuthService.login($scope.email, $scope.password)
        .then(function() {
          console.log('I am inside!');
           window.history.back();          
          //$state.go('add-review');
        },  
        function(err, token)
        {
          console.log(err);
          $scope.loginMsg = err.data.error.message;
        });
    };
  }])
  .controller('AuthLogoutController', ['$scope', 'AuthService', 'Helper',
      function($scope, AuthService, Helper) {
    AuthService.logout()
      .then(function() {
        //$state.go('all-reviews');
      });
  }])

rashmihunt:
[...]

PR #1496, opened 2015-07-01
Fail with "401 Unauthorized" when token is not provided to User.logout()

### LINKED ISSUE #1669: User.logout() return 500 error if auth token not found
I think this is related to https://github.com/strongloop/loopback/issues/1489 but it seems to me that if an auth token is not found, Strongloop should assume the request is unauthorized and return http code 401. Currently, it returns a 500 error which implies the server is broken.

**EDIT**: upon further inspection of the code, it looks like the error is happening explicitly when `User.logout()` is called and the access token does not exist in the database. I can see how a `500` error could make sense in this situation but it might make more sense for the method to simply return success.

ritch:
Connect to https://github.com/strongloop/strong-arc/issues/1414

Fix #1669

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -292,13 +292,23 @@ module.exports = function(User) {
 
   User.logout = function(tokenId, fn) {
     fn = fn || utils.createPromiseCallback();
-    this.relations.accessTokens.modelTo.findById(tokenId, function(err, accessToken) {
+
+    if (!tokenId) {
+      var err = new Error(g.f('{{accessToken}} is required to logout'));
+      err.status = 401;
+      process.nextTick(fn, err);
+      return fn.promise;
+    }
+
+    this.relations.accessTokens.modelTo.destroyById(tokenId, function(err, info) {
       if (err) {
         fn(err);
-      } else if (accessToken) {
-        accessToken.destroy(fn);
+      } else if ('count' in info && info.count === 0) {
+        var err = new Error(g.f('Could not find {{accessToken}}'));
+        err.status = 401;
+        fn(err);
       } else {
-        fn(new Error(g.f('could not find {{accessToken}}')));
+        fn();
       }
     });
     return fn.promise;
@@ -753,10 +763,10 @@ module.exports = function(User) {
       {
         description: 'Logout a user with access token.',
         accepts: [
-          {arg: 'access_token', type: 'string', required: true, http: function(ctx) {
+          {arg: 'access_token', type: 'string', http: function(ctx) {
             var req = ctx && ctx.req;
             var accessToken = req && req.accessToken;
-            var tokenID = accessToken && accessToken.id;
+            var tokenID = accessToken ? accessToken.id : undefined;
 
             return tokenID;
           }, description: 'Do not supply this argument, it is automatically extracted ' +
--- a/test/user.integration.js
+++ b/test/user.integration.js
@@ -106,6 +106,17 @@ describe('users - integration', function() {
             .expect(200, done);
         });
     });
+
+    it('returns 401 on logout with no access token', function(done) {
+      this.post('/api/users/logout')
+        .expect(401, done);
+    });
+
+    it('returns 401 on logout with invalid access token', function(done) {
+      this.post('/api/users/logout')
+        .set('Authorization', 'unknown-token')
+        .expect(401, done);
+    });
   });
 
   describe('sub-user', function() {
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1193,6 +1193,22 @@ describe('User', function() {
       }
     });
 
+    it('fails when accessToken is not provided', function(done) {
+      User.logout(undefined, function(err) {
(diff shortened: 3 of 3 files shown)


common/models/user.js
```diff
@@ -269,6 +269,14 @@ module.exports = function(User) {
...
 
   User.logout = function(tokenId, fn) {
+    if(!tokenId) {
+      process.nextTick(function() {
+        var err = new Error('access token is required to logout');
+        err.status = 400;
+        fn(err);
+      });
+      return;
+    }
```
raymondfeng: Should https://github.com/strongloop/loopback/blob/fix/logout-without-token/common/models/user.js#L286 throw an error with statusCode set to 404 or 400 too?

raymondfeng:
Connect to #1489

bajtos:
I find it surprising that the error was not caught by strong-remoting, since the argument is marked as required. The solution proposed here is fine though.

common/models/user.js
```diff
@@ -269,6 +269,14 @@ module.exports = function(User) {
...
     this.relations.accessTokens.modelTo.findById(tokenId, function(err, accessToken) {
```
raymondfeng: Can this just be `this.accessTokens.destroyById(tokenId, function(err) {...}`? I also wondering if we need to provide feedback if the access_token is invalid.

raymondfeng:
@ritch Assigning back to you to address the comments.

kevintechie:
According to the HTTP 1.1 RFC, if the Authorization header is missing, the response must be 401 Unauthorized and respond with an appropriate WWW-Authenticate header. If the Authorization header is  rejected (i. e. the access token is not found in the database), the response must also be 401.

Loopback returns a 401 for other REST requests that require authorization if the Authorization header is missing. It should return 401 for user/logout as well. Loopback does not respond with a WWW-Authenticate header as required, however.

Loopback also returns a 401 for other REST requests if the Authorization header is invalid. It should also return 401 for user/logout if the Authorization header is invalid.

According to the RFC, the response can include diagnostic information as to why the Authorization was rejected: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2 Perhaps Loopback should as well -- similar to how Oauth 2.0 Bearer tokens are implemented.

bajtos:
Hi @kevintechie, thank you for the description of what a correct behaviour should be per the HTTP spec. 

Would you mind taking this patch over, rebasing it on top of the current master and submitting a new PR where I can help you to get the code into the shape where it can be landed? Or you can start from a scratch and treat this patch as a source of inspiration only, if that's easier for you.

seanmangar:
Is this still open? I implemented this by overriding the base ACLs for user and applying the ACL below. I think this could be directly done in the loopback common/models/user.json.

```
    {
      "principalType": "ROLE",
      "principalId": "$authenticated",
      "permission": "ALLOW",
      "property": "logout"
    },
```

bajtos:
I think this problem is quite important and we should fix it. I am reassigning the pull request to myself.

bajtos:
@raymondfeng I have addressed your comments, does the patch LGTY now?

> I also wondering if we need to provide feedback if the access_token is invalid.

We should be returning 401 based on @kevintechie's https://github.com/strongloop/loopback/pull/1496#issuecomment-157469342 and that's what I have implemented here.

@kevintechie @seanmangar Please feel free to review my proposed changes and give me feedback what to fix/improve before merging.

common/models/user.js
```diff
@@ -752,10 +762,10 @@ module.exports = function(User) {
       {
         description: 'Logout a user with access token.',
         accepts: [
-          {arg: 'access_token', type: 'string', required: true, http: function(ctx) {
```
bajtos: With `required:true`, logout without any access tokens returns 400 because strong-remoting rejects the request as missing required arguments.

common/models/user.js
```diff
@@ -752,10 +762,10 @@ module.exports = function(User) {
...
+          {arg: 'access_token', type: 'string', http: function(ctx) {
             var req = ctx && ctx.req;
             var accessToken = req && req.accessToken;
-            var tokenID = accessToken && accessToken.id;
+            var tokenID = accessToken ? accessToken.id : undefined;
```
bajtos: The original code returned `null` when there was no token provided. strong-remoting does not accept `null` as a `string` value, therefore logout with no access token failed with 400 "missing required argument".

common/models/user.js
raymondfeng: Why is 400 worse than 401 if the token is not provided? IMO, it's better to report it as missing argument.

kevintechie: The RFC says that it should be 401. Client frameworks expect 401 and subsequent WWW-Authenticate header to respond correctly. Also, other loopback APIs respond correctly with 401 when the token is missing.

raymondfeng: Got it. +1 to be consistent.

bajtos:
Landed 🎉

Issue #564, opened 2014-09-16
Suppress stack traces from `loopback.errorHandler` output in production


bajtos:
strong-remoting was recently fixed to exclude the stack trace from the error response when `process.env.NODE_ENV === 'development'`.

However, `loopback.errorHandler` used by the scaffolded `server/server.js` does not support this feature.

We should modify `loopback.errorHandler` to return middleware that does not print stack traces when running in production.

chinuy:
Is there any timeline you want to put this implemented?

bajtos:
Not really. Would you mind contributing this feature yourself?

pasindud:
shouldn't this be set by a config.json ? or any other flag

bajtos:
> shouldn't this be set by a config.json ? or any other flag

Now that the error-handling middleware is configured in `server/middleware.json`, it should be reasonably easy to modify the handler factory to accept an `options` argument which would specify the behaviour.

digitalsadhu:
So looking around a bit, I understand that express errorhandler is for development only and provides no options for disabling stack traces. It looks to me that 1 idea might be to wrap express errorhandler to either return the errorhandler middleware if we are not in production or to provide a custom piece of errorhandler middleware (without stack traces) if we are. Does that sound like the right approach @bajtos ?

PR #1502, opened 2015-07-05
Suppress error handler stack traces in production#564

digitalsadhu:
Creates a new loopback.errorHandler function that wraps the existing express errorhandler module and essentially just unsets the stack trace when the app is in production.

Note, a modification to middleware.json file generation will need to occur for this approach to work.

```json
"final:after": {
  "errorhandler": {}
}
```
will need to change to:
```json
"final:after": {
  "loopback#errorHandler": {}
}
```
Fixes strongloop/loopback#564 

@bajtos This is my attempt to solve this issue. Thoughts?
If you like the approach I can put some more effort into writing tests and documentation. If you would prefer another approach, please advise.


slnode:

bajtos:
Hi @digitalsadhu, thank you for the pull request.

I would prefer to expose a configuration option instead of hard-coding this behaviour. That way we can preserve backwards compatibility.

```js
function errorHandler(options) {
  if (options.includeStack) {
    // original error handler
  } else {
    // remove stack
  }
}
```

The configuration - `server/middleware.production.json`:

```json
{
"final:after": {
  "loopback#errorHandler": {
    "params": {
      "includeStack": false
    }
  }
}
```

Thoughts?

bajtos:
I have edited my comment and fixed the condition to behave in backwards-compatible way.

digitalsadhu:
Yes thats nicer, will make the change

digitalsadhu:
will look into fixing tests shortly

digitalsadhu:
Tests are passing locally, under iojs, 0.10 and 0.12 so not sure whats going on with Travis.

@bajtos How is this looking now?

bajtos:
> Tests are passing locally, under iojs, 0.10 and 0.12 so not sure whats going on with Travis.

Did you run `npm test` too? That will run the tests in PhantomJS too, that's the part failing on Travis.

lib/loopback.js
```diff
@@ -158,7 +158,9 @@ if (loopback.isServer) {
     });
 
   loopback.urlNotFound = loopback['url-not-found'];
+  loopback.errorHandler = loopback['error-handler'];
   delete loopback['url-not-found'];
+  delete loopback['error-handler'];
```
bajtos: Nitpick: group by property name, not by operations.

```
loopback.urlNotFound = loopback['url-not-found'];
delete loopback['url-not-found'];

loopback.errorHandler = loopback['error-handler'];
delete loopback['error-handler'];
```

In this particular case, I'd personally keep `loopback[error-handler]` so that it can be loaded in middleware.json via `loopback#error-handler`. Not a big deal though.

test/error-handler.test.js
```diff
@@ -0,0 +1,43 @@
+var loopback = require('../');
+var app;
+var MockReq = require('mock-req');
+var MockRes = require('mock-res');
+var assert = require('assert');
```
bajtos: Please don't mock req/res objects, express maintainers are discouraging that. Setup a full loopback app and make a real request using supertest instead.

```js
var app = loopback();
app.use(loopback.urlNotFound());
app.use(loopback.errorHandler(/* opts */));

request(app).get('/url-does-not-exist')
  // verify that the response does or does not contain stack
```

bajtos:
Thank you for the update @digitalsadhu. I am starting a long (3 weeks+) paternity leave on Monday, it's likely that I won't be able to continue reviewing this patch as timely as I'd like. Sorry for that.

bajtos:
@slnode ok to test

digitalsadhu:
Thanks @bajtos. I'm on paternity leave 3 days/week myself. Enjoy :)
Anyway, that buys me some time to make fixes. 
Thanks for the feedback

bajtos:
@digitalsadhu ping, what's the status of this pull request? Are you still keen to get it finished & landed?

digitalsadhu:
@bajtos sorry for the delay, I have made fixes based on your comments, rebased loopback master onto my branch and pushed. How are we looking?

Also, question. Express is logging to the console during tests which is a bit ugly. Whats the best way to disable this?

bajtos:
> sorry for the delay, I have made fixes based on your comments, rebased loopback master onto my branch and pushed. How are we looking?

Thank you, the patch LGTM now.

>  Express is logging to the console during tests which is a bit ugly. Whats the best way to disable this?

I believe you can pass `log:false` in the options, could you try that out please? See https://github.com/expressjs/errorhandler#log

server/middleware/error-handler.js
```diff
@@ -0,0 +1,15 @@
+var expressErrorHandler = require('errorhandler');
```
bajtos: One more thing to address. `errorhandler` is a `devDependency`, thus this line will thrown when loopback is installed in a project that does not have `errorhandler` in project's dependencies.

Please make `errorhandler` a regular dependency in loopback's `package.json`

bajtos:
@digitalsadhu I am looking forward to land this patch! Please remove console logs from unit tests and fix the comment above first, or let me know whether I should make these two small changes myself.

digitalsadhu:
@bajtos 
- moved the dep from dev deps to deps 
- attempted to disable logging and discovered I wasn't passing the options object through to express errorHandler. Have added a test for this case and fixed. Logging now disabled in console for tests

bajtos:
> discovered I wasn't passing the options object through to express errorHandler. Have added a test for this case and fixed.

Haha, good catch! I appreciate you added a unit-test for that too :+1:

server/middleware/error-handler.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = errorHandler;
+
+function errorHandler(options) {
+  if (!options || options.includeStack !== false) {
+    return expressErrorHandler(options);
+  } else {
+    return function errorHandler(err, req, res, next) {
+      delete err.stack;
+      return expressErrorHandler(options)(err, req, res, next);
```
bajtos: Creating a new `expressErrorHandler(options)` for each request is suboptimal, it's better to call it only once. I'll fix this myself to speed things up.

bajtos:
Landed via 57e40cf, thank you for the contribution!

bajtos:
Note: when `includeStack` is `false`, the stack is removed from the error messages printed to the console too.

PR #1503, opened 2015-07-07
allow for unencoded access token for oauth2.0 using options.unencoded…

cooperjbrandon:
…TokenReferers

Signed-off-by: Brandon Cooper cooperjbrandon@gmail.com

This allows for the access bearer token to not be base64 encoded...Some front-end libraries don't base64 encode the auth token. For example, ember-simple-auth-oauth2 does not base64 encode the token: 
https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authorizers/oauth2.js#L33

The user would pass in an array of referers that it would not expect to encode the token, like so:

```
app.use(loopback.token({
  unencodedTokenReferers: ['localhost:4200', 'someotherurl.com']
}));
```

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -5,6 +5,7 @@
 var loopback = require('../../lib/loopback');
 var assert = require('assert');
 var uid = require('uid2');
+var url = require('url');
 var DEFAULT_TOKEN_LEN = 64;
 
 /**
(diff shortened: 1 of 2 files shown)


cooperjbrandon:
thoughts on this one?

raymondfeng:
@ritch Can you review?

cooperjbrandon:
Hi there, would it be possible to get some feedback on this?

PR #1517, opened 2015-07-10
Feature/mandrill template

aleho70:
Support for sending user verify emails with templates also via lb-connector-mandrill to Mandrill API.
It allows to use local EJS templates or remote Mandrill templates.
- If template attribute is string, then is expected that it is file name of local EJS template. 
- If template is object with name attribute then is expected that email transport connector is converting template to html itselfs, so in case of lb-connector-mandrill it could be for example Handlebars Mandril template.
  - NOTE: All atribute names in options in camelCase format will be converted to lowercase format, i.e. options.verifyHref to options.verifyhref

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -406,7 +406,8 @@ module.exports = function(User) {
         if (options.template.name) {
           // if template is object with name attribute then is expected that email transport connector is converting template to html itselfs
           // Following code is adapted to use options attributes (e.g. verifyHref) in mandrill templates via nodemailer-mandrill-transport
-          var options_clone = extend({}, options);
+          // NOTE: all atribute names in options in camelCase format will be converted to lowercase format, i.e. options.verifyHref to options.verifyhref
+          var options_clone = JSON.parse(JSON.stringify(options)); // to prevent error "converting circular structure to JSON"
           options.global_merge_vars = [{
             name: 'options',
             content: options_clone
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -6,6 +6,7 @@ var loopback = require('../../lib/loopback');
 var path = require('path');
 var SALT_WORK_FACTOR = 10;
 var crypto = require('crypto');
+var extend = require('util')._extend;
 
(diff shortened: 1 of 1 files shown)


slnode:

teone:
Hi @aleho70  
I'm facing this issue right now.. Any suggestion on how to use a custom `.ejs` template without this patch?

I was trying something like 

```javascript
let options = {
    type: 'email',
    port: '80',
    to: ac.email,
    from: app.get('emailFrom'),
    subject: 'Thanks for registering.',
    verifyHref: verifyHref,
    template: null,
    user: ac
};

var template = loopback.template(path.resolve(__dirname, '../../server/views/verify.ejs'));
options.html = template(options);
```

But it seems that loopback set a default template if none is specified and then `lb-mandrill-connector` trhows `Uncaught AssertionError: template name should be defined`

Thanks in advance for any help

bajtos:
@slnode ok to test

bajtos:
@aleho70 thank you for the pull request. Could you please add a unit-test to verify the implementation?

@raymondfeng @ritch you are more familiar with this part of loopback than I am, could one of you please review & land this patch?

brainwatch:
Hi

I'm stuck with the same problem of Mandrill with loopback. Any update on this ?

Thanks

bajtos:
@ritch you seem to be the author of this part, could you please review?

bajtos:
Hello! We have implemented a different solution in #2937, where users can provide a custom `templateFn` function that can use a custom templating engine to render the HTML body (for example). Also since #3004, we are removing `options.template` after `options.html` was constructed, in order to prevent Email transport errors.

IIUC, lb-mandrill-connector allows users to perform template->html conversion by the server sending emails (Mandrill in this case). I understand how this use-case can be useful. If there is a volunteer willing to rework the patch proposed in this pull request on top of the latest master, and include unit-tests to cover this new feature, then please open a new pull request.

I am closing this one as it's too outdated.

PR #1545, opened 2015-07-24
Add route based ACL support

raymondfeng:
/to @ritch 
/cc @superkhau 

This PR adds route based ACL as an express middleware.


raymondfeng:
Connect to strongloop/strong-gateway#38

server/middleware/acl.js
```diff
@@ -0,0 +1,543 @@
...
+      }
+      if (route.regexp.test(url)) {
+        matchedScopes.push(route.scope);
+      }
+    }
+  }
+  return matchedScopes;
+}
+
+/**
```
ritch: Mark this as private. Shouldn't include this in the docs.

ritch: > Calculate the ACL score based its specifics of principalType and permission

Should be 

> Calculate the ACL score based **on** its specifics of principalType and permission

test/route-acl.test.js
```diff
@@ -0,0 +1,202 @@
...
+  var handler;
+
+  before(function(done) {
+    handler = routeAcl(options);
+    var ds = loopback.createDataSource({
+      connector: 'memory',
+      name: 'db',
+      defaultForType: 'db'
+    });
+    loopback.Application.attachTo(ds);
```
ritch: We shouldn't use this soon to be deprecated version of core models... Use the new `registry` approach.

ritch: A good rule of thumb is to not use `loopback.Application` directly in your test app.

test/route-acl.test.js
```diff
@@ -0,0 +1,202 @@
...
+      'methods': 'GET',
+      'path': '/api/catalog'
+    }],
+    order: [{
+      methods: 'ALL',
+      path: '/api/orders'
+    }],
+    cancelOrder: [{
+      methods: 'delete',
+      'path': '/api/orders/:id'
```
ritch: inconsistent quoting???

ritch:
@raymondfeng 

Can you post an example and explain how this would work along side the existing ACL behavior?

Also I do not think we should add this into core. It looks like it might be convenient now, but its a brand new feature that I imagine will change quite a bit. If we land it in core now, we cannot make those changes until loopback 3. That approach (landing it in core now) will make this very difficult to maintain.

I recommend we create a `loopback-acl` module that exports this middleware. We can still land the `ACL` changes as is.

test/route-acl.test.js
```diff
@@ -0,0 +1,202 @@
...
+      permission: 'ALLOW'
+    },
+    {
+      principalType: 'ROLE',
+      principalId: 'cs',
+      scopes: ['cancelOrder'],
+      permission: 'ALLOW'
+    }
+  ],
+  roles: {
```
ritch: Why do we have to include roles here? Is this an advantage somehow?

raymondfeng: Yes, this allows us to define role mappings without DB.

test/route-acl.test.js
raymondfeng: Can you point me to an example? Please note that my test doesn't boot up an app.

test/route-acl.test.js
```diff
@@ -0,0 +1,202 @@
...
+      },
+      query: {
+        filter: {
+          limit: 100
+        }
+      }
+    };
+    var res = {};
+    handler(req, res, function(err) {
+      if (err) return done();
```
ritch: this should assert the `res` has a status code or something similar... `err` could be included for many unrelated reasons / unrelated errors

test/route-acl.test.js
```diff
@@ -0,0 +1,202 @@
...
+        'Accept': 'application/json'
+      },
+      accessToken: {
+        userId: 'john',
+        appId: 'demo-app'
+      }
+    };
+    var res = {};
+    handler(req, res, function(err) {
+      if (err) return done();
```
ritch: same as comment above... assert something about the `res`... `err` isn't enough

ritch:
I'm closing this since it should not be merged. Instead we should create a new module that exports this middleware.

Also the tests should not use `loopback.*` classes directly.

raymondfeng:
See https://github.com/strongloop-internal/scrum-loopback/issues/423

PR #1561, opened 2015-08-04
Allow methods filter for middleware config

raymondfeng:
/to @ritch 
/cc @bajtos 

This PR allows `methods` in middleware configuration so that the request can be filtered by the http methods. It's a requirement from declarative policies.

### DIFF
--- a/lib/server-app.js
+++ b/lib/server-app.js
@@ -65,6 +65,34 @@ proto.middlewareFromConfig = function(factory, config) {
   }
 
   var handler = factory.apply(null, params);
+
+  // Check if methods/verbs filter exists
+  var verbs = config.methods || config.verbs;
+  if (Array.isArray(verbs)) {
+    verbs = verbs.map(function(verb) {
+      return verb && verb.toUpperCase();
+    });
+    if (verbs.indexOf('ALL') === -1) {
+      var originalHandler = handler;
+      if (handler.length <= 3) {
+        // Regular handler
+        handler = function(req, res, next) {
+          if (verbs.indexOf(req.method.toUpperCase()) === -1) {
+            return next();
+          }
+          originalHandler(req, res, next);
+        };
+      } else {
+        // Error handler
+        handler = function(err, req, res, next) {
+          if (verbs.indexOf(req.method.toUpperCase()) === -1) {
+            return next(err);
+          }
+          originalHandler(err, req, res, next);
+        };
+      }
+    }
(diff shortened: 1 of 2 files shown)


lib/server-app.js
```diff
@@ -65,6 +65,29 @@ proto.middlewareFromConfig = function(factory, config) {
...
   var handler = factory.apply(null, params);
+
+  // Check if methods/verbs filter exists
+  var verbs = config.methods || config.verbs;
+  if (Array.isArray(verbs)) {
+    if (verbs.indexOf('all') === -1) {
+      var originalHandler = handler;
+      if (handler.length <= 3) {
+        handler = function(req, res, next) {
+          if (verbs.indexOf(req.method.toLowerCase()) === -1) {
```
bajtos: Is it worth optimising this lookup (together with L83) and replace `Array.prototype.indexOf` with object-property lookup? 

I think it would be also good to normalise method/verbs that are coming from the JSON config (apply `toLowerCase`).

```js
var verbs = Object.create(null);
(config.methods || config.verbs).forEach(function(val) { verbs[val.toLowerCase()] = true; });
// later in the handler:
if (verbs[req.method.toLowerCase]) 
```

test/app.test.js
```diff
@@ -438,12 +438,20 @@ describe('app', function() {
         params: null
       });
 
+      app.middlewareFromConfig(handlerFactory, {
+        enabled: true,
+        phase: 'routes:before',
+        methods: ['get', 'head'],
+        params: {x: 1}
+      });
```
bajtos: I believe this test will pass even without the implementation above. To verify that the method/verb filter works as intended, you need to test the negative case too (i.e. send a request with a method that is not allowed).

raymondfeng: Fixed. Added a negative test with `post`

lib/server-app.js
raymondfeng: For a tiny array, indexOf() should be very performed.

bajtos: > For a tiny array, indexOf() should be very performed.

Fair enough.

How about converting the method/verbs values specified by the config to lowercase? I think it's more likely that a user writes `POST` instead of `post` in the configuration, compared to  `req.method` being upper-case.

raymondfeng: Fixed

bajtos:
LGTM.

PR #1584, opened 2015-08-11
Enhance the ACL related models

raymondfeng:
/to @ritch 

The PR cleans up how related models are referenced and adds two utilities to ACL.

### DIFF
--- a/common/models/role-mapping.js
+++ b/common/models/role-mapping.js
@@ -17,6 +17,15 @@ module.exports = function(RoleMapping) {
   RoleMapping.APP = RoleMapping.APPLICATION = 'APP';
   RoleMapping.ROLE = 'ROLE';
 
+  RoleMapping.resolveRelatedModels = function() {
+    if (!this.userModel) {
+      var reg = this.registry;
+      this.roleModel = reg.getModelByType(loopback.Role);
+      this.userModel = reg.getModelByType(loopback.User);
+      this.applicationModel = reg.getModelByType(loopback.Application);
+    }
+  };
+
   /**
    * Get the application principal
    * @callback {Function} callback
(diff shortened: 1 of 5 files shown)


common/models/role-mapping.js
```diff
@@ -17,6 +17,15 @@ module.exports = function(RoleMapping) {
   RoleMapping.APP = RoleMapping.APPLICATION = 'APP';
   RoleMapping.ROLE = 'ROLE';
 
+  RoleMapping.getRelatedModels = function() {
+    if (!this.userModel) {
```
ritch: This is weird... Why can't we just do this in `setup()`? I don't like that we aren't returning anything and just modifying the `RoleMapping` model.

common/models/role-mapping.js
```diff
@@ -24,11 +33,10 @@ module.exports = function(RoleMapping) {
    * @param {Application} application
    */
   RoleMapping.prototype.application = function(callback) {
-    var registry = this.constructor.registry;
+    this.constructor.getRelatedModels();
```
ritch: This is really weird...

common/models/role.js
```diff
@@ -31,19 +28,19 @@ module.exports = function(Role) {
     return new Date();
   };
 
+  Role.getRelatedModels = function() {
```
ritch: same concern as `RoleMapping` version of this... I'm guessing there is something preventing from doing this is a more standard way... Can you elaborate?

common/models/scope.js
```diff
@@ -13,6 +13,13 @@ var loopback = require('../../lib/loopback');
  */
 
 module.exports = function(Scope) {
+  Scope.getRelatedModels = function() {
```
ritch: Same concern as elsewhere...

ritch:
I'd like to see the comments addressed here somehow. The rest of the code looks good.

common/models/role-mapping.js
raymondfeng: `setup` is still too early to resolve the related models as they are guaranteed to be available in the registry when `setup` is called.

I'll rename the method to `resolveRelatedModels`.

raymondfeng:
@ritch I renamed the method to be `resolveRelatedModels` to reflect its purpose. We'll revisit it after a new hook is introduced to resolve the model dependencies after all models are defined and registered in the registry.

ritch:
Can you squash and reword the commit so they make sense in the changelog? Also the build is failing.

PR #1632, opened 2015-08-27
Allow EJS templates to use includes

gausie:
Simple change that allows templates to use includes.


slnode:

bajtos:
@slnode ok to test

bajtos:
Landed, thank you for the contribution.

Issue #1322, opened 2015-04-21
Rest API: DELETE method always returns 204


Kampfgnom:
When calling DELETE on a resource, the response code is always 204 with an empty body, independently from whether the resource exists or not, or if it has been deleted or not (headers omitted for brevity):

````
$ curl -i http://localhost:3000/api/TodoLists
HTTP/1.1 200 OK
[]

$ curl -i -X PUT http://localhost:3000/api/TodoLists
HTTP/1.1 200 OK
{"id":"27e2ea50-e814-11e4-8782-29c27147e8f9"}

$ curl -i http://localhost:3000/api/TodoLists
HTTP/1.1 200 OK
[{"title":null,"id":"27e2ea50-e814-11e4-8782-29c27147e8f9","systemUserId":null}]

$ curl -i -X DELETE http://localhost:3000/api/TodoLists/asd
HTTP/1.1 204 No Content

$ curl -i http://localhost:3000/api/TodoLists
HTTP/1.1 200 OK
[{"title":null,"id":"27e2ea50-e814-11e4-8782-29c27147e8f9","systemUserId":null}]

$ curl -i -X DELETE http://localhost:3000/api/TodoLists/27e2ea50-e814-11e4-8782-29c27147e8f9
HTTP/1.1 204 No Content

$ curl -i http://localhost:3000/api/TodoLists
HTTP/1.1 200 OK
[]

$ curl -i -X DELETE http://localhost:3000/api/TodoLists/27e2ea50-e814-11e4-8782-29c27147e8f9
HTTP/1.1 204 No Content
````

I see the same behaviour in the API-explorer and my (currently in development) Qt-Client.

Thanks for any help.

superkhau:
@Kampfgnom I've verified your observations and 204 is always returned as you described. Escalating to "bug" status.

fabien:
@raymondfeng we should really get this fixed, see also: https://github.com/strongloop/loopback/issues/1543#issuecomment-124330089

I would suggest a config option: `deleteInfo` as well as `Model.settings.deleteInfo` or something like that to enable it. Ideally I would like such option to work like this: https://github.com/strongloop/loopback-connector-mongodb/pull/164#issue-96759406

raymondfeng:
Sure. @superkhau Can you take this one? It's very similar as the work you did for update.

superkhau:
Assigned to myself so we can prioritize during sprint planning.

/cc @chandadharap

superkhau:
[...]

PR #1667, opened 2015-09-11
Return HTTP 200 when deleting a resource via REST

superkhau:
Connect to #1322

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -618,6 +618,12 @@ module.exports = function(registry) {
       description: 'Delete all matching records.',
       accessType: 'WRITE',
       accepts: {arg: 'where', type: 'object', description: 'filter.where object'},
+      returns: {
+        arg: 'count',
+        type: 'object',
+        description: 'The number of instances deleted',
+        root: true
+      },
       http: {verb: 'del', path: '/'},
       shared: false
     });
@@ -632,6 +638,12 @@ module.exports = function(registry) {
         {arg: 'data', type: 'object', http: {source: 'body'},
           description: 'An object of model property name/value pairs'},
       ],
+      returns: {
+        arg: 'count',
+        description: 'The number of instances updated',
+        type: 'object',
+        root: true
+      },
       http: {verb: 'post', path: '/update'}
     });
 
@@ -641,7 +653,8 @@ module.exports = function(registry) {
       accessType: 'WRITE',
       accepts: {arg: 'id', type: 'any', description: 'Model id', required: true,
                 http: {source: 'path'}},
-      http: {verb: 'del', path: '/:id'}
+      http: {verb: 'del', path: '/:id'},
+      returns: {arg: 'count', type: 'object', root: true}
     });
 
     setRemoting(PersistedModel, 'count', {
--- a/test/remoting.integration.js
+++ b/test/remoting.integration.js
@@ -117,14 +117,15 @@ describe('remoting - integration', function() {
         'findById(id:any,filter:object):store GET /stores/:id',
         'find(filter:object):store GET /stores',
         'findOne(filter:object):store GET /stores/findOne',
-        'updateAll(where:object,data:object) POST /stores/update',
-        'deleteById(id:any) DELETE /stores/:id',
+        'updateAll(where:object,data:object):object POST /stores/update',
+        'deleteById(id:any):object DELETE /stores/:id',
         'count(where:object):number GET /stores/count',
-        'prototype.updateAttributes(data:object):store PUT /stores/:id'
+        'prototype.updateAttributes(data:object):store PUT /stores/:id',
+        'createChangeStream(options:object):ReadableStream POST /stores/change-stream'
       ];
 
       // The list of methods is from docs:
-      // http://docs.strongloop.com/display/LB/Exposing+models+over+a+REST+API
+      // https://docs.strongloop.com/display/public/LB/Exposing+models+over+REST
       expect(methods).to.include.members(expectedMethods);
     });
 
--- a/test/rest.middleware.test.js
+++ b/test/rest.middleware.test.js
@@ -1,3 +1,5 @@
+var path = require('path');
+
 describe('loopback.rest', function() {
   var MyModel;
   beforeEach(function() {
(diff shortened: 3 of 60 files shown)


superkhau:
@raymondfeng Review please.

lib/persisted-model.js
```diff
@@ -641,7 +641,8 @@ module.exports = function(registry) {
       accessType: 'WRITE',
       accepts: {arg: 'id', type: 'any', description: 'Model id', required: true,
                 http: {source: 'path'}},
-      http: {verb: 'del', path: '/:id'}
+      http: {verb: 'del', path: '/:id'},
+      returns: {arg: 'count', type: 'number', root: true}
```
raymondfeng: Is it an object with the count property or just the plain number?

superkhau: It's an object with the count property, ie:

```
{count: 1}
```

test/rest.middleware.test.js
```diff
@@ -14,6 +14,17 @@ describe('loopback.rest', function() {
...
   });
 
+  it('should report 200 for DELETE /:id found', function(done) {
+    app.set('legacyExplorer', false);
+    app.model(MyModel);
+    app.use(loopback.rest());
+    MyModel.create({name: 'm1'}, function(err, inst) {
+      request(app)
+        .del('/mymodels/' + inst.id)
+        .expect(200, done);
```
raymondfeng: We should test the http body too to make sure it returns the correct count.

lib/persisted-model.js
raymondfeng: Then the return type is an object instead of number.

raymondfeng: What about updateAll? It should have similar return value.

superkhau: >Then the return type is an object instead of number.

I see, I thought it was describing the `count` property. Changing to `object`.

>What about updateAll? It should have similar return value.

Beyond the scope of my current task, but IMO yes and we should probably check the others too. Like the same can go for destroyAll (count of items deleted), updateAll, bulkUpdate, maybe create a new issue to find out what needs updating?

test/rest.middleware.test.js
superkhau: +1

test/remoting.integration.js
```diff
@@ -118,13 +118,14 @@ describe('remoting - integration', function() {
         'find(filter:object):store GET /stores',
         'findOne(filter:object):store GET /stores/findOne',
         'updateAll(where:object,data:object) POST /stores/update',
-        'deleteById(id:any) DELETE /stores/:id',
+        'deleteById(id:any):number DELETE /stores/:id',
```
raymondfeng: IIRC, `deleteById` should callback with `true` or `false`.

raymondfeng: The `count` should be used for updateAll/deleteAll.

superkhau: OK, I'm going to update everything according to your feedback and reassign to you for review after.

superkhau: I assume I need to change the memory connector impl to return boolean instead of object with a count property. This will probably break a bunch of tests on the way, so then I will have to fix all those down the chain too I assume.

@chandadharap Increasing scope for this story. FYI.

rmg:
@slnode test please

ritch:
Nice work. This must be released as a minor version, do not merge without a plan for a minor release. Otherwise we might hold up patch releases.

kraman:
@superkhau What will this return when someone tries to delete an instance that has already been deleted, i.e. does not exist?. A return of 200 with a count of 0 would be preferred on my end.

superkhau:
@kraman This I'm not sure about yet. The request was successful, but the document was not found. So I'm assuming HTTP 200 and a count of 0 like you said makes sense. @raymondfeng What do you think?

raymondfeng:
If the `strictDelete` flag is set to true, `404` will be returned. Otherwise `200 + {count : 0}`.

superkhau:
@raymondfeng All the tests pass except these 3. It passes when running it standalone, but fails when I run it together. I have no idea why, but I think there is some test pollution going on. Any ideas?

```
  users - integration
    base-user
      1) should create a new user
      2) should log into the user
      3) should create post for a given user

events.js:72
        throw er; // Unhandled 'error' event
              ^
AssertionError: expected [] to have a property 'length' of 1, but got 0
```

superkhau:
@slnode test please

superkhau:
@raymondfeng @ritch Review please.

lib/persisted-model.js
```diff
@@ -618,6 +618,7 @@ module.exports = function(registry) {
       description: 'Delete all matching records.',
       accessType: 'WRITE',
       accepts: {arg: 'where', type: 'object', description: 'filter.where object'},
+      returns: {arg: 'count', type: 'object', root: true},
```
raymondfeng: Add a `description`?

superkhau: I don't understand. The `description` is already there in line 618?

raymondfeng: It's a description for the return object.

superkhau: ICIC. Fixed.

lib/persisted-model.js
```diff
@@ -618,6 +618,12 @@ module.exports = function(registry) {
       description: 'Delete all matching records.',
       accessType: 'WRITE',
       accepts: {arg: 'where', type: 'object', description: 'filter.where object'},
+      returns: {
+        arg: 'count',
+        type: 'object',
+        description: 'The number of instances deleted',
```
superkhau: Fixed here.

lib/persisted-model.js
```diff
@@ -632,6 +638,12 @@ module.exports = function(registry) {
         {arg: 'data', type: 'object', http: {source: 'body'},
           description: 'An object of model property name/value pairs'},
       ],
+      returns: {
+        arg: 'count',
+        description: 'The number of instances updated',
```
superkhau: @raymondfeng And here.

superkhau:
@raymondfeng Please review again.

raymondfeng:
LGTM

superkhau:
@slnode test please

superkhau:
@rmg How can I tell is this is safe to merge?

sachinhub:
What is the release plan for this patch? I am looking for a feature to disable certain remote methods.

superkhau:
@sachinhub It's already been merged via #1667. This has nothing to do with disabling certain remote methods though.

superkhau:
>@superkhau I am horrified by this...

Yeah when I look back now, it does look bad (to be fair, I got the LGTM from numerous reviews). Creating the app model programmatically is definitely better than scaffolding the entire project to as a fixture). This can definitely use a refactor. ;)

superkhau:
@bajtos See https://github.com/strongloop-internal/scrum-loopback/issues/791. Please prioritize it as you see fit during planning.

PR #1716, opened 2015-10-08
Set application's id property only if it's empty.

### LINKED ISSUE #1689: Application.resetKeys will create a new instance instead of updating a record
When I want to reset keys for a application use `Application.resetKeys`. It will create a new instance instead of updating this instance.

This caused by 'before save' operation hook.
https://github.com/strongloop/loopback/blob/master/common/models/application.js#L95

Maybe it should set the id property just it's empty.

wusuopu:
to @ritch

Close #1689

### DIFF
--- a/common/models/application.js
+++ b/common/models/application.js
@@ -92,7 +92,9 @@ module.exports = function(Application) {
 
     var app = ctx.instance;
     app.created = app.modified = new Date();
-    app.id = generateKey('id', 'md5');
+    if (!app.id) {
+      app.id = generateKey('id', 'md5');
+    }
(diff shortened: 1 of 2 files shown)


slnode:

ritch:
Ok to test

ritch:
This LGTM

bajtos:
@wusuopu could you please add a unit-test to verify your implementation and prevent regressions in the future?

wusuopu:
@bajtos it has added now

bajtos:
Landed via  14e6ec5, thank you for the contribution!

Issue #1769, opened 2015-10-24
Error using connector-remote datasource for AccessToken model


danilat:
Using loopback to develop a couple of microservices I found a bug when I try to use a common microservice to auth validate access-tokens using remote-connector. 

I have this configuration in the datasource:

```
{
  "db": {
    "database": "mydb",
    "username": "root",
    "name": "db",
    "connector": "mysql"
  },
  "auth": {
    "url": "http://myauthservice/api",
    "name": "auth",
    "connector": "remote"
  }
}
```

And all related authentication/authorization stuff in the model-config use the auth datasource. For example:

```
"AccessToken": {
    "dataSource": "auth",
    "public": false
  }
```

When I use an existing access-token I get this error:

```
/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback/common/models/access-token.js:104
          token.validate(function(err, isValid) {
                ^
TypeError: undefined is not a function
    at HttpInvocation.<anonymous> (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback/common/models/access-token.js:104:17)
    at /Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/rest-adapter.js:131:18
    at execStack (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/remote-objects.js:460:7)
    at RemoteObjects.execHooks (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/remote-objects.js:464:10)
    at HttpInvocation.<anonymous> (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/rest-adapter.js:129:15)
    at HttpInvocation.transformResponse (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/http-invocation.js:337:12)
    at Request._callback (/Users/dani/workspace/bierasolutions/accounts/node_modules/loopback-connector-remote/node_modules/strong-remoting/lib/http-invocation.js:261:10)
    at Request.self.callback (/Users/dani/workspace/bierasolutions/accounts/node_modules/request/request.js:198:22)
    at Request.emit (events.js:110:17)
[...]

PR #1770, opened 2015-10-24
Fix issue 1769

danilat:
This is the explanation of the bug I found https://github.com/strongloop/loopback/issues/1769. In my PR I have a ugly patch in this commit https://github.com/danilat/loopback/commit/e028b56aad42b0828a9bdd10dcaa9e8892344c34.

Now I'm working in a workaround with some changes in connector-remote parsing all the valid date-format strings in the respose objects. What do you think about this?

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -101,7 +101,7 @@ module.exports = function(AccessToken) {
         if (err) {
           cb(err);
         } else if (token) {
-          token.validate(function(err, isValid) {
+          AccessToken.validate(token, function(err, isValid) {
             if (err) {
               cb(err);
             } else if (isValid) {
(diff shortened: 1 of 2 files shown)


slnode:

superkhau:
Can you fix the PR so they pass CI?

danilat:
Fixed

common/models/access-token.js
```diff
@@ -132,26 +137,26 @@ module.exports = function(AccessToken) {
    * @param {Boolean} isValid
    */
 
-  AccessToken.prototype.validate = function(cb) {
+  AccessToken.validate = function(token, cb) {
```
bajtos: This is a breaking change, we need to add a backwards-compatible wrapper method:

```js
AccessToken.prototype.validate = function(cb) {
  AccessToken.validate(this, cb); // or this.toObject()?
};
```

bajtos: @danilat pardon me if my answer is obvious, but why is it necessary to change this method from a prototype method to a static method?

common/models/access-token.js
```diff
@@ -101,7 +101,12 @@ module.exports = function(AccessToken) {
         if (err) {
           cb(err);
         } else if (token) {
-          token.validate(function(err, isValid) {
+          var possibleDate = new Date(token.created);
+          var isAValidDate = !isNaN(possibleDate.getTime());
+          if (isAValidDate) {
+            token.created = new Date(token.created);
+          }
```
bajtos: This looks suspicious. What part of the system is producing `token.created` that is not already a `Date` instance? I'd rather fix the problem at the source.

/cc @ritch @kraman

danilat: I have an application using connector-remote datasource for AccessToken, this was a temporal fix but I fixed the source in the connector module https://github.com/danilat/loopback-connector-remote/commit/db9446705dedabea6811a68faf4ce33e48a93beb

common/models/access-token.js
danilat: In the connector-remote ds scenario for AccessToken model the prototype method doesn't work, changing the implementation to static I was the solution that I found, I tried a couple of workarounds but nothing of them worked. I'm sure that a more elegant implementation could be done.

bajtos:
I am afraid this is fixing the wrong place. Ideally, loopback-connector-remote should support remote methods and seamlessly convert (coerce) response properties to the correct type. See my https://github.com/strongloop/loopback/issues/1769#issuecomment-265717471

Issue #1723, opened 2015-10-10
Too strict validation on User.email


daankets:
I get the following validation error when trying to persist a user in my local test environment:

```
ValidationError: The `User` instance is not valid. Details: `email` Must provide a valid email (value: **"test@localhost"**).
    at ...(not relevant)
```

According to me, this is a too strict validation for an email address. The spec states that an email address should comply to the pattern <userOrMailBox>@<smtpHostName>. There is no requirement whatsoever that the hostname must also contain a domain name. Of course it is possible to override this by re-implementing the User model, but it would be nice if the current validation was a bit less strict (it is always possible to add a more strict validation on top of it).

vdust:
I confirm this issue, but it goes beyond being too strict: it even accepts some invalid email addresses as valid. I created a fork of loopback-sandbox with some examples of addresses that should be accepted  (but aren't) and addresses that should be rejected (but aren't):

https://github.com/vdust/loopback-sandbox

The truth is that email addresses *can't* be validated by a single regular expression. RFC 822 and RFC 2822 describe what valid email addresses are (the 'addr-spec' definition in RFC 2822 covers both) and it's huge work to implement it properly. But some effort should be done in that direction as it would benefit a lot of projects.

daankets:
[This library](https://github.com/chriso/validator.js) comes pretty close I believe...

superkhau:
@daankets @vdust Would one of you guys like to contribute a patch?

vdust:
First, I must correct what i stated in my previous comment: It's [RFC5321](http://tools.ietf.org/html/rfc5321) and [RFC5322](http://tools.ietf.org/html/rfc5322) that define what valid email addresses are in the context that concerns us here. As a result, some of the *valid* addresses i added in my test code (e.g. ``local."quoted word"@example.com``) use deprecated syntax and are probably safe to reject since the semantics of such forms are the same as the recommended fully-quoted local part (``"local.quoted word"@example.com``).

@daankets I saw this library, it but makes some trivial mistakes like assuming a local part is always properly quoted as a single unit (see https://github.com/chriso/validator.js/blob/master/validator.js#L202 ), matching a lot of invalid addresses (i.e. ``"a test@example.com``)

[...]

PR #1782, opened 2015-10-28
Use isemail for email validation in user model

vdust:
instead of the regular expression that did not correctly validate all
email addresses (both false positives and false negatives).

Should fix issue #1723

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -91,7 +91,7 @@ describe('User', function() {
         assert.equal(err.details.context, User.modelName);
         assert.deepEqual(err.details.codes.email, [
           'presence',
-          'format.null'
+          'custom.isemail'
         ]);
 
         done();
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -7,6 +7,7 @@ var utils = require('../../lib/utils');
 var path = require('path');
 var SALT_WORK_FACTOR = 10;
 var crypto = require('crypto');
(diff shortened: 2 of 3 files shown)


slnode:

superkhau:
@rmg There seems to be an issue with io.js. Can you take a look?

rmg:
@slnode ok to test

rmg:
The `.travis.yml` file needs to be updated to replace `iojs` with `4`.

bajtos:
+1 for using `isemail` instead of a hand-written regexp.

test/user.test.js
```diff
@@ -91,7 +91,7 @@ describe('User', function() {
         assert.equal(err.details.context, User.modelName);
         assert.deepEqual(err.details.codes.email, [
           'presence',
-          'format.null'
+          'custom'
```
bajtos: Could you please improve your patch to report a unique error code, e.g. `isemail`? The validation codes are typically used by front-end to display a localised error message. The code `custom` can mean anything...

vdust: I'm not sure how to do this. The Validatable api in loopback-datasource-juggler/lib/validations.js doesn't seem to allow override of the validator name. Should i move the logic in a new Validatable.validateEmail function there and then simply use it in the user model ?

bajtos:
Closing in favour of #2716.

PR #1804, opened 2015-11-09
Add case-sensitive email option for User model

### LINKED ISSUE #1150: loopback auth treats User.email case sensitive
And it goes the same for User.username, but that depends on taste...

My use case: I can't login when my iOS automatically sets the first character to uppercase...but please look beyond this ;)

richardpringle:
Have not created a test file for this yet. @ritch could you review this please (if you get a chance)?

Fix #1150
Connect to #1150

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -58,6 +58,7 @@ var debug = require('debug')('loopback:user');
  * @property {String} settings.realmDelimiter When set a realm is required.
  * @property {Number} settings.resetPasswordTokenTTL Time to live for password reset `AccessToken`. Default is `900` (15 minutes).
  * @property {Number} settings.saltWorkFactor The `bcrypt` salt work factor. Default is `10`.
+ * @property {Boolean} settings.caseSensitiveEmail Enable case sensitive email.
  *
  * @class User
  * @inherits {PersistedModel}
@@ -573,6 +574,14 @@ module.exports = function(User) {
     this.settings.maxTTL = this.settings.maxTTL || DEFAULT_MAX_TTL;
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
+    UserModel.setter.email = function(value) {
+      if (!UserModel.settings.caseSensitiveEmail) {
+        this.$email = value.toLowerCase();
+      } else {
+        this.$email = value;
+      }
+    };
+
     UserModel.setter.password = function(plain) {
       if (typeof plain !== 'string') {
         return;
@@ -585,6 +594,14 @@ module.exports = function(User) {
         this.$password = this.constructor.hashPassword(plain);
       }
     };
+
+    // Access token to normalize email credentials
+    UserModel.observe('access', function normalizeEmailCase(ctx, next) {
+      if (!ctx.Model.settings.caseSensitiveEmail && ctx.query.where && ctx.query.where.email) {
+        ctx.query.where.email = ctx.query.where.email.toLowerCase();
+      }
+      next();
+    });
 
     // Make sure emailVerified is not set by creation
     UserModel.beforeRemote('create', function(ctx, user, next) {
--- a/common/models/user.json
+++ b/common/models/user.json
@@ -29,6 +29,9 @@
     "created": "date",
     "lastUpdated": "date"
   },
+  "options": {
+    "caseSensitiveEmail": true
+  },
   "hidden": ["password"],
   "acls": [
     {
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -15,6 +15,7 @@ describe('User', function() {
   var validCredentialsEmailVerifiedOverREST = {email: 'foo2@bar.com', password: 'bar2', emailVerified: true};
   var validCredentialsWithTTL = {email: 'foo@bar.com', password: 'bar', ttl: 3600};
   var validCredentialsWithTTLAndScope = {email: 'foo@bar.com', password: 'bar', ttl: 3600, scope: 'all'};
+  var validMixedCaseEmailCredentials = {email: 'Foo@bar.com', password: 'bar'};
   var invalidCredentials = {email: 'foo1@bar.com', password: 'invalid'};
   var incompleteCredentials = {password: 'bar1'};
 
@@ -63,6 +64,26 @@ describe('User', function() {
         assert(!err);
         assert(user.id);
         assert(user.email);
+        done();
+      });
+    });
+
+    it('Create a new user (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: 'F@b.com', password: 'bar'}, function(err, user) {
+        if (err) return done(err);
+        assert(user.id);
+        assert.equal(user.email, user.email.toLowerCase());
+        done();
+      });
+    });
+
+    it('Create a new user (email case-sensitive)', function(done) {
+      User.create({email: 'F@b.com', password: 'bar'}, function(err, user) {
+        if (err) return done(err);
+        assert(user.id);
+        assert(user.email);
+        assert.notEqual(user.email, user.email.toLowerCase());
         done();
       });
     });
@@ -119,6 +140,27 @@ describe('User', function() {
       User.create({email: 'a@b.com', password: 'foobar'}, function() {
         User.create({email: 'a@b.com', password: 'batbaz'}, function(err) {
           assert(err, 'should error because the email is not unique!');
+          done();
+        });
+      });
+    });
+
+    it('Requires a unique email (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: 'A@b.com', password: 'foobar'}, function(err) {
+        if (err) return done(err);
+        User.create({email: 'a@b.com', password: 'batbaz'}, function(err) {
+          assert(err, 'should error because the email is not unique!');
+          done();
+        });
+      });
+    });
+
+    it('Requires a unique email (email case-sensitive)', function(done) {
+      User.create({email: 'A@b.com', password: 'foobar'}, function(err, user1) {
+        User.create({email: 'a@b.com', password: 'batbaz'}, function(err, user2) {
+          if (err) return done(err);
+          assert.notEqual(user1.email, user2.email);
           done();
         });
       });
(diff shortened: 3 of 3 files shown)


common/models/user.js
```diff
@@ -573,6 +575,19 @@ module.exports = function(User) {
     this.settings.maxTTL = this.settings.maxTTL || DEFAULT_MAX_TTL;
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
+    // user email case sensitivity default to true
+    if (this.settings.caseSensitiveEmail === undefined) {
```
superkhau: `if (this.settings.caseSensitiveEmail)`

superkhau: Actually, can't you just set this default to true somewhere before? ie. When configs are read in, etc.

common/models/user.json
```diff
@@ -23,6 +23,7 @@
       "type": "string",
       "required": true
     },
+    "emailCaeSensitive": "boolean",
```
superkhau: Case is spelled wrong.

common/models/user.js
```diff
@@ -119,11 +120,12 @@ module.exports = function(User) {
    * @returns {Object} The normalized credential object
    */
   User.normalizeCredentials = function(credentials, realmRequired, realmDelimiter) {
+    var caseSensitiveEmail = this.settings.caseSensitiveEmail;
```
superkhau: Move this down under line 127 so the variable doesn't always initialize if `credentials.email` doesn't exist.

superkhau:
Added comments. Also, can you write a test that fails and then make it pass?

common/models/user.js
richardpringle: I couldn't change the setting in a boot script if I set it to true in user.js. Maybe I'm just trying to set it in the wrong place? Do you have a line number you could refer me to?

common/models/user.js
richardpringle: Yeah, but then I would have to add the same line under 137 as well (if a realm is required). Still worth it?

richardpringle:
@superkhau, I actually had some issues with `npm test`. Who should I talk to about that.

.strong-pm/env.json
```diff
@@ -0,0 +1 @@
+{}
```
bajtos: Please remove.

common/models/user.json
```diff
@@ -23,6 +23,7 @@
       "type": "string",
       "required": true
     },
+    "emailCaseSensitive": "boolean",
```
bajtos: `emailCaseSensitive` is a model setting, not a model property. Please remove this line.

bajtos:
I am not very comfortable with modifying the email values to achieve case-insensitive comparison. It may lead to confusing situations, e.g.

```js
await User.create({ email: 'mbajtos@cz.ibm.com', password: 'pass' });
var list = await User.find({ where: { email: 'mbajtos@cz.ibm.com' });
// no user found
```

IMO, we should use case-insensitive query operator instead. IIRC it was you @superkhau who implemented it?

ritch:
> It may lead to confusing situations, e.g.

As long as this is clear in the documentation, I think its fine.

> I want to preserve the casing of the user input!

 - no one ever

superkhau:
>IMO, we should use case-insensitive query operator instead. IIRC it was you @superkhau who implemented it?

Yeah, we have the regex operator, which can take a a `/i` flag for case-insensitive regexes. However, the impl for each connector varies and the feature may not be supported, but you'll get a warning if it doesn't.

common/models/user.js
superkhau: No, I didn't notice 128. Leave it there, it's good! ;)

superkhau:
>I actually had some issues with npm test. Who should I talk to about that?

Not sure who the repo expert is for this particular repo. @ritch?

richardpringle:
@slnode test please

test/user.test.js
```diff
@@ -15,6 +15,7 @@ describe('User', function() {
   var validCredentialsEmailVerifiedOverREST = {email: 'foo2@bar.com', password: 'bar2', emailVerified: true};
   var validCredentialsWithTTL = {email: 'foo@bar.com', password: 'bar', ttl: 3600};
   var validCredentialsWithTTLAndScope = {email: 'foo@bar.com', password: 'bar', ttl: 3600, scope: 'all'};
+  var validCaseInsensitiveEmailCredentials = {email: 'Foo@bar.com', password: 'bar'};
```
richardpringle: @superkhau @bajtos, can you guys review my changes to this test file?

bajtos:
> > It may lead to confusing situations, e.g.

> As long as this is clear in the documentation, I think its fine.

Most people don't read documentation. I personally wouldn't even know where (in which part of the documentation) to look for this information.

> > I want to preserve the casing of the user input for email!

> .... said no one ever.

I think this is subjective, but let's say you are right and people won't mind LoopBack sanitising the email address on input.

In that case I would like LoopBack to also sanitise email addresses in all queries, so that my example keeps working:

```js
await User.create({ email: 'MBajtos@cz.ibm.com', password: 'pass' });
  // email is stored as "mbajtos@cz.ibm.com"
var list = await User.find({ where: { email: 'MBajtos@cz.ibm.com' });
  // the query is modified to email: "mbajtos@cz.ibm.com"
```

test/user.test.js
```diff
@@ -67,6 +68,26 @@ describe('User', function() {
       });
     });
 
+   it('Create a new user (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: 'F@b.com', password: 'bar'}, function(err, user) {
+        assert(!err);
```
bajtos: Please use this form: `if (err) return done(err)`. It produces stack traces pointing to the original error, as opposed to `assert(!err)` which points to this line in tests. Please apply this rule to all new tests you are adding in this patch (no need to change existing code).

test/user.test.js
```diff
@@ -124,6 +145,27 @@ describe('User', function() {
       });
     });
 
+    it('Requires a unique email (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: 'A@b.com', password: 'foobar'}, function() {
```
bajtos: Missing error handler.

test/user.test.js
```diff
@@ -124,6 +145,27 @@ describe('User', function() {
...
+        User.create({email: 'a@b.com', password: 'batbaz'}, function(err) {
+          assert(err, 'should error because the email is not unique!');
+          done();
+        });
+      });
+    });
+
+    it('Requires a unique email (email case-sensitive)', function(done) {
+      User.settings.caseSensitiveEmail = true;
+      User.create({email: 'A@b.com', password: 'foobar'}, function(err, user1) {
```
bajtos: Missing error handler.

bajtos:
> @richardpringle can you guys review my changes to this test file?

You are on the right track :+1: , I left few comments on what should be improved.

test/user.test.js
richardpringle: I followed the same conventions at the other tests in that `describe` block. Should I replace `assert(!err)` in all cases then?

test/user.test.js
```diff
@@ -223,6 +265,23 @@ describe('User', function() {
...
+      User.login(validCaseInsensitiveEmailCredentials, function(err, accessToken) {
+        assert(accessToken.userId);
+        assert(accessToken.id);
+        assert.equal(accessToken.id.length, 64);
+        done();
+      });
+    });
+
+    it('Try to login with invalid email case', function(done) {
+      User.login(validCaseInsensitiveEmailCredentials, function(err, accessToken) {
```
superkhau: This case is confusing. You're trying to log in with `validCaseInsenstiveEmailCredentials` and expect an error?

test/user.test.js
bajtos: > I followed the same conventions at the other tests in that describe block.

It's an old convention that I would like us to eventually get rid of.

> Should I replace assert(!err) in all cases then?

I don't mind either. However, if you decide to fix existing tests, then make it in a standalone commit. This will make future inspection (code archaeology) much easier.

richardpringle:
@slnode test please

test/user.test.js
bajtos: Good catch, @superkhau. I think the name `validCaseInsensitiveEmailCredentials` is confusing. The variable contains credentials that are invalid by default but valid when case-insensitive check is made.

bajtos:
@richardpringle could you please add an `access` hook to convert email fields to lower case in queries too, as I proposed in https://github.com/strongloop/loopback/pull/1804#issuecomment-156429476?

richardpringle:
test please

richardpringle:
test please

bajtos:
@rmg the build is failing on Ubuntus with the following message:

```
ERROR: No Xvfb installations defined, please define one in the configuration. Once defined you’ll need to choose one under Advanced options for Xvfb plugin job settings and save job configuration.
```

It looks like a problem with the build slave VM image, could you PTAL?

bajtos:
@slnode test please

richardpringle:
@bajtos 
> If you have ACLs enabled then `GET /api/users` will be rejected. Let's try `Users.find().then(...)` instead.
I'm not quite sure what you mean. Make another remote method?

bajtos:
@slnode test please

bajtos:
> I'm not quite sure what you mean. Make another remote method?

Sorry for the confusion. Just check that when you disable `caseSensitiveEmail ` and call `User.find()` (with no query), your access hook does not crash on accessing a property of `undefined`.

richardpringle:
test please

test/user.test.js
```diff
@@ -212,6 +254,25 @@ describe('User', function() {
     });
   });
 
+  describe('Access-hook for queries with email NOT case-sensitive', function() {
+    it('Should not throw an error if the query does not contain {where: }', function(done) {
+      User.find({}, function(err) {
+        if (err) done(err);
+        done();
+      });
+    });
```
richardpringle: @bajtos, does this look good? Wasn't too sure about case description...

richardpringle:
@elkorep, Could you please take a look at the failed checks here? Everything passes on my machine...

bajtos:
@richardpringle LGTM :+1: , please squash the commits to a single one.

elkorep:
@richardpringle I am using this as one of my investigations. Look here for some reports on failed checks https://github.com/strongloop-internal/scrum-loopback/issues/630

richardpringle:
Thanks @elkorep, looks like most (if not all) of the issues are concerning the npm registry or the modules within. Is that accurate?

elkorep:
@richardpringle it seems like the failures are caused by npm and the ones with Error status are being aborted for some reason from what I have investigated so far.

bajtos:
@slnode test please

bajtos:
@slnode test please

bajtos:
I am going to ignore failing downstream builds and land this now.

bajtos:
@richardpringle could you please work with @crandmck to get this new feature described in the documentation (docs or apidocs or both)?

PR #1805, opened 2015-11-10
Fix cookie-parser error

superkhau:
`slc loopback` errors out with:

```
Error loading module "loopback":
 [Error: The middleware loopback.cookieParser is not installed.
Run `npm install --save cookie-parser` to fix the problem.]
```


superkhau:
@ritch Merging without review. Simple dependency change in `package.json`.

/cc @cgole

superkhau:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/579

PR #1851, opened 2015-12-03
Hide verificationToken

gausie:
We should never be showing this publically.

### DIFF
--- a/common/models/user.json
+++ b/common/models/user.json
@@ -32,7 +32,7 @@
   "options": {
     "caseSensitiveEmail": true
   },
-  "hidden": ["password"],
+  "hidden": ["password", "verificationToken"],
   "acls": [
     {
       "principalType": "ROLE",
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1320,6 +1320,12 @@ describe('User', function() {
         });
       });
 
+      it('should hide verification tokens from user JSON', function(done) {
+        var user = new User({email: 'bar@bat.com', password: 'bar', verificationToken: 'a-token' });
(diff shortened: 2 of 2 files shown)


gausie:
I don't really understand how the automated tests work - none of them have actually failed.

bajtos:
Hi @gausie, could you please rebase your patch on top of the current master and add a unit-test covering your change? See https://github.com/strongloop/loopback/commit/5b50a99eb366f2af36c6a5b6479adacc04fc1547 for inspiration.

As for the automated tests, our Jenkins CI server is not very stable these days. Make sure all tests pass on your machine and on Travis CI.

bajtos:
@ritch @raymondfeng you are more familiar with user verification process than I am. Do you see any reason why it may be a bad idea to hide "verificationToken" from User JSON?

gausie:
Rebased and unit test added.

bajtos:
@gausie could you please squash the commits into a single one?

Note that `!userJson.verificationToken` is true when the token has a falsy value. You should use `!('verificationToken' in userJson)` instead.

I am wondering if we need to go through the whole `verify` process in the unit-test, isn't it enough to set the property and then check it's not serialized to JSON?

```js
var user = new User({email: 'bar@bat.com', password: 'bar', verificationToken: 'a-token' });
var data = user.toJSON();
assert(!('verificationToken' in data));
```

gausie:
@bajtos, have followed both of your recommendations

bajtos:
@slnode test please

bajtos:
Landed, thank you @gausie for your contribution!

PR #1860, opened 2015-12-07
Fix replication performance

bajtos:
1) Modify `Change.rectify()` to not make any changes to the Change instance
(most notably to not modify the `checkpoint` field) when the tracked
model instance was not changed.

This should improve the performance of change replication as it reduces
the number of unnecessary replications.

For example, before this commit, every run of `rectifyAll` would
trigger a full sync of all clients, because all change instances would
be moved to the current checkpoint.

2) Modify `Change.rectify` to look up the current checkpoint only when
there was actually some change made.

This should improve the performance of `rectifyAll` when called from a
regular timer and there were no changes made since the last call.
Before this commit, `rectifyAll` would perform N calls of
`Checkpoint.current` where N is the number of model instances. With
this commit in place, no call is made.

Connect to strongloop-internal/scrum-cs/issues/397

/to @ritch please review

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -175,24 +175,31 @@ module.exports = function(Change) {
       if (err) throw new Error(err);
     };
 
-    async.parallel([
-      function getCurrentCheckpoint(next) {
-        change.constructor.getCheckpointModel().current(next);
-      },
-      function getCurrentRevision(next) {
-        change.currentRevision(next);
-      }
-    ], doRectify);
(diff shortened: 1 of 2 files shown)


ritch:
Not a fan of console.warn over debug... Otherwise LGTM

bajtos:
> Not a fan of console.warn over debug... 

Fixed in 62d2b0b

PR #1894, opened 2015-12-17
PersistedModel: log rectify/rectifyAll triggers

bajtos:
Add debug logging to identify what operations trigger rectify and rectifyAll calls.

Related: strongloop-internal/scrum-cs#397

/to @ritch please review
/cc @Amir-61


ritch:
LGTM

PR #1896, opened 2015-12-17
[SEMVER-MAJOR] Always use bluebird as promise library

jannyHou:
This pr is created for updating files to use `bluebird` as promise library.
See details here:
Connect to strongloop-internal/scrum-loopback#615

Related: https://github.com/strongloop/loopback-datasource-juggler/pull/790

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,17 +1,9 @@
 exports.createPromiseCallback = createPromiseCallback;
+var Promise = require('bluebird');
 
 function createPromiseCallback() {
   var cb;
-
-  if (!global.Promise) {
-    cb = function() {};
-    cb.promise = {};
-    Object.defineProperty(cb.promise, 'then', { get: throwPromiseNotDefined });
-    Object.defineProperty(cb.promise, 'catch', { get: throwPromiseNotDefined });
-    return cb;
-  }
-
-  var promise = new global.Promise(function(resolve, reject) {
+  var promise = new Promise(function(resolve, reject) {
     cb = function(err, data) {
       if (err) return reject(err);
       return resolve(data);
@@ -19,10 +11,4 @@ function createPromiseCallback() {
   });
   cb.promise = promise;
   return cb;
-}
-
-function throwPromiseNotDefined() {
-  throw new Error(
-    'Your Node runtime does support ES6 Promises. ' +
-    'Set "global.Promise" to your preferred implementation of promises.');
 }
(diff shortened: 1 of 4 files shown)


superkhau:
LGTM but @bajtos should confirm this one before you merge. Also, why so many fails on CI?

test/support.js
```diff
@@ -48,6 +48,3 @@ assert.isFunc = function(obj, name) {
   assert(typeof obj[name] === 'function', name + ' is not a function');
 };
 
-if (!('Promise' in global)) {
```
bajtos: Please remove the trailing empty line at L50.

bajtos:
@jannyHou code changes LGTM. Please rebase your patch on top of the current master and add an entry to `3.0-RELEASE-NOTES.md`. The idea is write a short text explaining loopback users (developers) what this change means for their applications and provide instructions how to upgrade. Also include a link to this pull request to make it easy to look up more details if needed in the future.

@superkhau could you please help here?

package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {
     "async": "^0.9.0",
     "bcryptjs": "^2.1.0",
+    "bluebird": "^2.9.9",
```
bajtos: Please use `3.x` version - the same as we do in juggler (https://github.com/strongloop/loopback-datasource-juggler/pull/790)

bajtos:
> code changes LGTM

"mostly good" would be a better description - we should upgrade to `bluebird@3.x` as part of this patch, see the comment above

package.json
superkhau: Didn't notice the older version of bluebird here. As suggested by @bajtos, please use 3.1.1 (latest on NPM - I checked for you).

lib/utils.js
```diff
@@ -1,17 +1,9 @@
...
-    Object.defineProperty(cb.promise, 'then', { get: throwPromiseNotDefined });
-    Object.defineProperty(cb.promise, 'catch', { get: throwPromiseNotDefined });
-    return cb;
-  }
-
-  var promise = new global.Promise(function(resolve, reject) {
+  var promise = new Promise(function(resolve, reject) {
     cb = function(err, data) {
       if (err) return reject(err);
       return resolve(data);
```
superkhau: I don't think the return is necessary here.

superkhau:
>@superkhau could you please help here?

Sure. @jannyHou Ping me when you're working on this.

lib/utils.js
superkhau: It's fine to leave this since it's in the original code and not part of your changes anyways.

Issue #1893, opened 2015-12-17
bulkUpdate triggers rectifyAll


bajtos:
The current implementation of `bulkUpdate` triggers expensive `rectifyAll` operation whenever a model instance was updated or deleted.

TL;DR: we need to improve [getIdFromWhereByModelId()](https://github.com/strongloop/loopback/blob/3b66fc1c8839f21d25c8af05d6073c83378666ae/lib/persisted-model.js#L1477-L1490) to detect filters that can match either one record with a known id, or no records at all.

**"UPDATE" operations**

To ensure we don't accidentally override changes made while the replication is in progress, we update the record only if all fields have the expected values, see [lib/persisted-model.js#L1245](https://github.com/strongloop/loopback/blob/3b66fc1c8839f21d25c8af05d6073c83378666ae/lib/persisted-model.js#L1245):

```js
Model.updateAll(current.toObject(), data, function(err, result) {
 // ...
});
```

This triggers "after save" hook with a query containing lots of properties in the "where" part, e.g.

```json
{
  "where": {
    "id":"278d7610-a4bb-11e5-afb5-b3fd49f730c6",
    "title": "buy eggs",
    "completed": false,
    "created": 1450355943409
  }
}
```

A filter like this is not recognised by our heuristics detecting whether a single model instance is affected, see [getIdFromWhereByModelId](https://github.com/strongloop/loopback/blob/3b66fc1c8839f21d25c8af05d6073c83378666ae/lib/persisted-model.js#L1477-L1490):

```js
    var whereKeys = Object.keys(where);
    if (whereKeys.length != 1) return undefined;

    var idName = Model.getIdName();
    if (whereKeys[0] !== idName) return undefined;
    // etc.
```

**"DELETE" operations**

The situation is similar as above, we delete the record only if all properties have expected values, see [lib/persisted-model.js#L1327](https://github.com/strongloop/loopback/blob/3b66fc1c8839f21d25c8af05d6073c83378666ae/lib/persisted-model.js#L1327):

```js
Model.deleteAll(current.toObject(), function(err, result) {
  // ...
});
```

This triggers "after save" hook with a query containing lots of properties in the "where" part, e.g.

```json
{
  "id":"278d7610-a4bb-11e5-afb5-b3fd49f730c6",
  "title": "buy eggs",
  "completed": false,
  "created": 1450355943409
}
```

Such query is not recognised as affecting a single model instance only, and thus a full scan (rectifyAll) is triggered.

[...]

PR #1899, opened 2015-12-19
Fix bulkUpdate to not trigger rectifyAll 

Amir-61:
#### Initial issue:
`bulkUpdate` was triggering `rectifyAllChanges`; now it just calls specific updated model instances through calling `rectifyChange`.

#### Related GitHub issue:

Connect to:https://github.com/strongloop/loopback/issues/1893

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -1488,11 +1488,8 @@ module.exports = function(registry) {
   }
 
   function getIdFromWhereByModelId(Model, where) {
-    var whereKeys = Object.keys(where);
-    if (whereKeys.length != 1) return undefined;
-
     var idName = Model.getIdName();
-    if (whereKeys[0] !== idName) return undefined;
+    if (!(idName in where)) return undefined;
 
     var id = where[idName];
     // TODO(bajtos) support object values that are not LB conditions
--- a/test/replication.test.js
+++ b/test/replication.test.js
@@ -52,6 +52,115 @@ describe('Replication / Change APIs', function() {
     };
   });
 
+  describe('optimization check rectifyChange Vs rectifyAllChanges', function() {
+    beforeEach(function initialData(done) {
+      var data = [{name: 'John', surname: 'Doe'}, {name: 'Jane', surname: 'Roe'}];
+      async.waterfall([
+        function(callback) {
+          SourceModel.create(data, callback);
+        },
+        function(data, callback) {
+          SourceModel.replicate(TargetModel, callback);
+        }], function(err, result) {
+        done(err);
+      });
+    });
+
+    it('should call rectifyAllChanges if no id is passed for rectifyOnDelete', function(done) {
+      SourceModel.rectifyChange = function() {
+        return done(new Error('Should not call rectifyChange'));
+      };
+      SourceModel.rectifyAllChanges = function() {
+        return done();
+      };
+      SourceModel.destroyAll({name: 'John'}, function(err, data) {
+        if (err)
+          return done(err);
+      });
+    });
+
+    it('should call rectifyAllChanges if no id is passed for rectifyOnSave', function(done) {
+      SourceModel.rectifyChange = function() {
+        return done(new Error('Should not call rectifyChange'));
+      };
+      SourceModel.rectifyAllChanges = function() {
+        return done();
+      };
+      var newData = {'name': 'Janie'};
+      SourceModel.update({name: 'Jane'}, newData, function(err, data) {
+        if (err)
+          return done(err);
+      });
+    });
+
+    it('rectifyOnDelete for Delete should call rectifyChange instead of rectifyAllChanges', function(done) {
+      TargetModel.rectifyChange = function() {
+        return done();
+      };
+      TargetModel.rectifyAllChanges = function() {
+        return done(new Error('Should not call rectifyAllChanges'));
+      };
+
+      async.waterfall([
+        function(callback) {
+          SourceModel.destroyAll({name: 'John'}, callback);
+        },
+        function(data, callback) {
+          SourceModel.replicate(TargetModel, callback);
+          // replicate should call `rectifyOnSave` and then `rectifyChange` not `rectifyAllChanges` through `after save` operation
+        }
+      ], function(err, results) {
+        if (err)
+          return done(err);
+      });
+    });
+
+    it('rectifyOnSave for Update should call rectifyChange instead of rectifyAllChanges', function(done) {
+      TargetModel.rectifyChange = function() {
+        return done();
+      };
+      TargetModel.rectifyAllChanges = function() {
+        return done(new Error('Should not call rectifyAllChanges'));
+      };
+
+      var newData = {'name': 'Janie'};
+      async.waterfall([
+        function(callback) {
+          SourceModel.update({name: 'Jane'}, newData, callback);
+        },
+        function(data, callback) {
+          SourceModel.replicate(TargetModel, callback);
+          // replicate should call `rectifyOnSave` and then `rectifyChange` not `rectifyAllChanges` through `after save` operation
+        }], function(err, result) {
+        if (err)
+          return done(err);
+      });
+    });
+
+    it('rectifyOnSave for Create should call rectifyChange instead of rectifyAllChanges', function(done) {
+      TargetModel.rectifyChange = function() {
+        return done();
+      };
+      TargetModel.rectifyAllChanges = function() {
+        return done(new Error('Should not call rectifyAllChanges'));
+      };
+
+      var newData = [{name: 'Janie', surname: 'Doe'}];
+      async.waterfall([
+        function(callback) {
+          SourceModel.create(newData, callback);
+        },
+        function(data, callback) {
+          SourceModel.replicate(TargetModel, callback);
+          // replicate should call `rectifyOnSave` and then `rectifyChange` not `rectifyAllChanges` through `after save` operation
+        }
+      ], function(err, result) {
+        if (err)
+          return done(err);
+      });
+    });
+  });
+
   describe('Model.changes(since, filter, callback)', function() {
     it('Get changes since the given checkpoint', function(done) {
       var test = this;


Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
Hi @bajtos,

I have added some test cases for [this issue](https://github.com/strongloop/loopback/issues/1893); could you please take a look?

Thanks a lot!

test/replication.test.js
```diff
@@ -52,6 +52,130 @@ describe('Replication / Change APIs', function() {
...
+          return done(err);
+      });
+    });
+
+    it('rectifyOnSave for Update should call rectifyChange instead of rectifyAllChanges', function(done) {
+      var TargetModel = this.TargetModel;
+      var SourceModel = this.SourceModel;
+      var count = 0;
+
+      TargetModel.rectifyChange = function() {
```
Amir-61: Should this be reverted back in `afterEach()`?

test/replication.test.js
```diff
@@ -52,6 +52,121 @@ describe('Replication / Change APIs', function() {
...
+            SourceModel.replicate(TargetModel, callback);
+          }]
+      }, function(err, results) {
+        done(err);
+      });
+    });
+
+    it('rectifyOnDelete for Delete should call rectifyChange instead of rectifyAllChanges', function(done) {
+      var count = 0;
+      TargetModel.rectifyChange = function() {
```
Amir-61: Should this be reverted back in `afterEach`?

lib/persisted-model.js
```diff
@@ -1489,7 +1489,6 @@ module.exports = function(registry) {
 
   function getIdFromWhereByModelId(Model, where) {
     var whereKeys = Object.keys(where);
-    if (whereKeys.length != 1) return undefined;
 
     var idName = Model.getIdName();
     if (whereKeys[0] !== idName) return undefined;
```
bajtos: This assumes the id is the first key, which is not always the case.

Consider the following: `where: { name: 'Miroslav', id: 123 }`. IIUC, such filter would not be recognised as updating a single property only.

bajtos: I think you need to rewrite the whole block up as follows:

```js
var idName = Model.getIdName();
if (!(idName in where)) return undefined;
```

test/replication.test.js
```diff
@@ -52,6 +52,121 @@ describe('Replication / Change APIs', function() {
     };
   });
 
+  describe('optimization check rectifyChange Vs rectifyAllChanges', function() {
+    beforeEach(function initialData(done) {
+      var data = [{name: 'foo1', lname: 'bar1'}, {name: 'foo2', lname: 'bar2'}];
+      async.auto({
```
bajtos: I think `async.auto` is a bit of overkill here, why is `async.waterfall` not enough?

test/replication.test.js
```diff
@@ -52,6 +52,121 @@ describe('Replication / Change APIs', function() {
...
+    it('rectifyOnDelete for Delete should call rectifyChange instead of rectifyAllChanges', function(done) {
+      var count = 0;
+      TargetModel.rectifyChange = function() {
+        // based on destroyAll where condition it should call rectifyChange two times
+        if (++count == 2) {
+          TargetModel.find(function(err, data) {
+            expect(data.length).to.equal(0);
+            return done(err);
+          });
+        }
```
bajtos: This test mixes test setup with assertions, which makes it more difficult to understand.

Here is a better solution (a mock-up to illustrate the idea):

```js
var calls = [];
TargetModel.rectifyChange = function() { calls.push('rectifyChange'); };
TargetModel.rectifyAllChanges = function() { calls.push('rectifyAllChanges'); };

// run destroy & replicate

expect(calls).to.eql(['rectifyChange', 'rectifyChange']);
```

test/replication.test.js
```diff
@@ -52,6 +52,121 @@ describe('Replication / Change APIs', function() {
...
+      };
+      TargetModel.rectifyAllChanges = function() {
+        return done(new Error('Should not call rectifyAllChanges'));
+      };
+
+      async.auto({
+        destroySourceModel: function(callback) {
+          SourceModel.destroyAll({'or': [{name: 'foo1'}, {lname: 'bar2'}]}, callback);
```
bajtos: IIRC, the current implementation does not support `or` in the filters. Why does this query trigger two `rectifyChange` calls? Also `name` and `lname` do not look like id properties. If that's true, then `rectifyAllChange` should have been triggered instead. We cannot run `rectifyChange` when we don't know the exact model id.

bajtos: Oh, I see, these two new tests are operating at a higher level, verifying that replication of Update/Delete triggers `rectifyChange` in the target model.

In that case you should not use a complex filter for `SourceModel` operation, because that filter is not passed to `TargetModel`. Always use the simplest solution that can reproduce the scenario you want to test. Otherwise readers can get confused into thinking your test is covering more than it does, as happened to me right now.

Here is what your test does under the hood currently:

```
delete all SourceModel instances matching {'or': [{name: 'foo1'}, {lname: 'bar2'}]}
  - delete instance id#1
  - delete instance id#2
replicate changes
  - update TargetModel - delete instance id#1
  - update TargetModel - delete instance id#2
```

It is good idea to keep these two high-level tests in place, please add also another one for CREATE, to get a complete coverage.

To give use higher confidence that `getIdFromWhereByModelId` works as expected for complex filters, you need to add unit-tests mocking out `SourceModel.rectifyChange` and `SourceModel.rectifyAllChanges`.

```js
var calls = [];
SourceModel.rectifyChange = function() { calls.push('rectifyChange'); };
SourceModel.rectifyAllChanges = function() { calls.push('rectifyAllChanges'); };

await SourceModel.destroyAll({'or': [{name: 'foo1'}, {lname: 'bar2'}]}

expect(calls).to.eql(['rectifyAllChanges']); // no id was specified in the query
```

Last but not least, I find the property name `lname` confusing, it looks like a typo. Could you please use `surname` or perhaps `age` instead? Also use real names instead of `bar1` and `bar2`, e.g. `John` and `Jane`.

lib/persisted-model.js
raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:25:50 AM.

This is the only notification you will receive while this person is away.

test/replication.test.js
raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:27:13 AM.

This is the only notification you will receive while this person is away.

test/replication.test.js
raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:31:20 AM.

This is the only notification you will receive while this person is away.

lib/persisted-model.js
raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:25:01 AM.

This is the only notification you will receive while this person is away.

test/replication.test.js
raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:33:22 AM.

This is the only notification you will receive while this person is away.

raymondfeng: I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re: [loopback] WIP:
Rectify change vs rectify all changes [Optimization] (#1899)" sent on
12/21/15 9:45:28 AM.

This is the only notification you will receive while this person is away.

Amir-61: Hi Mirolsav,

> Always use the simplest solution that can reproduce the scenario you want to test.
Sure I make it simpler...

> It is good idea to keep these two high-level tests in place, please add also another one for CREATE, to get a complete coverage.
I already added the test case for CREATE -> https://github.com/strongloop/loopback/pull/1899/files#diff-1884851c88410a9acc23f7bb4dc6c6cfR132

>```
var calls = [];
SourceModel.rectifyChange = function() { calls.push('rectifyChange'); };
SourceModel.rectifyAllChanges = function() { calls.push('rectifyAllChanges'); };

await SourceModel.destroyAll({'or': [{name: 'foo1'}, {lname: 'bar2'}]}

expect(calls).to.eql(['rectifyAllChanges']); // no id was specified in the query
```
Sorry I did not get what you mean by `await` here... Is this a sudo code block or `await` is something I'm not familiar with...?

File: test/replication.test.js (hunk 5 above)
Comment (Amir-61) [inline test/replication.test.js]: In fact if I simplify the fllowing 
```
SourceModel.destroyAll({'or': [{name: 'foo1'}, {lname: 'bar2'}]}, callback);
```
To:
```
SourceModel.destroyAll({name: 'John'}, callback);
```
Where I create/update/delete just one item we don't need this trick either...

**Comment by Amir-61** [comment]
@slnode test please

File: lib/persisted-model.js [hunk 7]
```diff
@@ -1488,12 +1488,9 @@ module.exports = function(registry) {
...
 
   function getIdFromWhereByModelId(Model, where) {
-    var whereKeys = Object.keys(where);
-    if (whereKeys.length != 1) return undefined;
-
     var idName = Model.getIdName();
-    if (whereKeys[0] !== idName) return undefined;
-
+    if (!(idName in where))
+      return undefined;
```
bajtos: The original code used empty lines to visually delimit blocks of code (sub-steps). Please don't remove them!

bajtos:
@Amir-61 I have fixed two spacing problems, squashed all commits into a single one and rewrote the commit message.

Our Jenkins CI is rubbish today, I am going to go ahead and land this without waiting for the results.

bajtos:
Landed.

Amir-61:
@bajtos Thanks Miroslav!

Issue #1864, opened 2015-12-08
Speed up Checkpoint implementation


bajtos:
The current `Checkpoint` implementation keeps all past checkpoints around. When looking up the current checkpoint, we run the query `this.find({ limit: 1, order: 'seq DESC' })` which is not backed up by any index and thus triggers a full table scan.

We should rework the implementation to keep a single global `Checkpoint` instance and update it in-place when bumping up `seq`.

Related: strongloop-internal/scrum-cs#397

bajtos:
Workaround: schedule a regular cleanup of old checkpoints.

Code example:

```js
var app = require('./');
var Model = app.models.Todo; // the change-tracked model
var Change = Model.getChangeModel();
var Checkpoint = Change.getCheckpointModel();

Checkpoint.current(function(err, cp) {
  if (err) return console.error('Cannot obtain current checkpoint:', err);
  console.log('Current checkpoint:', cp);

  Checkpoint.deleteAll({ seq: { lt: cp } }, function(err, info) {
    if (err) return console.error('Cannot delete old checkpoints:', err);
    console.log('Deleted %s old checkpoints', info.count);
  });
});
```

Amir-61:
Connect to: https://github.com/strongloop/loopback/pull/1908

PR #1908, opened 2015-12-22
Checkpoint speedup

Amir-61:
@bajtos This is still a WIP. Please let me know if I'm going the right direction.

#### Problem:
The previous checkpoint implementation was storing all of the previous checkpoints in DB and every time we needed to get the current checkpoint we needed to query all the checkpoints, sort them in descending order and finally get the first checkpoint; in other words, the algorithm was naive and very slow, for every time we needed to query all the checkpoints, which are not indexed nor are they all needed.

#### Fix:
Now we just save one global variable in DB for checkpoint and update it whenever needed. 

Connect to strongloop/loopback#1864

#### Other notes:
We obviously expect the test cases to fail since findOrCreate for memory connector is implemented in another PR under `loopback-datasource-juggler` - see https://github.com/strongloop/loopback-datasource-juggler/pull/815

Thanks!

### DIFF
--- a/common/models/checkpoint.js
+++ b/common/models/checkpoint.js
@@ -27,43 +27,45 @@ module.exports = function(Checkpoint) {
    * Get the current checkpoint id
    * @callback {Function} callback
    * @param {Error} err
-   * @param {Number} checkpointId The current checkpoint id
+   * @param {Number} checkpoint The current checkpoint seq
    */
-
   Checkpoint.current = function(cb) {
     var Checkpoint = this;
-    this.find({
-      limit: 1,
-      order: 'seq DESC'
-    }, function(err, checkpoints) {
-      if (err) return cb(err);
-      var checkpoint = checkpoints[0];
-      if (checkpoint) {
-        cb(null, checkpoint.seq);
-      } else {
-        Checkpoint.create({ seq: 1 }, function(err, checkpoint) {
-          if (err) return cb(err);
-          cb(null, checkpoint.seq);
-        });
-      }
+    Checkpoint._getSingleton(function(err, cp) {
+      cb(err, cp.seq);
     });
   };
 
-  Checkpoint.observe('before save', function(ctx, next) {
-    if (!ctx.instance) {
-      // Example: Checkpoint.updateAll() and Checkpoint.updateOrCreate()
-      return next(new Error('Checkpoint does not support partial updates.'));
-    }
+  Checkpoint._getSingleton = function(cb) {
+    var query = {limit: 1}; // match all instances, return only one
+    var initialData = {seq: 1};
+    this.findOrCreate(query, initialData, cb);
+  };
 
-    var model = ctx.instance;
-    if (!model.getId() && model.seq === undefined) {
-      model.constructor.current(function(err, seq) {
-        if (err) return next(err);
-        model.seq = seq + 1;
-        next();
+  /**
+   * Increase the current checkpoint if it already exists otherwise initialize it
+   * @callback {Function} callback
+   * @param {Error} err
+   * @param {Object} checkpoint The current checkpoint
+   */
+  Checkpoint.bumpLastSeq = function(cb) {
+    var Checkpoint = this;
+    Checkpoint._getSingleton(function(err, cp) {
+      if (err) return cb(err);
+      var originalSeq = cp.seq;
+      cp.seq++;
+      // Update the checkpoint but only if it was not changed under our hands
+      Checkpoint.updateAll({id: cp.id, seq: originalSeq}, {seq: cp.seq}, function(err, info) {
+        if (err) return cb(err);
+        // possible outcomes
+        // 1) seq was updated to seq+1 - exactly what we wanted!
+        // 2) somebody else already updated seq to seq+1 and our call was a no-op.
+        //   That should be ok, checkpoints are time based, so we reuse the one created just now
+        //  3) seq was bumped more than once, so we will be using a value that is behind the latest seq.
+        //    @bajtos is not entirely sure if this is ok, but since it wasn't handled by the current implementation either,
+        //    he thinks we can keep it this way.
+        cb(null, cp);
       });
-    } else {
-      next();
-    }
-  });
+    });
+  };
 };
--- a/test/checkpoint.test.js
+++ b/test/checkpoint.test.js
@@ -1,20 +1,22 @@
 var async = require('async');
 var loopback = require('../');
+var expect = require('chai').expect;
 
-// create a unique Checkpoint model
 var Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');
 
-var memory = loopback.createDataSource({
-  connector: loopback.Memory
-});
-Checkpoint.attachTo(memory);
-
 describe('Checkpoint', function() {
-  describe('current()', function() {
+  describe('bumpLastSeq() and current()', function() {
+    beforeEach(function() {
+      var memory = loopback.createDataSource({
+        connector: loopback.Memory
+      });
+      Checkpoint.attachTo(memory);
+    });
+
     it('returns the highest `seq` value', function(done) {
       async.series([
-        Checkpoint.create.bind(Checkpoint),
-        Checkpoint.create.bind(Checkpoint),
+        Checkpoint.bumpLastSeq.bind(Checkpoint),
+        Checkpoint.bumpLastSeq.bind(Checkpoint),
         function(next) {
           Checkpoint.current(function(err, seq) {
             if (err) next(err);
@@ -23,6 +25,60 @@ describe('Checkpoint', function() {
           });
         }
       ], done);
+    });
+
+    it('Should be no race condition for current() when calling in parallel', function(done) {
+      async.parallel([
+        function(next) { Checkpoint.current(next); },
+        function(next) { Checkpoint.current(next); }
+      ], function(err, list) {
+        if (err) return done(err);
+        Checkpoint.find(function(err, data) {
+          if (err) return done(err);
+          expect(data).to.have.length(1);
+          done();
+        });
+      });
+    });
+
+    it('Should be no race condition for bumpLastSeq() when calling in parallel', function(done) {
+      async.parallel([
+        function(next) { Checkpoint.bumpLastSeq(next); },
+        function(next) { Checkpoint.bumpLastSeq(next); }
+      ], function(err, list) {
+        if (err) return done(err);
+        Checkpoint.find(function(err, data) {
+          if (err) return done(err);
+          // The invariant "we have at most 1 checkpoint instance" is preserved
+          // even when multiple calls are made in parallel
+          expect(data).to.have.length(1);
+          // There is a race condition here, we could end up with both 2 or 3 as the "seq".
+          // The current implementation of the memory connector always yields 2 though.
+          expect(data[0].seq).to.equal(2);
+          // In this particular case, since the new last seq is always 2, both results
+          // should be 2.
+          expect(list.map(function(it) {return it.seq;}))
+            .to.eql([2, 2]);
+          done();
+        });
+      });
+    });
+
+    it('Checkpoint.current() for non existing checkpoint should initialize checkpoint', function(done) {
+      Checkpoint.current(function(err, seq) {
+        expect(seq).to.equal(1);
+        done(err);
+      });
+    });
+
+    it('bumpLastSeq() works when singleton instance does not exists yet', function(done) {
+      Checkpoint.bumpLastSeq(function(err, cp) {
+        // We expect `seq` to be 2 since `checkpoint` does not exist and
+        // `bumpLastSeq` for the first time not only initializes it to one,
+        // but also increments the initialized value by one.
+        expect(cp.seq).to.equal(2);
+        done(err);
+      });
     });
   });
 });
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -898,12 +898,7 @@ module.exports = function(registry) {
 
   PersistedModel.checkpoint = function(cb) {
     var Checkpoint = this.getChangeModel().getCheckpointModel();
-    this.getSourceId(function(err, sourceId) {
-      if (err) return cb(err);
-      Checkpoint.create({
-        sourceId: sourceId
-      }, cb);
-    });
+    Checkpoint.bumpLastSeq(cb);
   };
 
   /**


common/models/checkpoint.js
```diff
@@ -27,43 +28,59 @@ module.exports = function(Checkpoint) {
...
+          cb(null, cp.seq);
+        } else {
+          Checkpoint.init(function(err, data) {
+            cb(err, data.seq);
+          });
+        }
       }
-    });
+    ],
+    cb);
```
bajtos: This particular part is pretty simple and can be coded without `async`. Keep in mind that `async` adds a bit of overhead, so it's better to not overuse it.

Could you please rewrite this without `async`?

common/models/checkpoint.js
```diff
@@ -27,43 +28,59 @@ module.exports = function(Checkpoint) {
...
+
+    async.waterfall([
+      function(cb) {
+        Checkpoint.findOne(cb);
+      },
+      function(cp, cb) {
+        if (cp) {
+          cb(null, cp.seq);
+        } else {
+          Checkpoint.init(function(err, data) {
```
bajtos: I am little bit afraid that this can create a race condition - what if there are two calls of `Checkpoint.current` called at the same time? Both will run `findOne` and get back `null`, and then they will go ahead and create two new checkpoints.

This test may be able to reproduce the race condition (or not, it depends on memory connector internal details). If it does, then it should be added to the test suite:

```js
async.parallel([
  function(next) { Checkpoint.current(next); },
  function(next) { Checkpoint.current(next); }
], function(err, list) {
  if (err) return done(err);
  Checkpoint.find(function(err) {
    if (err) return done(err);
    expect(list).to.have.length(1);
    done();
  });
});
```


I am proposing to use `findOrCreate` instead. Here is a mockup implementation:

```js
Checkpoint.current = function(cb) {
  var Checkpoint = this;
  var query = { limit: 1 }; // match all instances, return only one
  var initialData = { seq: 1 };
  Checkpoint.findOrCreate(query, initialData, function(err, list) {
    if (err) return cb(err);
    if (list.length != 1) {
     err = new Error('Assertion failed, there are ' + list.length + ' Checkpoint instances. Expected exactly 1.');
     return cb(err);
    }
    cb(list[0].seq);
  });
};
```

Thoughts?

The [memory connector](https://github.com/strongloop/loopback-datasource-juggler/blob/dfda177ff7a2f3f6dce924e00c479b546f68f420/lib/connectors/memory.js) does not implement atomic `findOrCreate` now, we will need to fix that as part of this work too. As part of that, this [test setup](https://github.com/strongloop/loopback-datasource-juggler/blob/dfda177ff7a2f3f6dce924e00c479b546f68f420/test/memory.test.js#L815-L823) should be fixed to not provide a dummy implementation (that is actually not atomic!).

common/models/checkpoint.js
```diff
@@ -27,43 +28,59 @@ module.exports = function(Checkpoint) {
...
+      },
+      function(cp, cb) {
+        if (!cp) {
+          Checkpoint.init(cb);
+        } else {
+          cp.seq++;
+          cp.save(cb);
+        }
+      }
+    ], cb);
```
bajtos: Hmm, this will be tricky to implement in an atomic fashion. I wish we had a method like `updateOrCreate(filter, { seq: { $inc: 1 } }, { seq: 1})`.

I think this should be good enough for now:

```js
// todo - this is shared with .current(), refactor it to a shared private function, e.g. _getSingleton()
Checkpoint.findOrCreate({}, { seq: 1 }, function(err, cp) {
  // todo - handle errors

  var originalSeq = cp.seq;
  cp.seq++;

  // Update the checkpoint but only if it was not changed under our hands
  Checkpoint.updateAll({ id: cp.id, seq: originalSeq }, { seq: cp.seq }, function(err, info) {
    // todo handle errors

    // possible outcomes - please preserve this info in a code comment
    // 1) seq was updated to seq+1 - exactly what we wanted!
    // 2) somebody else already updated seq to seq+1 and our call was a no-op. 
    //   That should be ok, checkpoints are time based, so we reuse the one created just now
   //  3) seq was bumped more than once, so we will be using a value that is behind the latest seq.
   //    I am not entirely sure if this is ok, but since it wasn't handled by the current implementation either,
   //    I think we can keep it this way.
    cb(null, cp);
}
```

test/checkpoint.test.js
```diff
@@ -1,28 +1,43 @@
...
-        Checkpoint.create.bind(Checkpoint),
-        Checkpoint.create.bind(Checkpoint),
+        Checkpoint.bump.bind(Checkpoint),
+        Checkpoint.bump.bind(Checkpoint),
         function(next) {
           Checkpoint.current(function(err, seq) {
-            if (err) next(err);
-            expect(seq).to.equal(3);
-            next();
+            expect(seq).to.equal(2);
```
bajtos: Hmm, this looks like a bug to me. When you start with `seq:1` and increment it twice, then you surely should end up with `seq:3`?

common/models/checkpoint.js
Amir-61: >I am little bit afraid that this can create a race condition

Oh you are right! It is prone to race condition!

>The memory connector does not implement atomic findOrCreate now, we will need to fix that as part of this work too. 

I just looked at the implementation of `upsert` (`updateOrCreate`) in memory connector and it seems there is a possibility of race condition there as well  please see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/connectors/memory.js#L224)

I just wanted to know your opinion about using lock via Redis, Semaphore or... to cope with these race conditions... Thought?

common/models/checkpoint.js
Amir-61: Right; this semantic  (`updateOrCreate(filter, { seq: { $inc: 1 } }, { seq: 1})`)  seems the only one which solves the problem of race condition in our case...

Amir-61:
Hey @bajtos,

Could you please review this patch?

Thanks a lot!

common/models/checkpoint.js
```diff
@@ -27,43 +27,46 @@ module.exports = function(Checkpoint) {
...
+  };
 
-    var model = ctx.instance;
-    if (!model.getId() && model.seq === undefined) {
-      model.constructor.current(function(err, seq) {
-        if (err) return next(err);
-        model.seq = seq + 1;
-        next();
+  /**
+   * Increase the current checkpoint If it already exists otherwise initialize it
```
bajtos: typo, should be `... checkpoint if it ...` (lower-case `i`)

common/models/checkpoint.js
```diff
@@ -27,43 +27,46 @@ module.exports = function(Checkpoint) {
...
+   * @callback {Function} callback
+   * @param {Error} err
+   * @param {Object} checkpoint The current checkpoint
+   */
+  Checkpoint.bump = function(cb) {
+    var Checkpoint = this;
+    Checkpoint._getSingleton(function(err, cp) {
+      if (err)  return cb(err);
```
bajtos: spacing, use only one space between `)` and `return`

common/models/checkpoint.js
```diff
@@ -27,43 +27,46 @@ module.exports = function(Checkpoint) {
...
+      var originalSeq = cp.seq;
+      cp.seq++;
+      // Update the checkpoint but only if it was not changed under our hands
+      Checkpoint.updateAll({id: cp.id, seq: originalSeq}, {seq: cp.seq}, function(err, info) {
+        if (err)
+          return cb(err);
```
bajtos: Since you are already using the single-line style above, please stay consistent and do the same here too:

```js
if (err) return cb(err);
```

common/models/checkpoint.js
```diff
@@ -27,43 +27,46 @@ module.exports = function(Checkpoint) {
...
+        // possible outcomes
+        // 1) seq was updated to seq+1 - exactly what we wanted!
+        // 2) somebody else already updated seq to seq+1 and our call was a no-op.
+        //   That should be ok, checkpoints are time based, so we reuse the one created just now
+        //  3) seq was bumped more than once, so we will be using a value that is behind the latest seq.
+        //    I am not entirely sure if this is ok, but since it wasn't handled by the current implementation either,
+        //    I think we can keep it this way.
```
bajtos: Perhaps mention that `I` refers to `@bajtos` here (L66, L67).

test/checkpoint.test.js
```diff
@@ -1,28 +1,43 @@
...
-Checkpoint.attachTo(memory);
+var expect = require('chai').expect;
 
 describe('Checkpoint', function() {
-  describe('current()', function() {
+  var Checkpoint;
+  describe('bump and current', function() {
+    beforeEach(function() {
+      // create a unique Checkpoint model
+      Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');
```
bajtos: Hmm, this does not work exactly the way you think it works. There is only one global model registry, this code is rewriting the entry under the key `TestCheckpoint`. If you really want unique model, then you have to use an unique model name.

However, I think in this particular case, it's ok to share the same model, just attach it to a unique data source in each test.

```js
var Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');
//...
beforeEach(function() {
  var memory = loopback.createDataSource(/*...*/);
  Checkpoint.attachTo(memory);
});
```

common/models/checkpoint.js
```diff
@@ -27,43 +27,46 @@ module.exports = function(Checkpoint) {
...
-        if (err) return next(err);
-        model.seq = seq + 1;
-        next();
+  /**
+   * Increase the current checkpoint If it already exists otherwise initialize it
+   * @callback {Function} callback
+   * @param {Error} err
+   * @param {Object} checkpoint The current checkpoint
+   */
+  Checkpoint.bump = function(cb) {
```
bajtos: I feel `Checkpoint.bump` may not be descriptive enough. How about `Checkpoint.bumpLastSeq` or `Checkpoint.createNextSeq`?

/cc @ritch could you suggest a better name?

test/checkpoint.test.js
```diff
@@ -1,28 +1,43 @@
...
-        Checkpoint.create.bind(Checkpoint),
-        Checkpoint.create.bind(Checkpoint),
+        Checkpoint.bump.bind(Checkpoint),
+        Checkpoint.bump.bind(Checkpoint),
         function(next) {
           Checkpoint.current(function(err, seq) {
-            if (err) next(err);
             expect(seq).to.equal(3);
-            next();
+            next(err);
```
bajtos: I think this may lead to difficult-to-troubleshoot situations. When `Checkpoint.current` fails with an error, your version will report an assertion failure (`expected 3, got 1`) and the error details will be lost. Could you please revert this change? Unless I am missing something...

test/checkpoint.test.js
```diff
@@ -1,28 +1,43 @@
...
+    it('`Checkpoint.current()` for non existing `checkpoint` should initialize `checkpoint`', function(done) {
+      Checkpoint.current(function(err, seq) {
+        expect(seq).to.equal(1);
+        done(err);
+      });
+    });
+
+    it('`bump()` for non existing `checkpoint` should initialize `checkpoint`', function(done) {
+      Checkpoint.bump(function(err, cp) {
+        expect(cp.seq).to.equal(2);
```
bajtos: Hmm, I find this confusing from reader's point of view. Could you please add a comment explaining why `seq` should be `2`? Also perhaps modify test name to something like `bump() works when singleton instance does not exists yet`

bajtos:
@Amir-61 reviewed, see the comment above. You are heading in the right direction!

bajtos:
@Amir-61 could you please add a test to verify that `current` does not create more than one checkpoint, as I outlined in my earlier comment https://github.com/strongloop/loopback/pull/1908#discussion_r48239870?

```js
async.parallel([
  function(next) { Checkpoint.current(next); },
  function(next) { Checkpoint.current(next); }
], function(err, list) {
  if (err) return done(err);
  Checkpoint.find(function(err) {
    if (err) return done(err);
    expect(list).to.have.length(1);
    done();
  });
});
```

bajtos:
Possibly add a test to verify what happens when `bump` is called in parallel too. There should be still only one checkpoint instance, and the `seq` number should have been increased either by 1 or 2 (both outcomes are valid according the reasoning behind our implementation).

Amir-61:
>@Amir-61 could you please add a test to verify that current does not create more than one checkpoint, as I outlined in my earlier comment #1908 (comment)?
```
async.parallel([
  function(next) { Checkpoint.current(next); },
  function(next) { Checkpoint.current(next); }
], function(err, list) {
  if (err) return done(err);
  Checkpoint.find(function(err) {
    if (err) return done(err);
    expect(list).to.have.length(1);
    done();
  });
});
```
I guess you meant this assertion: `expect(data).to.have.length(1);` Am I right?

test/checkpoint.test.js
```diff
@@ -24,5 +25,36 @@ describe('Checkpoint', function() {
...
+        done(err);
+      });
+    });
+
+    it('bumpLastSeq() works when singleton instance does not exists yet', function(done) {
+      Checkpoint.bumpLastSeq(function(err, cp) {
+        // We expect `seq` to be 2 since `checkpoint` does not exist and
+        // `bumpLastSeq` for the first time not only initializes it to one,
+        // but also increments the initialized value by one.
+        expect(cp.seq).to.equal(2);
```
Amir-61: @bajtos, Please note I can check if the `checkpoint` is a new instance or not and if it is new I can just initialize it to 1 as opposed to initialized it and increment the initialized value by one; something like:

```
  Checkpoint.bumpLastSeq = function(cb) {
    var Checkpoint = this;
    Checkpoint._getSingleton(function(err, cp, isNew) {
      if (err) return cb(err);
      if (isNew) return cb(null, cp);
      
      var originalSeq = cp.seq;
      cp.seq++;
      // Update the checkpoint but only if it was not changed under our hands
      Checkpoint.updateAll({id: cp.id, seq: originalSeq}, {seq: cp.seq}, function(err, info) {
        if (err) return cb(err);
        cb(null, cp);
      });
    });
  };
```

But it depends what behavior your would expect in this case...

Amir-61:
Hey @bajtos,

Thanks so much for the review, patience and constructive feedback! I applied your feedback and left a comment; I'm assigning this PR to you now. Please review.

Also it is worth mentioning that test cases pass on local machine using `npm link` for the support for `findOrCreate` in `loopback-datasource-juggler` and here we obviously expect the test cases to fail since `findOrCreate` for memory connector is implemented in another PR under loopback-datasource-juggler - see strongloop/loopback-datasource-juggler#815

Thanks a lot!

Amir-61:
>Possibly add a test to verify what happens when bump is called in parallel too. There should be still only one checkpoint instance, and the seq number should have been increased either by 1 or 2 (both outcomes are valid according the reasoning behind our implementation).

I don't think if this test case is that much helpful; as you already mentioned in the code there is a possibility of race condition in this case for `bumpLastSeq()`, which should not be a big problem as you explained; in other words, the race condition may or may not happen; hence we can not expect a specific value in this case...But again for the reasons you illustrated in the code they should be OK for now...

>Hmm, this will be tricky to implement in an atomic fashion. I wish we had a method like updateOrCreate(filter, { seq: { $inc: 1 } }, { seq: 1}).

:+1:

Amir-61:
>Possibly add a test to verify what happens when bump is called in parallel too. There should be still only one checkpoint instance, and the seq number should have been increased either by 1 or 2 (both outcomes are valid according the reasoning behind our implementation).
I don't think if this test case is that much helpful; as you already mentioned in the code there is a possibility of race condition in this case for bumpLastSeq(), which should not be a big problem as you explained; in other words, the race condition may or may not happen; hence we can not expect a specific value in this case...But again for the reasons you illustrated in the code they should be OK for now...
Please correct me if I'm wrong...

Oh no I'm wrong; it is still safe to always expect `seq` to be 2 based on the current implementation if we have at most 2 parallel calls of `bumpLastSeq` where we are sure we never encounter the third outcome you mentioned in the code 

>3) seq was bumped more than once, so we will be using a value that is behind the latest seq.

I believe if we have more than 2 parallel calls of `bumpLastSeq` there is a possibility of the third outcome; however the odds are low.

Please correct me if I'm wrong...

. Thanks!

test/checkpoint.test.js
```diff
@@ -24,5 +25,51 @@ describe('Checkpoint', function() {
...
+
+    it('Should be no race condition for bumpLastSeq() when calling in parallel', function(done) {
+      async.parallel([
+        function(next) { Checkpoint.bumpLastSeq(next); },
+        function(next) { Checkpoint.bumpLastSeq(next); }
+      ], function(err, list) {
+        if (err) return done(err);
+        Checkpoint.find(function(err, data) {
+          if (err) return done(err);
+          expect(data[0].seq).to.equal(2);
```
bajtos: I am not entirely sure that the current implementation guarantees the value will be exactly `2`, I think it may be possible to arrive at `3`, depending on timing. Perhaps the memory connector guarantees `2`, but I think it won't be the case with external databases.

Could you please add a comment to explain a bit more what we are asserting here?

I would also move the assertion of `data.length` at the top and add a comment too.

We should also check that the values returned by `bumpLastSeq` are all the same (or lower) than the new checkpoint seq.

```js
Checkpoint.find(function(err, data) {
  if (err) return done(err);
  // The invariant "we have at most 1 checkpoint instance" is preserved
  // even when multiple calls are made in parallel
  expect(data).to.have.length(1);

  // There is a race condition here, we could end up with both 2 or 3 as the "seq".
  // The current implementation of the memory connector always yields 2 though.
  expect(data[0].seq).to.equal(2);

  // All checkpoints returned by bumpLastSeq must be 
  //   greater than the old last seq and
  //   lower-or-equal than the new last seq.
  // In this particular case, since the new last seq is always 2, both results
  // should be 2.
  expect(list.map(function(it) { return it.seq; })
    .to.eql([2, 2]);
```

test/checkpoint.test.js
bajtos: Not all connectors report whether `findOrCreate` creates a new instance or not. Let's keep the current implementation as it simpler. I think the extra cost caused by the more complex version is not worth the little gain it brings.

test/checkpoint.test.js
```diff
@@ -1,20 +1,21 @@
...
-
-var memory = loopback.createDataSource({
-  connector: loopback.Memory
-});
-Checkpoint.attachTo(memory);
+var expect = require('chai').expect;
 
 describe('Checkpoint', function() {
-  describe('current()', function() {
+  var Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');
```
bajtos: Would you mind moving this line back to L5 (see above), to keep the diff smaller?

```js
var async = require('async');
var loopback = require('../');
var expect = require('chai').expect;

var Checkpoint = loopback.Checkpoint.extend('TestCheckpoint');

describe('Checkpoint', function() {
  // ...
});
```

bajtos:
@Amir-61 Two more minor comments to address, otherwise the patch LGTM.

I am proposing the following steps:
 - Address the two comments above, no further review is necessary. (You can check with Ritchie in the improbable case that you would run into questions/issues.) 
 - Wait until https://github.com/strongloop/loopback-datasource-juggler/pull/815 is landed to both master and 2.x
 - Land this patch to both `master` and `2.x`

Amir-61:
Applied last feedback and I will merge the code accordingly once strongloop/loopback-datasource-juggler#815 is landed to both master and 2.x

Amir-61:
@slnode test please

bajtos:
@0candy why is this assigned to Raymond? I believe no further reviews are necessary, we are waiting for @Amir-61 to land the other patch in juggler@2.x. Can we assign this back to @Amir-61?

Amir-61:
@bajtos Sorry I forgot to assign it back to myself; I just need to land it to master and `2.x`

Amir-61:
@slnode test please

superkhau:
@Amir-61 This is the one causing CI to fail. http://ci.strongloop.com/job/loopback/2806/

Amir-61:
> @Amir-61 This is the one causing CI to fail. http://ci.strongloop.com/job/loopback/2806/ 

I chatted with @superkhau and apparently failure is not due to this PR. This PR got merged on Jan 19th morning ET and before that the CI was already failing and talked to @bajtos  and we figured out it is a [known issue](https://github.com/strongloop/loopback/pull/1946) and after merging this patch we see some green lights...

superkhau:
Yes, it's related to some windows issues at https://github.com/strongloop/loopback/pull/1946 after speaking with @rmg.

PR #1934, opened 2015-12-31
Fix failing tests

superkhau:
JSCS is detecting improper whitespaces lib/persisted-model.js.


superkhau:
@bajtos Merging without review as its only simple comment changes to fix linting.

Issue #275, opened 2014-05-22
Get rid of peer dependency on loopback-datasource-juggler


bajtos:
LoopBack has a peer dependency on loopback-datasource-juggler. As explained in an older [comment](https://github.com/strongloop/loopback/pull/85#issuecomment-30466138):

> Before peerDependency is adopted, we end up with the following tree:

  ```
  myapp 
    --> loopback --> loopback-datasource-juggler
    --> loopback-connector-mongodb --> loopback-datasource-juggler
  ```

Having multiple instances of juggler is a problem because then there are multiple instances of the global Model registry.

I have looked at several connectors (mysql, postgresql, oracle, rest, soap) and found the following usages of loopback-datasource-juggler:

 - Mysql connector calls `juggler.ModelBuilder.registerType` to register a custom type `Point`. The call is made from `initialize`. 

 - All SQL connectors are extending `juggler.BaseSQL`

----

I am proposing to make the following changes:

 - Move `BaseSQL` and `Connector` (which is super class of BaseSQL) to a new module (e.g. `loopback-connector-base` or `loopback-connector`). Modify connectors to use this new module instead of juggler.
 - To preserve backwards compatibility, juggler should probably require `loopback-connector-base` and expose `BaseSQL` and `Connector` classes. This can be removed in the next major version.
 - Modify the signature of `initialize` function from `initializeDataSource(dataSource, callback)` to `initializeDataSource(dataSource, ModelBuilder, callback)`.
 

The new dependency graph would look like this:

```
myapp 
  --> loopback --> loopback-datasource-juggler -> loopback-connector-base
  --> loopback-connector-mongodb --> loopback-connector-base
```

/to @ritch @raymond what's your opinion on this? I am happy to implement these changes once I get green light from you.

Note: if we are going to release juggler 2.x then it should wait until this change is made, as it allows us to drop juggler dependency on connector base.

---

Connector check list:
 - [x] mysql (v1.3.0)
 - [x] mongodb (v1.3.0)
 - [x] [rest](https://github.com/strongloop/loopback-connector-rest)  (v1.2.1)
 - [x] [soap](https://github.com/strongloop/loopback-connector-soap) (v1.1.0)
 - [x] [postgresql](https://github.com/strongloop/loopback-connector-postgresql) (v1.1.0)
 - [x] [oracle](https://github.com/strongloop/loopback-connector-oracle) (v1.4.0)
[...]

PR #1943, opened 2016-01-06
[SEMVER-MAJOR] Make juggler a regular dependency

bajtos:
Originally, we (ab)used peer dependencies to ensure there is only one instance
of loopback-datasource-juggler in the dependency tree, so that there is only
one singleton instance of model registry. This was very fragile and might not
have worked in certain edge cases.

We have reworked loopback-datasource-juggler and connectors to not rely on
a single juggler instance anymore. As the last step, juggler became a regular
dependency.

Connect to https://github.com/strongloop/loopback/issues/275

I tested this change by installing the new loopback version in loopback-example-app and loopback-example-offline-sync to verify it works correctly with both database and remote connector(s).

/to @raymondfeng please review
/cc @ritch

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -4,3 +4,18 @@ All breaking changes must be described here. When adding a new entry,
 always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
+## loopback-datasource-juggler was moved from peerDependencies to dependencies
+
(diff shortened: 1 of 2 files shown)


raymondfeng:
LGTM

bajtos:
@slnode test please

bajtos:
Windows tests are failing on the master too, I'll ignore those failures and land this patch.

Issue #1939, opened 2016-01-04
Wrong documentation in user.hasPassword


alilloig:
In the [api docs](https://apidocs.strongloop.com/loopback/#user-prototype-haspassword), it is said that the user.hasPassword method returns a boolean, but the method [actually](https://github.com/strongloop/loopback/blob/6d040a98aea0945a7b52d577b67d94705b0c5a7d/common/models/user.js#L307) returns a promise/callback that will resolve to/return a boolean.

jannyHou:
Thank you @alilloig ! You are right, we would correct it's description as soon as possible!

crandmck:
Also, can take a second arg that is a function - see https://github.com/strongloop/loopback/blob/master/common/models/user.js#L307

jannyHou:
Yeah, thank you @crandmck !

jannyHou:
Since https://github.com/strongloop/loopback/pull/1953 is merged, I am closing issue here. Thanks!

Vaelor:
Still not correct in the newest version.
https://apidocs.strongloop.com/loopback/v/2.27.0/#user-prototype-haspassword
Or am I missing something there, this not being the newest version?

jannyHou:
@Vaelor I think it's merged into 3.x, but the latest apidocs version is still 2.x. I will see cherry pick it into 2.x branch.

crandmck:
Thanks, @jannyHou!

gdeckert:
Still incorrect in 2.29.0.

crandmck:
@jannyHou This is still not fixed. 
Somehow your fix in #1953 got clobbered, because it's no longer in `master`.  Perhaps you could try again?

jannyHou:
@crandmck sure let me try again.

crandmck:
ping...?

crandmck:
Thas was done for 2.x in 8fe77b2a062bf0de945f9b899f9e71af6d23546f
@bajtos Would it be possible to publish 2.29.1 so it gets published to http://apidocs.strongloop.com/loopback/?

bajtos:
Done: `loopback@2.29.1`

crandmck:
Confirmed: https://apidocs.strongloop.com/loopback/v/2.29.1/#user-prototype-haspassword

PR #1953, opened 2016-01-11
Fix description for User.prototype.hasPassword

jannyHou:
Connect to: strongloop/loopback#1939
Apply the fix mentioned here:
https://github.com/strongloop/loopback/issues/1939

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -301,7 +301,9 @@ module.exports = function(User) {
    * Compare the given `password` with the users hashed password.
    *
    * @param {String} password The plain text password
-   * @returns {Boolean}
+   * @callback {Function} callback Callback function
+   * @param {Error} err Error object
+   * @param {Boolean} isMatch Returns true if the given `password` matches record
    */
 
   User.prototype.hasPassword = function(plain, fn) {


common/models/user.js
```diff
@@ -301,7 +301,9 @@ module.exports = function(User) {
    * Compare the given `password` with the users hashed password.
    *
    * @param {String} password The plain text password
-   * @returns {Boolean}
+   * @callback {Function} callback Callback function
+   * @param {Error} err Error object
+   * @param {Boolean} isMatch Returns true if the given `password` matches record
```
superkhau: The docs are strange here. hasPassword has two @params, plain and fn, your docs should look something like:

```
@param {String} plain Some description here
@param {Function} fn Some description here
```

Etc.

jannyHou: thank you @superkhau and yeah I have the same question when I edit the docs, I am confused with the naming, like in function [`login`](https://github.com/strongloop/loopback/blob/doc/user-hasPassword/common/models/user.js#L169) it also uses `callback` but the the parameter in `login` called `fn`
And same here in [`logout`](https://github.com/strongloop/loopback/blob/doc/user-hasPassword/common/models/user.js#L282)

And I use `password` as it original is.
I guess the doc names are easy to understand by users.

superkhau: Refer to https://github.com/strongloop-internal/strong-docs#jsdoc-annotations

Issue #1951, opened 2016-01-09
nestRemoting() throws TypeError without {hooks:false} hack


timneedham:
Hi,

Really loving Loopback and the community... but just can't crack this. Spent a few days in the docs and can't find anything obviously wrong... 

If I call ```nestRemoting()``` on a model with the option ```hooks:false``` set, then things are fine.
Having to provide options isn't mentioned in the docs... so I guess this is masking a deeper problem?

If **planet** is the top-level model, which ```hasMany``` **moon[s]**, which in-turn  ```hasMany``` **crater[s]**...

And we've ```planet.nestRemoting('moons')```, then posting to ```/api/planets```, with the ```hooks:false``` option **omitted**, results in:

```
TypeError: Cannot read property 'moon' of undefined
    at EventEmitter.<anonymous> (/home/tim/GithubProjects/nest-remoting-issue/node_modules/loopback/lib/model.js:809:58)
    at EventEmitter.g (events.js:260:16)
    at emitThree (events.js:97:13)
    at EventEmitter.emit (events.js:175:7)
    at /home/tim/GithubProjects/nest-remoting-issue/node_modules/loopback/lib/application.js:279:22
    at Array.forEach (native)
    at EventEmitter.app.handler (/home/tim/GithubProjects/nest-remoting-issue/node_modules/loopback/lib/application.js:278:21)
    at /home/tim/GithubProjects/nest-remoting-issue/node_modules/loopback/server/middleware/rest.js:65:20
    at /home/tim/GithubProjects/nest-remoting-issue/node_modules/loopback/server/middleware/rest.js:72:7
    at iterate (/home/tim/GithubProjects/nest-remoting-issue/node_modules/async/lib/async.js:146:13)
```

Which seems to point to something not being quite right with the way our remotes are set-up? This is the line in ```Model.nestRemoting``` that fails:

```var beforeListeners = remotes.listenerTree.before[toModelName] || {};```

* ...because ```remotes.listenerTree``` is an empty object with no ```before``` key.

* **I've made a repo to replicate this issue [here](https://github.com/timneedham/nest-remoting-issue)**.

* And this might be a clue: once that first post fails, any subsequent posts to the same endpoint work just fine!

I guess I'm doing something stupid/in the wrong order/wrong timing when booting, but just can't find it.

We would really appreciate some assistance as this is (hopefully) the last snag before we commit to Loopback in our next project.

Thanks in advance, keep up the great work!

Tim

[...]

PR #1966, opened 2016-01-17
Should use copy of remotes.listenerTree?

timneedham:
Hi,

_This is my first ever pull request (apologies if it sucks)..._

As part of investigating [Loopback issue #1951] (https://github.com/strongloop/loopback/issues/1951) with @gunjpan's assitance (whereby creating models/relations programatically appears to cause `model.nestRemoting()` to throw a TypeError in the absence of a remotes listenerTree) I think I found something here:

https://github.com/strongloop/loopback/blob/master/lib/model.js#L805-L810

* From what I can tell, a local copy has been made of  `remotes.listenerTree`, which has been defaulted to handle the absence of a `listenerTree`, but shortly-after the original un-defaulted value is used, and not the local copy. That's when I hit the `TypeError`.

A quick change to use the local copy (as per the PR) fixes the immediate issues I was facing in #1951.

Hope this helps?

Tim

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -806,8 +806,8 @@ module.exports = function(registry) {
         listenerTree.before = listenerTree.before || {};
         listenerTree.after = listenerTree.after || {};
 
-        var beforeListeners = remotes.listenerTree.before[toModelName] || {};
-        var afterListeners = remotes.listenerTree.after[toModelName] || {};
+        var beforeListeners = listenerTree.before[toModelName] || {};
+        var afterListeners = listenerTree.after[toModelName] || {};
 
         sharedClass.methods().forEach(function(method) {
           var delegateTo = method.rest && method.rest.delegateTo;


slnode:

bajtos:
@slnode ok to test

slnode:

bajtos:
Hi @timneedham, thank you for the pull request. Could you please add a unit-test to reproduce the problem and verify that your change fixed it?

timneedham:
Hi @bajtos,

I think I've hit my current ceiling at the moment... I did a repo to illustrate the problem here:

https://github.com/timneedham/nest-remoting-issue

...which @gunjpan confirmed in #1951  : apologies, I have tried, expressing this as a unit-test is a bit out of my grasp at the moment.

Kind regards

Tim

slnode:

bajtos:
@timneedham apparently this bug is not possible to reproduce in our Mocha-based test suite :( I opened a new PR to make sure nothing breaks in the current code base, and also reworded the commit message - see #2245.

PR #1989, opened 2016-01-25
[SEMVER-MAJOR] Remove "loopback.autoAttach()"

### LINKED ISSUE #294: Remove `autoAttach`
Details TBD.

This was a failed attempt at allowing models to declare a dependency. As of 2.0 this dependency is more explicit (see `PersistedModel`). This change will require similarly explicit dependency for `Email`, `Push` and similar models.

bajtos:
The method was deprecated since LoopBack 2.0, there is no need to keep it around in 3.0.

Close #294

/to @ritch please review

### DIFF
--- a/lib/loopback.js
+++ b/lib/loopback.js
@@ -363,43 +363,6 @@ loopback.createDataSource = function(name, options) {
 loopback.memory = function(name) {
   return this.registry.memory.apply(this.registry, arguments);
 };
-
-/**
- * Set the default `dataSource` for a given `type`.
- * @param {String} type The datasource type.
- * @param {Object|DataSource} dataSource The data source settings or instance
- * @returns {DataSource} The data source instance.
- *
- * @header loopback.setDefaultDataSourceForType(type, dataSource)
- */
-
-loopback.setDefaultDataSourceForType = function(type, dataSource) {
-  return this.registry.setDefaultDataSourceForType.apply(this.registry, arguments);
-};
-
-/**
- * Get the default `dataSource` for a given `type`.
- * @param {String} type The datasource type.
- * @returns {DataSource} The data source instance
- */
-
-loopback.getDefaultDataSourceForType = function(type) {
-  return this.registry.getDefaultDataSourceForType.apply(this.registry, arguments);
-};
-
-/**
- * Attach any model that does not have a dataSource to
- * the default dataSource for the type the Model requests
- */
-
-loopback.autoAttach = function() {
-  return this.registry.autoAttach.apply(this.registry, arguments);
-};
-
-loopback.autoAttachModel = function(ModelCtor) {
-  return this.registry.autoAttachModel.apply(this.registry, arguments);
-};
-
 /*!
  * Built in models / services
  */
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -118,11 +118,6 @@ Registry.prototype.createModel = function(name, properties, options) {
 
   this._defineRemoteMethods(model, options.methods);
 
-  // try to attach
-  try {
-    this.autoAttachModel(model);
-  } catch (e) {}
-
   return model;
 };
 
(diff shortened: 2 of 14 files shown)


lib/registry.js
```diff
@@ -361,7 +356,7 @@ Registry.prototype.createDataSource = function(name, options) {
   };
 
   if (ds.settings && ds.settings.defaultForType) {
-    this.setDefaultDataSourceForType(ds.settings.defaultForType, ds);
+    console.warn('DataSource option "defaultForType" is no longer supported');
```
ritch: I think we should `assert` or throw... How would the app function with this setting?

lib/loopback.js
```diff
@@ -363,43 +363,6 @@ loopback.createDataSource = function(name, options) {
...
-/**
- * Set the default `dataSource` for a given `type`.
- * @param {String} type The datasource type.
- * @param {Object|DataSource} dataSource The data source settings or instance
- * @returns {DataSource} The data source instance.
- *
- * @header loopback.setDefaultDataSourceForType(type, dataSource)
- */
-
-loopback.setDefaultDataSourceForType = function(type, dataSource) {
```
ritch: Maybe we should deprecate this instead of removing it... to make it easier to catch?

lib/loopback.js
```diff
@@ -363,43 +363,6 @@ loopback.createDataSource = function(name, options) {
...
-  return this.registry.getDefaultDataSourceForType.apply(this.registry, arguments);
-};
-
-/**
- * Attach any model that does not have a dataSource to
- * the default dataSource for the type the Model requests
- */
-
-loopback.autoAttach = function() {
-  return this.registry.autoAttach.apply(this.registry, arguments);
```
ritch: Same comment as above: perhaps throw here?

ritch:
Made a few comments. LGTM either way. Feel free to merge as is.

lib/registry.js
bajtos: The setting will be ignored. If there are any models depending on `autoAttach`, then these models will return an error whenever a DAO method is invoked, something like "The model is not attached to any datasource".

On the second thought, I agree with you it's better to throw an error here, so that the problem is detected as soon as possible (at startup). Will rework.

lib/loopback.js
bajtos: I think the default error message thrown by V8 is good enough:

```
> loopback.setDefaultDataSourceForType()
TypeError: loopback.setDefaultDataSourceForType is not a function
    at repl:1:10
```

bajtos:
Uff, this change is much more tricky than I initially thought. The unit-tests are depending on `autoAttach` a lot. I'd like to clean up that as part of this PR, but it requires quite some effort. I'll post another comment when I make progress.

bajtos:
I have cleaned up the tests in 2fad45c and ff75613. Along the way, I discovered two issues:

 1. `role.isOwner` did not support app-local model registry, this is fixed by 6b699b9
 2. One of the tests was passing only because `User` model was already attached to a datasource, which may not have been the intention (it's hard to tell). I have fixed the test to pass without autoAttach to keep this patch smaller, we can discuss what's the correct behaviour outside of this PR. See https://github.com/strongloop/loopback/pull/1989/commits/da3991f846a6837d1df7415a36b89482fd9685a1#diff-bef842b910c6b8becd88271f314e19f6R343

@ritch could you please review again?

bajtos:
@ritch Also my intention is to back-port all commits except the last one (Remove "loopback.autoAttach()") to 2.x, as they are backwards-compatible. Any objections?

/cc @superkhau @raymondfeng

superkhau:
LGTM

ritch:
LGTM as well. 

> @ritch Also my intention is to back-port all commits except the last one (Remove "loopback.autoAttach()") to 2.x, as they are backwards-compatible. Any objections?

Go for it 👍

bajtos:
@slnode test please

bajtos:
For posterity, I had to make few tweaks to get CI passing on all machines: 8dc9c7e, edda89b and 957db84. I have also added unrelated change 9b39a59 which I needed to be able to troubleshoot the CI failures.

bajtos:
@slnode test please

PR #1999, opened 2016-01-27
Promisify change

jannyHou:
Connect to strongloop-internal/scrum-loopback#751
Promisify methods of model `Change`

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -4,6 +4,7 @@
 
 var PersistedModel = require('../../lib/loopback').PersistedModel;
 var loopback = require('../../lib/loopback');
+var utils = require('../../lib/utils');
 var crypto = require('crypto');
 var CJSON = {stringify: require('canonical-json')};
 var async = require('async');
@@ -77,6 +78,8 @@ module.exports = function(Change) {
     var Change = this;
     var errors = [];
 
+    callback = callback || utils.createPromiseCallback();
+
     var tasks = modelIds.map(function(id) {
       return function(cb) {
         Change.findOrCreateChange(modelName, id, function(err, change) {
@@ -111,6 +114,7 @@ module.exports = function(Change) {
       }
       callback();
     });
+    return callback.promise;
   };
 
   /**
@@ -138,6 +142,7 @@ module.exports = function(Change) {
 
   Change.findOrCreateChange = function(modelName, modelId, callback) {
     assert(loopback.findModel(modelName), modelName + ' does not exist');
+    callback = callback || utils.createPromiseCallback();
     var id = this.idForModel(modelName, modelId);
     var Change = this;
 
@@ -155,6 +160,7 @@ module.exports = function(Change) {
         Change.updateOrCreate(ch, callback);
       }
     });
+    return callback.promise;
   };
 
   /**
@@ -171,9 +177,7 @@ module.exports = function(Change) {
 
     change.debug('rectify change');
 
-    cb = cb || function(err) {
-      if (err) throw new Error(err);
-    };
+    cb = cb || utils.createPromiseCallback();
 
     change.currentRevision(function(err, rev) {
       if (err) return cb(err);
@@ -194,6 +198,7 @@ module.exports = function(Change) {
         }
       );
     });
+    return cb.promise;
 
     function doRectify(checkpoint, rev) {
       if (rev) {
@@ -248,6 +253,7 @@ module.exports = function(Change) {
    */
 
   Change.prototype.currentRevision = function(cb) {
+    cb = cb || utils.createPromiseCallback();
     var model = this.getModelCtor();
     var id = this.getModelId();
     model.findById(id, function(err, inst) {
@@ -258,6 +264,7 @@ module.exports = function(Change) {
         cb(null, null);
       }
     });
+    return cb.promise;
   };
 
   /**
@@ -390,8 +397,11 @@ module.exports = function(Change) {
    */
 
   Change.diff = function(modelName, since, remoteChanges, callback) {
+    callback = callback || utils.createPromiseCallback();
+
     if (!Array.isArray(remoteChanges) || remoteChanges.length === 0) {
-      return callback(null, {deltas: [], conflicts: []});
+      callback(null, {deltas: [], conflicts: []});
+      return callback.promise;
     }
     var remoteChangeIndex = {};
     var modelIds = [];
@@ -455,6 +465,7 @@ module.exports = function(Change) {
         conflicts: conflicts
       });
     });
+    return callback.promise;
   };
 
   /**
--- a/test/change.test.js
+++ b/test/change.test.js
@@ -82,6 +82,38 @@ describe('Change', function() {
     });
   });
 
+  describe('Change.rectifyModelChanges - promise variant', function() {
+    describe('using an existing untracked model', function() {
+      beforeEach(function(done) {
+        var test = this;
+        Change.rectifyModelChanges(this.modelName, [this.modelId])
+          .then(function(trackedChanges) {
+            done();
+          })
+          .catch(done);
+      });
+
+      it('should create an entry', function(done) {
+        var test = this;
+        Change.find()
+          .then(function(trackedChanges) {
+            assert.equal(trackedChanges[0].modelId, test.modelId.toString());
+            done();
+          })
+          .catch(done);
+      });
+
+      it('should only create one change', function(done) {
+        Change.count()
+          .then(function(count) {
+            assert.equal(count, 1);
+            done();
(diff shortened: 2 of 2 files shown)


jannyHou:
@slnode test please

jannyHou:
Please review :) @bajtos thank you.
I promisified 4 functions this time. I will work on the remain ones if I implement it and test it in the right direction.

common/models/change.js
```diff
@@ -96,7 +99,9 @@ module.exports = function(Change) {
     });
 
     async.parallel(tasks, function(err) {
-      if (err) return callback(err);
+      if (err) {
+        callback(err);
+      }
```
bajtos: No need to change this, the `return` is for the anonymous function passed to `async.parallel`.

Please revert this block.

common/models/change.js
```diff
@@ -107,10 +112,11 @@ module.exports = function(Change) {
         var msg = 'Cannot rectify ' + modelName + ' changes:\n' + desc;
         err = new Error(msg);
         err.details = { errors: errors };
-        return callback(err);
+        callback(err);
```
bajtos: ditto.

common/models/change.js
bajtos: Actually the return is important here! If you don't return from the function, then the code below is run even when there was an error.

common/models/change.js
```diff
@@ -138,11 +144,12 @@ module.exports = function(Change) {
 
   Change.findOrCreateChange = function(modelName, modelId, callback) {
     assert(loopback.findModel(modelName), modelName + ' does not exist');
+    callback = callback || utils.createPromiseCallback();
     var id = this.idForModel(modelName, modelId);
     var Change = this;
 
     this.findById(id, function(err, change) {
-      if (err) return callback(err);
+      if (err) callback(err);
```
bajtos: The return is important here! If you don't return from the function, then the code below is run even when there was an error.

Please revert.

test/change.test.js
```diff
@@ -82,6 +82,36 @@ describe('Change', function() {
...
+    describe('using an existing untracked model', function() {
+      beforeEach(function(done) {
+        var test = this;
+        Change.rectifyModelChanges(this.modelName, [this.modelId])
+          .then(function(trackedChanges) {
+            done();
+          })
+          .catch(function(err) {
+            return done(err);
+          });
```
bajtos: Mocha supports promises OOTB, no need to install dummy handlers. Notice that the function passed to `beforeEach` does not take any arguments.

```js
beforeEach(function() {
  return Change.rectifyModelChanges(this.modelName, [this.modelId]);
});
```

BTW for future reference, `then` function takes an optional second argument `errorHandler`, so if we wanted to install both success & error handler, this is a better way:

```js
Change.rectifyModelChanges(...).then(
  function onSuccess() { done(); },
  function onError(err) { done(err); });

// which can be further simplified to

Change.rectifyModelChanges(...).then(
  function onSuccess(trackedChanges) { done(); },
  done);
```

test/change.test.js
```diff
@@ -82,6 +82,36 @@ describe('Change', function() {
...
+      });
+
+      it('should create an entry', function(done) {
+        var test = this;
+        Change.find(function(err, trackedChanges) {
```
bajtos: AFAICT, this is using the old call-back based API. Please rework it to use promises.

test/change.test.js
```diff
@@ -82,6 +82,36 @@ describe('Change', function() {
...
+          assert.equal(trackedChanges[0].modelId, test.modelId.toString());
+          done();
+        });
+      });
+
+      it('should only create one change', function(done) {
```
bajtos: AFAICT, this is using the old call-back based API. Please rework it to use promises. There are more places like these in the diff, please fix them too.

test/change.test.js
bajtos: Hang on. Since these tests are checking whether the promise support works, we must be prepared to correctly handle the case when the support is broken, so that we get helpful test failures (assertion messages).

My example above would fail in a hard-to-debug way.

```js
beforeEach(function() {
  return Change.rectifyModelChanges(this.modelName, [this.modelId]);
});
```

When `Change.rectifyModelChanges` does not return a promise, Mocha will think the beforeEach hook was synchronous and will immediately invoke the next test case. That test case would start before the test environment was set up, it will fail and most likely with a confusing error message.

@jannyHou So you are right in using `function(done)` approach here, please keep it using in all tests you are adding as part of this patch.

However, please do simplify it as I shown above.

```js
beforeEach(function(done) {
  Change.rectifyModelChanges(this.modelName, [this.modelId])
    .then(function(trackedChanges) { done(); }, done);
});
```

test/change.test.js
```diff
@@ -104,6 +134,29 @@ describe('Change', function() {
...
+      beforeEach(function(done) {
+        var test = this;
+        Change.findOrCreateChange(this.modelName, this.modelId)
+        .then(function(result) {
+          test.result = result;
+          done();
+        })
+        .catch(function(err) {
+          return done(err);
+        });
```
bajtos: Please simplify - pass `done` as the second arg of `then`, remove the `catch` clause.

test/change.test.js
```diff
@@ -234,6 +287,25 @@ describe('Change', function() {
...
+      });
+
+      change.currentRevision()
+      .then(function(rev) {
+        assert.equal(rev, test.revisionForModel);
+        done();
+      })
+      .catch(function(err) {
+        done(err);
+      });
```
bajtos: In this particular case, we cannot pass `done` as the second argument of `then`, because the error thrown by `assert.equal` would not be passed to that error handler.

So just simplify the `catch` block here by passing in `done` as the argument, no need to create a dummy argument-forwarding function.

```js
.catch(done);
```

test/change.test.js
```diff
@@ -374,6 +446,27 @@ describe('Change', function() {
...
+
+      Change.diff(this.modelName, 0, remoteChanges)
+        .then(function(diff) {
+          assert.equal(diff.deltas.length, 1);
+          assert.equal(diff.conflicts.length, 1);
+          done();
+        })
+        .catch(function(err) {
+          done(err);
+        });
```
bajtos: Ditto, use `.catch(done)`

test/change.test.js
bajtos: The more I think about this, the more I am leaning towards using `.catch` everywhere.

Deciding whether it's ok to pass `done` as the second argument of `then` or whether a new `catch` call is needed is tricky and may lead to subtle bugs. Consider this example:

```js
.then(function(modelInstance) {
  return modelInstance.myMethod();
}, errorHandler);
```

One may incorrectly assume that the `then` callback will never throw, but what if `myMethod` is not a function, perhaps because `modelInstance` is `null` or `undefined`?

Therefore I am arguing we should be using `.catch(done)` everywhere.

test/change.test.js
bajtos: In the light of https://github.com/strongloop/loopback/pull/1999/files#r51441636, just simplify the `catch` part.

```js
.then(function(result) {
  // ...
})
.catch(done);
```

jannyHou:
Hi @bajtos I applied your comment and promisified one more function `rectify`, could you review it, thanks!

test/change.test.js
jannyHou: Yeah I agree with using `.catch(done)`, thank you!

test/change.test.js
```diff
@@ -219,6 +272,92 @@ describe('Change', function() {
...
+        .then(function(ch) {
+          assert.equal(ch.rev, test.revisionForModel);
+          done();
+        })
+        .catch(done);
+    });
+
+    // This test is a low-level equivalent of the test in replication.test.js
+    // called "replicates multiple updates within the same CP"
+    it('should merge updates within the same checkpoint', function(done) {
```
bajtos: Please remove this test, it's a complex one that checks internal behaviour of checkpoints, not the public API. It's better to not duplicate a complex code like this, plus it's very unlikely this would fail just because of a promise API.

test/change.test.js
```diff
@@ -219,6 +272,92 @@ describe('Change', function() {
...
+
+        model.name += 'updated';
+        model.save(function(err) {
+          test.revisionForModel = Change.revisionForInst(model);
+          next(err);
+        });
+      }
+    });
+
+    it('should not change checkpoint when rev is the same', function(done) {
```
bajtos: ditto.

bajtos:
@jannyHou the patch looks mostly good. Please remove the two promise-based tests per my comments above, squash the commits and you can land this without further review. Also please back-port it to 2.x branch.

PR #1946, opened 2016-01-07
WIP: Safer tests

rmg:
I started cleaning up the tests in order to debug the windows failures, but haven't had time to finish yet.

@superkhau and @jannyHou have each contacted me separately about the failures, so I'm opening this PR to get more eyes.

PR #2017, opened 2016-01-31
Determine which part of test causes error

jannyHou:
Ralated to : https://github.com/strongloop/loopback/pull/1946


jannyHou:
@slnode test please

PR #2030, opened 2016-02-04
Safer tests - step 1

bajtos:
Because the work on #1946 is taking quite long, I would like to land at least the original changes proposed by Ryan, as they are IMO clearly improving the test suite and I don't see any reason why they could not be landed even if they don't completely fix the CI.

/cc @rmg @jannyHou @richardpringle


jannyHou:
@bajtos Sure I agree with land these change first :)

superkhau:
+1 for landing first. :shipit:

PR #2034, opened 2016-02-04
Fix race condition in error handler test

bajtos:
I think I found the last issue causing CI build failure: a race condition in one of the tests.

/cc @richardpringle @jannyHou

### DIFF
--- a/test/error-handler.test.js
+++ b/test/error-handler.test.js
@@ -41,15 +41,23 @@ describe('loopback.errorHandler(options)', function() {
     //arrange
     var app = loopback();
     app.use(loopback.urlNotFound());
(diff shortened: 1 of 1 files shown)


bajtos:
For anybody looking at this pull request and scratching they head.

Here is a simplified version of the original test code:

```js
    app.use(loopback.errorHandler({
      includeStack: false,
      log: function() { done(); }
    }));
  
   request(app).get('/url-does-not-exist').end();
 ```

When the error is logged, error handler calls `done` and ends the test. I think this somehow stop the HTTP server (based on the observed log messages) and thus triggers a failure in supertest. At that time we are running a completely different test, which just adds to confusion.

rmg:
nice catch :+1:

PR #2035, opened 2016-02-04
test: remove errant console.log from test

rmg:
Using console.log like this can result in invalid xml when the xunit
reporter is used.


superkhau:
LGTM

bajtos:
Landed, thank you @rmg

Issue #1998, opened 2016-01-27
Update loopback-core API docs for promises


davidcheung:
Parent: strongloop/loopback/issues/1409
Related: strongloop-internal/strong-docs/issues/63, strongloop-internal/scrum-loopback/issues/615

`common/models/user.js`
Instance Methods
- [x] [User.prototype.hasPassword](https://apidocs.strongloop.com/loopback/#user-prototype-haspassword)
- [x] [User.prototype.verify](https://apidocs.strongloop.com/loopback/#user-prototype-verify) Need update callback fn
- [x] [User.prototype.createAccessToken](https://apidocs.strongloop.com/loopback/#user-prototype-createaccesstoken)

Static Methods
- [x] [User.login](https://apidocs.strongloop.com/loopback/#user-login)
- [x] [User.logout](https://apidocs.strongloop.com/loopback/#user-logout) no param except err
- [x] [User.confirm](https://apidocs.strongloop.com/loopback/#user-confirm) no param except err
- [x] [User.resetPassword](https://apidocs.strongloop.com/loopback/#user-resetpassword) no param except err

`common/models/application.js`
- [x] [Application.register](https://apidocs.strongloop.com/loopback/#application-register) Need update callback fn, `this.create(props, cb);`
- [x] [Application.resetKeys](https://apidocs.strongloop.com/loopback/#application-resetkeys) no param except err, `this.save(cb);`
- [x] [Application.authenticate](https://apidocs.strongloop.com/loopback/#application-authenticate)

`lib/persisted-model.js`
- [x] [PersistedModel.replicate](https://apidocs.strongloop.com/loopback/#persistedmodel-replicate) how to read Array with datatype like `Change []` from callback?

jannyHou:
I am closing it since https://github.com/strongloop/loopback/pull/2040 merged.

PR #2040, opened 2016-02-04
Feature/add promise doc

jannyHou:
Connect to strongloop/loopback#1998

### DIFF
--- a/common/models/application.js
+++ b/common/models/application.js
@@ -108,7 +108,9 @@ module.exports = function(Application) {
    * @param {String} owner Owner's user ID.
    * @param {String} name  Name of the application
    * @param {Object} options  Other options
-   * @param {Function} callback  Callback function
+   * @callback {Function} callback  Callback function
+   * @param {Error} err
+   * @promise
    */
   Application.register = function(owner, name, options, cb) {
     assert(owner, 'owner is required');
@@ -150,6 +152,7 @@ module.exports = function(Application) {
    * @param {Any} appId
    * @callback {Function} callback
    * @param {Error} err
+   * @promise
    */
   Application.resetKeys = function(appId, cb) {
     cb = cb || utils.createPromiseCallback();
@@ -176,7 +179,7 @@ module.exports = function(Application) {
    * - restApiKey
    * - windowsKey
    * - masterKey
-   *
+   * @promise
    */
   Application.authenticate = function(appId, key, cb) {
     cb = cb || utils.createPromiseCallback();
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -79,6 +79,7 @@ module.exports = function(User) {
    * @callback {Function} cb The callback function
    * @param {String|Error} err The error string or object
    * @param {AccessToken} token The generated access token object
+   * @promise
    */
   User.prototype.createAccessToken = function(ttl, options, cb) {
     if (cb === undefined && typeof options === 'function') {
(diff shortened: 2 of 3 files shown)


jannyHou:
hey @bajtos take your time to review it :)
I add @promise doc for each function in https://github.com/strongloop/loopback/issues/1998

Some questions:
1) If a method only has `error` message in callback, `sdocs --preview` will give warning like 
> Resolve object not found in User.confirm()

But doc page looks fine:
Handler | Argument | Description
resolve  | undefined | The resolve handler does not receive any arguments. 

2) promise annotation cannot read Array with datatype like Change [] from callback, so in [PersistedModel.replicate](https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L927)
I am thinking of overwriting the description derived from callback, eg:
```
@promise {Conflict[]} conflicts A list of changes that could not be replicated due to conflicts.
{Object] checkpoints The new checkpoints to use as the "since" argument for the next replication.
```

bajtos:
> >  1) If a method only has `error` message in callback, `sdocs --preview` will give warning like 

> Resolve object not found in User.confirm()

> But doc page looks fine:
Handler | Argument | Description
resolve  | undefined | The resolve handler does not receive any arguments. 

@hacksparrow could you PTAL?

bajtos:
> 2) promise annotation cannot read Array with datatype like Change [] from callback, so in [PersistedModel.replicate](https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L927)
I am thinking of overwriting the description derived from callback, eg:
```
@promise {Conflict[]} conflicts A list of changes that could not be replicated due to conflicts.
{Object] checkpoints The new checkpoints to use as the "since" argument for the next replication.
```

According to [jsdoc](http://usejsdoc.org/tags-param.html), `{Conflict[]}` is a valid syntax.

OTOH, [Google Closure Compiler](https://developers.google.com/closure/compiler/docs/js-for-compiler#types) uses `{Array<Conflict>}` to describe typed array.

@hacksparrow do you happen to know why "promise annotation cannot read Array with datatype like Change[] from callback"? Is this something that can be fixed in strong-docs?

common/models/application.js
```diff
@@ -108,7 +108,9 @@ module.exports = function(Application) {
    * @param {String} owner Owner's user ID.
    * @param {String} name  Name of the application
    * @param {Object} options  Other options
-   * @param {Function} callback  Callback function
+   * @callback {Function} callback  Callback function
+   * @param {Error} err
```
bajtos: @hacksparrow  Is it necessary to add `@param {Error} err` to callbacks that have at most one argument? I.e. `cb(err)` on error, `cb()` on success.

I would prefer if strong-docs was able to add this automatically.

bajtos:
@jannyHou the changes look good to me so far. Please let me know when this is no longer "work in progress".

jannyHou:
@bajtos Thanks for reviewing so quickly, and sorry I put work in progress when I created it but forgot to delete them :) it's done now excepts the problem I mentioned above.
I will wait for Yaapa's reply, I guess he's also quite busy these days.

common/models/application.js
hacksparrow: @bajtos In my opinion, we should make it explicit. Reasons:

1. Helps to give more details about the error, if required.
2. When a human being looks at the source code, if the error param is missing, it will be interpreted as the callback returns nothing. Only `sdocs` will be able to add the missing info. We should avoid magic.

bajtos: Fair enough.

bajtos:
@jannyHou LGTM. Please rebase on top of the current master and squash the commits before landing.

PR #2069, opened 2016-02-16
Validate ownership if parent model userId matches userId

jsheely:
If a model belongsTo a model that then belongsTo a user that userId should be enough to validate $owner ACL

Perhaps in an ideal world we would walk the belongsTo tree until we get to the top or hit the first derived class of User.

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -212,7 +212,7 @@ module.exports = function(Role) {
         // Try to follow belongsTo
         for (var r in modelClass.relations) {
           var rel = modelClass.relations[r];
-          if (rel.type === 'belongsTo' && isUserClass(rel.modelTo)) {
+          if (rel.type === 'belongsTo') {
             debug('Checking relation %s to %s: %j', r, rel.modelTo.modelName, rel);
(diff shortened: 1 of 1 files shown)


fabien:
@jsheely it looks like this PR would require the `fromKey` to be `userId` - however, this is not always the case, since you can subclass User (say, as Account). Your patch would prevent this from working.

jsheely:
I'm not sure what the ```fromKey``` is exactly. I assume it has something to do with the user being able to set the foreign key to whatever they want. 

In either case I refactored the process as I noted originally to walk the ```belongsTo``` tree in order to find a suitable user model and then use that to match the ```user.id``` against the ```userId``` as it was working originally.

bajtos:
@slnode ok to test

slnode:

bajtos:
@jsheely thank you for the pull request. Please add unit-tests to demonstrate that your implementation works as intended and also to prevent regressions in the future.

I am personally not familiar with this part of code. @raymondfeng could you please review this change yourself?

bajtos:
Closing as abandoned.

PR #2108, opened 2016-02-29
Allow built-in token middleware to run repeatedly

### LINKED ISSUE #2106: allow built-in token middleware to be run multiple times
I'm loading the built-in token middleware once globally (all verbs, all paths) during the "auth" phase. However, for the use-case of linking multiple accounts (e.g. via OAuth) - and due to the fact, that my primary auth mechanism is not cookie-based - I'm also loading the token middleware specifically on the `callbackPath` for my auth providers. The latter instance of the middleware is configured to only check for a provider-specific cookie (which is set when accessing the `authPath`) - and is supposed to use that to fill `req.accessToken`. The cookie expires within a configurable amount of minutes and gets removed when the `callbackPath` is requested.

Digging [in the code](https://github.com/strongloop/loopback/blob/master/server/middleware/token.js#L99), I found that the token middleware skips when `req.accessToken !== undefined`. Unfortunately, if the first instance of the middleware can not find valid credentials, it assigns `req.accessToken = null` (which is `!== undefined`).

In your [tests](https://github.com/strongloop/loopback/blob/master/test/access-token.test.js#L182) you say that it "should skip when req.token is already present" - which is not the case when set to `null`.

Am I missing a switch somewhere?
Would you guys be okay with an PR to change the [condition](https://github.com/strongloop/loopback/blob/master/server/middleware/token.js#L99) from `req.accessToken !== undefined` to `req.accessToken && req.accessToken.id`?

benkroeger:
Add two new options:

  - When `enableDoublecheck` is true, the middleware will run
    even if a previous middleware has already set `req.accessToken`
    (possibly to `null` for anonymous requests)

  - When `overwriteExistingToken` is true (and `enableDoublecheck` too),
    the middleware will overwrite `req.accessToken` set by a previous
    middleware instances.

---

fixes #2106 
For details, please see issue description #2106 
@richardpringle

### DIFF
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -1,7 +1,7 @@
 var loopback = require('../');
 var extend = require('util')._extend;
 var Token = loopback.AccessToken.extend('MyToken');
-var ds = loopback.createDataSource({connector: loopback.Memory});
+var ds = loopback.createDataSource({ connector: loopback.Memory });
 Token.attachTo(ds);
 var ACL = loopback.ACL;
 
@@ -32,28 +32,27 @@ describe('loopback.token(options)', function() {
   });
 
   it('should not search default keys when searchDefaultTokenKeys is false',
-  function(done) {
-    var tokenId = this.token.id;
-    var app = createTestApp(
-      this.token,
-      { token: { searchDefaultTokenKeys: false } },
-      done);
-    var agent = request.agent(app);
+    function(done) {
+      var tokenId = this.token.id;
+      var app = createTestApp(
+        this.token, { token: { searchDefaultTokenKeys: false } },
+        done);
+      var agent = request.agent(app);
 
-    // Set the token cookie
-    agent.get('/token').expect(200).end(function(err, res) {
-      if (err) return done(err);
+      // Set the token cookie
+      agent.get('/token').expect(200).end(function(err, res) {
+        if (err) return done(err);
 
-      // Make a request that sets the token in all places searched by default
-      agent.get('/check-access?access_token=' + tokenId)
-        .set('X-Access-Token', tokenId)
-        .set('authorization', tokenId)
-        // Expect 401 because there is no (non-default) place configured where
-        // the middleware should load the token from
-        .expect(401)
-        .end(done);
+        // Make a request that sets the token in all places searched by default
+        agent.get('/check-access?access_token=' + tokenId)
+          .set('X-Access-Token', tokenId)
+          .set('authorization', tokenId)
+          // Expect 401 because there is no (non-default) place configured where
+          // the middleware should load the token from
+          .expect(401)
+          .end(done);
+      });
     });
-  });
 
   it('should populate req.token from an authorization header with bearer token', function(done) {
     var token = this.token.id;
@@ -144,7 +143,7 @@ describe('loopback.token(options)', function() {
         .set('authorization', id)
         .end(function(err, res) {
           assert(!err);
-          assert.deepEqual(res.body, {userId: userId});
+          assert.deepEqual(res.body, { userId: userId });
           done();
         });
     });
@@ -159,7 +158,7 @@ describe('loopback.token(options)', function() {
         .set('authorization', id)
         .end(function(err, res) {
           assert(!err);
-          assert.deepEqual(res.body, {userId: userId, state: 1});
+          assert.deepEqual(res.body, { userId: userId, state: 1 });
           done();
         });
     });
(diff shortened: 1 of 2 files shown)


slnode:

bajtos:
@slnode ok to test

bajtos:
@benkroeger thank you for the pull request. The code changes look reasonable, I appreciate you included a unit-test too!

I have one concern though.

> Digging in the code, I found that the token middleware skips when `req.accessToken !== undefined`. Unfortunately, if the first instance of the middleware can not find valid credentials, it assigns `req.accessToken = null` (which is `!== undefined`).

I think our original assumption was that all instances of the token middleware use the same auth method and if the first instance of the token middleware does find any valid credentials, then the second and all subsequent instances will not find any valid credentials either, and thus we can skip them to improve performance.

Now your use case is different and our assumption no longer holds.

Can we find a way how to support both cases? For example, what if we added a configuration option for the `token` middleware that will instruct it to run even if other token middleware was not able to find any valid credentials?

/cc @raymondfeng @ritch thoughts?

benkroeger:
@bajtos sure, making this a middleware prop shouldn't be a problem. suggestions on the naming? how 'bout `enableDoublecheck`?

benkroeger:
now that I think about it - I actually don't want a subsequent instance to overwrite any existing `req.accessToken` that was set by a previous instance. Meaning, if a user provides a token on the "*regular*" way, that one supersedes any other tokens.

bajtos:
> now that I think about it - I actually don't want a subsequent instance to overwrite any existing req.accessToken that was set by a previous instance. Meaning, if a user provides a token on the "regular" way, that one supersedes any other tokens.

Sure, that makes sense to me as well.

I am thinking about the following rule:

```
req.accessToken is defined and truthy 
   => return next();
req.accessToken is defined but falsy && the new feature is NOT in effect 
   => return next();
go ahead, extract & verify the access token
```

As for the option name, now about `overwriteTokenNotFound`?

benkroeger:
I've implemented a two-step decision now:

It checks for `enableDoublecheck` to be truthy and, it also a token exists already, `overwriteExistingToken` must be truthy as well.

```
if (req.accessToken && req.accessToken.id) {
      rewriteUserLiteral(req, currentUserLiteral);
      return next();
    if (req.accessToken !== undefined) {
      if (!enableDoublecheck) {
        // req.accessToken is defined already (might also be "null" or "false") and enableDoublecheck
        // has not been set --> skip searching for credentials
        rewriteUserLiteral(req, currentUserLiteral);
        return next();
      }
      if (req.accessToken.id && !overwriteExistingToken) {
        // req.accessToken.id is defined, which means that some other middleware has identified a valid user.
        // when overwriteExistingToken is not set to a truthy value, skip searching for credentials.
        rewriteUserLiteral(req, currentUserLiteral);
        return next();
      }
      // continue normal operation (as if req.accessToken was undefined)
    }
```

richardpringle:
@slnode test please!

slnode:

richardpringle:
ok to test

benkroeger:
Since there hasn't been any updates on this issue from the maintainers, I thought it might be because the tests on travis fail. Judging from the indications, it looks like "grunt" does not get installed correctly during the tests for Node.js 5.0. All of them state "grunt: command not found". Thoughts? Any chance we can get this PR moving again?

richardpringle:
@rmg should your latest CI fix have resolved this?
> it looks like "grunt" does not get installed correctly during the tests for Node.js 5.0. All of them state "grunt: command not found". Thoughts?

Node 5 is failing on all platforms.

rmg:
@slnode test please

rmg:
The Node 5 failure looks like it happened before I removed shelljs from the cache (which was the source of the source of the problem when installing deps with npm3). The current failure on node-0.12 on Windows looks like a run of the mill flaky test.

rmg:
@slnode test please

benkroeger:
looks like now tests are failing that previously were successful - without further commits / code change. npm-debug.log says something about failing to install phantom.js

richardpringle:
@slnode test please

benkroeger:
r u guys adding more and more tests? What's about all the dependant tests? Didn't see those in other PRs. And the ones I looked at are failing due to timeout reasons (e.g. dependant strong-arc after > 5 mins).

richardpringle:
@benkroeger, we have always checked the dependants to make sure that nothing is broken down stream. I believe that the only two dependants that failed that we need to worry about are the angular sdk and the angular live-set.

@bajtos PTAL. Are we still having issues with CI?

test/access-token.test.js
```diff
@@ -32,28 +32,27 @@ describe('loopback.token(options)', function() {
...
+
+        // Make a request that sets the token in all places searched by default
+        agent.get('/check-access?access_token=' + tokenId)
+          .set('X-Access-Token', tokenId)
+          .set('authorization', tokenId)
+          // Expect 401 because there is no (non-default) place configured where
+          // the middleware should load the token from
+          .expect(401)
+          .end(done);
+      });
```
bajtos: This looks like pure spacing/formatting change to me, unrelated to the changes made in this patch. Could you please revert?

test/access-token.test.js
```diff
@@ -144,7 +143,7 @@ describe('loopback.token(options)', function() {
         .set('authorization', id)
         .end(function(err, res) {
           assert(!err);
-          assert.deepEqual(res.body, {userId: userId});
+          assert.deepEqual(res.body, { userId: userId });
```
bajtos: Unrelated space change, please revert.

test/access-token.test.js
```diff
@@ -159,7 +158,7 @@ describe('loopback.token(options)', function() {
         .set('authorization', id)
         .end(function(err, res) {
           assert(!err);
-          assert.deepEqual(res.body, {userId: userId, state: 1});
+          assert.deepEqual(res.body, { userId: userId, state: 1 });
```
bajtos: Unrelated space change, please revert.

test/access-token.test.js
```diff
@@ -174,30 +173,83 @@ describe('loopback.token(options)', function() {
...
```
bajtos: Unrelated space change, please revert.

test/access-token.test.js
```diff
@@ -174,30 +173,83 @@ describe('loopback.token(options)', function() {
...
+      app.get('/', function(req, res, next) {
+        res.send(req.accessToken);
       });
+
+      request(app).get('/')
+        .set('Authorization', token.id)
+        .expect(200)
+        .end(function(err, res) {
+          if (err) return done(err);
+          expect(res.body.userId).to.eql(token.userId);
```
bajtos: Why can't we do `expect(res.body).to.eql(token)` here, as we already do in other test cases? What am I missing?

test/access-token.test.js
```diff
@@ -238,16 +290,19 @@ describe('AccessToken', function() {
...
 
           // express helpers
-          param: function(name) { return this._params[name]; },
-          header: function(name) { return this.headers[name]; }
+          param: function(name) {
+            return this._params[name];
+          },
+          header: function(name) {
+            return this.headers[name];
+          }
```
bajtos: Unrelated formatting/space changes, please revert. Please check also all modified lines below, the same comment applies there too.

test/access-token.test.js
benkroeger: because `token` is a full token object including all the properties starting with `__` (e.g. `__dataSource`, `__persisted`).
`res.body` on the other hand is a `JSON.stringify`ed version of this token object and doesn't have those properties.

when `expect(res.body).to.eql(token), you get this:
```
Uncaught AssertionError: expected { Object (id, ttl, ...) } to deeply equal { Object (__cachedRelations, __data, ...) }
  + expected - actual

   {
  -  "created": "2016-03-29T07:27:27.166Z"
  +  "__cachedRelations": {}
  +  "__data": {
  +    "created": [Date: 2016-03-29T07:27:27.166Z]
  +    "id": "BZTUdV1JCTLEnQVtUDkuM4kjZZhiq5igzeHvXi2M9ap69tt4oTYA4EdMZZ7Ay9xP"
  +    "ttl": 1209600
  +    "userId": 123
  +  }
  +  "__dataSource": [undefined]
  +  "__persisted": true
  +  "__strict": false
     "id": "BZTUdV1JCTLEnQVtUDkuM4kjZZhiq5igzeHvXi2M9ap69tt4oTYA4EdMZZ7Ay9xP"
  -  "ttl": 1209600
  -  "userId": 123
   }
```

bajtos: Ah, I see. How about `expect(res.body).to.eql(token.toObject())` or `token.toJSON()`, would that work?

benkroeger: nope, token.toJSON() applies a different transformation to the `created` key (which is of type `Date`)  

````
Uncaught AssertionError: expected { Object (id, ttl, ...) } to deeply equal { Object (id, ttl, ...) }
      + expected - actual

       {
      -  "created": "2016-03-30T08:07:41.452Z"
      +  "created": "Wed Mar 30 2016 10:07:41 GMT+0200 (CEST)"
         "id": "KBZH080uNqZJsfEUReNATJ7tLmZ9AxbSmwMLgtI00jJPnBceln1AnPHQ5WdOz71z"
         "ttl": 1209600
         "userId": 123
       }
```

bajtos:
@benkroeger please let me know when this patch is ready for another round of review. Note that github does not send any notifications when you add a commit, one has to leave a comment for that (ideally including the GH handle of the reviewer).

benkroeger:
alright @bajtos this PR is ready for another review. I reverted the `blank` changes and am using a better comparison on the token object in my unit test

test/access-token.test.js
bajtos: I see. Let's not dwell too much on details and keep what you have now.

Consider this comment thread as resolved.

bajtos:
@slnode test please

bajtos:
Landed via a9c7801380, thank you @benkroeger for your contribution and the patience with us :)

richardpringle:
Finally! :tada:

Agreed, thanks a lot @benkroeger, awesome job!

benkroeger:
Woohoo!
Thanks guys!
@richardpringle @bajtos I just saw that my last commit didn't make it into the merge. It solves a potential "can not read property of null" error that I occurred during one of my experiments. Any chance to get that one in, too?

benkroeger:
forget my previous comment. it's in a9c78013806b54e7733eebddb21b1da5d3af580c
sorry for the confusion!

bajtos:
@benkroeger ouch! Could you please add a unit-test to a9c7801 that demonstrates/reproduces the bug in the current implementation and send a new PR? Please mention my GH handle in the PR description, I'll get it landed ASAP.

PR #2174, opened 2016-03-28
Remove constraint making isStatic required

0candy:
Connect to: https://github.com/strongloop-internal/scrum-loopback/issues/617

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -33,3 +33,33 @@ you will have to check all places where you are using non-standard promise API
 and update them to use Bluebird API instead.
 
 Please see [Related code change](https://github.com/strongloop/loopback/pull/1896) here.
+
+## new method of defining remoting metadata
+
+In 2.0, remote methods were defined as:
+```
+methods: {
+  staticMethod: {
+    isStatic: true,
+    http: { path: '/static' }
+  },
+  instanceMethod: {
+    isStatic: false,
+    http: { path: '/instance' }
+  }
+}
+```
+
+For 3.0, the isStatic flag is no longer required and will be determined from the method name.
+Method name starting with "prototype." will be the same as having isStatic flag set to false.
+```
+methods: {
+  staticMethod: {
+    http: { path: '/static' }
+  },
+  'prototype.instanceMethod': {
+  http: { path: '/instance' }
+}
+```
+
+Please see [related code change](https://github.com/strongloop/loopback/pull/2174) here.
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -4,6 +4,7 @@ var juggler = require('loopback-datasource-juggler');
 var debug = require('debug')('loopback:registry');
 var DataSource = juggler.DataSource;
 var ModelBuilder = juggler.ModelBuilder;
+var deprecated = require('depd')('strong-remoting');
 
 module.exports = Registry;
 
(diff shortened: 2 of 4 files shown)


lib/registry.js
```diff
@@ -258,12 +259,21 @@ Registry.prototype._defineRemoteMethods = function(ModelCtor, methods) {
...
     if (typeof meta.isStatic !== 'boolean') {
-      console.warn('Remoting metadata for "%s.%s" is missing "isStatic" ' +
-        'flag, the method is registered as an instance method.',
-        ModelCtor.modelName,
-        key);
-      console.warn('This behaviour may change in the next major version.');
+      key = isStatic ? key : m[1];
+      meta.isStatic = isStatic;
+    } else {
+      if (meta.isStatic && m) {
```
bajtos: Nitpick: the new block `{...}` starting on L268 is not necessary, it adds extraneous nesting. Here is a version that I consider cleaner:

```js
if (typeof meta.isStatic !== 'boolean') {
  // ...
} else if (meta.isStatic && m) {
  // ...
} else {
  // ...
}
```

test/app.test.js
```diff
@@ -740,7 +740,7 @@ describe('app', function() {
 
       expect(function() {
         app.dataSource('bad-ds', { connector: 'throwing' });
-      }).to.throw(/bad-ds.*throwing/);
+      }).to.throw(/bad-ds.*expected test error/);
```
bajtos: This is suspicious, as it does not relate to the rest of the changes made in this patch. Could you please check? If the tests are failing on the current master, then let's fix them in a different pull request.

In this particular test, it's important to check that the data source name (`throwing`) is included in the error message. The connector error (`expected test error`) is already included by loopback-datasource-juggler.

test/loopback.test.js
```diff
@@ -631,4 +631,73 @@ describe('loopback', function() {
...
+          },
+          'prototype.instanceMethod': {
+            http: { path: '/instance' }
+          }
+        }
+      });
+
+      var methodNames = TestModel.sharedClass.methods().map(function(m) {
+        return m.stringName.replace(/^[^.]+\./, ''); // drop the class name
+      });
```
bajtos: Could you please extract these three lines of code and reuse in all places where it's duplicated now?

Example usage:

```js
var methodNames = getAllMethodNamesWithoutClassName(TestModel);
```

test/loopback.test.js
```diff
@@ -631,4 +631,73 @@ describe('loopback', function() {
...
+
+      expect(methodNames).to.include.members([
+        'staticMethod',
+        'prototype.instanceMethod'
+      ]);
+    });
+
+    it('Incorrectly defined remote method', function() {
```
bajtos: Please make the test read as an English sentence, where the initial `it` stands for the name from the parent `describe` blocks, in this case `... new remote method configuration`. A good test name should describe 1) what scenario/conditions are tested 2) what is the expected outcome. That way the list of tests can be read as a functional specification.

For example:

```js
it('throws when "isStatic" is "true" and method name starts with "prototype."', ...);
```

test/loopback.test.js
```diff
@@ -631,4 +631,73 @@ describe('loopback', function() {
       });
     });
   });
+
+  describe('new remote method configuration', function() {
+    it('Remote method - isStatic specified in method name', function() {
```
bajtos: Per my previous comment, this would be a better test name IMO:

```js
it('treats method names starting with "prototype." as "isStatic:false"', ...)
```

bajtos: Oh, is this test checking both static and prototype methods? In general, it's better to have one test for each different test scenario ("one assert per unit test"). If you think writing (and maintaining) two tests is not worth it, then perhaps this would be a better test name?

```js
it('parses "isStatic" flag from method name', ...)
```

test/loopback.test.js
```diff
@@ -631,4 +631,73 @@ describe('loopback', function() {
...
+          'prototype.instanceMethod': {
+            isStatic: true,
+            http: { path: '/instance' }
+          }
+        }
+      });}).to.throw(Error, new Error('Remoting metadata for' + TestModel.modelName +
+      ' "isStatic" does not match new method name-based style.'));
+    });
+
+    it('Remote method - isStatic match method name', function() {
```
bajtos: Perhaps

```js
it('allows explicit "isStatic" flag when the value matches method name', ...)
```

bajtos:
@0candy one more thing, please add an entry to 3.0-RELEASE-NOTES.

3.0-RELEASE-NOTES.md
```diff
@@ -32,4 +32,32 @@ via `global.Promise`,
...
+Method name starting with "prototype." will be the same as having isStatic flag set to false.
+```
+methods: {
+  staticMethod: {
+    http: { path: '/static' }
+  },
+  'prototype.instanceMethod': {
+  http: { path: '/instance' }
+}
+```
```
bajtos: Please add a link to this pull request, to make it easier for readers to look up discussions that led to this change. AFAICT, other entries include a link to the relevant pull request too.

bajtos:
@0candy one last comment to address. When you fix it, please squash the commits and land the patch, no need for more reviews.

Some of the CI builds are failing, please check that these failures are not caused by your change.

bajtos:
The Travis build is failing because we haven't released a new juggler version with https://github.com/strongloop/loopback-datasource-juggler/commit/0d11186. We are withholding 2.x releases until @Amir-61 finishes his work on replaceOrCreate feature, to ensure we won't need to change the implementation that is already there in a backwards-incompatible way.

lehni:
@0candy, @bajtos please note that this change has cause a new issue, see #3529

PR #2215, opened 2016-04-09
Fix for remote calls to PersistedModel.diff

jouke:
A remote POST call to diff never received the 'since' parameter. By specifying that it is in the request body it works again


slnode:

jouke:
Sorry, but this does not fix it. Preparing another fix

PR #2227, opened 2016-04-13
add missing unit tests for #2108

### LINKED ISSUE #2226: Add missing unit test for token middleware
as discussed in #2108

benkroeger:
subsequent token middleware tries to read `token.id` when `enableDoublecheck: true`. That caused a "Cannot read property `id` of `null`" error when the first middleware didn't actually find a valid accessToken.

@bajtos 

fixes #2226

### DIFF
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -249,6 +249,39 @@ describe('loopback.token(options)', function() {
         });
     });
 
+    it('should overwrite invalid existing token (is !== undefined and has no "id" property) ' +
+      ' when enableDoubkecheck is true',
+    function(done) {
+      var token = this.token;
+
+      app.use(function(req, res, next) {
+        req.accessToken = null;
+        next();
+      });
+
+      app.use(loopback.token({
+        model: Token,
+        enableDoublecheck: true,
+      }));
+
+      app.get('/', function(req, res, next) {
+        res.send(req.accessToken);
+      });
+
+      request(app).get('/')
+        .set('Authorization', token.id)
+        .expect(200)
+        .end(function(err, res) {
+          if (err) return done(err);
+          expect(res.body).to.eql({
+            id: token.id,
+            ttl: token.ttl,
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
@slnode ok to test

test/access-token.test.js
```diff
@@ -249,6 +249,39 @@ describe('loopback.token(options)', function() {
...
+      app.get('/', function(req, res, next) {
+        res.send(req.accessToken);
+      });
+
+      request(app).get('/')
+        .set('Authorization', token.id)
+        .expect(200)
+        .end(function(err, res) {
+          if (err) return done(err);
+          expect(res.body).to.eql({
```
bajtos: Is it necessary to check all token properties here? I am worried that if we add a new Token property in the future, we will have to fix way too many tests that don't really depend on specific Token properties.

It think it is enough to check that `res.body.id` is the same as `token.id`, isn't it?

test/access-token.test.js
```diff
@@ -249,6 +249,39 @@ describe('loopback.token(options)', function() {
...
+            id: token.id,
+            ttl: token.ttl,
+            userId: token.userId,
+            created: token.created.toJSON(),
+          });
+          done();
+        });
+    });
```
bajtos: Please add an empty line after the test case to serve as a visual delimiter.

bajtos:
@benkroeger thank you for the pull request. The patch looks good in general, I pointed out two details to improve in the comments above.

test/access-token.test.js
benkroeger: in my last unit test, I did exactly that - and you commented that only checking on `id` wouldn't be enough.
Let me know which one you prefer

bajtos: I see, sorry for me not being consistent! Let's keep what you have now.

bajtos:
Landed via 103935c, thank you for the contribution!

Issue #1612, opened 2015-08-20
Support returning promises in registerResolver instead of using "done" callback


gausie:
As in title, this would be a great alternative.

superkhau:
Would you like to submit a pull request?

ritch:
The offending line: https://github.com/strongloop/loopback/blob/06cece038ec39db1fd61184aef3a9a0ca2140d3a/common/models/role.js#L295

Also added to the promise support issue: https://github.com/strongloop/loopback/issues/418

bajtos:
Relabelled as "Beginner Friendly" + "Feature".

@gausie I am afraid this particular issue is not very high in our list of priorities, would you mind contributing the improvement yourself? See https://github.com/strongloop/strong-remoting/commit/a7f5d12c0129932236cfcc7dc6fb824edf40d8a4 for inspiration.

gausie:
Yeah I'll take a look into it, I understand that it's not super important.

superkhau:
@jannyHou Reassigning to you as this may be related to the promise work you're doing.

jannyHou:
@superkhau Thank you I am choosing the model to be promisified next, then I will take `Role`.

jannyHou:
@gausie @bajtos @superkhau @ritch I am trying to implement this feature but got confused by it's requirement, could anyone take a look? Thanks a lot:

From my understanding promisify means a function either executes the callback or create a promise then return the promise, like:
```js
myFunction(param1, param2, callback) {
  callback = callback || createPromise();
  callback;
  return callback.promise;
}
```
So I feel reasonable to promisify functions in model `Role` like:
[Roles.isOwner](https://github.com/strongloop/loopback/blob/88d41f84fd8eac997a31388ccdbd3ce026ab0dfe/common/models/role.js#L177)
[Roles.isInRole](https://github.com/strongloop/loopback/blob/88d41f84fd8eac997a31388ccdbd3ce026ab0dfe/common/models/role.js#L279)
[Roles.getRoles](https://github.com/strongloop/loopback/blob/88d41f84fd8eac997a31388ccdbd3ce026ab0dfe/common/models/role.js#L370)

But I am confused with how to promisify [registerResolver](https://github.com/strongloop/loopback/blob/88d41f84fd8eac997a31388ccdbd3ce026ab0dfe/common/models/role.js#L126)

```js
Roles.registerResolver(role, resolver) {
    The resolver function is defined by user, and from the doc, 
[...]

PR #2259, opened 2016-04-26
Resolver support return promise

jannyHou:
Connect to strongloop/loopback#1612

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -120,8 +120,9 @@ module.exports = function(Role) {
   /**
    * Add custom handler for roles.
    * @param {String} role Name of role.
-   * @param {Function} resolver Function that determines if a principal is in the specified role.
-   * Signature must be `function(role, context, callback)`
+   * @param {Function} resolver Function that determines
+   * if a principal is in the specified role.
+   * Should provide a callback or return a promise.
    */
   Role.registerResolver = function(role, resolver) {
     if (!Role.resolvers) {
@@ -289,7 +290,14 @@ module.exports = function(Role) {
     var resolver = Role.resolvers[role];
     if (resolver) {
       debug('Custom resolver found for role %s', role);
-      resolver(role, context, callback);
+
+      var promise = resolver(role, context, callback);
+      if (promise && typeof promise.then === 'function') {
+        promise.then(
+          function(result) { callback(null, result); },
+          callback
+        );
+      }
       return;
     }
 
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -8,6 +8,7 @@ var Application = loopback.Application;
 var ACL = loopback.ACL;
 var async = require('async');
 var expect = require('chai').expect;
+var Promise = require('bluebird');
 
 function checkResult(err, result) {
   // console.log(err, result);
@@ -172,6 +173,14 @@ describe('role model', function() {
   });
 
   it('should support owner role resolver', function() {
+    Role.registerResolver('returnPromise', function(role, context) {
+      return new Promise(function(resolve) {
+        process.nextTick(function() {
+          resolve(true);
+        });
+      });
+    });
+
     var Album = ds.createModel('Album', {
       name: String,
       userId: Number,
@@ -186,10 +195,16 @@ describe('role model', function() {
     });
 
     User.create({ name: 'Raymond', email: 'x@y.com', password: 'foobar' }, function(err, user) {
+      Role.isInRole('returnPromise', { principalType: ACL.USER, principalId: user.id },
+      function(err, yes) {
+        assert(!err && yes);
+      });
+
       Role.isInRole(Role.AUTHENTICATED, { principalType: ACL.USER, principalId: user.id },
       function(err, yes) {
         assert(!err && yes);
       });
+
       Role.isInRole(Role.AUTHENTICATED, { principalType: ACL.USER, principalId: null },
       function(err, yes) {
         assert(!err && !yes);


jannyHou:
@bajtos Please review :) Thanks.

common/models/role.js
```diff
@@ -240,6 +241,15 @@ module.exports = function(Role) {
     Role.isAuthenticated(context, callback);
   });
 
+  Role.registerResolver(Role.EVERYONE_PROMISE, function(role, context) {
```
bajtos: I find this suspicious. Where is `EVERYONE_PROMISE` coming from? I would not expect that adding promise support would require a new role. Am I missing something?

common/models/role.js
```diff
@@ -289,7 +299,11 @@ module.exports = function(Role) {
     var resolver = Role.resolvers[role];
     if (resolver) {
       debug('Custom resolver found for role %s', role);
-      resolver(role, context, callback);
+
+      var promise = resolver(role, context, callback);
+      if (promise instanceof Promise) {
```
bajtos: This will not work when `resolver` uses a different promise implementation. Promises were designed for interoperability, basically the only requirement is to provide a `then` method. See https://github.com/strongloop/loopback/issues/1612#issuecomment-214748981

```
      if (result && typeof result.then === 'function') {
        result.then(function() { next(); }, next);
      }
```

common/models/role.js
```diff
@@ -289,7 +299,11 @@ module.exports = function(Role) {
...
+        promise.then(callback).catch(callback);
```
bajtos: This version is subtly different from my proposal above: when `callback` throws an error, it will be invoked again with the error thrown:

```js
callback = function(err) {
  console.log(err || 'ok');
  throw new Error('bang!');
}
```

The result: resolver is called only once, but two messages are printed:

```
ok
Error: bang!
```

The implementation I proposed above should treat the error thrown from callback as an unhandled error (promise rejection), which is IMO more correct.

test/role.test.js
```diff
@@ -172,6 +173,14 @@ describe('role model', function() {
   });
 
   it('should support owner role resolver', function() {
+    Role.registerResolver(Role.EVERYONE_PROMISE, function(role, context) {
+      process.nextTick(function() {
```
bajtos: This role resolver returns `undefined`, which is probably not what you wanted. The trick is to create the promise first:

```js
return new Promise(function(resolve, reject) {
  process.nextTick(resolve);
}
```

bajtos: Oh, I did not notice the promise needs to be resolved with a boolean arg.

```js
return new Promise(function(resolve, reject) {
  process.nextTick(function() { 
    resolve(null, true); 
  });
}

Comment (bajtos) [inline test/role.test.js]: Also notice that Promise `resolve` function does not have error-first signature, you pass only the result.

```js
// wrong
resolve(null, true);
resolve(err);

// correct
resolve(true);
reject(err);
```

common/models/role.js
bajtos: I think you should not create a named constant like `Role.EVERYONE_PROMISE`, and simply use an anonymous string in the tests.

```js
Role.registerResolver('TEST_PROMISE_EVERYONE', /*...*/);
Role.isInRole('TEST_PROMISE_EVERYONE', /*...*/);
```

test/role.test.js
jannyHou: Oops, that's a careless error, yeah it shouldn't resolve error msg.

common/models/role.js
jannyHou: @bajtos I registered a new resolver since we don't have any existed resolver that returns promise.
They all use this signature: [function(role, context, callback)](https://github.com/strongloop/loopback/blob/master/common/models/role.js#L124)

But my test need to verify signature that ignore callback but return a promise.

bajtos: > But my test need to verify signature that ignores callback but return a promise.

Can you register such resolver in the test only?

jannyHou: Sure, sorry I didn't notice that it's also in role.js, it should only be defined in test.

jannyHou:
@bajtos All macos servers are failing, I am confirming with Ryan is it a problem with macos config. Since all green ci builds skips those macos servers.

common/models/role.js
```diff
@@ -2,6 +2,7 @@ var loopback = require('../../lib/loopback');
 var debug = require('debug')('loopback:security:role');
 var assert = require('assert');
 var async = require('async');
+var Promise = require('bluebird');
```
bajtos: AFAICT, this is no longer used, please remove.

test/role.test.js
```diff
@@ -186,10 +195,16 @@ describe('role model', function() {
     });
 
     User.create({ name: 'Raymond', email: 'x@y.com', password: 'foobar' }, function(err, user) {
+      Role.isInRole('returnPromise', { principalType: ACL.USER, principalId: user.id },
+      function(yes) {
```
bajtos: Note that `Role.isInRole` is keeps callback in Node's error-first style. This should be `function(err, yes)` with the same body as in other tests below.

common/models/role.js
```diff
@@ -289,7 +290,11 @@ module.exports = function(Role) {
     var resolver = Role.resolvers[role];
     if (resolver) {
       debug('Custom resolver found for role %s', role);
-      resolver(role, context, callback);
+
+      var promise = resolver(role, context, callback);
+      if (promise && typeof promise.then === 'function') {
+        promise.then(callback, callback);
```
bajtos: This will report `resolve` result the same way as `reject` result.

The correct version:

```js
promise.then(
  function(result) { callback(null, result),
  callback // this is the (almost) the same as function(err) { callback(err); }
);
```

jannyHou:
@slnode test please

common/models/role.js
jannyHou: :( this explains why macos all fail. Changed.

jannyHou:
@bajtos Please review:)

bajtos:
Lovely, LGTM.

Do we need to update any (API) documentation as part of this change?

jannyHou:
@bajtos sure, I updated a similar one in strong-remoting in https://github.com/strongloop/strong-remoting/pull/291
So will update this change as well.

jannyHou:
@superkhau could you help review the jsdoc part of this pr? Thanks. @bajtos already reviews the code part.

common/models/role.js
```diff
@@ -121,7 +121,11 @@ module.exports = function(Role) {
    * Add custom handler for roles.
    * @param {String} role Name of role.
    * @param {Function} resolver Function that determines if a principal is in the specified role.
-   * Signature must be `function(role, context, callback)`
+   * Signature could be using callback: `function(role, context, callback)`
+   * or returning promise:
+   * `function(role, context) {
+   *    return new Promise(function(resolve, reject){});
+   *  }`
```
superkhau: A function that determines if a principle is in the specified
role. Should provide a callback or return a promise.

common/models/role.js
```diff
@@ -121,7 +121,11 @@ module.exports = function(Role) {
    * Add custom handler for roles.
    * @param {String} role Name of role.
    * @param {Function} resolver Function that determines if a principal is in the specified role.
```
superkhau: Wrap to 80 chars.

superkhau:
@jannyHou commented

bajtos:
@jannyHou @superkhau I'll leave the jsdoc tweaks to you, no further review from me is needed.

bajtos:
Please back-port this change to `2.x` after it lands on `master`.

common/models/role.js
jannyHou: @superkhau Thank you:) Applied.

jannyHou:
@superkhau Please review:)

superkhau:
LGTM

PR #2262, opened 2016-04-26
[DO-NOT-MERGE][WIP] Test validation should inherit to submodels

davidcheung:
connect to strongloop/loopback/issues/1485

test cases created for this fix https://github.com/strongloop/loopback-datasource-juggler/pull/913


bajtos:
@davidcheung please find a way how to write these tests in juggler, so that they can accompany the implementation. We can keep some of them in loopback too, to make sure the feature works here, but the change should be tested mainly in the juggler.

davidcheung:
validation should no longer be inherited, closing PR

Issue #2111, opened 2016-03-01
The name attribute on the "Role" model should be unique


javefang:
For now I can create many Role instances with the same name. But since roles are referenced by name in the ACLs of the model JSON definition files. It doesn't make sense to allow creating roles with the same name.

Looking at the example https://docs.strongloop.com/display/public/LB/Creating+a+default+admin+user, If I run the code example `/server/boot/script.js` every time API starts, it's going to create a new role with name "admin" (but different id) every time.

Plus, the implementation for `Role.isInRole()` (https://github.com/strongloop/loopback/blob/master/common/models/role.js#L325), seems dangerous to use `findOne` to look up role by name without adding uniqueness constraint on the `name` attribute in the first place.

javefang:
@richardpringle apologize if I am wrong, but shouldn't this issue be considered as a bug? surely if can create two roles with the same name and break correct behavior when referencing the role in ACL?

richardpringle:
@javefang, absolutely no need to apologize! I would say this is a feature simply because of the implementation. It is assumed that the developer would not create multiple *roles* with the same name. I can't imagine any scenario where one would do so (by accident or not). However, I agree that it could be a good idea to implement the Role Class is such a way that naming is enforced instead of assumed.

The behavior is as expected and therefore it is not a bug. It could however be improved, and thus a feature! Make sense? In any case, it's a matter of semantics.

@raymondfeng, do you have any thoughts here?

@javefang, if you would like to see this feature implemented right away, the best thing to do would be to make the change yourself! [Here's a link for instructions on code contribution](https://github.com/strongloop/loopback/wiki/Contributing-code) if you haven't done so before. If you do choose to open a PR, mention me in a comment and I will make sure that it gets reviewed as soon as possible.

supasate:
Hi @richardpringle @raymondfeng @javefang, let's me take this feature.

PR #2269, opened 2016-04-29
Add feature to not allow duplicate role name

supasate:
This PR is for #2111.

### DIFF
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -87,6 +87,22 @@ describe('role model', function() {
     });
   });
 
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      if (err) return done(err);
+
+      Role.create({ name: 'userRole' }, function(err, role) {
+        expect(err).to.exist;
+        expect(err).to.have.property('name', 'ValidationError');
+        expect(err).to.have.deep.property('details.codes.name');
+        expect(err.details.codes.name).to.contain('uniqueness');
+        expect(err).to.have.property('statusCode', 422);
+
+        done();
+      });
+    });
+  });
+
   it('should automatically generate role id', function() {
     User.create({ name: 'Raymond', email: 'x@y.com', password: 'foobar' }, function(err, user) {
       // console.log('User: ', user.id);
@@ -240,6 +256,7 @@ describe('role model', function() {
         password: 'jpass',
       }, function(err, u) {
         if (err) return done(err);
+
         user = u;
         User.create({
           username: 'mary',
@@ -247,15 +264,18 @@ describe('role model', function() {
           password: 'mpass',
         }, function(err, u) {
           if (err) return done(err);
+
           Application.create({
             name: 'demo',
           }, function(err, a) {
             if (err) return done(err);
+
             app = a;
             Role.create({
               name: 'admin',
             }, function(err, r) {
               if (err) return done(err);
+
               role = r;
               var principals = [
                 {
@@ -279,7 +299,9 @@ describe('role model', function() {
     it('should resolve user by id', function(done) {
       ACL.resolvePrincipal(ACL.USER, user.id, function(err, u) {
         if (err) return done(err);
+
         expect(u.id).to.eql(user.id);
+
         done();
       });
     });
@@ -287,7 +309,9 @@ describe('role model', function() {
     it('should resolve user by username', function(done) {
       ACL.resolvePrincipal(ACL.USER, user.username, function(err, u) {
         if (err) return done(err);
+
         expect(u.username).to.eql(user.username);
+
         done();
       });
     });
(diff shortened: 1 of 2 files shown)


slnode:

supasate:
@bajtos Could you help verify this PR?

test/role.test.js
```diff
@@ -87,6 +87,20 @@ describe('role model', function() {
     });
   });
 
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      expect(err).to.be.null;
```
superkhau: if (err) return done(err);

superkhau:
LGTM, @bajtos for final OK and merge.

test/role.test.js
supasate: Fixed in a new squashed commit :)

supasate:
@superkhau Normally, do you 'squash commits in each revised code' **or** 'make new commits for each revised code and squash them when there's no more comment'? Which one is the preferred workflow?

superkhau:
@supasate Normally we squash commits, but in your case I asked to do a separate commit because the indent changes have nothing to do with the actual feature. I also didn't want to put you through the work of submitting a separate PR for the cleanup.

test/role.test.js
```diff
@@ -87,6 +87,20 @@ describe('role model', function() {
     });
   });
 
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      if (err) return done(err);
```
superkhau: Add an empty line to separate logic.

test/role.test.js
```diff
@@ -87,6 +87,20 @@ describe('role model', function() {
...
+      Role.create({ name: 'userRole' }, function(err, role) {
+        expect(err).to.exist;
+        expect(err).to.have.property('name', 'ValidationError');
+        expect(err).to.have.property('message');
+        expect(err.message).to.have.string('`name` already exists');
+        expect(err).to.have.property('statusCode', 422);
```
superkhau: Add empty line here too.

supasate: Between which line numbers? I followed convention from lines 293 - 372.

superkhau: Nah, its fine. It's just a nitpick. We actually need to fix all the other ones. There should never be two empty lines, just single empty lines that group logical items (the lines that are related) together. Like returning the error is separate from calling Role.create, etc.

supasate: I've fixed them and also clean up some unnecessary comments left in the code.

test/role.test.js
```diff
@@ -287,6 +298,7 @@ describe('role model', function() {
     it('should resolve user by username', function(done) {
       ACL.resolvePrincipal(ACL.USER, user.username, function(err, u) {
         if (err) return done(err);
+
         expect(u.username).to.eql(user.username);
         done();
```
superkhau: can you put a space btwn the dones too? completing the test has nothing to do with the assertion logic.

superkhau:
@supasate Thanks for the fixes. Please add the spaces between the dones and squash the commits. Should be good after that. ;)

superkhau:
@supasate Don't squash the unrelated changes together though. I assume you know this (don't squash the commit with the console log removals).

supasate:
@superkhau Done. Please check.

test/role.test.js
```diff
@@ -87,9 +84,24 @@ describe('role model', function() {
...
 
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      if (err) return done(err);
+
+      Role.create({ name: 'userRole' }, function(err, role) {
+        expect(err).to.exist;
+        expect(err).to.have.property('name', 'ValidationError');
+        expect(err).to.have.property('message');
+        expect(err.message).to.have.string('`name` already exists');
```
bajtos: Rather than checking exact error message, which may change in the future and thus break this test, it's better to check the error code, which should remain the same. The code can be found in the array in `error.details.codes.name`.

You can print the error object to see what I mean, I think it should look like this:

```js
{
  name: 'ValidationError',
  message: '`name` already exists',
  details: {
    codes: {
      name: [
        'uniquenes'
      ]
    }
  },
  messages: {
    name: [
      'already exists'
    ]
  }
}
```

bajtos:
One more nitpick, the PR looks good to me otherwise. @superkhau I'll let you to finish the review and land the patch. I think it should be back-ported to 2.x too, what do you think?

test/role.test.js
supasate: I agree. Already fixed in a new commit.

supasate:
@superkhau I've fixed the test based on the @bajtos's comment.

superkhau:
>I think it should be back-ported to 2.x too, what do you think?

I agree, no harm in backporting this too.

superkhau:
@supasate Can you squash the done commit with the feature commit (and leave the remove comments separate)?

supasate:
@superkhau Done.

superkhau:

PR #2283, opened 2016-05-03
travis: drop node@5, add node@6

bajtos:
Add `node@6` as a replacement for `node@5`. `node@5` will become unsupported in few weeks/months and besides it's covered by Jenkins CI, thus I think it's safe to remove.

@superkhau please review
/cc @strongloop/loopback-dev


superkhau:
LGTM

PR #2266, opened 2016-04-27
Emit a remoteMethodDisabled event when disabling a remote method

supasate:
This PR for #686 to make loopback emits a remoteMethodDisabled event. I'll make another PR for loopback-component-explorer to register for this event.

PR #2284, opened 2016-05-03
[DO NOT MERGE] - PoC - Controller Spike

richardpringle:
Connect to strongloop-internal/scrum-loopback#784

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -395,6 +395,17 @@ app.enableAuth = function(options) {
   this.isAuthEnabled = true;
 };
 
+app.controller = function(controllerCtor, options) {
+  var self = this;
+  SharedClass = RemoteObjects.SharedClass;
+
+  controllerCtor._sharedClass = new SharedClass(controllerCtor.name);
+  controllerCtor._sharedClass.resolve = function() {
+    return new controllerCtor(app, remotingContext);
+  };
+  app.remotes().addClass(controllerCtor._sharedClass);
+};
+
 app.boot = function(options) {
   throw new Error(
     '`app.boot` was removed, use the new module loopback-boot instead');
--- a/lib/application.js
+++ b/lib/application.js
@@ -395,17 +395,6 @@ app.enableAuth = function(options) {
   this.isAuthEnabled = true;
 };
 
-app.controller = function(controllerCtor, options) {
-  var self = this;
-  SharedClass = RemoteObjects.SharedClass;
-
-  controllerCtor._sharedClass = new SharedClass(controllerCtor.name);
-  controllerCtor._sharedClass.resolve = function() {
-    return new controllerCtor(app, remotingContext);
-  };
-  app.remotes().addClass(controllerCtor._sharedClass);
-};
-
 app.boot = function(options) {
   throw new Error(
     '`app.boot` was removed, use the new module loopback-boot instead');
@@ -578,3 +567,49 @@ app.listen = function(cb) {
 
   return server;
 };
+
+/* START CONTROLLER CODE*/
+
+// Implementation without wrapper
+
+// app.controller = function(controllerCtor, options) {
+//   var self = this;
+//   SharedClass = RemoteObjects.SharedClass;
+
+//   controllerCtor._sharedClass = new SharedClass(controllerCtor.name, controllerCtor);
+//   controllerCtor._sharedClass.resolve = function() {
+//     return new controllerCtor(app, remotingContext);
+//   };
+//   app.remotes().addClass(controllerCtor._sharedClass);
+// };
+
+// Implementation with wrapper
+
+app.controller = function(ctor, options) {
+  var controller = new SharedController(ctor);
+  app.remotes().addClass(wrapper.sharedClass);
+  app.controllers[ctor.name] = wrapper;
+  return wrapper;
+};
+
+function SharedController(ctor, options) {
+  options = options || {};
+  this.name = ctor.name;
+  this.ctor = ctor;
+  this.sharedClass = new SharedClass(this.name, ctor, options);
+};
+
+SharedController.remoteMethod = function(name, options) {
+  if (options.isStatic === undefined) {
+    var m = name.match(/^prototype\.(.*)$/);
+    options.isStatic = !m;
+    name = options.isStatic ? name : m[1];
+  }
+  this.sharedClass.defineMethod(name, options);
+};
+
+SharedController.disableRemoteMethod = function() {
+
+};
+
+/* END CONTROLLER CODE */
--- a/lib/application.js
+++ b/lib/application.js
@@ -285,6 +285,7 @@ app.handler = function(type, options) {
   var handler = this._handlers[type] = remotes.handler(type, options);
 
   remotes.classes().forEach(function(sharedClass) {
+    /* HACK!!! */
     if (sharedClass.name !== 'ctrl') {
       sharedClass.ctor.emit('mounted', app, sharedClass, remotes);
     }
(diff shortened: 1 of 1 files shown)


richardpringle:
cc/ @ritch @raymondfeng @bajtos 

Please review.

lib/application.js
```diff
@@ -284,7 +285,10 @@ app.handler = function(type, options) {
   var handler = this._handlers[type] = remotes.handler(type, options);
 
   remotes.classes().forEach(function(sharedClass) {
-    sharedClass.ctor.emit('mounted', app, sharedClass, remotes);
+    /* HACK!!! */
+    if (sharedClass.name !== 'ctrl') {
```
ritch: What is the reason for this?

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
 
   return server;
 };
+
+/* START CONTROLLER CODE*/
+
+// this is similar to app.model
+// at the moment I haven't really used 'options'
+app.controller = function(ctor, options) {
```
ritch: Missing the JSDoc comments (see other methods) eg. `app.model`.

lib/application.js
richardpringle: I need the controller-ctor to inherit the EventEmitter functions and was trying to figure out a clean way of doing so.

I've seen this done before as follows:
```node
function ctor() {
    EventEmitter.call(this);
}

util.inherits(ctor, EventEmitter);
```

but in this case, `ctor` is the constructor created by the loopback-developer. I do not want the lb-developer to have to write this line, `EventEmitter.call(this);` inside their controller constructor.

Is there another way to do it?

bajtos: I think there are two options:

1) Don't emit `mounted` for controllers that don't provide EventEmitter mixed into the constructor. (BTW this is rather unusual, to emit events on a constructor - a Function instance. Typically, events are emitted on object instances created by a constructor function.) Instead of checking for controller name, check whether the constructor provides `emit` method.

```js
if (sharedClass.ctor && sharedClass.ctor.emit) {
  sharedClass.ctor.emit('mounted', app, sharedClass, remotes);
}
```

2) Automatically add EventEmitter when a controller is registered via `app.controller`. See [this code](https://github.com/strongloop/loopback-datasource-juggler/blob/2189b9a746071b599d6a129313b2cdd11052e36b/lib/model-builder.js#L184-L206) in juggler for an example, just ignore the deprecation warnings.

```js
app.controller = function(ctor, options) {
  if (!ctor.emit) {
    // mix in EventEmitter into the constructor (not the instances!)
    var events = new EventEmitter();
    for (var f in EventEmitter.prototype) {
      if (typeof EventEmitter.prototype[f] === 'function') {
        ctor[f] = EventEmitter.prototype[f].bind(events);
      }
    }
  }

  var self = this;
  // etc.
};
```

---

Personally, I am proposing to use the first approach and don't emit `mounted` method for controllers. Here are my arguments:

 - It's less code/functionality to maintain.
 - We can always add it later.
 - Controllers work differently from models. There is always a new controller instance created for each request and this instance receives `app` in the constructor. There is no need to pass the `app` object via `mounted` event.

@richardpringle @ritch thoughts?

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+    this.disableRemoteMethod = function(name, isStatic) {
+      this.sharedClass.disableMethod(name, isStatic || false);
+      // this.emit('remoteMethodDisabled', this.sharedClass, name);
+    };
+  };
+};
+
+// similar to this app.models
+// the controllers still need to registered in the Registy though
+app.controllers = function() {
```
bajtos: `app.models` has dual API: 

1) it's a function you can call to receive an array of models:

```js
app.models().forEach(function(model) {
  // do something with a model
});
```

2) And at the same time, it's a name-based lookup table allowing quick access to any model:

```js
var MyUser = app.models['my-user'];
// app.models.myUser and app.models.MyUser point to the same model ctor
```

This duality is causing a bit of confusion, one of our colleagues was surprised to learn that `Object.keys(app.models)` contains each model two or three times.

I think it would be great if we can come up with a better API for controllers.

Questions to consider:
 - Do we need a name-based lookup for controllers?
 - If the answer is yes, do we need API for getting a list of controllers, or is it enough to build this list via `Object.keys(app.controllers)`?
 - Can we require controller names to be unique? Name-based lookup won't work otherwise.
 - I am proposing to remove classify/camelize aliases and always use the original controller name.
 - Since it's not possible to change function name after it was defined, we need to use a different property name to allow code-generated controller functions to provide custom name. For example, models use `modelName` property.
 
@richardpringle @ritch @raymondfeng thoughts?

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+    this.name = ctor.name;
+    this.ctor = ctor;
+    // TODO - right now there's a hack on line 289 to skip this
+    // util.inherits(ctor, EventEmitter);
+
+    this.sharedClass = new RemoteObjects.SharedClass(this.name, ctor, options);
+    // the resolveScope method instantiates the controller upon request
+    // (please correct my terminology if it's wrong)
+    this.sharedClass.resolveScope = function(remotingContext) {
+      // self is the app - should it be something more obvious like appInstance?
```
bajtos: I think this shows that the current design, where a new function `SharedController` is defined for each `app.controller` call, is suboptimal.

I would prefer if `SharedController` was a standalone class, defining methods on a prototype, and receiving the `app` or `appInstance` object via constructor.

```js
// goes to lib/shared-controller.js
function SharedController(ctor, app, options) {
  // ...
  this.sharedClass.resolveScope = function(remotingContext) {
     return new ctor(app, remotingContext);
  };
}

SharedController.prototype.remoteMethod = function(name, options) {
  // ...
};

// etc.
```

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+      return new ctor(self, remotingContext);
+    };
+
+    // similar to model.remoteMethod
+    this.remoteMethod = function(name, options) {
+      // I didn't dive into why, but this only worked when I set isStatic to true
+      // whether the controller methods were prototype methods or not
+      options.isStatic = true;
```
bajtos: Let me explain. When `isStatic:true`, strong-remoting will call the method on the scope provided by `resolveScope` function. When `isStatic:false`, strong-remoting will use the scope as a constructor function and call the method on an instance created by this constructor. (Obviously, that assumes that the scope is a constructor function, which is true for LoopBack models, but it's not true for controllers).

Because SharedController returns a controller instance as the scope, all remote methods defined on a controller must be marked as static, even though they are defined on a prototype like instance methods. I admit this may be a little bit confusing.

Recently, @0candy was changing the API for working with remote methods, see https://github.com/strongloop/loopback/issues/597 and https://github.com/strongloop-internal/scrum-loopback/issues/617. I am could not quickly find a comprehensive overview of our discussions, here is the essence: Instead of identifying methods by name+isStatic flag, we encode this flag in the method name. For example, a static method is "find" and a prototype method is "prototype.updateAttributes". See RELEASE NOTES in [loopback](https://github.com/strongloop/loopback/blob/363bc4d6c182067cf02fdf964ee4ea38ce408f35/3.0-RELEASE-NOTES.md#new-method-of-defining-remoting-metadata) and [strong-remoting](https://github.com/strongloop/strong-remoting/blob/aa391d2e65460196ae2a59493f9dd4844622f35d/3.0-RELEASE-NOTES.md#invocation-for-remote-method-has-been-changed).

>```js
{ 
methods: {
  staticMethod: {
    http: { path: '/static' }
  },
  'prototype.instanceMethod': {
  http: { path: '/instance' }
}}
```

I think this is fortunate for controllers, because there is no `isStatic` flag around to confuse users. It also means that you should not need to set it here, as long as you are using the latest strong-remoting version from the master.

File: lib/application.js [hunk 6]
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+      this.sharedClass.defineMethod(name, options);
+    };
+
+    // have not tested this method yet but it follows model.disableRemoteMethod
+    this.disableRemoteMethod = function(name, isStatic) {
```
bajtos: Please follow the new strong-remoting API `disableMethodByName`.

```js
SharedController.prototype.disableRemoteMethodByName = function(name) {
  this.sharedClass.disableMethodByName(name);
}
```

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+      this.sharedClass.disableMethod(name, isStatic || false);
+      // this.emit('remoteMethodDisabled', this.sharedClass, name);
```
bajtos: The event `remoteMethodDisabled` is used by loopback-component-explorer to rebuild the list of methods displayed in the UI, see https://github.com/strongloop/loopback/pull/2266/files and https://github.com/strongloop/loopback-component-explorer/pull/157/files.

Perhaps we can emit the event directly on the `app` object?

Alternatively, we can change `SharedController` to inherit from `EventEmitter`, then here we can emit the event on the shared-controller instance and finally `app.controller` can install a listener for this event that will emit app-wide event, similarly to what we do in https://github.com/strongloop/loopback/pull/2266/files.

Thoughts?

bajtos:
@richardpringle Reviewed, this is great stuff!

lib/application.js
richardpringle: @bajtos, I agree, controllers shouldn't really be mounted since they are resolved on a per request basis. However, I'm not aware if there are any other parts of LoopBack that use the "mounted" event such as the SDKs. Thoughts?

lib/application.js
richardpringle: Yes, will add those for the actual implementation (this is just a PoC)

lib/application.js
richardpringle: I saw the code for that but wasn't aware of the issue. Makes sense.

lib/application.js
richardpringle: Emitting the event on the `SharedController` makes more sense off the top of my head. I am not opposed to emitting it directly off the `app` object though.

lib/application.js
bajtos: I don't think we use the "mounted" event anywhere else besides explorer as described below, but I am not sure. Can you do a quick `grep` to confirm?

lib/application.js
richardpringle: 1. I don't see a reason not to.
2. I honestly prefer using vanilla JS or the node API wherever possible but that's a personal preference.
3. Yes, I don't think it would ever make any sense to have multiple controllers with the same name.
4. Makes sense.
5. Agree.

lib/application.js
richardpringle: Looks like it's used in the [loopback/lib/model.js file](https://github.com/strongloop/loopback/blob/363bc4d6c182067cf02fdf964ee4ea38ce408f35/lib/model.js#L810) to add the listener tree.

In theory, there's no reason to have a beforeRemote/afterRemote on a controller method right?

lib/application.js
```diff
@@ -567,3 +571,65 @@ app.listen = function(cb) {
...
+  // SharedController class contains strong-remoting objects to set up the metadata
+  // and register the endpoint, etc
+  function SharedController(ctor, options) {
+    options = options || {};
+    this.name = ctor.name;
+    this.ctor = ctor;
+    // TODO - right now there's a hack on line 289 to skip this
+    // util.inherits(ctor, EventEmitter);
+
+    this.sharedClass = new RemoteObjects.SharedClass(this.name, ctor, options);
```
richardpringle: As per [@bajtos' comment](https://github.com/strongloop/loopback-sandbox/pull/7/files/6c4d633f9b5b55710980069a4efaee2ebe06c2c1#r63170059):
> why would you like to set the controller name in the first place, I mean what do you use the name for?

By default, the path that the SharedClass is registered to the path `'/' + this.name`. The user must set `ctor.http.path` to explicitly define the path at which to register the controller.

bajtos: I think `this.name` is not very useful, because `SharedClass` is created right from the constructor, and thus it's not possible to customize it. How about allowing `options.name`?

```js
this.name = options.name || ctor.name;
```

Not a big deal, we can always add this later.

bajtos: Taking into account https://github.com/strongloop/loopback/pull/2284/files#r62626126, this may be a good solution:

```js
this.name = ctor.controllerName = options.name || ctor.controllerName || ctor.name;
```

PR #2296, opened 2016-05-05
Backport/do not allow duplicate role names

superkhau:
Added jshint ignore line to make pass linting.

### DIFF
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -15,7 +15,6 @@ var async = require('async');
 var expect = require('chai').expect;
 
 function checkResult(err, result) {
-  // console.log(err, result);
   assert(!err);
 }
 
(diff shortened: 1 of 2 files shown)


test/role.test.js
```diff
@@ -93,11 +90,25 @@ describe('role model', function() {
...
-  it('should automatically generate role id', function() {
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      if (err) return done(err);
 
-    User.create({name: 'Raymond', email: 'x@y.com', password: 'foobar'}, function(err, user) {
-      // console.log('User: ', user.id);
-      Role.create({name: 'userRole'}, function(err, role) {
+      Role.create({ name: 'userRole' }, function(err, role) {
+        expect(err).to.exist; // jshint ignore:line
```
superkhau: @bajtos Do you know if there is a better way to fix this (besides `...to.exist()`).

test/role.test.js
```diff
@@ -93,11 +90,25 @@ describe('role model', function() {
...
-  it('should automatically generate role id', function() {
+  it('should not allow duplicate role name', function(done) {
+    Role.create({ name: 'userRole' }, function(err, role) {
+      if (err) return done(err);
 
-    User.create({name: 'Raymond', email: 'x@y.com', password: 'foobar'}, function(err, user) {
-      // console.log('User: ', user.id);
-      Role.create({name: 'userRole'}, function(err, role) {
+      Role.create({ name: 'userRole' }, function(err, role) {
+        expect(err).to.exist; //jshint ignore:line
```
superkhau: @bajtos Do you know if there is a better way to fix this (besides ...to.exist())?

superkhau:
@bajtos Review please.

test/role.test.js
bajtos: That's fine with me, just squash the commit adding this line with the one you are fixing.

I think we should fix our eslint-config-loopback to reject lines like `expect(err).to.exist` too.

bajtos: > I think we should fix our eslint-config-loopback to reject lines like `expect(err).to.exist` too.

https://github.com/strongloop/eslint-config-loopback/pull/6

PR #2297, opened 2016-05-05
Separate error-checking and next/done logic from other logic

supasate:
Based on @superkhau's suggestion that we should add blank lines to separate error-checking logic, done logic from other logic. I found most test files still don't follow this convention. So, I've refactored them.

I also changed a few error checking conditions to be one-liner;
From
```
if (err) {
  return done(err);
}
````
to
`
if (err) return done(err);
`


superkhau:

PR #2301, opened 2016-05-05
Remove unused condition for checking non-existent model

davidcheung:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/770

This behavior is actually handled by [registry.getModel()](https://github.com/strongloop/loopback/blob/1559db2ca3f5cf75606a09a806428140e9c967d6/lib/registry.js#L305)

added a test case to make sure it throws the error. 
This actually is no longer a breaking change because it already threw error

### DIFF
--- a/test/registries.test.js
+++ b/test/registries.test.js
@@ -4,6 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 describe('Registry', function() {
+  describe('createModel', function() {
+    it('should throw error upon extending non-exist base model', function() {
+      var app = loopback();
+      var props = {};
+      var opts = { base: 'nonexistent' };
+      expect(function() { app.registry.createModel('aModel', props, opts); })
+        .to.throw(/model\s`aModel`(.*)unknown\smodel\s`nonexistent`/);
+    });
+  });
+
   describe('one per app', function() {
     it('should allow two apps to reuse the same model name', function(done) {
       var appFoo = loopback();
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -109,11 +109,10 @@ Registry.prototype.createModel = function(name, properties, options) {
 
   if (typeof BaseModel === 'string') {
     var baseName = BaseModel;
-    BaseModel = this.getModel(BaseModel);
-
-    if (BaseModel === undefined) {
-      console.warn('Model `%s` is extending an unknown model `%s`. ' +
-        'Using `PersistedModel` as the base.', name, baseName);
+    BaseModel = this.findModel(BaseModel);
+    if (!BaseModel) {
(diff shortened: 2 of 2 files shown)


bajtos:
> This behavior is actually handled by registry.getModel()
> This actually is no longer a breaking change because it already threw error

Ouch, it makes me wonder whether we have accidentally introduced a breaking change in the past.

Anyways, I think the error thrown by `getModel()` is not very helpful, as it does not tell you what is the name of the model that was trying to inherit from a non-existing model. The second name is important because it's the name you need to find the JSON file containing invalid "base" model name.

Here is a mockup of what would IMO work better:

```js
BaseModel = this.findModel(BaseModel);
if (!BaseModel) {
  // throw a helpful error mentioning both "baseName" and "name"
 }
```

test/registries.test.js
```diff
@@ -4,6 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 describe('Registry', function() {
+  describe('models', function() {
+    it('should throw error upon extending non-exist base model', function(done) {
+      var app = loopback();
+      var props = {};
+      var opts = { base: 'nonexistent' };
+      expect(function() { app.registry.createModel('aModel', props, opts); })
+        .to.throw();
```
bajtos: Please check that the error message contains both the name of the model being created (`aModel`) and the non-existing base model (`nonexistent`).

davidcheung:
Fixed, the error should now show as 
>Model not found: model `aModel` is extending an unknown model `nonexistent`.
@bajtos PTAL

test/registries.test.js
```diff
@@ -4,6 +4,16 @@
...
 
 describe('Registry', function() {
+  describe('models', function() {
+    it('should throw error upon extending non-exist base model', function(done) {
+      var app = loopback();
+      var props = {};
+      var opts = { base: 'nonexistent' };
+      expect(function() { app.registry.createModel('aModel', props, opts); })
+        .to.throw(/model\s`aModel`(.*)unknown\smodel\s`nonexistent`/);
+      done();
```
bajtos: Since this test completes on the first tick, please remove the `done` argument and its invocation on this line.

test/registries.test.js
```diff
@@ -4,6 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 describe('Registry', function() {
+  describe('models', function() {
```
bajtos: Please use the function name under test as the suite name.

```
describe('createModel', function() {
  // ...
});
```

We don't have a unified policy about the exact format of the test name, in some places we would have 
```js
describe('createModel()'`);
```
 or even
```js
describe('createModel(name, properties, options)');
```

Pick the style you like most :)

bajtos:
To details to fix, the rest LGTM. No more reviews are needed.

Issue #2306, opened 2016-05-06
Replace all instances of {href} in user.verify.


ambrt:
I just saw that user.verify() replaces only first instance of {href} in text here:
https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/common/models/user.js#L436

It would be better if instead of:
`options.text = options.text.replace('{href}', options.verifyHref);`

we would use:
` options.text = options.text.replace(/\{href\}/g, options.verifyHref);`

That way we could send email with live link using `<a href="{href}">verify</a>`
and in case someone wouldn't have html email then just plain text in one go.

superkhau:
Would you like to submit a PR?

ambrt:
Its my firs PR :) i hope i made it right  https://github.com/strongloop/loopback/pull/2310

richardpringle:
#2310 was merged to solve this issue: closing.

PR #2167, opened 2016-03-24
Replace all the "{href}" placeholders in verify email template

aitoraznar:
Hello,

I needed to send an email with the link in a button and an alternative link in plain text.
So I used a regex to replace all the "{href}" placeholders in verify email template.

Regards

PR #2310, opened 2016-05-08
Update user.js

ambrt:
allow to change all {href} instances in user.verify() mail into generated url instead of just one

connect to #2306


slnode:

ambrt:
It was requested here also https://github.com/strongloop/loopback/pull/2167

richardpringle:
ok to test

superkhau:

superkhau:
@ambrt I forgot to check before merging this, can you write a unit test for this change to prevent regressions in the future?

Issue #2307, opened 2016-05-06
Redirect url should be urlencoded


ambrt:
The user verify link is made in such way that it provides redirect path in the middle of whole link.

The problem is that when one would use `redirect=/#/my-angular-path` then the token is treated as hash valu and isn't read by api (to verify) and api throws error.
Redirect is added on line [407](https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/common/models/user.js#L407) and token on [431](https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/common/models/user.js#L431).

I would just remove:

```
   '&redirect=' +
      options.redirect;
```

from line 407 and put it behind 
`options.verifyHref += '&token=' + user.verificationToken;`
on line 431

superkhau:
Can you provide a sample repo for us to reproduce your issue?

ambrt:
@superkhau  Here's a sample of it https://github.com/ambrt/loopback-sandbox

superkhau:
@davidcheung Reassigning to you as this is your FA.

davidcheung:
@ambrt if we also url encode the redirectUrl would this already not be a problem? it is altering the url?

ambrt:
Hi @davidcheung 
Yes, I think that url encode  [here ](https://github.com/strongloop/loopback/blob/master/common/models/user.js#L425) and url decode [here](https://github.com/strongloop/loopback/blob/master/common/models/user.js#L790) would solve the problem.

davidcheung:
actually i believe that would also mean if the url has parameters it will also be lost, we should fix this

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/3185.

bajtos:
Fixed.

PR #2311, opened 2016-05-08
Allow hash path (/#/path ) for redirect Url

ambrt:
Changed adding `'&redirect=' +options.redirect` to verifyHref, from inside `.verfiy()` method to insde of `sendEmail()` method, to allow hash paths for redirect url. From issue at https://github.com/strongloop/loopback/issues/2307

connect to #2307

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -426,8 +426,8 @@ module.exports = function(User) {
 
     // TODO - support more verification types
     function sendEmail(user) {
-      options.verifyHref += '&token=' + user.verificationToken+ 
-      '&redirect=' + options.redirect;
+      options.verifyHref += '&token=' + user.verificationToken +
+        '&redirect=' + options.redirect;
 
       options.text = options.text || 'Please verify your email by opening ' +
         'this link in a web browser:\n\t{href}';
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -403,9 +403,7 @@ module.exports = function(User) {
       userModel.http.path +
       userModel.sharedClass.findMethodByName('confirm').http.path +
       '?uid=' +
-      options.user.id +
-      '&redirect=' +
-      options.redirect;
+      options.user.id;
 
     // Email model
     var Email = options.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
@@ -428,7 +426,8 @@ module.exports = function(User) {
 
     // TODO - support more verification types
     function sendEmail(user) {
-      options.verifyHref += '&token=' + user.verificationToken;
+      options.verifyHref += '&token=' + user.verificationToken+ 
+      '&redirect=' + options.redirect;
 
       options.text = options.text || 'Please verify your email by opening ' +
         'this link in a web browser:\n\t{href}';


slnode:

richardpringle:
ok to test

superkhau:
LGTM. Thanks for the contribution! ;) @bajtos For final review and merge.

bajtos:
I am not very familiar with this part of LoopBack. @ritch could you please review?

However, I would like to see a unit-test covering the change, to demonstrate the new behaviour and prevent regressions in the future.

ambrt:
Two things to note: the test [Should report 302 when redirect url is set](https://github.com/strongloop/loopback/blob/a6f8c07301e02c4094c5cdf244bd0a6558beed9a/test/user.test.js#L1420) fails when the redirect is having hash value like:
```
  it('Should report 302 when redirect url is set', function(done) {
        testConfirm(function(result, done) {
          request(app)
            .get('/test-users/confirm?uid=' + (result.uid) +
              '&token=' + encodeURIComponent(result.token) +
              '&redirect=http://foo.com/bar/#/')
            .expect(302)
            .expect('Location', 'http://foo.com/bar/#/')
            .end(done);
        }, done);
      });
```
with error:
` Error: expected "Location" of "http://foo.com/bar/#/", got "http://foo.com/bar/"` which is expected because server can't read hash values.

The second thing is that test `"Confirm a user verification"` encodes the /# into %2F%23 [here](https://github.com/strongloop/loopback/blob/a6f8c07301e02c4094c5cdf244bd0a6558beed9a/test/user.test.js#L1410)
so test passes even when the redirect is set with hash (like `redirect: 'http://foo.com/bar/#/foo'`), 
while the email link will be broken.


My idea to solve it would be to  `encodeURIComponent` when  [creating verifyHref](https://github.com/strongloop/loopback/blob/a6f8c07301e02c4094c5cdf244bd0a6558beed9a/common/models/user.js#L405).
and decode it after verifying email.
This way test would pass and email link would work.

I'm just not sure where the redirecting is happening.

ambrt:
ok here i made a pull request https://github.com/strongloop/loopback/pull/2341

ritch:
test please

slnode:

slnode:

slnode:

slnode:

ritch:
ok to test

bajtos:
> Error: expected "Location" of "http://foo.com/bar/#/", got "http://foo.com/bar/" which is expected because server can't read hash values.
> The second thing is that test "Confirm a user verification" encodes the /# into %2F%23 here
so test passes even when the redirect is set with hash (like redirect: 'http://foo.com/bar/#/foo'),
while the email link will be broken
> My idea to solve it would be to encodeURIComponent when creating verifyHref.
and decode it after verifying email.

Yes please, we should be applying `encodeURIComponent` to all query string values.

bajtos:
@ambrt I see this patch is rather old, sorry for our long silence! Are you still keen to work on this change?

ambrt:
@bajtos I made new pull request with solution https://github.com/strongloop/loopback/pull/3035

bajtos:
> @ambrt  I made new pull request with solution #3035

I am closing this pull request then.

In the future, please use `git commit --amend` or `git rebase -i` together with `git push -f` and keep the work in the same pull request. That way all discussion stays in one place.

Issue #1979, opened 2016-01-20
Expose replace API endpoints


Amir-61:
Form @bajtos:
> I'd like to discuss the naming scheme for these new methods (replaceOrCreate, replaceAttributes). I feel that `replaceAttributes` is not describing the intent precisely, and the difference between `replaceAttributes` and `updateAttributes` may not be clear enough.

> At the moment, we have (among others):
 - `MyModel.create`
 - `MyModel.updateOrCreate`
 - `myModelInstance.updateAttributes`
 - `myModelInstance.updateAttribute` (a single attribute)

> I am proposing to come up with new names while we are at adding support for REPLACE operations:
 - `MyModel.create` (remains same)
 - `MyModel.patchOrCreate` (with an alias `MyModel.updateOrCreate` for backwards-compatibility)
 - `myModelInstance.patchAttributes` (with an alias `myModelInstance.updateAttributes` for backwards-compatibility)
 -  `myModelInstance.updateAttribute`
 - `MyModel.replaceOrCreate`
 - `myModelInstance.replace` or perhaps one of 
  - `myModelInstance.updateTo`
  - `myModelInstance.updateAllAttributes` 
  - `myModelInstance.setTo`

Amir-61:
From @bajtos:

@raymondfeng let's discuss the naming of the new methods. You wrote:

I'm debating if replace should be a prototype method. Maybe a static method such as replaceById is more consistent with updateById and destroyById. If that's agreed, I prefer to use replace as an alias to replaceAll.
My initial thinking was to make prototype.replace similar to existing prototype.updateAttributes and eventually use it inside instance.save. Note we don't have updateById yet (I think patchById would be a better name BTW).

OTOH, replaceById elegantly solves the naming problem I have outlined before, and it also allows a more efficient implementation (no need to fetch the instance when we are going to replace all attributes anyways). I have a small concern about ACLs, whether the current mechanism is able extract the id information from replaceById arguments to apply correct ACL checks. I guess it has to, if destroyById is remoted. Besides this, I am ok to run with your proposal.

I personally dislike replace as an alias to replaceAll - to me it's not clear whether replace is manipulating a single instance or multiple instances. Also note that replaceAll is out of scope of this work, or at least that was our plan.

Compiling the discussion above, here is my proposal:

Add the following new methods:

MyModel.replaceOrCreate
MyModel.replaceById
Add two new aliases:

[...]

PR #2316, opened 2016-05-09
Expose `Replace*` methods

Amir-61:
##### Description:

* Re-mapping `updateAttributes` and `updateOrCreate` endpoints to use `PATCH` verb by default for `3.X`.
* Exposing `replaceById` and `replaceORCreate` via `PUT` verbs by default for `3.X`.
* Add a flag `model.settings.options.updateOnPUT` for `3.X`.

**Note:** This patch is for `3.X`

Related to https://github.com/strongloop/loopback/pull/2330 (Implementation for `2.x`)

Connect to https://github.com/strongloop/loopback/issues/1979

### DIFF
--- a/common/models/user.json
+++ b/common/models/user.json
@@ -66,6 +66,12 @@
     },
     {
       "principalType": "ROLE",
+      "principalId": "$owner",
+      "permission": "ALLOW",
+      "property": "replaceById"
+    },  
+    {
+      "principalType": "ROLE",
       "principalId": "$everyone",
       "permission": "ALLOW",
       "property": "confirm"
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -111,7 +111,8 @@ module.exports = function(registry) {
    * @param {Object} model Updated model instance.
    */
 
-  PersistedModel.upsert = PersistedModel.updateOrCreate = function upsert(data, callback) {
+  PersistedModel.upsert = PersistedModel.updateOrCreate = PersistedModel.patchOrCreate =
+  function upsert(data, callback) {
     throwNotAttached(this.modelName, 'upsert');
   };
 
@@ -494,7 +495,8 @@ module.exports = function(registry) {
    * @param {Object} instance Updated instance.
    */
 
-  PersistedModel.prototype.updateAttributes = function updateAttributes(data, cb) {
+  PersistedModel.prototype.updateAttributes = PersistedModel.prototype.patchAttributes =
+  function updateAttributes(data, cb) {
     throwNotAttached(this.modelName, 'updateAttributes');
   };
 
@@ -600,6 +602,9 @@ module.exports = function(registry) {
     var typeName = PersistedModel.modelName;
     var options = PersistedModel.settings;
 
+    // This is just for LB 3.x
+    options.replaceOnPUT = options.replaceOnPUT !== false;
+
     function setRemoting(scope, name, options) {
       var fn = scope[name];
       fn._delegate = true;
@@ -616,15 +621,35 @@ module.exports = function(registry) {
       http: { verb: 'post', path: '/' },
     });
 
-    setRemoting(PersistedModel, 'upsert', {
-      aliases: ['updateOrCreate'],
-      description: 'Update an existing model instance or insert a new one into the data source.',
+    var upsertOptions = {
+      aliases: ['upsert', 'updateOrCreate'],
+      description: 'Patch an existing model instance or insert a new one into the data source.',
       accessType: 'WRITE',
       accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description:
         'Model instance data' },
       returns: { arg: 'data', type: typeName, root: true },
-      http: { verb: 'put', path: '/' },
-    });
+      http: [{ verb: 'patch', path: '/' }],
+    };
+
+    if (!options.replaceOnPUT) {
+      upsertOptions.http.push({ verb: 'put', path: '/' });
+    }
+    setRemoting(PersistedModel, 'patchOrCreate', upsertOptions);
+
+    var replaceOrCreateOptions = {
+      description: 'Replace an existing model instance or insert a new one into the data source.',
+      accessType: 'WRITE',
+      accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'Model instance data' },
+      returns: { arg: 'data', type: typeName, root: true },
+      http: [{ verb: 'post', path: '/replaceOrCreate' }],
+    };
+
+    if (options.replaceOnPUT) {
+      replaceOrCreateOptions.http.push({ verb: 'put', path: '/' });
+    }
+
+    setRemoting(PersistedModel, 'replaceOrCreate', replaceOrCreateOptions);
 
     setRemoting(PersistedModel, 'exists', {
       description: 'Check whether a model instance exists in the data source.',
@@ -670,6 +695,26 @@ module.exports = function(registry) {
       http: { verb: 'get', path: '/:id' },
       rest: { after: convertNullToNotFoundError },
     });
+
+    var replaceByIdOptions = {
+      description: 'Replace attributes for a model instance and persist it into the data source.',
+      accessType: 'WRITE',
+      accepts: [
+        { arg: 'id', type: 'any', description: 'Model id', required: true,
+          http: { source: 'path' }},
+          { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'Model instance data' },
+      ],
+      returns: { arg: 'data', type: typeName, root: true },
+      http: [{ verb: 'post', path: '/:id/replace' }],
+    };
+
+    if (options.replaceOnPUT) {
+      replaceByIdOptions.http.push({ verb: 'put', path: '/:id' });
+    }
+
+    setRemoting(PersistedModel, 'replaceById', replaceByIdOptions);
+
 
     setRemoting(PersistedModel, 'find', {
       description: 'Find all instances of the model matched by filter from the data source.',
@@ -741,13 +786,20 @@ module.exports = function(registry) {
       http: { verb: 'get', path: '/count' },
     });
 
-    setRemoting(PersistedModel.prototype, 'updateAttributes', {
-      description: 'Update attributes for a model instance and persist it into the data source.',
+    var updateAttributesOptions = {
+      aliases: ['updateAttributes'],
+      description: 'Patch attributes for a model instance and persist it into the data source.',
       accessType: 'WRITE',
       accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description: 'An object of model property name/value pairs' },
       returns: { arg: 'data', type: typeName, root: true },
-      http: { verb: 'put', path: '/' },
-    });
+      http: [{ verb: 'patch', path: '/' }],
+    };
+
+    setRemoting(PersistedModel.prototype, 'patchAttributes', updateAttributesOptions);
+
+    if (!options.replaceOnPUT) {
+      updateAttributesOptions.http.push({ verb: 'put', path: '/' });
+    }
 
     if (options.trackChanges || options.enableRemoteReplication) {
       setRemoting(PersistedModel, 'diff', {
--- a/test/access-control.integration.js
+++ b/test/access-control.integration.js
@@ -111,7 +111,13 @@ describe('access control - integration', function() {
           assert.equal(user.password, undefined);
         });
       });
+
+      // user has replaceOnPUT = false; so then both PUT and PATCH should be allowed for update
       lt.describe.whenCalledRemotely('PUT', '/api/users/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
+
+      lt.describe.whenCalledRemotely('PATCH', '/api/users/:id', function() {
         lt.it.shouldBeAllowed();
       });
     });
@@ -163,7 +169,7 @@ describe('access control - integration', function() {
     }
   });
 
-  describe('/accounts', function() {
+  describe('/accounts with replaceOnPUT true', function() {
     var count = 0;
     before(function() {
       var roleModel = loopback.getModelByType(loopback.Role);
@@ -177,47 +183,67 @@ describe('access control - integration', function() {
       });
     });
 
-    lt.beforeEach.givenModel('account');
+    lt.beforeEach.givenModel('accountWithReplaceOnPUTtrue');
 
-    lt.it.shouldBeDeniedWhenCalledAnonymously('GET', '/api/accounts');
-    lt.it.shouldBeDeniedWhenCalledUnauthenticated('GET', '/api/accounts');
-    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'GET', '/api/accounts');
+    lt.it.shouldBeDeniedWhenCalledAnonymously('GET', '/api/accounts-replacing');
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('GET', '/api/accounts-replacing');
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'GET', '/api/accounts-replacing');
 
     lt.it.shouldBeDeniedWhenCalledAnonymously('GET', urlForAccount);
     lt.it.shouldBeDeniedWhenCalledUnauthenticated('GET', urlForAccount);
     lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'GET', urlForAccount);
 
-    lt.it.shouldBeDeniedWhenCalledAnonymously('POST', '/api/accounts');
-    lt.it.shouldBeDeniedWhenCalledUnauthenticated('POST', '/api/accounts');
-    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'POST', '/api/accounts');
+    lt.it.shouldBeDeniedWhenCalledAnonymously('POST', '/api/accounts-replacing');
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('POST', '/api/accounts-replacing');
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'POST', '/api/accounts-replacing');
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('POST', urlForReplaceAccountPOST);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('POST', urlForReplaceAccountPOST);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'POST', urlForReplaceAccountPOST);
 
     lt.it.shouldBeDeniedWhenCalledAnonymously('PUT', urlForAccount);
     lt.it.shouldBeDeniedWhenCalledUnauthenticated('PUT', urlForAccount);
     lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'PUT', urlForAccount);
 
+    lt.it.shouldBeDeniedWhenCalledAnonymously('PATCH', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('PATCH', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'PATCH', urlForAccount);
+
     lt.describe.whenLoggedInAsUser(CURRENT_USER, function() {
+      var actId;
       beforeEach(function(done) {
         var self = this;
-
         // Create an account under the given user
-        app.models.account.create({
+        app.models.accountWithReplaceOnPUTtrue.create({
           userId: self.user.id,
           balance: 100,
         }, function(err, act) {
-          self.url = '/api/accounts/' + act.id;
-
+          actId = act.id;
+          self.url = '/api/accounts-replacing/' + actId;
           done();
         });
       });
 
-      lt.describe.whenCalledRemotely('PUT', '/api/accounts/:id', function() {
+      lt.describe.whenCalledRemotely('PATCH', '/api/accounts-replacing/:id', function() {
         lt.it.shouldBeAllowed();
       });
-      lt.describe.whenCalledRemotely('GET', '/api/accounts/:id', function() {
+      lt.describe.whenCalledRemotely('PUT', '/api/accounts-replacing/:id', function() {
         lt.it.shouldBeAllowed();
       });
-      lt.describe.whenCalledRemotely('DELETE', '/api/accounts/:id', function() {
+      lt.describe.whenCalledRemotely('GET', '/api/accounts-replacing/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
+      lt.describe.whenCalledRemotely('DELETE', '/api/accounts-replacing/:id', function() {
         lt.it.shouldBeDenied();
+      });
+      describe('replace on POST verb', function() {
+        beforeEach(function(done) {
+          this.url = '/api/accounts-replacing/' + actId + '/replace';
+          done();
+        });
+        lt.describe.whenCalledRemotely('POST', '/api/accounts-replacing/:id/replace', function() {
+          lt.it.shouldBeAllowed();
+        });
       });
     });
 
@@ -226,7 +252,76 @@ describe('access control - integration', function() {
     lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'DELETE', urlForAccount);
 
     function urlForAccount() {
-      return '/api/accounts/' + this.account.id;
+      return '/api/accounts-replacing/' + this.accountWithReplaceOnPUTtrue.id;
+    }
+    function urlForReplaceAccountPOST() {
+      return '/api/accounts-replacing/' + this.accountWithReplaceOnPUTtrue.id + '/replace';
+    }
+  });
+
+  describe('/accounts with replaceOnPUT false', function() {
+    lt.beforeEach.givenModel('accountWithReplaceOnPUTfalse');
+    lt.it.shouldBeDeniedWhenCalledAnonymously('POST', urlForReplaceAccountPOST);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('POST', urlForReplaceAccountPOST);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'POST', urlForReplaceAccountPOST);
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('PUT', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('PUT', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'PUT', urlForAccount);
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('PATCH', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('PATCH', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'PATCH', urlForAccount);
+
+    lt.describe.whenLoggedInAsUser(CURRENT_USER, function() {
+      var actId;
+      beforeEach(function(done) {
+        var self = this;
+        // Create an account under the given user
+        app.models.accountWithReplaceOnPUTfalse.create({
+          userId: self.user.id,
+          balance: 100,
+        }, function(err, act) {
+          actId = act.id;
+          self.url = '/api/accounts-updating/' + actId;
+          done();
+        });
+      });
+
+      lt.describe.whenCalledRemotely('PATCH', '/api/accounts-updating/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
+
+      lt.describe.whenCalledRemotely('PUT', '/api/accounts-updating/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
+      lt.describe.whenCalledRemotely('GET', '/api/accounts-updating/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
+      lt.describe.whenCalledRemotely('DELETE', '/api/accounts-updating/:id', function() {
+        lt.it.shouldBeDenied();
+      });
+
+      describe('replace on POST verb', function() {
+        beforeEach(function(done) {
+          this.url = '/api/accounts-updating/' + actId + '/replace';
+          done();
+        });
+        lt.describe.whenCalledRemotely('POST', '/api/accounts-updating/:id/replace', function() {
+          lt.it.shouldBeAllowed();
+        });
+      });
+    });
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('DELETE', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('DELETE', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'DELETE', urlForAccount);
+
+    function urlForAccount() {
+      return '/api/accounts-updating/' + this.accountWithReplaceOnPUTfalse.id;
+    }
+    function urlForReplaceAccountPOST() {
+      return '/api/accounts-updating/' + this.accountWithReplaceOnPUTfalse.id + '/replace';
     }
   });
 });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -624,9 +624,14 @@ describe.onServer('Remote Methods', function() {
       var methodNames = [];
       metadata.methods.forEach(function(method) {
         methodNames.push(method.name);
-        methodNames = methodNames.concat(method.sharedMethod.aliases || []);
+        var aliases = method.sharedMethod.aliases;
+        if (method.name.indexOf('prototype.') === 0) {
+          aliases = aliases.map(function(alias) {
+            return 'prototype.' + alias;
+          });
+        }
+        methodNames = methodNames.concat(aliases || []);
       });
-
       expect(methodNames).to.have.members([
         // NOTE(bajtos) These three methods are disabled by default
         // Because all tests share the same global registry model
@@ -634,9 +639,11 @@ describe.onServer('Remote Methods', function() {
         // this test was seeing this method (with all aliases) as public
         // 'destroyAll', 'deleteAll', 'remove',
         'create',
-        'upsert', 'updateOrCreate',
+        'upsert', 'updateOrCreate', 'patchOrCreate',
         'exists',
         'findById',
+        'replaceById',
+        'replaceOrCreate',
         'find',
         'findOne',
         'updateAll', 'update',
@@ -644,7 +651,7 @@ describe.onServer('Remote Methods', function() {
         'destroyById',
         'removeById',
         'count',
-        'prototype.updateAttributes',
+        'prototype.patchAttributes', 'prototype.updateAttributes',
         'createChangeStream',
       ]);
     });
--- a/test/remoting.integration.js
+++ b/test/remoting.integration.js
@@ -74,60 +74,27 @@ describe('remoting - integration', function() {
   });
 
   describe('Model shared classes', function() {
-    function formatReturns(m) {
-      var returns = m.returns;
-      if (!returns || returns.length === 0) {
-        return '';
-      }
-      var type = returns[0].type;
-      return type ? ':' + type : '';
-    }
-
-    function formatMethod(m) {
-      return [
-        m.name,
-        '(',
-        m.accepts.map(function(a) {
-          return a.arg + ':' + a.type;
-        }).join(','),
-        ')',
-        formatReturns(m),
-        ' ',
-        m.getHttpMethod(),
-        ' ',
-        m.getFullPath(),
-      ].join('');
-    }
-
-    function findClass(name) {
-      return app.handler('rest').adapter
-        .getClasses()
-        .filter(function(c) {
-          return c.name === name;
-        })[0];
-    }
-
-    it('has expected remote methods', function() {
+    it('has expected remote methods with default model.settings.replaceOnPUT' +
+      'set to true (3.x)',
+    function() {
       var storeClass = findClass('store');
-      var methods = storeClass.methods
-        .filter(function(m) {
-          return m.name.indexOf('__') === -1;
-        })
-        .map(function(m) {
-          return formatMethod(m);
-        });
+      var methods = getFormattedMethodsExcludingRelations(storeClass.methods);
 
       var expectedMethods = [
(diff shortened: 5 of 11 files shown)


common/models/user.json
```diff
@@ -76,6 +76,12 @@
       "permission": "ALLOW",
       "property": "resetPassword",
       "accessType": "EXECUTE"
+    },
+    {
+      "principalType": "ROLE",
+      "principalId": "$owner",
+      "permission": "ALLOW",
+      "property": "replaceById"
```
bajtos: As mentioned elsewhere, please move this new ACL entry closer to the entry for `updateAttributes`.

lib/persisted-model.js
```diff
@@ -550,6 +577,20 @@ module.exports = function(registry) {
...
+    var configurableVerbs = {
+      updateAttributes: 'patch',
+      updateOrCreate: 'patch',
+      replaceOrCreate: 'put',
+      replaceById: 'put',
+    };
+
+    if (options.updateOnPUT) {
+      configurableVerbs.replaceById = 'post';
+      configurableVerbs.replaceOrCreate = 'post';
```
bajtos: When `updateOnPUT` is set, then you need to change the HTTP path too, not only the verb.

```
updateOnPUT=false

  replaceById     => PUT /:id
  replaceOrCreate => PUT /

updateOnPUT=true

  replaceById     => POST /:id/replaceById
  replaceOrCreate => POST /replaceOrCreate
```

bajtos:
I am afraid I don't know from the top of my head why the `lt` tests don't work as expected. You will have to troubleshoot and debug yourself, @Amir-61

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

lib/persisted-model.js
```diff
@@ -573,9 +605,41 @@ module.exports = function(registry) {
       accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description:
         'Model instance data' },
       returns: { arg: 'data', type: typeName, root: true },
-      http: { verb: 'put', path: '/' },
+      http: { verb: 'patch', path: '/' },
     });
 
+    if (!options.replaceOnPUT) {
+      setRemoting(PersistedModel, 'upsert', {
```
bajtos: I don't think it's correct to call `setRemoting` twice for the same method, expecting it to create two HTTP endpoints. The correct way is to provide an array of values in the `http` property. See how `exists` defines two HTTP endpoints:

```js
       http: [
         { verb: 'get', path: '/:id/exists' },
         { verb: 'head', path: '/:id' },
       ],
```

We should do the same here. For example:

```js
var upsertEndpoints = [{ verb: 'patch', path: '/' }];
if (!options.replaceOnPUT) {
  upsertEndpoints.push( {
    verb: 'put',
    patch: '/',
  });
}

setRemoting('PersistedModel', 'upsert', {
  // ...
  http: upsertEndpoints
});
```

Or you can use the approach you have in place for `replaceOrCreateOptions` below, both ways are fine.

Amir-61:
@slnode test please

Amir-61:
Please note `strong-remoting` does not support re-registering/re-exposing endpoints for models; that means after registering/exposing endpoints once the models get created then later on you cannot pass something like the following and re-register endpoints:

```
app.models.store.settings.replaceOnPUT = true;
app.models.store.setup();
```

So I changed the test suits to have two models (let's say for `store` model); one with `replaceOnPUT` set to true and one for `replaceOnPUT` set to false in JSON files fixtures.

It is almost done; I would say 90%; could you please confirm you agree with my approach?

Thanks Miroslav!

lib/persisted-model.js
```diff
@@ -600,6 +600,11 @@ module.exports = function(registry) {
     var typeName = PersistedModel.modelName;
     var options = PersistedModel.settings;
 
+    // This is just for LB 3.x
+    if (options.replaceOnPUT !== false || options.replaceOnPUT === undefined) {
```
bajtos: When `options.replaceOnPUT` is undefined, then `options.replaceOnPUT !== false` is truthy. Please simplify the condition to the following:

```js
if (options.replaceOnPUT !== false)
```

lib/persisted-model.js
```diff
@@ -600,6 +600,11 @@ module.exports = function(registry) {
...
+      options.replaceOnPUT = true;
```
bajtos: Perhaps this could be rewritten as follows?

```js
options.replaceOnPUT = options.replaceOnPUT !== false;
```

lib/persisted-model.js
```diff
@@ -616,15 +621,36 @@ module.exports = function(registry) {
...
+      description: 'Replace an existing model instance or insert a new one into the data source.',
+      accessType: 'WRITE',
+      accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'Model instance data' },
+      returns: { arg: 'data', type: typeName, root: true },
+      http: { verb: 'post', path: '/replaceOrCreate' },
+    };
+
+    if (options.replaceOnPUT) {
+      replaceOrCreateOptions.http.verb = 'put';
```
bajtos: Should we preserve `POST /replaceOrCreate` endpoint for people upgrading from LoopBack 2.x to 3.0?

lib/persisted-model.js
```diff
@@ -671,6 +697,28 @@ module.exports = function(registry) {
...
+          { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'Model instance data' },
+      ],
+      returns: { arg: 'data', type: typeName, root: true },
+      http: { verb: 'post', path: '/:id/replace' },
+      rest: { after: convertNullToNotFoundError },
+    };
+
+    if (options.replaceOnPUT) {
+      replaceByIdOptions.http.verb = 'put';
```
bajtos: Ditto, should we preserve `POST /:id/replace` for users upgrading from LoopBack 2.x to 3.0?

lib/persisted-model.js
```diff
@@ -671,6 +697,28 @@ module.exports = function(registry) {
...
+      accessType: 'WRITE',
+      accepts: [
+        { arg: 'id', type: 'any', description: 'Model id', required: true,
+          http: { source: 'path' }},
+          { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'Model instance data' },
+      ],
+      returns: { arg: 'data', type: typeName, root: true },
+      http: { verb: 'post', path: '/:id/replace' },
+      rest: { after: convertNullToNotFoundError },
```
bajtos: IMO, `replaceById` should return 404 error when the record is not found, and therefore this `rest.after` hook should not be needed at all.

test/access-control.integration.js
```diff
@@ -111,9 +111,15 @@ describe('access control - integration', function() {
...
       });
+
+      // user has replaceOnPUT = false; so then both PUT and PATCH should be allowed for update
       lt.describe.whenCalledRemotely('PUT', '/api/users/:id', function() {
         lt.it.shouldBeAllowed();
       });
+
+      lt.describe.whenCalledRemotely('PATCH', '/api/users/:id', function() {
+        lt.it.shouldBeAllowed();
+      });
```
bajtos: Please add test(s) for new replace* endpoints too, to verify that ACLs work correctly for them too.

bajtos: I see you are testing new replace* endpoints below for `replaceOnPUT:true`. I think we should test the endpoints for `replaceOnPUT:false` too.

test/access-control.integration.js
```diff
@@ -163,7 +169,7 @@ describe('access control - integration', function() {
     }
   });
 
-  describe('/accounts', function() {
+  describe('/accounts with replaceOnPUT true', function() {
```
bajtos: Modifying existing test suite and adding more tests into it is not a good practice. It makes it more difficult to understand the scope of this change, why are we changing existing tests not related to replace* methods to run against a model with `replaceOnPUT` set to true?

A better approach is to create a new `describe` block where you will use the new `accountWithReplaceOnPUTtrue` and then add only tests that are related to `replaceOnPUT` into this new block.

bajtos: Oh, I should have read the full patch before commenting. So you actually did what I proposed in `/accounts with replaceOnPUT false`. Can we please simplify that other describe block and keep there only tests relevant to `replaceOnPut`?

test/access-control.integration.js
```diff
@@ -226,7 +238,101 @@ describe('access control - integration', function() {
...
-      return '/api/accounts/' + this.account.id;
+      return '/api/accounts/' + this.accountWithReplaceOnPUTtrue.id;
+    }
+  });
+
+  describe('/accounts with replaceOnPUT false', function() {
+    var count = 0;
+    before(function() {
+      var roleModel = loopback.getModelByType(loopback.Role);
+      roleModel.registerResolver('$dummy', function(role, context, callback) {
```
bajtos: I think this is not relevant to `replaceOnPUT`, can we remove it?

test/access-control.integration.js
```diff
@@ -226,7 +238,101 @@ describe('access control - integration', function() {
...
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('GET', '/api/accounts2');
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'GET', '/api/accounts2');
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('GET', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('GET', urlForAccount);
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'GET', urlForAccount);
+
+    lt.it.shouldBeDeniedWhenCalledAnonymously('POST', '/api/accounts2');
+    lt.it.shouldBeDeniedWhenCalledUnauthenticated('POST', '/api/accounts2');
+    lt.it.shouldBeDeniedWhenCalledByUser(CURRENT_USER, 'POST', '/api/accounts2');
```
bajtos: ditto for all test cases above.

test/access-control.integration.js
```diff
@@ -226,7 +238,101 @@ describe('access control - integration', function() {
...
+      });
+      lt.describe.whenCalledRemotely('DELETE', '/api/accounts2/:id', function() {
+        lt.it.shouldBeDenied();
+      });
+
+      describe('replace on POST verb', function() {
+        beforeEach(function(done) {
+          var self = this;
+          // Create an account under the given user
+          app.models.accountWithReplaceOnPUTfalse.create({
```
bajtos: Why is it necessary to create another user account, when we already have one created by `beforeEach` block on L285?

test/fixtures/access-control/common/models/account2.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "accountWithReplaceOnPUTfalse",
+  "plural": "accounts2",
```
bajtos: Please use a more descriptive URL path than `accounts2` (and `accounts` for the other model). How about `accounts-replacing` and `accounts-updating`?

bajtos:
> This patch is almost done; I would say 90% is done; could you please confirm you agree with my approach?

The approach looks good to me.

My main concern is that we don't have any tests verifying the default mapping (when `replaceOnPUT` is not set). Can we perhaps use the `User` model for this?

test/remoting.integration.js
```diff
@@ -238,5 +208,154 @@ describe('remoting - integration', function() {
         ];
         expect(methods).to.include.members(expectedMethods);
       });
+
+    it('has expected remote methods with model.settings.replaceOnPUT set to true', function() {
+      app.models.store.settings.replaceOnPUT = true;
+      app.models.store.setup();
```
bajtos: @Amir-61 Based on your earlier comment, this is expected to not work correctly. Do you need to create a new model with `replaceOnPUT:true` too?

test/remoting.integration.js
```diff
@@ -238,5 +208,154 @@ describe('remoting - integration', function() {
...
+      // https://docs.strongloop.com/display/public/LB/Exposing+models+over+REST
+      expect(methods).to.include.members(expectedMethods);
+    });
+  });
+});
+
+describe('With model.settings.replaceOnPUT false' +
+  'set to false', function() {
+  lt.beforeEach.withApp(app);
+  lt.beforeEach.givenModel('storeWithreplaceOnPUTfalse');
```
bajtos: Possible typo: `storeWithreplaceOnPUTfalse` vs. `storeWithReplaceOnPUTfalse`

test/remoting.integration.js
```diff
@@ -238,5 +208,154 @@ describe('remoting - integration', function() {
...
+        return m.name.indexOf('__') === -1;
+      })
+      .map(function(m) {
+        return formatMethod(m);
+      });
+
+    var expectedMethods = [
+      'create(data:object):storeWithreplaceOnPUTfalse POST /stores2',
+      'upsert(data:object):storeWithreplaceOnPUTfalse PATCH /stores2',
+//        'upsert(data:object):storeWithreplaceOnPUTfalse PUT /stores2',
```
bajtos: Commented out code - please uncomment or remove.

test/remoting.integration.js
```diff
@@ -238,5 +208,154 @@ describe('remoting - integration', function() {
...
+      'exists(id:any):boolean GET /stores2/:id/exists',
+      'findById(id:any,filter:object):storeWithreplaceOnPUTfalse GET /stores2/:id',
+      'replaceById(id:any,data:object):storeWithreplaceOnPUTfalse POST /stores2/:id/replace',
+      'find(filter:object):storeWithreplaceOnPUTfalse GET /stores2',
+      'findOne(filter:object):storeWithreplaceOnPUTfalse GET /stores2/findOne',
+      'updateAll(where:object,data:object):object POST /stores2/update',
+      'deleteById(id:any):object DELETE /stores2/:id',
+      'count(where:object):number GET /stores2/count',
+      'prototype.updateAttributes(data:object):storeWithreplaceOnPUTfalse PATCH /stores2/:id',
+//        'prototype.updateAttributes(data:object):storeWithreplaceOnPUTfalse PUT /stores2/:id',
```
bajtos: Ditto

test/remoting.integration.js
```diff
@@ -238,5 +208,154 @@ describe('remoting - integration', function() {
...
+    var methods = storeClass.methods
+      .filter(function(m) {
+        return m.name.indexOf('__') === -1;
+      })
+      .map(function(m) {
+        return formatMethod(m);
+      });
+
+    var expectedMethods = [
+      'create(data:object):storeWithReplaceOnPUTtrue POST /stores3',
```
bajtos: Can we use more descriptive URL paths please? E.g. `/stores-default`, `/stores-replacing`, `/stores-updating`.

bajtos:
> My main concern is that we don't have any tests verifying the default mapping (when `replaceOnPUT` is not set). Can we perhaps use the `User` model for this?

Let me clarify that. There is extensive coverage in `test/remoting.integration.js`, which is great.

It makes me wonder whether we should test access control for default `replaceOnPUT`? Or perhaps if we preserve `POST /replaceOrCreate` and `POST /:id/replace` endpoints, then the access control tests can invoke only these endpoints that are same regardless of `replaceOnPUT` settings.

Thoughts?

test/access-control.integration.js
Amir-61: > I think we should test the endpoints for `replaceOnPUT:false` too.
I already have tests for `replaceOnPUT:false` Please see:
https://github.com/strongloop/loopback/pull/2316/files#diff-eeafd17907dda5058f6e881724ba8c6eR245

test/access-control.integration.js
Amir-61: Good question....That was tricky; I copy & paste the `create` method from above L285 and `create` method here for `replace on POST verb here` (because after applying feedbacks the lines keep changing);

create method from above L285:
```
        app.models.accountWithReplaceOnPUTfalse.create({
          userId: self.user.id,
          balance: 100,
        }, function(err, act) {
          self.url = '/api/accounts2/' + act.id;

          done();
        });
```

create method here for `replace on POST verb here`:
```
          app.models.accountWithReplaceOnPUTfalse.create({
            userId: self.user.id,
            balance: 100,
          }, function(err, act) {
            self.url = '/api/accounts2/' + act.id + '/replace';
            done();
          });
```

As you see after creating an account we store url on `this.url`and that is used in [whenCalledRemotely](https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/test/helpers/loopback-testing-helper.js#L157) in `loopback-testing-helper.js`; please note that the url for `replace` has `/replace` at the end, so thats why I created another account for testing `replace` method, to have the right url when we call `whenCalledRemotely` in `loopback-testing-helper.js`

test/fixtures/access-control/common/models/account2.json
Amir-61: >Please use a more descriptive URL path than accounts2

Yes I could not come up with a good name :( Not sure if I understood your suggestion... are you suggesting to use?:

|  model name           | model plural  |
|-------------:| -----:|
| accountWithReplaceOnPUTfalse | accounts-updating |
| accountWithReplaceOnPUTtrue      |  accounts-replacing |

I applied name changes based on this table ^

Amir-61:
Please note there is a bug in `strong-remoting` which does not support multiple http-methods and paths; please see [here](https://github.com/strongloop/strong-remoting/blob/ac3093dcfbb787977ca0229b0f672703859e52e1/lib/rest-adapter.js#L622-L631); they simply return just the first element of `HttpMethod` and `Path`; therefore, thats why I commented the some methods [here](https://github.com/strongloop/loopback/pull/2316/files#diff-0d70d1046bcba629cf096a0d0ef51cc6R229), [here](https://github.com/strongloop/loopback/pull/2316/files#diff-0d70d1046bcba629cf096a0d0ef51cc6R273) and [here](https://github.com/strongloop/loopback/pull/2316/files#diff-0d70d1046bcba629cf096a0d0ef51cc6R319), because `strong-remoting` is just returning the first `httpMethod`, `path` and others (if there are any) tey are simply ignored in [strong-remoting](https://github.com/strongloop/strong-remoting/blob/ac3093dcfbb787977ca0229b0f672703859e52e1/lib/rest-adapter.js#L622-L631). I just tried to fix it in [strong-remoting](https://github.com/strongloop/strong-remoting/pull/305), but apparently that breaks lots of logic and test cases... So it requires me to dig more into this bug in `strong-remoting`.

After having support for array of results in `getHttpMethod` and `getFullPath` from `strong-remoting` I believe I need to have the following changes in `remoting.integration.js`:

`formatMethod()` in `remoting.integration.js` will be changed to:

```
function formatMethod(m) {
  var arr = [];
  var httpMethods = m.getHttpMethod();
  var fullPath = m.getFullPath();
  for(var i=0; i< httpMethods.length; i++) {
    arr.push([
      m.name,
      '(',
      m.accepts.map(function(a) {
        return a.arg + ':' + a.type;
      }).join(','),
      ')',
      formatReturns(m),
      ' ',
      httpMethods[i],
      ' ',
      fullPath[i],
    ].join(''));
  }
  return arr;
}
```

And finally we need to use `filter`, `map` and `reduce` to get methods like the following:

```
      var methods = storeClass.methods
        .filter(function(m) {
          return m.name.indexOf('__') === -1;
        })
        .map(function(m) {
          return formatMethod(m);
        })
        .reduce(function(p,c) {
          return p.concat(c);
        });
```

##### Note: I just added the methods which need to be changed for my future reference; since I have tested them and they work with array of results from `getHttpMethod` and `getFullPath` in `strong-remoting`.

Amir-61:
Thanks for your thoughtful review of this patch! I applied all of your comment except adding tests when `replaceOnPUT` is not set. Please see below:

>My main concern is that we don't have any tests verifying the default mapping (when replaceOnPUT is not set). Can we perhaps use the User model for this?

@bajtos Yes I was thinking of adding tests when `replaceOnPUT` is not set as well, but that makes the implementation of patch different between `2.x` and `3.x` and we cannot simply backport the same patch. should I still add tests when `replaceOnPUT` is not set?


Also apart from that, there is another bug in `strong-remoting` which is explained [here](https://github.com/strongloop/loopback/pull/2316#issuecomment-221737602) and it requires me to dig more into it to fix it.

Amir-61:
>It makes me wonder whether we should test access control for default `replaceOnPUT`? Or perhaps if we preserve `POST /replaceOrCreate` and `POST /:id/replace` endpoints, then the access control tests can invoke only these endpoints that are same regardless of `replaceOnPUT` settings.

>>@bajtos Yes I was thinking of adding tests when replaceOnPUT is not set in access-control.integration.js as well, but apart from what you mentioned that makes the implementation of patch different between 2.x and 3.x and we cannot simply backport the same patch.

Oh I see now I understood what you meant... yes and after preserving `POST /replaceOrCreate` and `POST /:id/replace` endpoints  even I don't think if I need to have two separate test suits for [/account with replaceOnPUT false](https://github.com/strongloop/loopback/pull/2316/files#diff-eeafd17907dda5058f6e881724ba8c6eR268) Vs. [/accounts with replaceOnPUT true'](https://github.com/strongloop/loopback/pull/2316/files#diff-eeafd17907dda5058f6e881724ba8c6eR172); they both would be identical I believe and makes the tests cleaner. 


Also apart from that, there is another bug in `strong-remoting` which is explained [here](https://github.com/strongloop/loopback/pull/2316#issuecomment-221737602) and it requires me to dig more into it to fix it.

test/access-control.integration.js
bajtos: Could you please find a way how to share the same instance and modify only `this.url`?

test/fixtures/access-control/common/models/account2.json
bajtos: Yes please :+1:

bajtos:
@Amir-61 In LoopBack, we already have a method mapped to two HTTP endpoints, see [remoting metadata for `exists()`](https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/lib/persisted-model.js#L634-L637). It's puzzling me why we haven't discovered the issue you are fixing by https://github.com/strongloop/strong-remoting/pull/305 earlier.

I would like to resolve the strong-remoting issue/confusion first, before I continue with reviewing this pull request.

Amir-61:
From @bajtos:
>It makes me wonder whether we should test access control for default replaceOnPUT? Or perhaps if we preserve POST /replaceOrCreate and POST /:id/replace endpoints, then the access control tests can invoke only these endpoints that are same regardless of replaceOnPUT settings.

>Thoughts?

Yes I agree with this.

From @Amir-61:
>yes and after preserving POST /replaceOrCreate and POST /:id/replace endpoints even I don't think if I need to have two separate test suits for /account with replaceOnPUT false Vs. /accounts with replaceOnPUT true'; they both would be identical so I will remove one; that makes the tests cleaner as well.

On the second thought I believe I should still have two separate test suits for [/account with replaceOnPUT false](https://github.com/strongloop/loopback/pull/2316/files#diff-eeafd17907dda5058f6e881724ba8c6eR262) Vs. [/accounts with replaceOnPUT true](https://github.com/strongloop/loopback/pull/2316/files#diff-eeafd17907dda5058f6e881724ba8c6eR172), although the result would be the same. However, as you agreed we don't need to test `access control` for default replaceOnPUT which would be either false `2.x` or true `3.x`, because they are already tested.

This patch is ready for another round of review. I have addressed all the issues and had a refactoring based on [Strong-remoting PR#305](https://github.com/strongloop/strong-remoting/pull/305). Please just note that tests on this PR start passing once we land [Strong-remoting PR#305](https://github.com/strongloop/strong-remoting/pull/305).

Amir-61:
@slnode test please

Amir-61:
@slnode test please

lib/persisted-model.js
```diff
@@ -616,15 +619,35 @@ module.exports = function(registry) {
       http: { verb: 'post', path: '/' },
     });
 
-    setRemoting(PersistedModel, 'upsert', {
+    var upsertOptions = {
       aliases: ['updateOrCreate'],
       description: 'Update an existing model instance or insert a new one into the data source.',
```
bajtos: I am proposing to modify the description to use "patch" too.

```
descripton: 'Patch an existing model instance...'
```

Amir-61:
@slnode test please

lib/persisted-model.js
```diff
@@ -741,13 +784,19 @@ module.exports = function(registry) {
       http: { verb: 'get', path: '/count' },
     });
 
-    setRemoting(PersistedModel.prototype, 'updateAttributes', {
+    var updateAttributesOptions = {
       description: 'Update attributes for a model instance and persist it into the data source.',
```
bajtos: `Patch attributes...`

test/access-control.integration.js
```diff
@@ -111,9 +111,15 @@ describe('access control - integration', function() {
           assert.equal(user.password, undefined);
         });
       });
+
+      // user has replaceOnPUT = false; so then both PUT and PATCH should be allowed for update
       lt.describe.whenCalledRemotely('PUT', '/api/users/:id', function() {
```
bajtos: This test will have to be changed when back-ported to 2.x, because `replaceById` will be exposed at `/api/users/:id/replace` only. Could you please add a note so that we don't forget?

bajtos: Oh, only now I realised that you are explicitly setting `replaceOnPUT=false` for `User` model. In that case, I think we need to add a test case verifying that `POST /api/users/:id/replace` is allowed.

test/fixtures/simple-integration-app/common/models/storetrue.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "storeWithReplaceOnPUTtrue",
```
bajtos: The filename `storetrue` is difficult to understand, could you please use a better name? Also what's the difference between the `stores-replacing.json` and `storetrue.json`?

test/fixtures/simple-integration-app/common/models/stores-updating.json
```diff
@@ -0,0 +1,19 @@
+{
```
bajtos: Please use a singular name in model files - `store-replacing.json`, `store-updating.json`.

test/remoting.integration.js
```diff
@@ -73,60 +73,30 @@ describe('remoting - integration', function() {
...
-          return m.name.indexOf('__') === -1;
-        })
-        .map(function(m) {
-          return formatMethod(m);
-        });
+      var methods = getMethods(storeClass.methods);
 
+      // TODO: there is a bug in `strong-remoting which does not support multiple
+      // http-methods and paths; please see:
+      // `https://github.com/strongloop/strong-remoting/blob/ac3093dcfbb787977ca0229b0f672703859e52e1/lib/rest-adapter.js#L622-L631
```
bajtos: Since the bug will be fixed by the time this patch is landed, can we remove this comment?

test/remoting.integration.js
```diff
@@ -73,60 +73,30 @@ describe('remoting - integration', function() {
...
+    function() {
       var storeClass = findClass('store');
-      var methods = storeClass.methods
-        .filter(function(m) {
-          return m.name.indexOf('__') === -1;
-        })
-        .map(function(m) {
-          return formatMethod(m);
-        });
+      var methods = getMethods(storeClass.methods);
```
bajtos: I read this line as "get methods from storeClass methods", which sounds a bit unusual to me. How about

```js
var methdos = getFormattedMethodsExcludingRelations(storeClass);
```

test/remoting.integration.js
```diff
@@ -137,13 +107,7 @@ describe('remoting - integration', function() {
...
     it('has expected remote methods for scopes', function() {
       var storeClass = findClass('store');
-      var methods = storeClass.methods
-        .filter(function(m) {
-          return m.name.indexOf('__') === 0;
-        })
-        .map(function(m) {
-          return formatMethod(m);
-        });
+      var methods = getScopedMethods(storeClass.methods);
```
bajtos: `getFormattedScopeMethods(storeClass)`?

test/remoting.integration.js
```diff
@@ -158,13 +122,7 @@ describe('remoting - integration', function() {
...
       function() {
         var widgetClass = findClass('widget');
-        var methods = widgetClass.methods
-          .filter(function(m) {
-            return m.name.indexOf('prototype.__') === 0;
-          })
-          .map(function(m) {
-            return formatMethod(m);
-          });
+        var methods = getPrototypeMethods(widgetClass.methods);
```
bajtos: `getFormattedPrototypeMethods(widgetClass)`?

test/remoting.integration.js
```diff
@@ -238,5 +185,185 @@ describe('remoting - integration', function() {
...
         expect(methods).to.include.members(expectedMethods);
       });
+
+    it('has expected remote methods with model.settings.replaceOnPUT set to true', function() {
+      var storeClass = findClass('store');
+      var methods = getMethods(storeClass.methods);
+
+      // TODO: there is a bug in `strong-remoting which does not support multiple
+      // http-methods and paths; please see:
+      // `https://github.com/strongloop/strong-remoting/blob/ac3093dcfbb787977ca0229b0f672703859e52e1/lib/rest-adapter.js#L622-L631
```
bajtos: Can we remove this comment too?

test/remoting.integration.js
```diff
@@ -238,5 +185,185 @@ describe('remoting - integration', function() {
...
+        'findById(id:any,filter:object):store GET /stores/:id',
+        'replaceById(id:any,data:object):store PUT /stores/:id',
+        'replaceById(id:any,data:object):store POST /stores/:id/replace',
+        'find(filter:object):store GET /stores',
+        'findOne(filter:object):store GET /stores/findOne',
+        'updateAll(where:object,data:object):object POST /stores/update',
+        'deleteById(id:any):object DELETE /stores/:id',
+        'count(where:object):number GET /stores/count',
+        'prototype.updateAttributes(data:object):store PATCH /stores/:id',
+        'createChangeStream(options:object):ReadableStream POST /stores/change-stream',
```
bajtos: We are duplicating many methods that are not relevant to this test. I am proposing to trim down the list of methods to only the entries for the four methods we are interested in: replaceById, replaceOrCreate, upsert, prototype.updateAttributes.

test/remoting.integration.js
```diff
@@ -238,5 +185,185 @@ describe('remoting - integration', function() {
...
+    this.app.models.storeWithReplaceOnPUTfalse.destroyAll(done);
+  });
+
+  it('should have expected remote methods',
+  function() {
+    var storeClass = findClass('storeWithReplaceOnPUTfalse');
+    var methods = getMethods(storeClass.methods);
+
+    var expectedMethods = [
+      'create(data:object):storeWithReplaceOnPUTfalse POST /stores-updating',
```
bajtos: Ditto - https://github.com/strongloop/loopback/pull/2316/files#r66057931

test/remoting.integration.js
```diff
@@ -238,5 +185,185 @@ describe('remoting - integration', function() {
...
+      'createChangeStream(options:object):ReadableStream POST /stores-updating/change-stream',
+    ];
+
+    // The list of methods is from docs:
+    // https://docs.strongloop.com/display/public/LB/Exposing+models+over+REST
+    expect(methods).to.include.members(expectedMethods);
+  });
+});
+
+describe('With model.settings.replaceOnPUT true' +
```
bajtos: What is the difference between this test and `it('has expected remote methods with model.settings.replaceOnPUT set to true'` above? I think it's enough to keep only one of these tests, isn't it?

test/remoting.integration.js
```diff
@@ -238,5 +185,185 @@ describe('remoting - integration', function() {
...
+  'set to true', function() {
```
bajtos: The "describe" name is "With model.settings.replaceOnPUT true set to true". I think the last part "set to true" should be removed?

lib/persisted-model.js
```diff
@@ -616,15 +619,35 @@ module.exports = function(registry) {
...
       returns: { arg: 'data', type: typeName, root: true },
-      http: { verb: 'put', path: '/' },
-    });
+      http: [{ verb: 'patch', path: '/' }],
+    };
+
+    if (!options.replaceOnPUT) {
+      upsertOptions.http.push({ verb: 'put', path: '/' });
+    }
+    setRemoting(PersistedModel, 'upsert', upsertOptions);
```
bajtos: I am proposing to change the name to `patchOrCreate` and move `upsert` to the list of aliases.

lib/persisted-model.js
```diff
@@ -741,13 +784,19 @@ module.exports = function(registry) {
...
       description: 'Update attributes for a model instance and persist it into the data source.',
       accessType: 'WRITE',
       accepts: { arg: 'data', type: 'object', http: { source: 'body' }, description: 'An object of model property name/value pairs' },
       returns: { arg: 'data', type: typeName, root: true },
-      http: { verb: 'put', path: '/' },
-    });
+      http: [{ verb: 'patch', path: '/' }],
+    };
+
+    setRemoting(PersistedModel.prototype, 'updateAttributes', updateAttributesOptions);
```
bajtos: I am proposing to rename this method to `patchAttributes` and move `updateAttributes` to the list of aliases.

In 2.x, we should preserve `updateAttributes` and set `pachAttributes` as an alias only.

lib/persisted-model.js
bajtos: In 2.x, we should preserve `upsert` and set `pachOrCreate` as an alias only.

bajtos:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
The CI failures would be passed once this [PR#2375](https://github.com/strongloop/loopback/pull/2375) lands. Please see [this comment](https://github.com/strongloop/strong-remoting/pull/302#issuecomment-224393189)

test/access-control.integration.js
Amir-61: I guess we decided not to include ACL test for `replace` for user model, since `replace` for `user` model?

test/remoting.integration.js
```diff
@@ -240,3 +183,148 @@ describe('remoting - integration', function() {
...
+  return result = methods.filter(function(m) {
+    return m.name.indexOf('prototype.__') === 0;
+  })
+  .map(function(m) {
+    return formatMethod(m);
+  })
+  .reduce(function(p, c) {
+    return p.concat(c);
+  });
+}
```
Amir-61: Also please note I have changed the logic for `formatMethod`, `getFormattedMethodsExcludingRelations `, `getFormattedScopeMethods`,... based on the changes in [strong-remoting](https://github.com/strongloop/strong-remoting/pull/305)...

Amir-61:
@slnode test please

test/access-control.integration.js
bajtos: Fair enough.

test/model.test.js
```diff
@@ -634,17 +634,19 @@ describe.onServer('Remote Methods', function() {
...
+        'replaceOrCreate',
         'find',
         'findOne',
         'updateAll', 'update',
         'deleteById',
         'destroyById',
         'removeById',
         'count',
-        'prototype.updateAttributes',
+        'prototype.patchAttributes', 'updateAttributes',
```
bajtos: I think the second item/alias should have `prototype.` prefix too.

```js
 'prototype.patchAttributes', 'prototype.updateAttributes'
```

test/remoting.integration.js
```diff
@@ -241,3 +184,130 @@ describe('remoting - integration', function() {
...
+      'patchOrCreate(data:object):storeWithReplaceOnPUTfalse PATCH /stores-updating',
+      'patchOrCreate(data:object):storeWithReplaceOnPUTfalse PUT /stores-updating',
+      'replaceOrCreate(data:object):storeWithReplaceOnPUTfalse POST /stores-updating/replaceOrCreate',
+      'replaceById(id:any,data:object):storeWithReplaceOnPUTfalse POST /stores-updating/:id/replace',
+      'prototype.patchAttributes(data:object):storeWithReplaceOnPUTfalse PATCH /stores-updating/:id',
+      'prototype.patchAttributes(data:object):storeWithReplaceOnPUTfalse PUT /stores-updating/:id',
+    ];
+
+    // The list of methods is from docs:
+    // https://docs.strongloop.com/display/public/LB/Exposing+models+over+REST
```
bajtos: I think this comment is no longer relevant, since we are testing a subset of the methods only. Should we remove it?

test/remoting.integration.js
```diff
@@ -241,3 +184,130 @@ describe('remoting - integration', function() {
...
+    var expectedMethods = [
+      'patchOrCreate(data:object):storeWithReplaceOnPUTtrue PATCH /stores-replacing',
+      'replaceOrCreate(data:object):storeWithReplaceOnPUTtrue POST /stores-replacing/replaceOrCreate',
+      'replaceOrCreate(data:object):storeWithReplaceOnPUTtrue PUT /stores-replacing',
+      'replaceById(id:any,data:object):storeWithReplaceOnPUTtrue POST /stores-replacing/:id/replace',
+      'replaceById(id:any,data:object):storeWithReplaceOnPUTtrue PUT /stores-replacing/:id',
+      'prototype.patchAttributes(data:object):storeWithReplaceOnPUTtrue PATCH /stores-replacing/:id',
+    ];
+    // The list of methods is from docs:
+    // https://docs.strongloop.com/display/public/LB/Exposing+models+over+REST
```
bajtos: ditto

test/model.test.js
bajtos: I think we either need to fix [L627](https://github.com/strongloop/loopback/pull/2316/files#diff-e43576aabfd6aa741af8ac58f1b4a699R627) to prepend `prototype.` prefix to `isStatic:false` methods, or change remoting definition of `prototype.patchAttributes` to specify method alias as `prototype.updateAttributes`. This depends on how strong-remoting treats aliases of prototype methods, which is something I am not sure about. Could you please check? It is also worth checking how loopback-connector-remote handles these aliases, to ensure code calling `myModelInstance.updateAttributes` keeps working (assuming the model is attached to a remote-backed datasource).

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@bajtos I addressed your feedback... Please review :-)

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

bajtos:
LGTM 👏

PR #2329, opened 2016-05-11
Expose replace methods 2.x

Amir-61:


Amir-61:
https://github.com/strongloop/loopback/pull/2330

PR #2330, opened 2016-05-11
WIP:Expose replace methods 2.x

Amir-61:
##### Description:
* `updateAttributes` and `updateOrCreate` endpoints use `PATCH` (forward compatibility) and `PUT`(backward compatibility) verb by default .
* Exposed `replaceById` and `replaceORCreate` via `POST` verb by default.
* Add a flag `model.settings.options.replaceOnPUT` for `2.X` to get `3.X` behaviour

**Note:** This patch is for `2.X` to be backported.

###### TODO:
- [ ] Add a test which confirms the functionality of flag `replaceOnPUT` for `2.x`

Related to https://github.com/strongloop/loopback/pull/2316 (Implementation for `3.x`)

Connect to https://github.com/strongloop/loopback/issues/1979


bajtos:
Hey @Amir-61, before I review the code changes, I'd like to discuss the description:

> `updateAttributes` and `updateOrCreate` endpoints use `PATCH` verb by default .

I think they should be mapped on both PUT and PATCH. PUT is needed for backwards compatibility, PATCH is for forward compatibility with LB 3.0

Thoughts?

> Exposed `replaceById` and `replaceORCreate` via `PATCH` verb by default.

This does not make much sense to me. I would expect these two methods to be exposed as `POST /api/mymodel/:id/replace` and `POST /api/mymodel/replaceOrCreate`.

What do you think?

> Add a flag `model.settings.options.newMapping` for `2.X` to get `3.X` behaviour

In my opinion, `newMapping` does not describe the intent/effect very well. How about `replaceOnPUT`?

Also I think the flag should be set directly on `model.settings`, I don't think there is an `options` property.

lib/persisted-model.js
```diff
@@ -483,6 +495,19 @@ module.exports = function(registry) {
...
   /**
+   * Replace set of attributes.  Performs validation before replacing.
+   *
+   * @param {Object} data Data to replace.
+   * @callback {Function} callback Callback function called with `(err, instance)` arguments.  Required.
+   * @param {Error} err Error object; see [Error object](http://docs.strongloop.com/display/LB/Error+object).
+   * @param {Object} instance Repalced instance.
+   */
+
+  PersistedModel.replaceById = function replaceById(data, cb) {
```
bajtos: I think this methods has a different signature:

```js
function replaceById(id, data, cb);
```

Doesn't it? Please update jsdoc comment too.

lib/persisted-model.js
```diff
@@ -483,6 +495,19 @@ module.exports = function(registry) {
   };
 
   /**
+   * Replace set of attributes.  Performs validation before replacing.
+   *
+   * @param {Object} data Data to replace.
+   * @callback {Function} callback Callback function called with `(err, instance)` arguments.  Required.
+   * @param {Error} err Error object; see [Error object](http://docs.strongloop.com/display/LB/Error+object).
+   * @param {Object} instance Repalced instance.
```
bajtos: Typo, should be `Replaced`

lib/persisted-model.js
```diff
@@ -483,6 +495,19 @@ module.exports = function(registry) {
   };
 
   /**
+   * Replace set of attributes.  Performs validation before replacing.
```
bajtos: `Replace set of attributes` sounds a bit vague to me. How about 

```
Replace all data of a model instance identified by the given ID
```

or perhaps

```
Replace all attributes of a model instance identified by the given ID
```

lib/persisted-model.js
```diff
@@ -549,6 +574,21 @@ module.exports = function(registry) {
     var PersistedModel = this;
     var typeName = PersistedModel.modelName;
     var options = PersistedModel.settings;
+    // This is the defualt verb used for 2.X
+    var configurableVerb = {
+      updateAttributes: 'put',
+      updateOrCreate: 'put',
+      replaceOrCreate: 'patch',
+      replaceById: 'patch',
```
bajtos: As I commented earlier, I find it weird for `PATCH` to make a full replace. It also makes the surface of the breaking change between 2.x and 3.x bigger.

lib/persisted-model.js
```diff
@@ -619,6 +668,20 @@ module.exports = function(registry) {
...
     });
 
+    setRemoting(PersistedModel, 'replaceById', {
+      description: 'Replace attributes for a model instance and persist it into the data source.',
+      accessType: 'WRITE',
+      accepts: [
+        { arg: 'id', type: 'any', description: 'Model id', required: true,
+          http: { source: 'put' }},
+          { arg: 'data', type: 'object', http: { source: 'body' }, description:
+        'An object of model property name/value pairs' },
```
bajtos: Should the description say the same as in `replaceOrCreate`, i.e. `Model instance data`?

lib/persisted-model.js
```diff
@@ -619,6 +668,20 @@ module.exports = function(registry) {
...
+      ],
+      returns: { arg: 'data', type: typeName, root: true },
+      http: { verb: configurableVerb.replaceById, path: '/:id' },
+      rest: { after: convertNullToNotFoundError },
```
bajtos: Could you please check that `replaceById` returns `null` when the `id` was not found? I don't remember if we implemented this feature. I think this may have been the case of `prototype.updateAttributes` because of the way how shared constructor works.

For `replaceById`, I think it makes more sense to raise not-found-error directly from `replaceById` implementation, so that we don't need this extra bit here.

test/remoting.integration.js
```diff
@@ -122,12 +122,15 @@ describe('remoting - integration', function() {
         .map(function(m) {
           return formatMethod(m);
         });
-
+      // This is the list of expected default endpoints
+      // for LB 3.X (They are different for LB 2.X)
```
bajtos: Please fix this comment, we are in 2.x

lib/persisted-model.js
bajtos: Also please add an integration test to verify this case.

bajtos:
Please add some unit-tests calling the new REST API endpoints, and also tests to verify the new config options.

superkhau:
>Add a flag model.settings.options.newMapping for 2.X to get 3.X behaviour

@bajtos Do you think its worth maybe doing all 2.x settings in one setting for futureproofing? I'm proposing something like:

```
model.settings.options.v2.newMapping = ...

//and in the settings
...
"options": {
  "v2": {
      "newMapping": false
  }
}
...
```

Then we can list the options by default when scaffolding models, etc and have it self documenting. Users can simply change the config and get the behaviour without having to dig through too much documentation (only for what is "newMapping"). Thoughts?

Amir-61:
Closing this PR in favour of https://github.com/strongloop/loopback/pull/2316

Amir-61:
ummmm on the second thought, it seems this task requires very different implementation for `2.x`, so I re-open this PR to work on the implementation for `2.x`.

Amir-61:
@slnode test please

bajtos:
> Do you think its worth maybe doing all 2.x settings in one setting for futureproofing? I'm proposing something like.

No, I don't think that's a good idea. I'd like us to follow "feature flags" design pattern, where users are enabling/disabling features.

> Then we can list the options by default when scaffolding models, etc and have it self documenting. Users can simply change the config and get the behaviour without having to dig through too much documentation (only for what is "putMapping v2"). Thoughts?

In my opinion, it's your proposal that will lead to users wondering what the heck "put mapping 2.x" means. With a feature flag "replaceOnPUT" or "updateOnPUT", it should be very clear what feature (behaviour) is enabled/disabled.

@Amir-61 it's important to use the same name for this feature flag both in 2.x and 3.x branches, to make the upgrade process easier.

Here is what I am envisioning:

 1) Users preparing for upgrade from 2.x to 3.0 can temporarily update to 3.0 to find what's not working.
 2) One of the breaking change is the behaviour of PUT. Because there are too many other things to address, the user has an option to postpone fixing this part for later by adding `replaceOnPUT` flag to their models.
 3) Because the codebase is not ready for 3.0 yet, the version change is reverted. However, the feature flag can stay in place because it's supported by 2.x too.
 4) In the future, the application can be reworked to support the new 3.0 mapping, even while it's still running on LoopBack 2.x.

That way the update process can be easily split into many independent and incremental steps.

bajtos:
> ummmm on the second thought, it seems this task requires very different implementation between 2.x and 3.x

Why is that? What is the difference in the implementation? I would like to keep both implementations as close as possible, to simplify both maintenance (back-porting changes) and upgrade process for the users.

superkhau:
>No, I don't think that's a good idea. I'd like us to follow "feature flags" design pattern, where users are enabling/disabling features.
>
>In my opinion, it's your proposal that will lead to users wondering what the heck "put mapping 2.x" means. With a feature flag "replaceOnPUT" or "updateOnPUT", it should be very clear what feature (behaviour) is enabled/disabled.

Thanks for the feedback. I took a look at the "feature toggle" design pattern in some articles and what you said makes sense. +1 for sticking with feature flags, but we should also get rid of them as soon as possible when releasing new versions (ie. remove the LB2 flags ASAP when upgrading to LB3).

Issue #2324, opened 2016-05-10
By default, passwords can't be all numbers because of a weak type check in user.js


notbrain:
I would think a password that is all numbers would be a bit more secure than those that are of the well-known "letmein" or "password123" variety, especially if we enforce 8 digits.

Because of this line in user.js, passwords cannot be all numbers:

https://github.com/strongloop/loopback/blob/master/common/models/user.js#L610

This needs to be more intelligent and allow a full string of numbers for a password, or at the very least be highly publicized instead of buried inside the built-in code. As is, this is a bug that I had to go diving into the code to figure out why a bunch of our signups were being rejected for not having a password upon model creation, with no indication as to why the password was not making it all the way through to the database.

notbrain:
Is there any harm in adding an explicit conversion like this? Not sure of the security ramifications but password requirements should be up to the app owners. Kind of pointless to do a `toString()` and immediately after check that it's a string, so there's some thinking to do around this.

    UserModel.setter.password = function(plain) {

      plain = plain.toString();

      if (typeof plain !== 'string') {
        return;
      }
      if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
        // The password is already hashed. It can be the case
        // when the instance is loaded from DB
        this.$password = plain;
      } else {
        this.$password = this.constructor.hashPassword(plain);
      }
    };

richardpringle:
@notbrain, good question about security ramifications. I think it might be better to check for the following:
```node
if (typeof plain === 'number) {
  plain = plain.toString();
  // or at the very least throw a meaningful error
}
```

Do you want to submit the PR? You found the bug, you should get the credit for the fix!

notbrain:
Thanks, yeah I like checking for number type and only converting at that point.

notbrain:
https://github.com/strongloop/loopback/pull/2335

notbrain:
Ironically there is a test that tests for missing email that uses an all number password. Heh.

https://github.com/strongloop/loopback/blob/master/test/user.test.js#L114

Will try to carve out some time to add a test to include one that is valid with an all number password.

richardpringle:
Yeah weird. I only tried to reproduce the issue through *strong-remoting*. Once you write your test, we should know if a different change needs to be made.

bajtos:
As I commented in https://github.com/strongloop/loopback/pull/2335#issuecomment-247351225, I think this is a bug of strong-remoting that should have been already fixed in 3.0 (alpha).

PR #2335, opened 2016-05-11
Allow passwords to be all numbers

notbrain:
Allow number passwords by converting number to string before hashing.

connect to #2324

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -604,8 +604,8 @@ module.exports = function(User) {
     };
 
     UserModel.setter.password = function(plain) {
-      if (typeof plain !== 'string') {
-        return;
+      if (typeof plain === 'number') {
+        plain = plain.toString();
       }
       if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
         // The password is already hashed. It can be the case


slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

common/models/user.js
```diff
@@ -604,8 +604,8 @@ module.exports = function(User) {
     };
 
     UserModel.setter.password = function(plain) {
-      if (typeof plain !== 'string') {
-        return;
+      if (typeof plain === 'number') {
+        plain = plain.toString();
       }
```
richardpringle: I don't think that this check should be removed entirely. What if someone makes a mistake somewhere and `plain` is of type `object`? Instead, I think that you should add the check for the number above the check for string. 

This could also be cleaned up to one line:
```node
if (typeof plain !== 'string') return;
```

common/models/user.js
```diff
@@ -604,8 +604,8 @@ module.exports = function(User) {
     };
 
     UserModel.setter.password = function(plain) {
-      if (typeof plain !== 'string') {
-        return;
+      if (typeof plain === 'number') {
+        plain = plain.toString();
```
richardpringle: Not necessary but could be cleaned up to one line as suggested in my other comment.

common/models/user.js
reichert621: It's probably better to take advantage of the `validatePassword` function instead of just doing an early `return`... maybe by checking if `plain` is a string on line 610 below, which allows us to reach the `hashPassword` function where `validatePassword` is called:
```javascript
if (typeof plain === 'string' && plain.indexOf('$2a$') === 0 && plain.length === 60) {
  // ...
} else {
}
```
so if `plain` is of type `object` (e.g. , it will get passed along and validated:
```javascript
  this.validatePassword(plain);
  // ...
};

  if (typeof plain === 'string' && plain) {
    return true;
  }
  var err =  new Error('Invalid password: ' + plain);
  err.statusCode = 422;
  throw err;
};
```
at which point it will throw a better error message `Invalid password: { pw: 'abc123' }`, instead of what we currently get (`The User instance is not valid. Details: password can't be blank (value: undefined)`)

What do you think?

notbrain: @reichert621 I like it. But doesn't that still leave the loophole where all-number passwords are still not converted to strings? Your version will throw the error in that case, but I'm not sure it should be up to the framework to say if they should be allowed or not. `18181818` should be converted to `'18181818'` at some point, and in such a way that an object would *not* get converted to a string.

richardpringle: The enforcement of at least one non-numeric character is a bug as labelled in #2324. The hashing algorithm requires a string of characters but a number should be coerced to a string and that should be the outcome of this PR.

I just want to be clear that the current behavior is not what is expected.

@notbrain, looking forward to seeing the rest of your changes.

bajtos:
There is one thing I don't understand. How is it possible that the `password` argument is of type `number`? I think that points to bug in strong remoting that should have been fixed by https://github.com/strongloop/strong-remoting/pull/343.

Before landing this patch, could you @superkhau or @richardpringle please verify that the problem is present in LoopBack 3.0 too?

Here is the test case I am thinking about:
 - request `POST /api/users`
 - send a digits-only password in the body, e.g. 

  ```json
{
      "email": "test@example.com", 
      "password": "123"
}
```

In LB 2.x, I suspect that strong-remoting converts the `password` property to a number. In LB 3.x, the value should stays as a string.

slnode:

slnode:

slnode:

slnode:

superkhau:
>There is one thing I don't understand. How is it possible that the password argument is of type number?

Good catch, I believe this is wrong too, we need to make sure strings and only numbers work and coerce the numbers into strings as stated by @bajtos. @richardpringle Can you verify this?

bajtos:
> Good catch, I believe this is wrong too, we need to make sure strings and only numbers work and coerce the numbers into strings as stated by

No magic please. Number arguments should be rejected, it's the responsibility of the client to correctly encode model property values (send password as strings, not numbers).

superkhau:
@notbrain Are you still interesting in working on this? Please address the review and add some tests to verify your changes.

PR #2368, opened 2016-05-23
Explicitly specified indexes on the User and ACL models

penguinpowernz:
This change allows the default common models to work out of the box with the Redis loopback connector.  Without it, this error is shown:

```
Error: User: no indexes found for username impossible to sort and filter using redis connector
```

After this, normal operations can work properly.  I took this to mean that other connectors implicitly determine the indexes while the Redis connector cannot.

I understand that this may be a problem with the Redis connector not implicitly determining the correct indexes, but it is honestly a lot easier and a lot simpler to make the change here.

### DIFF
--- a/common/models/acl.json
+++ b/common/models/acl.json
@@ -3,13 +3,18 @@
   "properties": {
     "model": {
       "type": "string",
+      "index": true,
       "description": "The name of the model"
     },
     "property": {
       "type": "string",
+      "index": true,
       "description": "The name of the property, method, scope, or relation"
     },
-    "accessType": "string",
+    "accessType": {
+      "type": "string",
+      "index": true
+    },
     "permission": "string",
     "principalType": "string",
     "principalId": "string"
--- a/common/models/user.json
+++ b/common/models/user.json
@@ -5,7 +5,8 @@
       "type": "string"
     },
     "username": {
-      "type": "string"
+      "type": "string",
+      "index": true
     },
     "password": {
       "type": "string",
@@ -13,7 +14,8 @@
     },
     "email": {
       "type": "string",
-      "required": true
+      "required": true,
+      "index": true
     },
     "emailVerified": "boolean",
     "verificationToken": "string"


slnode:

penguinpowernz:
BUMP

slnode:

slnode:

slnode:

slnode:

penguinpowernz:
Hmmm is jenkins down?  Can't view those Details links...

superkhau:
LGTM. @raymondfeng I need a second opinion here though, can you take a quick look?

taoyuan:
+1

superkhau:
@bajtos Can I get a second opinion here? Do you know of any ramifications to defaulting with `index: true` to the some of the properties in this PR?

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode test please

bajtos:
@slnode ok to test

bajtos:
> Can I get a second opinion here? Do you know of any ramifications to defaulting with `index: true` to the some of the properties in this PR?

I think good indexes is something we should have implemented long time ago for all built-in models.

I have two questions that I would like to get answered to move this patch forward:

 - Is this patch building a single composite index (i.e. `(model, property, accessType)` for ACL) or a bunch of single-column indexes (i.e. an index for `model`, an index for `property`, an index for `accessType` in ACL)? 

 - How are other connectors, e.g. MySQL and MongoDB, going to interpret the updated model definitions? Will they create the correct indexes too?

penguinpowernz:
I have no idea what that means... :/

superkhau:
@bajtos I spoke with @raymondfeng and he agrees with default indexing too. However we need to be careful with regards to composite key indexing -- a requirement to land PRs that modify builtin models.

@penguinpowernz I think just adding index:true isn't going to work for acls (and maybe the other builtin models too). See https://docs.strongloop.com/display/public/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-CompositeIDs and use composite keys where necessary.

bajtos:
>  I think just adding index:true isn't going to work for acls (and maybe the other builtin models too).

@superkhau Let's not make our decisions on assumptions please. We should run the code, investigate the SQL/Redis/MongoDB queries being made by LoopBack and then figure out what are the right set of indices we need to create. Whether we need multiple indices (one per each column) or a single composite index instead.

>  See https://docs.strongloop.com/display/public/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-CompositeIDs and use composite keys where necessary.

This doesn't make sense to me. We are talking about a built-in LoopBack model that is being queried in multiple ways.

Sometimes we want to look up a user by their id, sometimes by their username, sometimes by their email. There is no way for a composite id to support all these use cases. To me, the part changing the `User` model makes perfect sense and should be landed. See also my https://github.com/strongloop/loopback/pull/2849#issuecomment-253452454 where I explain in more detail why we should use indices.

The changes in `ACL` is the part that I am not confident about, simply because I don't know what queries are typically made for that model.

cc @raymondfeng

superkhau:
>@superkhau Let's not make our decisions on assumptions please. We should run the code, investigate the SQL/Redis/MongoDB queries being made by LoopBack and then figure out what are the right set of indices we need to create. Whether we need multiple indices (one per each column) or a single composite index instead.

+1 for not making decisions based on assumptions. I'm on the same page as you regarding the changes to the User model being acceptible. I'm also not sure about the ACL model portion -- we should get @raymondfeng to chime in on that part.

@raymondfeng Can you chime in here regarding the composite key + ACL and how it relates?

bajtos:
The issue is more complex than it seems. I am closing this pull request for now, let's move the discussion to https://github.com/strongloop/loopback/issues/3011

PR #2375, opened 2016-05-24
[SEMVER-MAJOR] Fix remoting strong-error-handler

davidcheung:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/852

addresses test cases failing and stacktrace showing up on `npm test` 

should land with https://github.com/strongloop/strong-remoting/pull/302

### DIFF
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -613,6 +613,7 @@ describe('app', function() {
     var app, db;
     beforeEach(function() {
       app = loopback();
+      app.set('remoting', { errorHandler: { debug: true, log: false }});
       db = loopback.createDataSource({ connector: loopback.Memory });
     });
 
--- a/test/remote-connector.test.js
+++ b/test/remote-connector.test.js
@@ -13,6 +13,7 @@ describe('RemoteConnector', function() {
     beforeEach: function(done) {
       var test = this;
       remoteApp = loopback();
+      remoteApp.set('remoting', { errorHandler: { debug: true, log: false }});
       remoteApp.use(loopback.rest());
       remoteApp.listen(0, function() {
         test.dataSource = loopback.createDataSource({
--- a/test/replication.rest.test.js
+++ b/test/replication.rest.test.js
@@ -462,6 +462,7 @@ describe('Replication over REST', function() {
 
   function setupServer(done) {
     serverApp = loopback();
+    serverApp.set('remoting', { errorHandler: { debug: true, log: false }});
     serverApp.enableAuth();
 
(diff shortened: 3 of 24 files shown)


davidcheung:
@bajtos PTAL

test/app.test.js
```diff
@@ -613,6 +613,7 @@ describe('app', function() {
     var app, db;
     beforeEach(function() {
       app = loopback();
+      app.set('remoting', { errorHandler: { debug: false, log: false }});
```
bajtos: I think it's better to set `debug:true` in tests, so that we get more details in error messages. Thoughts?

test/remote-connector.test.js
```diff
@@ -13,6 +13,7 @@ describe('RemoteConnector', function() {
     beforeEach: function(done) {
       var test = this;
       remoteApp = loopback();
+      remoteApp.set('remoting', { errorHandler: { debug: false, log: false }});
```
bajtos: `debug:true`?

test/replication.rest.test.js
```diff
@@ -462,6 +462,7 @@ describe('Replication over REST', function() {
 
   function setupServer(done) {
     serverApp = loopback();
+    serverApp.set('remoting', { errorHandler: { debug: false, log: false }});
```
bajtos: `debug:true`?

bajtos:
@davidcheung the patch looks mostly good, my main concern is the inconsistency in `debug` setting across tests - most tests have `debug:true`, few have `debug:false`. Can we unify it and always set `debug:true`?

test/app.test.js
davidcheung: agreed, not sure what i was thinking

davidcheung:
@bajtos PTAL again please

bajtos:
@davidcheung the code changes LGTM, but the CI is all red. Please investigate. If it's something trivial, then feel free to land this without further reviews.

Also, could you please capture somewhere in release notes that by default, LoopBack 3.0-powered applications will log failed requests to stderr? We didn't log errors in 2.x, I think it may surprise people when they discover for the first time. The section in release notes should include instructions how to disable these log messages (revert back to 2.x behavior).

davidcheung:
@bajtos i will retrigger ci once strong-remoting lands, which should pass all the tests, the change in this PR is disabling stacktrace on new `rest-adapter`'s  error-handling, and removing the current ones

rmg:
@slnode test please

PR #2401, opened 2016-06-03
[SEMVER-MAJOR] Discard sugar method for model creation

gunjpan:
Current implementation of `app.model(modelName, settings)`
works as a sugar for model creation. In 3.0, this is
not supported anymore. This implementation reports an
error when sugar is used for model creation.
Includes:
 - Updated app.model() method
 - Fixed test cases reflecting the change

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/772

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -194,3 +194,12 @@ module:
 
 See also [loopback#2564](https://github.com/strongloop/loopback/pull/2564)
 and the official [documentation](https://docs.strongloop.com/display/APIC/Using+current+context)
+
+## Remove sugar for creating models
+
+`app.model(modelName, settings)`, a sugar for creating non-existing model, is
+now removed in favor of promoting use of:
+- `app.registry.createModel(modelName, properties, options)` to create new model
+- `app.model(modelCtor, config)` to update existing model and attach it to app
+
+Please see [related code change](https://github.com/strongloop/loopback/pull/2401) here.
--- a/example/colors/app.js
+++ b/example/colors/app.js
@@ -14,9 +14,9 @@ var schema = {
   name: String,
 };
 
-var Color = app.model('color', schema);
-
-app.dataSource('db', { adapter: 'memory' }).attach(Color);
+app.dataSource('db', { connector: 'memory' });
+var Color = app.registry.createModel('color', schema);
+app.model(Color, { dataSource: 'db' });
 
 Color.create({ name: 'red' });
 Color.create({ name: 'green' });
--- a/example/replication/app.js
+++ b/example/replication/app.js
@@ -5,9 +5,11 @@
 
 var loopback = require('../../');
 var app = loopback();
-var db = app.dataSource('db', { connector: loopback.Memory });
-var Color = app.model('color', { dataSource: 'db', options: { trackChanges: true }});
-var Color2 = app.model('color2', { dataSource: 'db', options: { trackChanges: true }});
+var db = app.dataSource('db', { connector: 'memory' });
+var Color = app.registry.createModel('color', {}, { trackChanges: true });
+app.model(Color, { dataSource: 'db' });
+var Color2 = app.registry.createModel('color2', {}, { trackChanges: true });
+app.model(Color2, { dataSource: 'db' });
 var target = Color2;
 var source = Color;
 var SPEED = process.env.SPEED || 100;
--- a/lib/application.js
+++ b/lib/application.js
@@ -101,7 +101,7 @@ app.disuse = function(route) {
  * app.model(User, { dataSource: 'db' });
  *```
  *
- * @param {Object|String} Model The model to attach.
+ * @param {Object} Model The model to attach.
  * @options {Object} config The model's configuration.
  * @property {String|DataSource} dataSource The `DataSource` to which to attach the model.
  * @property {Boolean} [public] Whether the model should be exposed via REST API.
@@ -114,29 +114,15 @@ app.model = function(Model, config) {
   var isPublic = true;
   var registry = this.registry;
 
+  if (typeof Model === 'string') {
+    var msg = 'app.model(modelName, settings) is no longer supported. ' +
+      'Use app.registry.createModel(modelName, definition) and ' +
+      'app.model(ModelCtor, config) instead.';
+    throw new Error(msg);
+  }
+
   if (arguments.length > 1) {
     config = config || {};
-    if (typeof Model === 'string') {
-      // create & attach the model - backwards compatibility
-
-      // create config for loopback.modelFromConfig
-      var modelConfig = extend({}, config);
-      modelConfig.options = extend({}, config.options);
-      modelConfig.name = Model;
-
-      // modeller does not understand `dataSource` option
-      delete modelConfig.dataSource;
-
-      Model = registry.createModel(modelConfig);
-
-      // delete config options already applied
-      ['relations', 'base', 'acls', 'hidden', 'methods'].forEach(function(prop) {
-        delete config[prop];
-        if (config.options) delete config.options[prop];
-      });
-      delete config.properties;
-    }
-
     configureModel(Model, config, this);
     isPublic = config.public !== false;
   } else {
@@ -186,7 +172,7 @@ app.model = function(Model, config) {
  * });
  * ```
  *
- * 2. Use `app.model` to access a model by name.
+ * 2. Use `app.models` to access a model by name.
  * `app.models` has properties for all defined models.
  *
  * The following example illustrates accessing the `Product` and `CustomerReceipt` models
@@ -201,8 +187,10 @@ app.model = function(Model, config) {
  *   }
  * });
  *
- * app.model('product', {dataSource: 'db'});
- * app.model('customer-receipt', {dataSource: 'db'});
+ * var productModel = app.registry.createModel('product');
+ * app.model(productModel, {dataSource: 'db'});
+ * var customerReceiptModel = app.registry.createModel('customer-receipt');
+ * app.model(customerReceiptModel, {dataSource: 'db'});
  *
  * // available based on the given name
  * var Product = app.models.Product;
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -393,19 +393,17 @@ describe('security ACLs', function() {
 });
 
 describe('access check', function() {
-  var app;
-  before(function() {
-    app = loopback();
-    app.set('remoting', { errorHandler: { debug: true, log: false }});
-    app.use(loopback.rest());
-    app.enableAuth();
-    app.dataSource('test', { connector: 'memory' });
-  });
-
   it('should occur before other remote hooks', function(done) {
-    var MyTestModel = app.model('MyTestModel', { base: 'PersistedModel', dataSource: 'test' });
+    var app = loopback();
+    var MyTestModel = app.registry.createModel('MyTestModel');
     var checkAccessCalled = false;
     var beforeHookCalled = false;
+
+    app.use(loopback.rest());
+    app.set('remoting', { errorHandler: { debug: true, log: false }});
+    app.enableAuth();
+    app.dataSource('test', { connector: 'memory' });
+    app.model(MyTestModel, { dataSource: 'test' });
 
     // fake / spy on the checkAccess method
     MyTestModel.checkAccess = function() {
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -611,11 +611,12 @@ describe('app', function() {
   });
 
   describe('app.model(Model)', function() {
-    var app, db;
+    var app, db, MyTestModel;
     beforeEach(function() {
       app = loopback();
       app.set('remoting', { errorHandler: { debug: true, log: false }});
       db = loopback.createDataSource({ connector: loopback.Memory });
+      MyTestModel = app.registry.createModel('MyTestModel');
     });
 
     it('Expose a `Model` to remote clients', function() {
@@ -626,7 +627,7 @@ describe('app', function() {
       expect(app.models()).to.eql([Color]);
     });
 
-    it('uses singlar name as app.remoteObjects() key', function() {
+    it('uses singular name as app.remoteObjects() key', function() {
       var Color = PersistedModel.extend('color', { name: String });
       app.model(Color);
       Color.attachTo(db);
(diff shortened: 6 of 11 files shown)


gunjpan:
@bajtos : Please review, thank you.

gunjpan:
@slnode test please

3.0-RELEASE-NOTES.md
```diff
@@ -127,3 +127,12 @@ removed properties together with the middleware module name to use instead:
 We have also removed `loopback.mime`, which was always set to `undefined`.
 
 See [loopback#2349](https://github.com/strongloop/loopback/pull/2394).
+
+## Remove sugar for creating models
+
+`app.model(modelName, settings)`, a sugar for creating non-existing model, is
+now removed in favor of promoting use of:
+- `app.registry.createModel(modelName, properties, options) to create new model
```
bajtos: Missing backslash character after `)`

example/colors/app.js
```diff
@@ -12,9 +12,9 @@ var schema = {
   name: String,
 };
 
-var Color = app.model('color', schema);
-
+var Color = app.registry.createModel('color', schema);
 app.dataSource('db', { adapter: 'memory' }).attach(Color);
+app.model(Color);
```
bajtos: While you are changing this example, please rework it to the following:

```js

var Color = app.registry.createModel('color', schema);
app.model(Color, { dataSource: 'db' });
```
var Color = app.

example/replication/app.js
```diff
@@ -6,8 +6,12 @@
 var loopback = require('../../');
 var app = loopback();
 var db = app.dataSource('db', { connector: loopback.Memory });
-var Color = app.model('color', { dataSource: 'db', options: { trackChanges: true }});
-var Color2 = app.model('color2', { dataSource: 'db', options: { trackChanges: true }});
+var Color = app.model(
+  app.registry.createModel('color', {}, { trackChanges: true }),
+  { dataSource: 'db' });
```
bajtos: Uff, there is too much stuff happening in a single statement. Please split this into two statements:

```js
var Color = app.registry.createModel('color', {}, { trackChanges: true });
app.model(Color, { dataSource: 'db' });
```

example/replication/app.js
```diff
@@ -6,8 +6,12 @@
...
+var Color2 = app.model(
+  app.registry.createModel('color2', {}, { trackChanges: true }),
+  { dataSource: 'db' });
```
bajtos: ditto

lib/application.js
```diff
@@ -115,24 +115,9 @@ app.model = function(Model, config) {
...
-
-      // delete config options already applied
-      ['relations', 'base', 'acls', 'hidden', 'methods'].forEach(function(prop) {
-        delete config[prop];
-        if (config.options) delete config.options[prop];
-      });
-      delete config.properties;
+      throw new Error('Invalid model type of string found: ' +
+        Model +
+        '. Provide a valid model object to attach.');
```
bajtos: I am proposing to reword the message a bit.

```js
var msg = '`app.model(modelName, settings)` is no longer supported. ' + 
  'Use `app.registry.createModel(modelName, definition)` and `app.model(modelCtor, config)` instead.';
throw new Error(msg);
```

test/acl.test.js
```diff
@@ -399,10 +399,13 @@ describe('access check', function() {
...
     app.enableAuth();
     app.dataSource('test', { connector: 'memory' });
+    var MyTestModel = app.registry.createModel('MyTestModel');
+    // attach model to app and datasource
+    app.model(MyTestModel, { dataSource: 'test' });
   });
 
   it('should occur before other remote hooks', function(done) {
-    var MyTestModel = app.model('MyTestModel', { base: 'PersistedModel', dataSource: 'test' });
+    var MyTestModel = app.models.MyTestModel;
```
bajtos: Please move the model creation from `before` hook to the test. Unless I am missing some reasons why it won't work?

test/app.test.js
```diff
@@ -687,76 +688,27 @@ describe('app', function() {
       });
     });
 
-    it('accepts null dataSource', function() {
-      app.model('MyTestModel', { dataSource: null });
-    });
```
bajtos: This test should be preserved. Just pass model ctor instead of the model name, keep the second config arg without changes.

test/app.test.js
```diff
@@ -610,10 +610,11 @@ describe('app', function() {
   });
 
   describe('app.model(Model)', function() {
-    var app, db;
+    var app, db, MyTestModel;
     beforeEach(function() {
       app = loopback();
       db = loopback.createDataSource({ connector: loopback.Memory });
+      MyTestModel = app.registry.createModel('MyTestModel');
```
bajtos: I take my comment back. I see that this model is used in several test, in which case it's best to reuse it.

test/app.test.js
```diff
@@ -687,76 +688,27 @@ describe('app', function() {
...
-    it('accepts false dataSource', function() {
-      app.model('MyTestModel', { dataSource: false });
+    it('accepts null dataSource', function(done) {
+      app.model(MyTestModel, { dataSource: null });
+      expect(MyTestModel.dataSource).to.eql(null);
+      done();
     });
 
-    it('should not require dataSource', function() {
-      app.model('MyTestModel', {});
```
bajtos: This test should be preserved. Just pass model ctor instead of the model name, keep the second config arg without changes.

test/app.test.js
bajtos: Hmm, github formatted the patch in such way that I did not realise this test is preserved below. Ignore my comment.

test/app.test.js
bajtos: Hmm, github formatted the patch in such way that I did not realise this test is preserved below. Ignore my comment.

test/app.test.js
```diff
@@ -687,76 +688,27 @@ describe('app', function() {
...
-
-    it('defaults config.public to be true', function() {
-      app.model('foo', {
-        dataSource: 'db',
-      });
-      expect(app.models.foo.app).to.equal(app);
-      expect(app.models.foo.shared).to.equal(true);
+    it('throws error if model typeof string is passed', function(done) {
+      var fn = function() { app.model('MyTestModel'); };
+      expect(fn).to.throw(Error);
```
bajtos: Please check that the error contains the expected message.

test/app.test.js
```diff
@@ -687,76 +688,27 @@ describe('app', function() {
...
+      done();
```
bajtos: Since this is sync test, there is no need to use `done`. Please remove the argument definition and invocation.

test/change-stream.test.js
```diff
@@ -9,7 +9,8 @@ describe('PersistedModel.createChangeStream()', function() {
       var test = this;
       var app = loopback({ localRegistry: true });
       var ds = app.dataSource('ds', { connector: 'memory' });
-      this.Score = app.model('Score', {
+      var Score = app.registry.createModel('Score', { base: 'PersistedModel' });
```
bajtos: I believe `PersistedModel` is used as a base model by default, why is it necessary to specify it explicitly?

test/integration.test.js
```diff
@@ -40,7 +40,9 @@ describe('loopback application', function() {
       loopback.ACL.attachTo(db);
       loopback.User.hasMany(loopback.AccessToken, { as: 'accessTokens' });
 
-      var Streamer = app.model('Streamer', { dataSource: 'db' });
+      var Streamer = app.registry.createModel('Streamer');
+      // attach model to app and datasource
```
bajtos: I think this comment is not needed, can we remove it? I mean we are following this pattern (createModel + model) everywhere else, why to comment in this place only?

test/registries.test.js
```diff
@@ -26,15 +26,18 @@ describe('Registry', function() {
       var dsFoo = appFoo.dataSource('dsFoo', { connector: 'memory' });
       var dsBar = appFoo.dataSource('dsBar', { connector: 'memory' });
 
-      var FooModel = appFoo.model(modelName, settings);
-      var FooSubModel = appFoo.model(subModelName, settings);
-      var BarModel = appBar.model(modelName, settings);
-      var BarSubModel = appBar.model(subModelName, settings);
+      var FooModel = appFoo.registry.createModel(modelName, {}, settings);
+      // attach model to app and ds
```
bajtos: ditto - https://github.com/strongloop/loopback/pull/2401/files#r66070652

test/relations.integration.js
```diff
@@ -31,24 +31,14 @@ describe('relations - integration', function() {
...
-        { properties: { name: 'string', imageableId: 'number', imageableType: 'string' },
-          dataSource: 'db',
-        }
-      );
+      var Team = app.registry.createModel('Team', { name: 'string' });
+      var Reader = app.registry.createModel('Reader', { name: 'string' });
+      var Picture = app.registry.createModel('Picture',
+        { name: 'string', imageableId: 'number', imageableType: 'string' });
+
+      app.model(Team, { dataSource: 'db' }); //attach model to app & datasource
```
bajtos: ditto - https://github.com/strongloop/loopback/pull/2401/files#r66070652

test/relations.integration.js
```diff
@@ -640,15 +630,19 @@ describe('relations - integration', function() {
...
-
+      var product = app.registry.createModel(
+         'product',
+         { id: 'string', name: 'string' }
       );
-      var category = app.model(
-        'category',
-        { properties: { id: 'string', name: 'string' }, dataSource: 'db' }
+      var category = app.registry.createModel(
+
```
bajtos: misplaced empty line, please remove

test/relations.integration.js
```diff
@@ -640,15 +630,19 @@ describe('relations - integration', function() {
...
+       'category',
+         { id: 'string', name: 'string' }
```
bajtos: the indentation looks a bit off to me, PTAL

test/relations.integration.js
```diff
@@ -877,17 +872,18 @@ describe('relations - integration', function() {
...
     before(function defineProductAndCategoryModels() {
-      var todoList = app.model(
+      var todoList = app.registry.createModel(
         'todoList',
-        { properties: { name: 'string' },
-          dataSource: 'db',
-          plural: 'todo-lists',
-        }
+        { name: 'string' },
+        {  plural: 'todo-lists' }
```
bajtos: extra space after `{`, please remove (there should be single space only)

test/remoting-coercion.test.js
```diff
@@ -12,7 +12,12 @@ describe('remoting coercion', function() {
     var app = loopback();
     app.use(loopback.rest());
 
-    var TestModel = app.model('TestModel', { base: 'Model', dataSource: null, public: true });
+    var TestModel = app.registry.createModel('TestModel',
+      {},
+      { base: 'Model', public: true }
```
bajtos: `public` should be passed to `app.model`, not `createModel`.

Here is the rule to make this easier to remember: `createModel()` takes settings from `common/models/my-model.json`, `app.model` takes settings from `server/model-config.json`.

```js
var TestModel = app.registry.createModel('TestModel',
  {},
  { base: 'Model' });
app.model(TestModel, { public: true });
```

IIRC, models are public by default, which would explains why the test works this way too.

bajtos:
@gunjpan reviewed, PTAL. Also make sure that `npm test` passes on your local machine.

lib/application.js
gunjpan: @bajtos: That's what I struggled with to come up with a decent message. Thanks for the update.

test/change-stream.test.js
gunjpan: yeah, `PersistedModel` is set as base bydefault. I added it at the time for readibility, but no need, agreed.

test/app.test.js
gunjpan: @bajtos: Isn't it the `best practice` to have `done()` to indicate end of the test?

test/relations.integration.js
gunjpan: @bajtos : Do we have any rule setup in eslint for this spacing issue?

lib/application.js
```diff
@@ -112,14 +112,15 @@ app.model = function(Model, config) {
   var isPublic = true;
   var registry = this.registry;
 
+  if (typeof Model === 'string') {
+    var msg = 'app.model(modelName, settings) is no longer supported. ' +
+      'Use app.registry.createModel(modelName, definition) and ' +
+      'app.model(modelCtor, config) instead.';
```
bajtos: `ModelCtor` please, we usually use Pascal Case for constructors (class names).

test/relations.integration.js
bajtos: I would hope so, I am surprised it did not catch it here.

test/app.test.js
bajtos: >  Isn't it the best practice to have done() to indicate end of the test?

I am not aware of such practice, but that does not mean it doesn't exist.

Note that Mocha supports promises too, in which case there is no `done` either:

```js
it('works', function() {
  return User.find().then(function(list) {
    expect(list).to.have.length(1);
  });
})
```

test/acl.test.js
bajtos: > Please move the model creation from `before` hook to the test. Unless I am missing some reasons why it won't work?

@gunjpan PTAL ^^^

bajtos: (I think this is different situation than in https://github.com/strongloop/loopback/pull/2401/files?short_path=e133443#r66069812, therefore it's ok to use different approach.)

bajtos:
@gunjpan two more comments, otherwise the patch LGTM.
 - https://github.com/strongloop/loopback/pull/2401#discussion_r66255868
 - https://github.com/strongloop/loopback/pull/2401#discussion_r66069606

No further review is necessary, please squash the commits before landing.

gunjpan:
@slnode test please

gunjpan:
@slnode test please

gunjpan:
@slnode test please

test/acl.test.js
```diff
@@ -393,20 +393,18 @@ describe('security ACLs', function() {
 });
 
 describe('access check', function() {
-  var app;
-  before(function() {
-    app = loopback();
-    app.set('remoting', { errorHandler: { debug: true, log: false }});
-    app.use(loopback.rest());
-    app.enableAuth();
-    app.dataSource('test', { connector: 'memory' });
```
bajtos: I think we misunderstood each other. Keep the `app` creation in the `before` hook, because that part is typically shared by all test. Keep the line `var MyTestModel = ...` in the test, because the model is specific to the test case.

gunjpan: @bajtos : In the previous commit, IIRC, _(can't quote as rebased)_, I had kept the `before()` part, but I was advised to remove it as there is only one case in that `describe()` block.

Do you still recommend removing it?

bajtos: I see. Let's not bike shed and keep the code as it is now.

bajtos:
@gunjpan LGTM

bajtos:
@gunjpan please rebase on top of the current master and check CI results before landing.

gunjpan:
@slnode test please

bajtos:
@gunjpan I have rebased your patch on top of the latest master. All CI builds are passing with the exception of `node=0.10,os=windows` which fails because of an unrelated problem (phantomjs cannot be started due to resource leaks).

I am going to land this patch.

Issue #2338, opened 2016-05-12
Remove loopback#errorHandler


bajtos:
AFAICT, the only purpose of `loopback#errorHandler` is to provide option `includeStack` that allows user to remove stack traces from error responses in production. However, that's not enough, as other error properties may contain sensitive information too.

We should remove the handler and tell users to use strong-error-handler instead.

To simplify upgrade from 2.x to 3.0, we should preserve the middleware factory but throw a descriptive error instead of returning a handler function.

```js
// server/middleware/error-handler.js
module.exports = function(options) {
  throw new Error('...');
};
```

bajtos:
Here is the implementation of `loopback#errorHandler`:

https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/server/middleware/error-handler.js

I am also cross-posting my comments from a private chat:

If we change `loopback#errorHandler` to use strong-error-handler, then we will couple the version of loopback with the version of strong-error-handler. Then when we make a breaking change in strong-error-handler, we won't be able to upgrade it in loopback until next loopback major version.

OTOH, when people use strong-error-handler directly as a dependency of their app, then they are in control of what version they like to use and when to upgrade to a newer major version. This is a lesson that many popular frameworks learned over time. Express stopped bundling middleware since v4, grunt came to the same conclusion in v1.

bajtos:
We also need to modify the templates in loopback-workspace, but that will be handled by https://github.com/strongloop-internal/scrum-loopback/issues/896.

loay:
connect to: https://github.com/strongloop/loopback/pull/2411

fizzl:
What do I need to do to use strong-error-handler?

I'm trying to upgrade an API project from 2.x to 3.x. I am getting the following error:

"Unhandled rejection Error: loopback.errorHandler is no longer available. Please use the module "strong-error-handler" instead."

bajtos:
@Fizzl

> What do I need to do to use strong-error-handler?
>  I'm trying to upgrade an API project from 2.x to 3.x. I am getting the following error:
>  "Unhandled rejection Error: loopback.errorHandler is no longer available. Please use the module "strong-error-handler" instead."

See https://loopback.io/doc/en/lb3/Migrating-to-3.0.html#update-loopback-version and https://loopback.io/doc/en/lb3/Migrating-to-3.0.html#update-use-of-rest-error-handler

fizzl:
[...]

PR #2411, opened 2016-06-08
[SEMVER-MAJOR] Remove `loopback#errorHandler` middleware

loay:
connect to https://github.com/strongloop/loopback/issues/2338

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -127,3 +127,29 @@ removed properties together with the middleware module name to use instead:
 We have also removed `loopback.mime`, which was always set to `undefined`.
 
 See [loopback#2349](https://github.com/strongloop/loopback/pull/2394).
+
+## Remove `loopback#errorhandler`
+
+We have removed `loopback#errorhandler` middleware, users should use `strong-error-handler` directly.
+
+```js
+// server/middleware.json
+{
+  // ...
+  "final:after": {
+    "strong-error-handler": {}
+  }
+}
+
+// server/middleware.development.json
+{
+  "final:after": {
+    "strong-error-handler": {
+      "params": {
+        "debug": true
+      }
+  }
+}
+```
+
+See also strong-error-handler's [options](https://github.com/strongloop/strong-error-handler#options) and the [related code change](https://github.com/strongloop/loopback/pull/2411).
--- a/server/middleware/error-handler.js
+++ b/server/middleware/error-handler.js
@@ -3,19 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-var expressErrorHandler = require('errorhandler');
-expressErrorHandler.title = 'Loopback';
-
-module.exports = errorHandler;
-
-function errorHandler(options) {
-  if (!options || options.includeStack !== false) {
-    return expressErrorHandler(options);
-  } else {
-    var baseHandler = expressErrorHandler(options);
-    return function errorHandler(err, req, res, next) {
-      delete err.stack;
-      return baseHandler(err, req, res, next);
-    };
-  }
-}
+module.exports = function(options) {
+  throw new Error('loopback.errorHandler is no longer available.' +
+  ' Please use the module "strong-error-handler" instead.');
+};
--- a/test/error-handler.test.js
+++ b/test/error-handler.test.js
@@ -7,61 +7,19 @@ var loopback = require('../');
 var app;
 var assert = require('assert');
 var request = require('supertest');
+var expect = require('chai').expect;
 
 describe('loopback.errorHandler(options)', function() {
-  it('should return default middleware when options object is not present', function(done) {
-    //arrange
-    var app = loopback();
-    app.use(loopback.urlNotFound());
-    app.use(loopback.errorHandler({ log: false }));
-
-    //act/assert
-    request(app)
-      .get('/url-does-not-exist')
-      .end(function(err, res) {
-        assert.ok(res.error.text.match(/<ul id="stacktrace"><li> &nbsp; &nbsp;at raiseUrlNotFoundError/));
-
-        done();
-      });
-  });
-
-  it('should delete stack when options.includeStack is false', function(done) {
-    //arrange
-    var app = loopback();
-    app.use(loopback.urlNotFound());
-    app.use(loopback.errorHandler({ includeStack: false, log: false }));
-
-    //act/assert
-    request(app)
-      .get('/url-does-not-exist')
-      .end(function(err, res) {
-        assert.ok(res.error.text.match(/<ul id="stacktrace"><\/ul>/));
-
-        done();
-      });
-  });
-
-  it('should pass options on to error handler module', function(done) {
-    //arrange
-    var app = loopback();
-    app.use(loopback.urlNotFound());
-
-    var errorLogged;
-    app.use(loopback.errorHandler({
-      includeStack: false,
-      log: function customLogger(err, str, req) {
-        errorLogged = err;
-      },
-    }));
-
-    //act
-    request(app).get('/url-does-not-exist').end(function(err) {
-      if (err) return done(err);
-      //assert
-      expect(errorLogged)
-        .to.have.property('message', 'Cannot GET /url-does-not-exist');
-
+  it('should throw a descriptive error', function(done) {
+    try {
+  //arrange
+      var app = loopback();
+      app.use(loopback.urlNotFound());
+      app.use(loopback.errorHandler({ log: false }));
+    } catch (e) {
+      expect(function() { loopback.errorHandler(); })
+      .to.throw(/no longer available.*strong-error-handler/);
       done();
-    });
+    }
   });
 });
--- a/test/fixtures/access-control/server/server.js
+++ b/test/fixtures/access-control/server/server.js
@@ -6,7 +6,7 @@
 var loopback = require('../../../..');
 var boot = require('loopback-boot');
 var app = module.exports = loopback();
-
+var errorHandler = require('strong-error-handler');
 boot(app, __dirname);
 
 var apiPath = '/api';
@@ -14,5 +14,5 @@ app.use(loopback.token({ model: app.models.accessToken }));
 app.use(apiPath, loopback.rest());
 
 app.use(loopback.urlNotFound());
-app.use(loopback.errorHandler());
+app.use(errorHandler());
 app.enableAuth();
--- a/test/fixtures/simple-integration-app/server/server.js
+++ b/test/fixtures/simple-integration-app/server/server.js
@@ -6,9 +6,10 @@
 var loopback = require('../../../../index');
 var boot = require('loopback-boot');
 var app = module.exports = loopback();
+var errorHandler = require('strong-error-handler');
 
 boot(app, __dirname);
 var apiPath = '/api';
 app.use(apiPath, loopback.rest());
 app.use(loopback.urlNotFound());
-app.use(loopback.errorHandler());
+app.use(errorHandler());
--- a/test/fixtures/user-integration-app/server/server.js
+++ b/test/fixtures/user-integration-app/server/server.js
@@ -6,10 +6,11 @@
 var loopback = require('../../../../index');
 var boot = require('loopback-boot');
 var app = module.exports = loopback();
+var errorHandler = require('strong-error-handler');
 app.enableAuth();
 boot(app, __dirname);
 app.use(loopback.token({ model: app.models.AccessToken }));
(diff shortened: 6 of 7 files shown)


bajtos:
@loay Could you please link this pull request with the GH issue where we discussed this change? I think it will be better to completely remove `loopback#errorHandler` middleware, what are the arguments for keeping it around?

bajtos:
@loay please add an entry into 3.0-RELEASE-NOTES.md, don't forget to include a link to this pull request - see other entries for inspiration.

loay:
@slnode test please

loay:
@slnode test please

loay:
In `test/relations.integration.js`, this line was failing the tests : ```expect(res.body.error.statusCode).to.equal(500);```, as it set to `undefined` so I changed it to 
```expect(res.body.error.status).to.equal(500);``` and it did run fine locally, as `res.body.error.status` was giving the expected value of 500. However it is still failing the ci testings for the same reason of being `undefined`. I commented out the line and it finally passed. 

@bajtos PTAL

File: 3.0-RELEASE-NOTES.md [hunk 1]
```diff
@@ -127,3 +127,16 @@ removed properties together with the middleware module name to use instead:
 We have also removed `loopback.mime`, which was always set to `undefined`.
 
 See [loopback#2349](https://github.com/strongloop/loopback/pull/2394).
+
+## remove `loopback.errorhandler`
```
bajtos: ## Remove `loopback#errorhandler`

We have removed `loopback#errorhandler` middleware, users should use `strong-error-handler` directly.

```js
// server/middleware.json
{
  // ...
  "final:after": {
    "strong-error-handler": {}
  }
}

// server/middleware.development.json
{
  "final:after": {
    "strong-error-handler": {
      "params": {
        "debug": true
      }
  }
}
```

See also strong-error-handler's [options](https://github.com/strongloop/strong-error-handler#options) and the [related code change](https://github.com/strongloop/loopback/pull/2411)

server/middleware/error-handler.js
```diff
@@ -3,19 +3,7 @@
...
-  } else {
-    var baseHandler = expressErrorHandler(options);
-    return function errorHandler(err, req, res, next) {
-      delete err.stack;
-      return baseHandler(err, req, res, next);
-    };
-  }
-}
+module.exports = function(options) {
+  throw new Error ('expressHandler is no longer coupled with loopback.\
```
bajtos:
```
loopback.errorHandler is no longer available. Please use the module "strong-error-handler" instead.
```

test/error-handler.test.js
```diff
@@ -9,59 +9,43 @@ var assert = require('assert');
...
-        assert.ok(res.error.text.match(/<ul id="stacktrace"><li> &nbsp; &nbsp;at raiseUrlNotFoundError/));
-
-        done();
-      });
+  it('should pass strong-error-handler message to migrate', function(done) {
+    try {
+  //arrange
+      var app = loopback();
+      app.use(loopback.urlNotFound());
+      app.use(loopback.errorHandler({ log: false }));
```
bajtos:
```js
expect(function() { loopback.errorHandler(); })
  .to.throw(/no longer available.*strong-error-handler/);
```

test/error-handler.test.js
```diff
@@ -9,59 +9,43 @@ var assert = require('assert');
...
-
-    //act/assert
-    request(app)
-      .get('/url-does-not-exist')
-      .end(function(err, res) {
-        assert.ok(res.error.text.match(/<ul id="stacktrace"><li> &nbsp; &nbsp;at raiseUrlNotFoundError/));
-
-        done();
-      });
+  it('should pass strong-error-handler message to migrate', function(done) {
```
bajtos: `it('should throw a descriptive error')`

test/error-handler.test.js
```diff
@@ -9,59 +9,43 @@ var assert = require('assert');
...
-
-    //act/assert
-    request(app)
-      .get('/url-does-not-exist')
-      .end(function(err, res) {
-        assert.ok(res.error.text.match(/<ul id="stacktrace"><\/ul>/));
-
-        done();
-      });
+  it('should pass strong-error-handler message to migrate', function(done) {
```
bajtos: Please remove, this is no longer relevant.

test/error-handler.test.js
```diff
@@ -9,59 +9,43 @@ var assert = require('assert');
...
-    }));
-
-    //act
-    request(app).get('/url-does-not-exist').end(function(err) {
-      if (err) return done(err);
-      //assert
-      expect(errorLogged)
-        .to.have.property('message', 'Cannot GET /url-does-not-exist');
-
+  it('should pass strong-error-handler message to migrate', function(done) {
```
bajtos: ditto

test/fixtures/access-control/server/server.js
```diff
@@ -14,5 +14,4 @@ app.use(loopback.token({ model: app.models.accessToken }));
 app.use(apiPath, loopback.rest());
 
 app.use(loopback.urlNotFound());
-app.use(loopback.errorHandler());
```
bajtos: add `strong-error-handler` instead

test/fixtures/user-integration-app/server/server.js
```diff
@@ -12,4 +12,3 @@ app.use(loopback.token({ model: app.models.AccessToken }));
 var apiPath = '/api';
 app.use(apiPath, loopback.rest());
 app.use(loopback.urlNotFound());
-app.use(loopback.errorHandler());
```
bajtos: ditto

test/relations.integration.js
```diff
@@ -1652,7 +1652,7 @@ describe('relations - integration', function() {
           expect(res.body).to.be.an('object');
           expect(res.body.error).to.be.an('object');
           expect(res.body.error.name).to.equal('Error');
-          expect(res.body.error.statusCode).to.equal(500);
+        //  expect(res.body.error.status).to.equal(500);
```
bajtos: Please check with @davidcheung and find out why this test is failing. I think this may be because your fixture app does not come with strong-error-handler configured?

test/fixtures/simple-integration-app/server/server.js
```diff
@@ -11,4 +11,3 @@ boot(app, __dirname);
 var apiPath = '/api';
 app.use(apiPath, loopback.rest());
 app.use(loopback.urlNotFound());
-app.use(loopback.errorHandler());
```
bajtos: ditto

3.0-RELEASE-NOTES.md
bajtos: ^^^ click on the "edit" icon of the comment above, copy the markdown source and use it to replace the content of the new section you are adding.

test/fixtures/access-control/server/server.js
bajtos: > add `strong-error-handler` instead

`strong-error-handler` should become a `devDependency` of loopback

bajtos:
@loay reviewed, PTAL

test/relations.integration.js
davidcheung: it might be because the loopback isnt linked to `strong-remoting:master` and the newest change(breaking) is not on npm yet, in the meanwhile you could probably point your registry to **strongloop CI**

loay: ```res.body.error.status``` works locally. 
```res.statusCode``` works locally too. 
Both do not work on jenkins. 
I will comment out the line till the most recent version of `strong-remoting` is published.

File: test/error-handler.test.js (hunk 3 above)
Comment (loay) [inline test/error-handler.test.js]: I hope you meant to use that inside the `catch`, since the test case fails without try/catch. 
It is inside the catch and working fine.

**Comment by loay** [comment]
@bajtos PTAL. Thanks

File: test/relations.integration.js (hunk 9 above)
Comment (loay) [inline test/relations.integration.js]: `res.body.error.status` . it is working now but failing only on travis which is using npm. That should be fixed once npm is updated.

File: test/relations.integration.js [hunk 11]
```diff
@@ -1652,7 +1652,7 @@ describe('relations - integration', function() {
           expect(res.body).to.be.an('object');
           expect(res.body.error).to.be.an('object');
           expect(res.body.error.name).to.equal('Error');
-          expect(res.body.error.statusCode).to.equal(500);
+          expect(res.statusCode).to.equal(500);
```
bajtos: Could you please check with @davidcheung whether this change is needed. I am surprised that `body.error.statusCode` is not available.

test/error-handler.test.js
bajtos: No no, there is no need to use `try/catch`. This is how the test case should look like:

```js
it('...', function() {
  expect(function() { loopback.errorHandler(); })
    .to.throw(/no longer available.*strong-error-handler/);
});
```

server/middleware/error-handler.js
```diff
@@ -3,19 +3,7 @@
...
-  } else {
-    var baseHandler = expressErrorHandler(options);
-    return function errorHandler(err, req, res, next) {
-      delete err.stack;
-      return baseHandler(err, req, res, next);
-    };
-  }
-}
+module.exports = function(options) {
+  throw new Error ('loopback.errorHandler is no longer available.\
```
bajtos: Style: no space between `Error` and `(`. Also don't use multi-line strings, use string concatenation instead.

server/middleware/error-handler.js
```diff
@@ -3,19 +3,7 @@
...
+   Please use the module "strong-error-handler" instead.');
```
bajtos: I think the indentation is off here, please check.

bajtos:
The code change LGTM, please make sure Travis CI builds and Jenkins builds (possibly excluding unstable dependants) are passing before landing.

bajtos:
@loay also squash the commits before landing :)

server/middleware/error-handler.js
```diff
@@ -3,19 +3,7 @@
...
-  } else {
-    var baseHandler = expressErrorHandler(options);
-    return function errorHandler(err, req, res, next) {
-      delete err.stack;
-      return baseHandler(err, req, res, next);
-    };
-  }
-}
+module.exports = function(options) {
+  throw new Error ('loopback.errorHandler is no longer available.' +
```
bajtos: We don't put space between function name and argument list, i.e. `Error(` 

Could you please add a rule to our eslint config to catch this?

loay:
@slnode test please

test/error-handler.test.js
bajtos: I didn't notice this comment was not addressed. I cleaned up the test in https://github.com/strongloop/loopback/commit/42c83f69a090664bb025d89f8dd5f5a7a3ffa34e

rickyngk:
I'n using strongErrorHandler with loopback 3.0.0-alpha.1. That's really a pain. It annoyed me with sort of message e.g: "Unhandled error for request GET /MyModel/action". How can I turn it off? I also disabled log & debug in both `middleware.json` and `middleware.development.json`, but no help

```JSON
"final:after": {
    "strong-error-handler": {
      "params": {
        "debug": true,
        "log": true
      }
    }
  }
```

Btw, I use loggy to store error message. For loopback 2.x, I send error log in config.local.js 

```Javascript
module.exports = {
  restApiRoot: '/api' + (version > 0 ? '/v' + version : ''),
  host: process.env.HOST || 'localhost',
  port: process.env.PORT || 3000,
  remoting: {
    errorHandler: {
      handler: function(err, req, res, next) {
        log.l('api-error', {
          method: req.method,
          originalUrl: req.originalUrl,
          statusCode: res.statusCode,
          stack: err.stack,
        });
        next(err);
      },
    },
  },
};
```

But err.stack is not available in 3.x. I have to change it to `req.remotingContext.error`

How can I do this officially?

ttothAISS:
Hi @rickyngk !

I have encountered the same issue.

The problem is the RestAdapter.errorHandler in the rest-adapter.js calling the new strongErrorHandler with ITS OWN option object, not the one the documentation suggest in the server/middleware.json.

To overcome and apply your desired params use the following in the server/config.json:

> "remoting": {
> ...
>     "errorHandler": {
>       "debug": false,
>       "log": false
> ...
>   }

@bajtos this should be mentioned in the migrate doc, because the full migration is not possible right now if you want to use options when using the strong error handler for the REST adapter.

T.

davidcheung:
@ttothAISS thanks for pointing it out, the migration guide for REST adapter error-handler should be [this](https://github.com/strongloop/strong-remoting/blob/ebf4a9a8d8fd1a787fb8b05fa36d68be8e73ce1b/3.0-RELEASE-NOTES.md#new-error-handler-for-rest-adapter) instead, will follow up and fix

Issue #2424, opened 2016-06-10
Confirming a user does not remove their verificationToken on postgres


JonForest:
Calling User.confirm does not remove the verificationToken

The line here https://github.com/strongloop/loopback/blob/master/common/models/user.js#L488
`user.verificationToken = undefined;`
Results in the verificationToken field not being part of the postgres UPDATE query.

Simply modifying to 
`user.verificationToken = null;`
fixes it

Will try to raise a PR this weekend.

davidcheung:
Thanks, this might be issue with generic connector too
https://github.com/strongloop/loopback-datasource-juggler/blob/058d9d46c09f249dd7aa14a09e3097aae8a7a21b/lib/dao.js#L546

https://github.com/strongloop/loopback-connector-postgresql/blob/0facead0dbb46c9e72c95800d3e656523ee3576b/test/postgresql.filterUndefined.test.js

PR #2425, opened 2016-06-10
Change the set of the verificationToken to null on User.confirm

JonForest:
This addresses issue https://github.com/strongloop/loopback/issues/2424

Simple change to setting the verificationToken to `null`, instead of `undefined`.  `undefined` causes the field to not be updated via the Postgres adapter.

Note: the tests have a single failure for me at the moment, but they're also failing on master branch.


slnode:

PR #2426, opened 2016-06-10
Change set verificationToken on User.confirm

JonForest:
This addresses issue #2424

Simple change to setting the verificationToken to null, instead of undefined. undefined causes the field to not be updated via the Postgres adapter.

Note: the tests have a single failure for me at the moment, but they're also failing on master branch.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -485,7 +485,7 @@ module.exports = function(User) {
         fn(err);
       } else {
         if (user && user.verificationToken === token) {
-          user.verificationToken = undefined;
+          user.verificationToken = null;
           user.emailVerified = true;
           user.save(function(err) {
             if (err) {


slnode:

davidcheung:
@slnode ok to test

davidcheung:
perhaps we should add a test case to prevent regression, not sure where we can add the test case tho, 
@Amir-61 any clues?  this is a user model behavior that is ignored before connectors ignore `undefined` upon `instance.save()`

Amir-61:
>@Amir-61 any clues? this is a user model behavior that is ignored before connectors ignore undefined upon instance.save()

I'm not sure if I understand you correctly. Is this the same question you are asking in slack or it is different?

>@amir: do u know if only SQL connectors ignore values that are `undefined` (upon ​*update*​)? or its both noSQL and SQL
something like `user.update({name:undefined, lastname:'cheung'});`
then it would ignore `name`  

It is an expected behaviour see: https://github.com/strongloop/loopback-datasource-juggler/blob/058d9d46c09f249dd7aa14a09e3097aae8a7a21b/test/manipulation.test.js#L515-L526
Also when you see they are removing undefined in `dao.js` it means it applies to all connectors not just NoSQL DBs or SQL DBs...

davidcheung:
Thanks @Amir-61 was trying to make sure the issue happens in all connectors, so the unit test could be added in loopback itself, if it was connector specific(**which its not**) maybe we would have needed to test on specific connector. 

@JonForest can you add a unit test to prevent regression, perhaps in this file
https://github.com/strongloop/loopback/blob/42c83f69a090664bb025d89f8dd5f5a7a3ffa34e/test/user.test.js#L1147

davidcheung:
related https://github.com/strongloop/loopback/pull/2440

JonForest:
Will take a look at that unit test this weekend.

davidcheung:
@JonForest #2440 is basically a duplicate task, I would actually like to propose to close this and follow up there

JonForest:
Works for me!  As long as it gets fixed, all good.

Issue #1676, opened 2015-09-15
Deprecate getCurrentContext()


ritch:
Well its been fun. And by fun I mean: not so fun. We've tried to create a simple method for getting at contextual information no matter where your code is executing. We did this by depending on a sophisticated module called [continuation-local-storage ](https://www.npmjs.com/package/continuation-local-storage).

> Continuation-local storage works like thread-local storage in threaded programming, but is based on chains of Node-style callbacks instead of threads.

Someday there will be a more reliable mechanism than the userland cls module, and we can re-add first class support for `getCurrentContext()`. But until then I am proposing we deprecate this method with the standard warning message.

What does this mean for users of `getCurrentContext()`? You will get an annoying message (configurable) every time you use the method. We will still support it, and we won't break semver. But we strongly encourage you to refactor. 

**For an immediate work around, take a look at https://github.com/strongloop/loopback/issues/1495. A more sustainable workaround will be added as a feature to the framework.**

/cc @bajtos @raymondfeng @superkhau @piscisaureus

fabien:
I'm concerned about this, as I have been using getCurrentContext extensively. And without any issues thusfar. I'm on mobile right now, so I don't have access to the whole discussion(s) - IIRC mainly regarding unreliable availability on different platforms - but what's the reasoning here?

Can we have a flag to silence the deprecation warning? ;-)

csvan:
I abandoned getCurrentContext pretty quickly since it led to unpredictable behavior. Notably, retrieving the Access Token using something like `getCurrentContext().get('accessToken)` seemed to work at random, if at all, so we worked around it by getting the accessToken from the request instead. Apart from that, we have not needed it anywhere.

bajtos:
I was thinking about the problems of getCurrentContext we are experiencing. It seems to me that most of the issues are caused by the fact that any connector using connection pool must be CLS aware and include few lines of code to switch to the right CLS context. I am wondering we can add this small piece into loopback-datasource-juggler, so that we don't depend on connector implementation anymore?

An example to illustrate my point:

```js
DAO.create = function(data, options, cb) {
  // setup instance, call hooks, etc.
  clsCtx = // save current CLS context;
  connector.create(/*args*/, function(err, result) {
    // restore CLS context from clsCtx
    // continue as before
  });
};
```

If that's not a viable approach, then I support @fabien's request for a feature flag. This flag can have three states:

 - `undefined` - print a warning (default)
[...]

PR #2439, opened 2016-06-16
Allow disabling node-continuation-local-storage initialization

seriousben:
node-continuation-local-storage and async-listener both install their hooks as soon as they are required.

This PR provides a way to disable those hooks.

It is very useful when not using cls/`getCurrentContext`in loopback removing some overhead.


seriousben:
Relates to #1676

Issue #414, opened 2014-07-28
User verification token not cleared after confirmation


johnsoftek:
User.confirm attempts to clear the verificationToken after confirmation but it does not work.

        user.verificationToken = undefined;
        user.emailVerified = true;
        user.save(function (err) {

It seems that the save function does not update fields with undefined value. Null works.

        user.verificationToken = null;

superkhau:
Please create a test project on Github to reproduce the issue as per https://github.com/strongloop/loopback/wiki/Issues#bug-report

Amir-61:
@johnsoftek Would you like to send a patch for this fix?

Thanks.

richardpringle:
@Amir-61 Did you reproduce this issue? I see that you are the one that put the bug label on the issue. Next time could you please add some details in a comment just in case someone else is going to make the fix.

Amir-61:
>@Amir-61 Did you reproduce this issue? 

I believe so... It was long time ago... Someone may have already solved it by now...

PR #2440, opened 2016-06-16
Fix verificationToken bug

loay:
connect to https://github.com/strongloop/loopback/issues/414

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1418,6 +1418,21 @@ describe('User', function() {
         }, done);
       });
 
+      it('sets verificationToken to null after confirmation', function(done) {
+        testConfirm(function(result, done) {
+          User.confirm(result.uid, result.token, false, function(err) {
+            if (err) return done(err);
+
+            // Verify by loading user data stored in the datasource
+            User.findById(result.uid, function(err, user) {
+              if (err) return done(err);
+              expect(user).to.have.property('verificationToken', null);
+              done();
+            });
+          });
+        }, done);
+      });
+
       it('Should report 302 when redirect url is set', function(done) {
         testConfirm(function(result, done) {
           request(app)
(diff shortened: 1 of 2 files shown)


loay:
@bajtos PTAL

davidcheung:
would also fix #2424

test/user.test.js
```diff
@@ -1418,6 +1418,21 @@ describe('User', function() {
...
+            User.findById(result.uid, function(err, user) {
+              if (err) return done(err);
+              expect(user).to.have.property('verificationToken', null);
+              done();
+            });
+          });
+        }, done);
+      });
+
       it('Should report 302 when redirect url is set', function(done) {
```
bajtos: You are calling done twice, see L1440. More importantly, this test passes even when common/models/user.json is not fixed.

_Never trust a test you haven't seen fail._

When I added console logs around `done()` calls, only the second done on L1440 was called. The event handler for `confirmed` even is never triggered. I think that's because you are adding the listener long time after `confirm()` method was invoked.

I am proposing to simplify this test and call `User.confirm` directly - the behaviour you are fixing is independent of the way how the method is invoked (from code or via REST API).

As I was playing with the test code to make it work correctly, I figured out I can as well commit my changes - see fc90538. This fixup commit should be rebased away before merging the PR.

bajtos: The line comment is misplaced because I submitted after pushing the fixup commit. Sorry for the confusion.

bajtos:
@loay I added a commit fixing the test, PTAL. If my changes LGTY and CI builds passes, then you can go ahead, squash commits and land the fix.

Please back-port this fix to 2.x, don't forget to include a reference to this PR in the commit message in 2.x.

test/user.test.js
loay: That's weird. I tested when it was set to `undefined` and it failed .. It must have been something else. 
Anyways, your changes LGTM. I tested with `undefined` and it failed then it passed with `null`. 
Thanks.

Issue #2445, opened 2016-06-17
Issue: 


sid2vicious:
Hi,

When I run an nsp scan on my code using loopback 2.29.0, I get the following vulnurablity error. Can this be patched in 2.29.0
```
Message:
    (+) 1 vulnerabilities found
┌───────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                                                                                      │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Name          │ negotiator                                                                                                                │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Installed     │ 0.5.3                                                                                                                     │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Vulnerable    │ <= 0.6.0                                                                                                                  │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Patched       │ >= 0.6.1                                                                                                                  │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Path          │ ##masked##@0.0.13 > loopback@2.29.0 > express@4.13.4 > accepts@1.2.13 > negotiator@0.5.3                             │
├───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/106                                                                                    │
└───────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘


```

loay:
Hi @sid2vicious 
Thanks for reporting the issue. Looking into it.

loay:
Fixed and updated. Closing.

PR #2446, opened 2016-06-17
update express version

loay:
connect to https://github.com/strongloop/loopback/issues/2445

Update express version which will update negotiator version to 0.6.1


bajtos:
LGTM :shipit:

PR #2471, opened 2016-06-26
adding additional check of string for case insentive emails

geekguy:
Issue:

This line doesn't work when we have `caseSensitiveEmail` set to `false` and we make queries like,

`User.findOne({where: {email: {inq: ["hello@test.com"]}}})`

Additional check for string is required.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -618,7 +618,8 @@ module.exports = function(User) {
 
     // Access token to normalize email credentials
     UserModel.observe('access', function normalizeEmailCase(ctx, next) {
-      if (!ctx.Model.settings.caseSensitiveEmail && ctx.query.where && ctx.query.where.email) {
+      if (!ctx.Model.settings.caseSensitiveEmail && ctx.query.where &&
+          ctx.query.where.email && typeof(ctx.query.where.email) === 'string') {
         ctx.query.where.email = ctx.query.where.email.toLowerCase();
       }
       next();


slnode:

slnode:

slnode:

slnode:

slnode:

davidcheung:
@geekguy thanks for your contribution! would you be able to add a test case to prevent regression on this fix?

slnode:

slnode:

slnode:

slnode:

bajtos:
@geekguy thank you for the pull request. As @davidcheung already mentioned, we need a unit test to verify your change.

BTW while the solution proposed here is a good short-term fix, for long term we should improve this code to correctly convert email values to lower case even for complex queries like the one shown above.

```js
User.findOne({where: {email: {inq: ["HELLO@test.com"]}}})
// should be converted to
{where: {email: {inq: ["hello@test.com"]}}}
```

geekguy:
@bajtos : Can you please point me to a file where I should add test cases ? 

I ll also check regarding converting complex queries to lower case.

davidcheung:
@geekguy perhaps like [this test case](https://github.com/strongloop/loopback/blob/ed76a3475a903dbce41aa600e1ff3a7b9501ea1c/test/user.test.js#L95-L105), but instead of using the returned object, it makes another call using the where-inq query, then assert the object returned?

davidcheung:
fixes #2522

bajtos:
> @geekguy  Can you please point me to a file where I should add test cases ?

Sure. You can start by looking at https://github.com/strongloop/loopback/pull/1804 which added the "access" hook and a test suite `describe('Access-hook for queries with email NOT case-sensitive')`, see
https://github.com/strongloop/loopback/blob/ed76a3475a903dbce41aa600e1ff3a7b9501ea1c/test/user.test.js#L437-L457

davidcheung:
taking over to rebase + add tests @geekguy on #2912 
cc/ @bajtos

PR #2489, opened 2016-07-05
Support 'alias' in mail transport config.

STRML:
Useful if you need to set up multiple transports of the same type.

For instance, we have to use a separate email service for some domains in China.


bajtos:
Landed, thank you for the contribution and sorry for taking so long to handle your pull request.

STRML:
:+1:

Issue #580, opened 2014-09-24
It's still possible create relation Objects of User when logged User was deleted.


bluestaralone:
When i'm using Explorer to test ACLs i see this case:
It's still possible create relation Objects of User when logged User was deleted.
Case:
- User has many Books (userId)
- Books belong to User (userId)

Step like below:
- Create User
- Login User
- Delete logged User (in case still logged - Token is set)
- Create Book (in case still logged - Token is set) - It's still can do this why ?

I don't know why when deleting User, you don't delete AccessToken so when i create Book it can use userId of Token to create Book but User of userId is not available. ?

bajtos:
@loay please back-port the fix to 2.x, the issue is not "done" until then.

loay:
- [ ] Fix security issue: https://github.com/strongloop/loopback/pull/2492
- [ ] backport: https://github.com/strongloop/loopback/pull/2540
-  [ ] address issue concerning test case.
- [ ] backport after fixing test cases.

PR #2492, opened 2016-07-07
fix security issue 580

loay:
connect to: https://github.com/strongloop/loopback/issues/580

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -301,6 +301,21 @@ module.exports = function(User) {
     return fn.promise;
   };
 
+  User.observe('before delete', function(ctx, next) {
+    var AccessToken = ctx.Model.relations.accessTokens.modelTo;
+    var pkName = ctx.Model.definition.idName() || 'id';
+    ctx.Model.find({ where: ctx.where, fields: [pkName] }, function(err, list) {
+      if (err) return next(err);
+
+      var ids = list.map(function(u) { return u[pkName]; });
+      ctx.where = {};
+      ctx.where[pkName] = { inq: ids };
+
+      AccessToken.destroyAll({ userId: { inq: ids }}, next);
+      console.log('Deleted token for users ', ids);
+    });
+  });
+
   /**
    * Compare the given `password` with the users hashed password.
    *
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -6,6 +6,7 @@
 require('./support');
 var loopback = require('../');
 var User, AccessToken;
+var async = require('async');
 
 describe('User', function() {
   var validCredentialsEmail = 'foo@bar.com';
@@ -212,6 +213,92 @@ describe('User', function() {
       assert(u1.password !== 'bar');
       var u2 = new User({ username: 'foo', password: u1.password });
       assert(u2.password === u1.password);
+    });
+
+    it('invalidates the user\'s accessToken when the user is deleted By id', function(done) {
+      var usersId;
+      async.series([
+        function(next) {
+          User.create({ email: 'b@c.com', password: 'bar' }, function(err, user) {
+            usersId = user.id;
+            next(err);
+          });
+        },
+        function(next) {
+          User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
+            if (err) return next(err);
+            assert(accessToken.userId);
+            next();
+          });
+        },
+        function(next) {
+          User.deleteById(usersId, function(err) {
+            next(err);
+          });
+        },
+        function(next) {
+          User.findById(usersId, function(err, userFound)  {
+            if (err) return next(err);
+            expect(userFound).to.equal(null);
+            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+              if (err) return next(err);
+              expect(tokens.length).to.equal(0);
+              next();
+            });
+          });
+        },
+      ], function(err) {
+        if (err) return done(err);
+        done();
+      });
+    });
+
+    it('invalidates the user\'s accessToken when the user is deleted all', function(done) {
+      var usersId, accessTokenId;
+      async.series([
+        function(next) {
+          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
+          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
+            usersId = user.id;
+            next(err);
+          });
+        },
+        function(next) {
+          User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
+            accessTokenId = accessToken.userId;
+            if (err) return next (err);
+            assert(accessTokenId);
+            next();
+          });
+        },
+        function(next) {
+          User.login({ email: 'd@c.com', password: 'bar' }, function(err, accessToken) {
+            accessTokenId = accessToken.userId;
+            if (err) return next (err);
+            assert(accessTokenId);
+            next();
+          });
+        },
+        function(next) {
+          User.deleteAll({ name: 'myname' }, function(err, user) {
+            next(err);
+          });
+        },
+        function(next) {
+          User.find({ where: { name: 'myname' }}, function(err, userFound)  {
+            if (err) return next (err);
+            expect(userFound.length).to.equal(0);
+            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+              if (err) return next(err);
+              expect(tokens.length).to.equal(0);
+              next();
+            });
+          });
+        },
+      ], function(err) {
+        if (err) return done(err);
+        done();
+      });
     });
 
     describe('custom password hash', function() {


test/user.test.js
```diff
@@ -214,6 +214,18 @@ describe('User', function() {
       assert(u2.password === u1.password);
     });
 
+    it('should invalidate accessToken once user is deleted', function(done) {
```
superkhau: `it('should invalidate the user\'s access token when the user is deleted', function(done) {`

test/user.test.js
```diff
@@ -214,6 +214,18 @@ describe('User', function() {
...
+      User.create({ email: 'b@c.com', password: 'bar' }, function(err, userInstance) {
```
superkhau: userInstance -> instance

test/user.test.js
```diff
@@ -214,6 +214,18 @@ describe('User', function() {
...
+        User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
+          assert(accessToken.id);
```
superkhau:
```
if (err) return done err;
assert(accessToken.id)
```

test/user.test.js
```diff
@@ -214,6 +214,18 @@ describe('User', function() {
...
+          User.deleteById(userInstance.id, function(err){
```
superkhau: space between ){

and if (err) return done(err);

superkhau:
Also, please run the linter to make sure your code follows conventions.

loay:
@bajtos PTAL

bajtos:
@loay there are many CI failures, PTAL.

common/models/user.js
```diff
@@ -301,6 +301,19 @@ module.exports = function(User) {
     return fn.promise;
   };
 
+  User.deleteById = function(tokenId, fn) {
```
bajtos: This is weird. You are overriding built-in `PersistedModel.deleteById` implementation provided by jugglers' DataAccessObject, and change the semantics so that it no longer deletes the User instance, but deletes an access token only. That does not look right to me.

I think the correct solution is to use a "before delete" hook and delete *all* access tokens of the user from the database.

test/user.test.js
```diff
@@ -214,6 +214,24 @@ describe('User', function() {
...
     });
 
+    it('invalidates the user\'s accessToken when the user is deleted', function(done) {
+      User.create({ email: 'b@c.com', password: 'bar' }, function(err, instance) {
+        User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
+          if (err) return done(err);
+          assert(accessToken.userId);
+          User.deleteById(instance.id, function(err) {
+            if (err) return done(err);
+            AccessToken.findById(accessToken.userId, function(err, notFound) {
```
bajtos: Please add a check that the user instance was deleted. I suspect such check would fail with your current implementation.

```js
User.findById(instance.id, function(err, user) {
  expect(user).to.equal(null);
  AccessToken.findById(...);
});
```

test/user.test.js
```diff
@@ -214,6 +214,24 @@ describe('User', function() {
       assert(u2.password === u1.password);
     });
 
+    it('invalidates the user\'s accessToken when the user is deleted', function(done) {
+      User.create({ email: 'b@c.com', password: 'bar' }, function(err, instance) {
```
bajtos: Please use a more descriptive name than `instance`, to make it clear it's an instance of User. E.g. `userInstance` or simply `user`.

test/user.test.js
bajtos: Also `AccessToken.findById(userId)` does not make sense to me - it says "find me an access token that has the same id as this user". You should find all tokens belonging to the user instead.

```js
AccessToken.find({ where: { userId: user.id } }, function(err, tokens) {
  // assert no tokens were found
});
```

test/user.test.js
bajtos: > userInstance -> instance

@superkhau in the context of this tests, I find the name `instance` confusing. What instance is it - User instance or AccessToken instance?

See https://github.com/strongloop/loopback/pull/2492#discussion_r70971983

test/user.test.js
superkhau: I'm fine with `user`, I agree instance can be misleading.

loay:
@bajtos PTAL. 
Also, please ignore all the `console.log`, as I will remove them once the code gets approved.

common/models/user.js
```diff
@@ -301,6 +301,25 @@ module.exports = function(User) {
     return fn.promise;
   };
 
+  User.observe('before delete', function(ctx, next) {
+    var AccessToken = ctx.Model.app.models.AccessToken;
```
bajtos: I believe it's possible to configure a subclass of `User` to use a subclass of `AccessToken` instead, in which case `app.models.AccessToken` may not exist.

A better solution is to use relation method `ctx.Model.accessTokens(true, cb)` instead (the first argument is called `condOrRefresh`, by setting it to `true` we bypass any cached results).

However, there is no need to find all related tokens and then delete them one by one. You should use `ctx.Model.accessTokens.destroyAll(cb)` instead.

```js
User.observe('before delete', function(ctx, next) {
  ctx.Model.accessTokens.destroyAll(next);
});
```

common/models/user.js
```diff
@@ -301,6 +301,25 @@ module.exports = function(User) {
...
+
+    AccessToken.find({
+      where: {
+        userId: ctx.where.id,
+      },
+    }, function(err, tokens) {
+      tokens.forEach(function(token) {
+        AccessToken.destroyById(token.id, function() {
```
bajtos: This code (`tokens.forEach` + `destroyById`) deletes tokens in background. It means that you would delete the user even if some of they tokens were not removed.

Also calling `destroyById` repeatedly is low performant. The best solution is to call `ctx.Model.accessTokens.destroyAll` as I commented above. For future reference, if you run into situation when that's not possible, then the second best option is to use [inq](https://docs.strongloop.com/display/LB/Where+filter#Wherefilter-inq) operator and delete multiple records at once, e.g. `AccessToken.deleteAll({ id: { inq: tokenIds } }, cb)`

test/user.test.js
```diff
@@ -214,6 +214,31 @@ describe('User', function() {
       assert(u2.password === u1.password);
     });
 
+    it('invalidates the user\'s accessToken when the user is deleted', function(done) {
+      User.create({ email: 'b@c.com', password: 'bar' }, function(err, user) {
```
bajtos: Unhandled `err`.

FWIW, I would personally rewrite this test to use promise API:

```js
it('...', function() {
  var userId;
  return User.create(...)
    .then(function(user) {
      userId = user.id;
      return User.login(...);
    })
    .then(function() {
      return User.deleteById(userrId);
    })
    // ...
});
```

If you prefer to stick to callbacks, then consider using `async.series` to avoid xmas tree in your code.

```js
it('...', function(done) {
  asyc.series([
    function(next) { User.create(...); },
    function(next) { User.login(...); },
    //...
  ], done);
});
```

common/models/user.js
bajtos: Oh, hang on, that won't work. `accessTokens` is a property of model instance, which is not available when running a bulk delete via `User.destroyAll`.

You need to get the AccessToken model via `ctx.Model.relations.accessTokens.modelTo`.

Now I think we should support the general use case, where
 - the User model can have primary key called differently than `id`, possibly using a primary key composed of several properties, 
 - `User.destroyAll` can be called with arbitrary filter, e.g. `where={isPreferred:false, country: 'CZ' }`

In such case, it will be very difficult to build an `AccessToken` query that would match all tokens belonging to (and only to) the users matched by `ctx.where`.

Therefore I am proposing to use a different approach, similar to what you did:
 - first, find all deleted users
 - then delete all access tokens of these users

There is a catch though: while we are deleting the access tokens, it may happen that more records matching the original `where` clause are added by other clients. As a result, we would delete these new users, but not their access token.

The solution is to rewrite `ctx.where` to use `inq` filter with the list of the ids of the users we found.

```js
var AccessToken = ctx.Model.relations.accessTokens.modelTo;
ctx.Model.find({ where: ctx.where }, function(err, list) {
  if (err) return next(err);

  var pkName = ctx.Model.definition.idName() || 'id';
  var ids = list.map(function(u) { return u[pkName]; });
  ctx.where = {};
  ctx.where[pkName] = { inq: ids };

  AccessToken.destroyAll({ userId: { inq: ids } }, next);
});
```

(The code may not work OOTB.)

Also please add a unit-test to verify that access tokens are deleted when users are removed via `User.deleteAll`.

bajtos: BTW the best long-term solution for this would be to implement support for [cascading delete](http://www.techonthenet.com/sql_server/foreign_keys/foreign_delete.php) in LoopBack relations. I am afraid that will require more work that we can afford right now. However, could you please open a feature request in juggler linked to this PR to keep track of this idea?

test/user.test.js
loay: I prefer the cb, as it looks organized in my opinion, as functions are defined and could be easily accessed for modification and testing.

common/models/user.js
loay: I made a minor modification to my code, in defining the `AccessToken`.  It doesn't look like it will delete the user, unless I missed a step or did something wrong. I could have tried it with destroyAll if destroybyId is  not a good practice, but I changed the code to cope with the strategy above, but when u have the time please look at the code below. 

```
  User.observe('before delete', function(ctx, next) {
    var AccessToken = ctx.Model.relations.accessTokens.modelTo;

    AccessToken.find({
      where: {
        userId: ctx.where.id,
      },
    }, function(err, tokens) {
      tokens.forEach(function(token) {
        AccessToken.destroyById(token.id, function() {
          console.log('Deleted token %s for user ', token.id, token.userId);
        });
      });
    });

    next();
  });
```

loay:
@bajtos 
I Changed the code to cope with destroyAll.
Added a 2nd test case. 
Also, please ignore  the logs. 
PTAL. Thanks

common/models/user.js
loay: created issue 1005
https://github.com/strongloop/loopback-datasource-juggler/issues/1005

common/models/user.js
```diff
@@ -301,6 +301,22 @@ module.exports = function(User) {
...
+    ctx.Model.find({ where: ctx.where }, function(err, list) {
+      if (err) return next(err);
+
+      var pkName = ctx.Model.definition.idName() || 'id';
+      var ids = list.map(function(u) { return u[pkName]; });
+      ctx.where = {};
+      ctx.where[pkName] = { inq: ids };
+
+      AccessToken.destroyAll({ userId: { inq: ids }}, next);
+      console.log('Deleted token for users ', ids);
```
bajtos: Please use `debug` instead of `console.log`, and also either 1) log "deleting tokens..." before calling destroyAll 2) or log from inside of the callback passed to destroyAll.

test/user.test.js
```diff
@@ -214,6 +215,95 @@ describe('User', function() {
...
+            accessTokenId = accessToken.userId;
+            if (err) return done (err);
+            assert(accessTokenId);
+            console.log('acessToken.userId: ', accessTokenId);
+            console.log('>>>user: ', usersId);
+            next();
+          });
+        },
+        function(next) {
+          User.deleteAll({ id: accessTokenId }, function(err) {
```
bajtos: This looks suspicous to me, you are deleting all users that have id same as access token id.

I think you should create two users with a property that has same value (e.g. `name` or similar) and then delete them via `User.deleteAll({ where: { name: '...' } })`

loay:
@bajtos 
It fails with 2 users with the same username.
However, if I return my old code (left the old one  commented out for comparing), it deletes the tokens and passes through.

loay:
@bajtos 
 corrected the test case
cleaned up the code. 
BTW, it works with old code still, but kept the implementation you suggested.

test/user.test.js
```diff
@@ -214,6 +215,98 @@ describe('User', function() {
       assert(u2.password === u1.password);
     });
 
+    it('invalidates the user\'s accessToken when the user is deleted By id', function(done) {
+      var usersId;
+      async.series([
+        function(next) {
+          User.create({ email: 'b@c.com', password: 'bar' }, function(err, user) {
+            usersId = user.id;
+            if (err) return done (err);
```
raymondfeng: Usually, we should call `next(err)` to fail the current task and `async` will abort with an error.

test/user.test.js
```diff
@@ -214,6 +215,98 @@ describe('User', function() {
...
+        },
+        function(next) {
+          User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
+            if (err) return done (err);
+            assert(accessToken.userId);
+            next();
+          });
+        },
+        function(next) {
+          User.deleteById(usersId, function(err) {
```
raymondfeng: Do we have a test for User.destroyAll?

test/user.test.js
loay: Thanks. Fixed it. 
But but I kept the `function (err)` at the end of the series array still, because it won’t work without it

loay:
@raymondfeng Update errorhandling. 
PTAL. Thanks.

test/user.test.js
```diff
@@ -214,6 +215,96 @@ describe('User', function() {
...
     });
 
+    it('invalidates the user\'s accessToken when the user is deleted By id', function(done) {
+      var usersId;
+      async.series([
+        function(next) {
+          User.create({ email: 'b@c.com', password: 'bar' }, function(err, user) {
+            usersId = user.id;
+            if (err) return next(err);
+            next(err);
```
raymondfeng: next() is good enough as the error branch has returned.

raymondfeng: Or you can just call `next(err)` to replace line 224 & 225.

test/user.test.js
```diff
@@ -214,6 +215,96 @@ describe('User', function() {
...
+            expect(userFound).to.equal(null);
+            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+              if (err) return next(err);
+              expect(tokens.length).to.equal(0);
+              next(err);
+            });
+          });
+        },
+      ], function(err) {
+        if (err) return next(err);
```
raymondfeng: Should be done(err)

loay:
@raymondfeng 
Fixed for both cases. Thanks.

common/models/user.js
```diff
@@ -301,6 +301,21 @@ module.exports = function(User) {
     return fn.promise;
   };
 
+  User.observe('before delete', function(ctx, next) {
+    var AccessToken = ctx.Model.relations.accessTokens.modelTo;
+    ctx.Model.find({ where: ctx.where }, function(err, list) {
```
raymondfeng: You should use `fields: [pkName]` so that only ids are returned.

common/models/user.js
```diff
@@ -301,6 +301,21 @@ module.exports = function(User) {
...
+      if (err) return next(err);
+
+      var pkName = ctx.Model.definition.idName() || 'id';
+      var ids = list.map(function(u) { return u[pkName]; });
+      ctx.where = {};
+      ctx.where[pkName] = { inq: ids };
+
+      AccessToken.destroyAll({ userId: { inq: ids }}, next);
```
raymondfeng: Please note that there is potential risk that number of ids exceeds the limit of `inq`.

raymondfeng: I guess we can defer this concern.

loay: I previously used this code, but Miroslav had his concerns about it in this comment: https://github.com/strongloop/loopback/pull/2492#discussion_r71323039
```
  User.observe('before delete', function(ctx, next) {
    var AccessToken = ctx.Model.relations.accessTokens.modelTo;

    AccessToken.find({
      where: {
        userId: ctx.where.id,
      },
    }, function(err, tokens) {
      tokens.forEach(function(token) {
        AccessToken.destroyById(token.id, function() {
          console.log('Deleted token %s for user ', token.id, token.userId);
        });
      });
    });

    next();
  });
```

common/models/user.js
loay: @raymondfeng 
do you mean: `ctx.Model.find({ where: ctx.where[pkName] }, function(err, list) {`?
this one fails the deleteById test case. 

and if I do it like that: 
`ctx.Model.find({ where: {fields:[pkName]} }, function(err, list) {`
it fails both and doesn't delete the token
Am I missing something here?

loay: NVM, I got it. Just ignore my previous comment.

common/models/user.js
raymondfeng: I was referring to use `inq` in batches. For example, 1000 ids are divided into 10 chunks with 100 ids in each chunk.

common/models/user.js
```diff
@@ -304,7 +304,7 @@ module.exports = function(User) {
   User.observe('before delete', function(ctx, next) {
     var AccessToken = ctx.Model.relations.accessTokens.modelTo;
     var pkName = ctx.Model.definition.idName() || 'id';
-    ctx.Model.find({ fields: ctx.where[pkName] }, function(err, list) {
+    ctx.Model.find({ fields:[pkName] }, function(err, list) {
```
raymondfeng: You need to keep `where` in addition to `fields`.

raymondfeng:
LGTM.

test/user.test.js
```diff
@@ -214,6 +215,92 @@ describe('User', function() {
...
+          });
+        },
+        function(next) {
+          User.deleteAll({ name: 'myname' }, function(err, user) {
+            next(err);
+          });
+        },
+        function(next) {
+          User.find({ where: { name: 'myname' }}, function(err, userFound)  {
+            if (err) return next (err);
```
Amir-61: @superkhau I just wonder why Linter had not complained about a space between `next` and `(` shouldn't it be like the following?
```
if (err) return next(err);
```

common/models/user.js
bajtos: @loay AFAICT, this comment hasn't been addressed. Could you please fix it in a new pull request ASAP?

test/user.test.js
```diff
@@ -214,6 +215,92 @@ describe('User', function() {
...
+        done();
+      });
+    });
+
+    it('invalidates the user\'s accessToken when the user is deleted all', function(done) {
+      var usersId, accessTokenId;
+      async.series([
+        function(next) {
+          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
+          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
```
bajtos: I find this formatting different to read, as it's not clear what are the items of the first array arg. How about this instead?

```js
User.create([
  { name: ... },
  { name: ... }.,
], function(err, users) {
});
```

test/user.test.js
```diff
@@ -214,6 +215,92 @@ describe('User', function() {
...
+            usersId = user.id;
```
bajtos: I find this suspicious, the `user` argument should contain an array containing two items (two users created). `user.id` will be `null` in such case.

test/user.test.js
```diff
@@ -214,6 +215,92 @@ describe('User', function() {
...
+        function(next) {
+          User.deleteAll({ name: 'myname' }, function(err, user) {
+            next(err);
+          });
+        },
+        function(next) {
+          User.find({ where: { name: 'myname' }}, function(err, userFound)  {
+            if (err) return next (err);
+            expect(userFound.length).to.equal(0);
+            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
```
bajtos: If I am right in my previous comment that `usersId` is `undefined`, then this query is effectively `{ where: {} }`, i.e. it matches all user records. I think the right thing to do here is to use `inq` operator to find access tokens belonging on any of the two test users.

test/user.test.js
superkhau: Yes it should, but I have no idea why it didn't catch. Please check and see if the rule listed at https://github.com/strongloop/eslint-config-loopback/blob/master/eslint.json and submit a PR with a fix if possible.

Issue #2104, opened 2016-02-26
Return type of `PersistedModel.updateAll` should be `number` instead of `object`?


hideya:
The return type of `PersistedModel.updateAll`is specified as `object` at the following line but I think it should have been `number`:
https://github.com/strongloop/loopback/blob/7252c7686cd4f804c47688ee97390436794d5a02/lib/persisted-model.js#L661

``` js
      returns: {
        arg: 'count',
        description: 'The number of instances updated',
        type: 'object',
        root: true
      },
```

It seems that the return type of `PersistedModel.destroyAll` and `deleteById` should also have been `number`:
 https://github.com/strongloop/loopback/blob/7252c7686cd4f804c47688ee97390436794d5a02/lib/persisted-model.js#L640
https://github.com/strongloop/loopback/blob/7252c7686cd4f804c47688ee97390436794d5a02/lib/persisted-model.js#L674

I found this while trying SDK generation using [swagger-codegen](https://github.com/swagger-api/swagger-codegen).  Because of the above, `slc loopback:export-api-def` generates swagger spec like the following:

``` yaml
      operationId: User.updateAll
      ...
      responses:
        '200':
          description: Request was successful
          schema:
            description: The number of instances updated
            type: object
```

This prevents swagger-codegen from generating expected code for `User.updateAll` that allows the user to access the returned number.  If I manually change this `object` to `number`, swagger-codegen generates expected code.

hideya:
Forgot to cc: @bajtos

bajtos:
> `PersistedModel.updateAll`, https://github.com/strongloop/loopback/blob/7252c7686cd4f804c47688ee97390436794d5a02/lib/persisted-model.js#L661
> ```js
returns: {
        arg: 'count',
        description: 'The number of instances updated',
        type: 'object',
        root: true
      },
```

This is rather confusing, the argument should have been called `info`. Here is an example response:

```json
{
  "count": 1
}
```

IMO, this is the correct remoting metadata (notice `root:true` must be omitted):

```js
returns: {
  arg: 'count',
  description: 'The number of instances updated',
  type: 'number',
},
```

bajtos:
> IMO, this is the correct remoting metadata (notice root:true must be omitted):
> ```js
returns: {
  arg: 'count',
  description: 'The number of instances updated',
  type: 'number',
```

Turns out I was wrong in my comment. Both `updateAll` and `deleteAll` pass `info` object to the callback. This `info` object contains a `count` property.

[...]

PR #2535, opened 2016-07-22
updateAll and destroyAll return number

kevdougful:
response to Issue #2104, updateAll and destroyAll return type is number with root=false

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -741,9 +741,8 @@ module.exports = function(registry) {
       accepts: { arg: 'where', type: 'object', description: 'filter.where object' },
       returns: {
         arg: 'count',
-        type: 'object',
+        type: 'number',
         description: 'The number of instances deleted',
-        root: true,
       },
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

bajtos:
I am afraid it's not enough to fix the remoting metadata in LoopBack, I think we need to fix `updateAll` and `destroyAll` methods to pass a different value to the callback.

Fortunately, loopback-swagger supports anonymous object types as of https://github.com/strongloop/loopback-swagger/pull/54.

I think the simplest solution is to rework remoting metadata to describe the root "returns" object as `{ type: { count: 'number' }}`.

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/2842

PR #2549, opened 2016-07-25
Fix test case error

loay:
https://github.com/strongloop/loopback/issues/580

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -254,19 +254,24 @@ describe('User', function() {
     });
 
     it('invalidates the user\'s accessToken when the user is deleted all', function(done) {
-      var usersId, accessTokenId;
+      var userIds = [];
+      var accessTokenId;
       async.series([
         function(next) {
-          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
-          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
-            usersId = user.id;
+          User.create([
+            { name: 'myname', email: 'b@c.com', password: 'bar' },
+            { name: 'myname', email: 'd@c.com', password: 'bar' },
+          ], function(err, users) {
+            userIds = users.map(function(u) {
+              return u.id;
+            });
             next(err);
           });
         },
         function(next) {
           User.login({ email: 'b@c.com', password: 'bar' }, function(err, accessToken) {
             accessTokenId = accessToken.userId;
-            if (err) return next (err);
+            if (err) return next(err);
             assert(accessTokenId);
             next();
           });
@@ -274,7 +279,7 @@ describe('User', function() {
         function(next) {
           User.login({ email: 'd@c.com', password: 'bar' }, function(err, accessToken) {
             accessTokenId = accessToken.userId;
-            if (err) return next (err);
+            if (err) return next(err);
             assert(accessTokenId);
             next();
           });
@@ -286,9 +291,9 @@ describe('User', function() {
         },
         function(next) {
           User.find({ where: { name: 'myname' }}, function(err, userFound)  {
-            if (err) return next (err);
+            if (err) return next(err);
             expect(userFound.length).to.equal(0);
-            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+            AccessToken.find({ where: { userId: { inq: userIds }}}, function(err, tokens) {
               if (err) return next(err);
               expect(tokens.length).to.equal(0);
               next();


loay:
@superkhau PTAL. 
I fixed the array issue that @bajtos raised:

- https://github.com/strongloop/loopback/pull/2492#discussion_r72075281
- https://github.com/strongloop/loopback/pull/2492#discussion_r72075391
- https://github.com/strongloop/loopback/pull/2492#discussion_r72075760

As for the array, is it ok in this format or you prefer that I don't specify an index  (e.g 0,1)? 
Thanks

common/models/user.js
```diff
@@ -312,6 +312,7 @@ module.exports = function(User) {
       ctx.where[pkName] = { inq: ids };
 
       AccessToken.destroyAll({ userId: { inq: ids }}, next);
+      console.log('tokens deleted for users %j', ids);
```
superkhau: I think we should use `debug` here instead.

test/user.test.js
```diff
@@ -254,27 +254,31 @@ describe('User', function() {
...
+      var usersId = [];
+      var accessTokenId;
       async.series([
         function(next) {
-          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
-          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
-            usersId = user.id;
+          User.create([
+          { name: 'myname', email: 'b@c.com', password: 'bar' },
+          { name: 'myname', email: 'd@c.com', password: 'bar' },
```
superkhau: indent one more level for the objects (is this using the latest eslint-config-loopback@4.0.0)? I thought we removed the spaces at the front and back.

test/user.test.js
```diff
@@ -254,27 +254,31 @@ describe('User', function() {
     });
 
     it('invalidates the user\'s accessToken when the user is deleted all', function(done) {
-      var usersId, accessTokenId;
+      var usersId = [];
```
superkhau: var userIds = [];

test/user.test.js
```diff
@@ -254,27 +254,31 @@ describe('User', function() {
...
         function(next) {
-          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
-          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
-            usersId = user.id;
+          User.create([
+          { name: 'myname', email: 'b@c.com', password: 'bar' },
+          { name: 'myname', email: 'd@c.com', password: 'bar' },
+          ], function(err, users) {
+            usersId[0] = users[0].id;
+            usersId[1] = users[1].id;
```
superkhau: users.forEach(function(user) {
  userIds.push(user.id);
});

test/user.test.js
```diff
@@ -286,9 +290,9 @@ describe('User', function() {
         },
         function(next) {
           User.find({ where: { name: 'myname' }}, function(err, userFound)  {
-            if (err) return next (err);
+            if (err) return next(err);
             expect(userFound.length).to.equal(0);
-            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+            AccessToken.find({ userId: { inq: usersId }}, function(err, tokens) {
```
superkhau: userIds

superkhau:
@loay Reviewed.

common/models/user.js
loay: I always remove as I did within the previous PR. I just need it for testing. 
It is not even safe in production mode, so I won't keep it at all.

test/user.test.js
loay: No this repo is still @ 1.0.0. 
I will do another PR to fix lint for this repo.

test/user.test.js
loay: I had this function but I removed for the same of a simple array of length of 2. 
I think this is better in case we need to add more users for testing.

loay:
@bajtos PTAL. 
Please ignore all the `console.log`, as they will be removed as usual before merging. 
Also, I will do another PR to update eslint version.

common/models/user.js
bajtos: > I just need it for testing. 

This is a code smell to me. The unit-tests should provide enough information when they fail so that no temporary console logs are needed.

test/user.test.js
```diff
@@ -254,27 +254,32 @@ describe('User', function() {
...
       async.series([
         function(next) {
-          User.create([{ name: 'myname', email: 'b@c.com', password: 'bar' },
-          { name: 'myname', email: 'd@c.com', password: 'bar' }], function(err, user) {
-            usersId = user.id;
+          User.create([
+            {name: 'myname', email: 'b@c.com', password: 'bar'},
+            {name: 'myname', email: 'd@c.com', password: 'bar'},
+          ], function(err, users) {
+            users.forEach(function(user) {
```
```js
userIds = users.map(function(u) { return u.id; });
```

test/user.test.js
```diff
@@ -286,10 +291,11 @@ describe('User', function() {
         },
         function(next) {
           User.find({ where: { name: 'myname' }}, function(err, userFound)  {
-            if (err) return next (err);
+            if (err) return next(err);
             expect(userFound.length).to.equal(0);
-            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+            AccessToken.find({ userId: { inq: userIds }}, function(err, tokens) {
```
```js
{ where: { userId: { inq: userIds } } }
```

bajtos:
Two more comments, the last one is crucial! Please remove the console logs before the final review.

loay:
@bajtos PTAL.  I fixed accordingly. 
Also, I had to keep the linting conform with version 1.0.0, otherwise all test cases will fail.  I will do a separate PR to update linting.

bajtos:
LGTM, please wait for the CI builds to verify all is green.

PR #2501, opened 2016-07-13
Set forceId = false for AccessToken

jannyHou:
Connect to strongloop/loopback-datasource-juggler#849

Some test cases need refactor after we set forceId as true by default.
Details related please see: https://github.com/strongloop/loopback-datasource-juggler/pull/982#issuecomment-227941053

PR #2551, opened 2016-07-26
test: use local registry in test fixtures

bajtos:
Use local registry in test fixtures to prevent collision in globally shared models.

Fix issues discoverd in auth implementation where the global registry was used instead of the correct local one.

@jannyHou please review.

Once this is landed, the number of places where we need `forceId:false` in #2501 can be reduced.

cc @raymondfeng @ritch


test/user.integration.js
```diff
@@ -75,7 +69,12 @@ describe('users - integration', function() {
 
           expect(res.body.title).to.be.eql('T1');
           expect(res.body.content).to.be.eql('C1');
-          expect(res.body.userId).to.be.eql(userId);
+          // HACK(bajtos) other tests are somehow screwing up the state of
+          // models and/or datasource, and as a result, body.userId is string
+          // while userId is a number.
+          // This cannot be reproduced when the tests in this file are run
+          // in isolation.
```
bajtos: The user integration test is really weird. I spent more than an hour trying to figure out what's going on, but with no success.

When running in isolation, the following post is created:

```
{ title: 'T1', content: 'C1', id: 1, userId: 1 }
```

When run together with other tests, then the following post is created:

```
{ title: 'T1',
  content: 'C1',
  id: '1413b820-5329-11e6-8c40-8d24a4878c7c',
  userId: '1' }
```

bajtos: I managed to track down the source of the conflict to "Remote Methods" suite in [test/model.test.js](https://github.com/strongloop/loopback/blob/8f7e032a01bdde54b195385dc5eff37dc4524739/test/model.test.js#L331-L706).

bajtos: Here is the offending line:

```js
User.hasMany(Post);
```

https://github.com/strongloop/loopback/blob/8f7e032a01bdde54b195385dc5eff37dc4524739/test/model.test.js#L105

Looks like when the user-integration-app defines "User hasMany Post" relation, the Post model defined here is looked up instead of the "post" model defined by that app.

/cc @raymondfeng @ritch any idea on why is that happening? I'll continue my investigation, I'm just wondering if you perhaps can save me some time.

bajtos: After another hour spent investigating this problem, I am giving up.

In this patch, I am renaming `post` to `Post` in `test/user-integration-app` to prevent collision with `test/model.test.js`.

I am not happy about it, but don't want to waste any more time on this issue.

bajtos:
@jannyHou PTAL

test/model.test.js
```diff
@@ -549,7 +544,7 @@ describe.onServer('Remote Methods', function() {
     it('Get the Change Model', function() {
       var UserChange = User.getChangeModel();
       var change = new UserChange();
-      assert(change instanceof Change);
+      assert(change instanceof app.registry.getModel('Change'));
```
jannyHou: Shall we delete the declaration of [`Change`](https://github.com/strongloop/loopback/blob/5b82f94ca1a0c6db183635898ea95844b275deb0/test/model.test.js#L9) since now it's replaced with `app.registry.getModel('Change')`?

jannyHou:
@slnode test please

jannyHou:
@bajtos one comment :) Other than that LGTM.

test/model.test.js
bajtos: Good catch, will remove.

bajtos:
@slnode test please

bajtos:
Looks like the build failures are not related to my changes, the master branch is failing on CI too. I am going to land this pull request.

benkroeger:
@bajtos Good thing I found this. Was about to submit a PR doing what you did in `common/models/acl.js` with the local registry. I found that when using 
```
const app =  loopback({ localRegistry: true, loadBuiltinModels: true }); 
```
ACLs defined in a `model-definition.json` are not found by `ACL.getStaticACLs()` since it would lookup the `modelName` in the global registry `loopback.findModel(modelName)`

This leads me to my actual problem: 
I'm using `app` with `local Registry` and `builtInModels`, no custom `Role` model.
Then I define a dynamic role 'admin':
```
const { registry } = app;
const Role = registry.getModel('Role');
Role.registerResolver('admin', (role, context, cb) => {
    process.nextTick(() => {
      console.log('in admin role resolver');
      cb(null, false);
    });
  });
```

On one of my models, I define the following acl:

```
"acls": [{
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    }, {
      "accessType": "READ",
      "principalType": "ROLE",
      "principalId": "admin",
      "permission": "ALLOW",
      "property": "find"
    }],
```

When calling the REST API, somewhere down the line `ACL.checkAccessForContext()` gets called. 
In `common/models/acl.js` you can find this:

```
var Role = loopback.Role;
...
    ...
    var roleModel = registry.getModelByType(Role);
    ...
    // this happens for each previously resolved alc
    roleModel.isInRole(...);
    ...
}
...
```

Due to the scenario described above (no custom models with `Role` as baseModel, `registry.getModelByTpe(Role)` returns the provided argument as-is, which is the global `loopback.Role`

When now calling `roleModel.isInRole('admin')`, the 'admin' resolver is not recognized since it was registered on the local registry of our app instance. `isInRole()` proceeds to calling `this.findOne()` in order to try to resolve a role named 'admin' from the attached dataSource and ultimately the application *crashes* since this global `loopback.Role` has never been attached to a dataSouce

Issue #382, opened 2014-07-18
Tighten security for resetting password


offlinehacker:
With `POST /user/{id}` user can change password without of presence of old password. This is security problem, because anyone who stoles any valid access token from user can change users password.

**Tasks**
- [x] Only allow resetPassword by email if email verification is required and done.
- [x] #3303 - Provide solution for finishing the registration when the verification link was not followed (opened) before it has expired.
- [x] Add a new method, let's call it `User.changePassword(id, oldPassword, newPassword, cb)`. This method should have ACLs configured to allow only the owner to execute it (i.e. users can change only their own passwords) - see https://github.com/strongloop/loopback/pull/3299
- [x] Scope the password-reset token so that it allows only changing User's password and nothing else. Regular access tokens must not allow changing User's password without providing the old one. 
  ~~Save the password-reset token the same way as the verification token i.e. with the user. That way you would know with which token you can change the password without knowing your old one. It would also mean that if you request another password reset it would delete/invalidate the previous token so that at max only one can be active at one time. Because the current implementation relies on the client to call `updateAttributes` with a temporary access token, I think we will need to implement a new method similar to `User.changePassword` that will accept the new reset-token instead of the old password.~~
  - [x] https://github.com/strongloop/loopback/pull/3313 Implement Access Token Scopes
  - [x] #3350 - Implement a new method for resetting the password. This method does not require the old password and can be invoked only with a password-reset token
- [x] #3350 - Disable changing the `password` property via other methods. To preserve backwards-compatibility, there should be a feature flag controlling this behaviour.

**Next**
- https://github.com/strongloop/loopback/issues/3349

shaunbruno:
Reading the comments, I realize this issue was closed as there may have been no further discussion. As @offlinehacker  and @TFaga have noted, a scope/type concept to differentiate the reset password token from a normal token would solve the issue.

Having read the code in my current version (see version info below), if I'm not mistaken the normal reset password scenario (without being logged in) has a fairly significant security flaw insofar that the AccessToken generated provides system wide access for 15 minute (simply by providing a valid email). 

The counter argument is perhaps that if the requester also has the email then they are indeed the owning user? If so, I would disagree and +1 expediting the differentiation of the AccessToken's type (so that a reset password user does not have the same full-access permissions as a normal user).

Version Info:
strongloop v2.10.3 (node v0.10.32)
├── strong-arc@1.1.0
├── strong-build@1.0.3 (5a43a57)
[...]

PR #2552, opened 2016-07-26
Allow resetPassword by email only if email verification was done

loay:
connect to https://github.com/strongloop/loopback/issues/382


loay:
@raymondfeng @bajtos PTAL. Thanks

common/models/user.js
```diff
@@ -560,19 +560,26 @@ module.exports = function(User) {
...
         return cb(err);
       }
-      // create a short lived access token for temp login to change password
-      // TODO(ritch) - eventually this should only allow password change
-      user.accessTokens.create({ ttl: ttl }, function(err, accessToken) {
-        if (err) {
-          return cb(err);
-        }
-        cb();
+      if (user && user.emailVerified) {
```
bajtos: `user` is always truthy, see L557 above. no need to check it here.

common/models/user.js
```diff
@@ -560,19 +560,26 @@ module.exports = function(User) {
...
         UserModel.emit('resetPasswordRequest', {
           email: options.email,
-          accessToken: accessToken,
```
bajtos: How are you going to construct an authenticated URL when there is no accessToken available?

bajtos:
@loay let's keep this pull request focused on the first task from #382, which is

> Only allow resetPassword by email if email verification is required and done.

The change proposed here goes in the opposite direction. Here is what we need:

1) if email is not verified, then return an error
2) if email was verified, then continue with the same behaviour as before

PR #2557, opened 2016-07-27
WIP -- Add KeyValue model

superkhau:
```
#static
get val by key
delete val by key
set val by key (optional ttl)
touch/expire key (optional ttl) -- renew ttl to now otherwise
optional - get remaining ttl/expire time

#prototype/instance
get() -- use id from inst to get the latest ver from backend
set([ttl]) -- use existing id and set ttl for the instance
touch/expire -- don't set any updates to the backend
  customer.set(10000) --> get list of props from inst together with ttl and call
  static set
  customer.touch(ttl) --> call touch/expire static and set ttl
delete()
getExpirationTime() -- return expiration time (UTC timestamp)
```

### DIFF
--- a/lib/keyvalue-model.js
+++ b/lib/keyvalue-model.js
@@ -1,0 +1,60 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Node module: loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/*!
+ * Module Dependencies.
+ */
+
+module.exports = KeyValueModel;
+
+var Model = KeyValueModel;
+
+function Model(registry) {
+  this.registry = registry;
+}
+
+// Static
+
+Model.get = function get(key) {
+  // ret val by key
+}
+
+Model.expire
+Model.expires = function() {
+  // ret remaining ttl/expires time
+}
+
+Model.delete =
+Model.remove = function(key) {
+  // del val by key
+}
+
+Model.set = function get(key) {
+  // set val by key
+}
+
+Model.touch = function touch(ttl) {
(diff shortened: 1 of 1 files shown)


bajtos:
It makes me wonder, should this model be defined in loopback repository, or loopback-datasource-juggler instead?

Perhaps we should create a new component to provide KV model and in-memory connector, that way this feature can be versioned independently. Having said that, I think doing that would require some work in loopback-boot to allow components to contribute both models and connectors.

In general, I think we should follow the practice we have for DAO in juggler, where the data access API, the in-memory connector and the test-suite validating connector implementation lives in the same module; and the test suite can be run by connectors to verify their implementation matches the contract expected by loopback/juggler.

@superkhau @raymondfeng @ritch Thoughts?

superkhau:
>It makes me wonder, should this model be defined in loopback repository, or loopback-datasource-juggler instead?

This I'm not sure of ATM, we already have persisted model in Jugger and I'm sure this will end up moving around once we decouple all the Juggler components as planned by @raymondfeng. I say leave it here for now and we can move it when we come to a concensus in the future.

>Perhaps we should create a new component to provide KV model and in-memory connector, that way this feature can be versioned independently. Having said that, I think doing that would require some work in loopback-boot to allow components to contribute both models and connectors.

I don't see any downsides except the time it would take to impl this.

>In general, I think we should follow the practice we have for DAO in juggler, where the data access API, the in-memory connector and the test-suite validating connector implementation lives in the same module; and the test suite can be run by connectors to verify their implementation matches the contract expected by loopback/juggler.

+1 for this.

bajtos:
Closing in favour of other pull requests in [#Epic: KV Connectors](https://github.com/strongloop/loopback/milestone/60).

Issue #1495, opened 2015-07-01
Injecting Remote Context via Options


ritch:
UPDATE 2016-12-22 We ended up implementing a different solution as described in https://github.com/strongloop/loopback/pull/3023.

**Tasks**
 - [x] Make the change in the `master` branch - see https://github.com/strongloop/loopback/pull/3023 (released in `loopback@3.2.0`)
 - [x] Back-port to `2.x` - see https://github.com/strongloop/loopback/issues/3048 (released in `loopback@2.37.0`)
 - [x] Update the documentation - see https://github.com/strongloop/loopback.io/pull/227
 - [x] Close other context-related issues and point users to this new solution.

**Original description for posterity**

The example below allows you to inject the remote context object (from strong remoting) into all methods that accept an `options` argument.

``` js

function inject(ctx, next) {
  var options = hasOptions(ctx.method.accepts) && (ctx.args.options || {});
  if(options) {
    options.remoteCtx = ctx;
    ctx.args.options = options;
  }
  next();
}

app.remotes().before('*.*', inject);

app.remotes().before('*.prototype.*', function(ctx, instance, next) {
  inject(ctx, next);
});

// unfortunately this requires us to add the options object
// to the remote method definition
app.remotes().methods().forEach(function(method) {
  if(!hasOptions(method.accepts)) {
    method.accepts.push({
      arg: 'options',
      type: 'object',
      injectCtx: true
    });
  }
});

function hasOptions(accepts) {
  for (var i = 0; i < accepts.length; i++) {
    var argDesc = accepts[i];
    if (argDesc.arg === 'options' && argDesc.injectCtx) {
      return true;
    }
  }
}
```

Why? So you can use the remote context in remote methods, operation hooks, connector implementation, etc.

``` js
MyModel.observe('before save', function(ctx, next) {
  console.log(ctx.options.remoteCtx.accessToken.userId); // current user
});
```

This approach is specifically designed to allow you to do what is possible with `loopback.getCurrentContext()` but without the dependency on `cls`. The injection approach is much simpler and should be quite a bit faster since `cls` adds some significant overhead.

@bajtos @fabien @raymondfeng

pulkitsinghal:
@ritch - the initial chunk of code sits in server.js correct?

ritch:
A boot script works fine... The goal is to add this to core

bajtos:
The mechanism of passing the context object via the `options` argument looks ok, it's consistent with what we already do for transactions.

[...]

PR #2559, opened 2016-07-28
Deprecate current-context API [2.x]

bajtos:
Deprecate all current-context APIs in favour of loopback-context-cls.

This is a follow-up for https://github.com/strongloop/loopback-context-cls/pull/1
> Based on our discussions, I think we should remove `getCurrentContext` from LoopBack 3.0. In order to simplify upgrade path and allow community to take over maintenance of CLS-based current context, I want to follow the same process we did with loopback-boot and that is to extract the functionality into a standalone module that can be versioned independently.

@raymondfeng @ritch please review
cc @strongloop/loopback-dev FYI


richardpringle:
Would this somehow make it any easier to solve the following issue without a hack?

https://github.com/strongloop/loopback/issues/1495

bajtos:
Probably not. My intention here is to allow us to back out of `getCurrentContext` and stop trying to fix it, while not breaking all apps that may be depending on this feature. I think `getCurrentContext` may work in certain cases, so why to break them if there is a reasonably easy workaround (proposed here).

lib/current-context.js
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * @returns {ChainedContext} The context object or null.
+   */
+  loopback.getCurrentContext = function() {
+    // NOTE(bajtos) ClsContext.getCurrentContext is overriden whenever
+    // the context changes, therefore we cannot simply assign
+    // ClsContext.getCurrentContext() to loopback.getCurrentContext()
+    deprecated('loopback.getCurrentContext() is deprecated. ' +
+      'Consider using different means of passing the context around. ' +
+      'As a short-term fix, use loopback-context-cls module.');
```
raymondfeng: This message is a bit confusing. Maybe we should say something like:
```
The current implementation is based on loopback-context-cls and it has some limitations.
```

bajtos: I am fine to improve the message, as long as it remains actionable. I.e. it should be clear to the user how to get rid of that deprecation warning.

IMO, your proposal @raymondfeng, describes the state but does not provide any clue to the user about how to proceed.

Which part of the message do you find confusing?

ritch: Can we link to a github issue or wiki (or other doc) so that we can keep the info up to date?

A deprecation warning will be sufficient to direct the person to a doc with some concrete info. We can also keep this up to date as we look into this further.

raymondfeng: The confusing part of `As a short-term fix, use loopback-context-cls module.` is that it seems to instruct a user to use `loopback-context-cls`, which is already being used.

bajtos: I have reworded deprecation warnings to point to our documentation page: https://docs.strongloop.com/display/APIC/Using%20current%20context

bajtos:
@raymondfeng @ritch PTAL again, LGTY now?

raymondfeng:
LGTM

bajtos:
I am waiting for https://github.com/strongloop/loopback-context/pull/3/files before landing this PR.

PR #2435, opened 2016-06-14
Expose `Replace*` methods for 2.x

Amir-61:
**Patch description:**

*Re-mapping `updateAttributes` endpoint to use
`PATCH` and `PUT`(configurable) verb
*Exposing `replaceById` and `replaceOrCreate` via
`POST` and `PUT`(configurable) verb

---

##### What are the differences between this ptach for `2.x` Vs. [patch for `3.x`](https://github.com/strongloop/loopback/pull/2316):

1) Default value for `replaceOnPUT` would be set to false; please see [here](https://github.com/strongloop/loopback/pull/2435/files#diff-3bf07e630a94e80eef9557e874f96f56R606)

2) In this patch for `2.x`, we preserve `upsert` and set `pachOrCreate` as an alias only and change tests accordingly.

3) In this patch for `2.x`, we preserve `updateAttributes` and set `pachAttributes` as an alias only and change tests accordingly.

4) *Question:* JSdoc for `PersistedModel Methods` are generated based on `lib\persisted-models.js`, so once this patch gets landed I believe docs for [replaceOrCreate](https://github.com/strongloop/loopback/pull/2435/files#diff-3bf07e630a94e80eef9557e874f96f56R120), [replaceAttributes](https://github.com/strongloop/loopback/pull/2435/files#diff-3bf07e630a94e80eef9557e874f96f56R504) and [replaceById](https://github.com/strongloop/loopback/pull/2435/files#diff-3bf07e630a94e80eef9557e874f96f56R519) gets added to our API docs, are we fine with that now?

PS: I squash my commits as one single commit after getting an LGTM. Thanks!

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/917

/to: @bajtos

PR #2560, opened 2016-07-28
Avoid calling deprecated methds

Amir-61:
*Avoid calling deprecated `getHttpMethod` and `getFullPath`

This is the fix for deprecated messages. But please let me bring up the status of https://github.com/strongloop/loopback/pull/2435 with Quentin in our group chat and if we are all good, we can land https://github.com/strongloop/loopback/pull/2435 instead.

/to: @bajtos 

Thanks!

### DIFF
--- a/test/remoting.integration.js
+++ b/test/remoting.integration.js
@@ -99,9 +99,9 @@ describe('remoting - integration', function() {
         ')',
         formatReturns(m),
         ' ',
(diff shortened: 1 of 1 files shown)


bajtos:
LGTM, :shipit: 

We can always land #2435 on top of this quick fix, cannot we?

Anyways, I'll leave the decision up to you.

PR #2564, opened 2016-07-29
[SEMVER-MAJOR] Remove current-context API

bajtos:
Change all current-context APIs to throw a helpful error.

This is a follow-up for https://github.com/strongloop/loopback-context/pull/1
> Based on our discussions, I think we should remove `getCurrentContext` from LoopBack 3.0. In order to simplify upgrade path and allow community to take over maintenance of CLS-based current context, I want to follow the same process we did with loopback-boot and that is to extract the functionality into a standalone module that can be versioned independently.

See also https://github.com/strongloop/loopback/pull/2559 which deprecates current-context API in 2.x.

@raymondfeng @ritch please review
cc @strongloop/loopback-dev FYI


raymondfeng:
LGTM

bajtos:
@slnode test please

bajtos:
I am waiting for https://github.com/strongloop/loopback-context/pull/3/files before landing this PR.

PR #2565, opened 2016-07-29
test: fix change-tracking setup

bajtos:
The remote-connector test has misconfigured the client (remote) model, where the client model was trying to keep track of changes. That's redundant because it's up to the server model (attached directly to the database) to track changes.

This patch fixes that configuration and also cleans up the test code a little bit to make it easier to distinguish between the remote (client) model and the server model.

/to @0candy Please review. Feel free to make any small improvements yourself, and also please back-port to 2.x if appropriate.

/cc @ritch

### DIFF
--- a/test/remote-connector.test.js
+++ b/test/remote-connector.test.js
@@ -25,13 +25,24 @@ describe('RemoteConnector', function() {
         done();
       });
     },
+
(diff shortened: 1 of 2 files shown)


0candy:
LGTM

PR #2580, opened 2016-08-03
Add bcrypt validation

loay:
connect https://github.com/strongloop-internal/scrum-loopback/issues/994

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -14,7 +14,7 @@ var utils = require('../../lib/utils');
 var path = require('path');
 var SALT_WORK_FACTOR = 10;
 var crypto = require('crypto');
-
+var MAX_PASSWORD_LENGTH = 72;
 var bcrypt;
 try {
   // Try the native module first
@@ -548,7 +548,6 @@ module.exports = function(User) {
     cb = cb || utils.createPromiseCallback();
     var UserModel = this;
     var ttl = UserModel.settings.resetPasswordTokenTTL || DEFAULT_RESET_PW_TTL;
-
     options = options || {};
     if (typeof options.email !== 'string') {
       var err = new Error(g.f('Email is required'));
@@ -558,6 +557,13 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    try {
+      if (options.password) {
+        UserModel.validatePassword(options.password);
+      }
+    } catch (err) {
+      return cb(err);
+    }
     UserModel.findOne({ where: { email: options.email }}, function(err, user) {
       if (err) {
         return cb(err);
@@ -596,14 +602,20 @@ module.exports = function(User) {
   };
 
   User.validatePassword = function(plain) {
-    if (typeof plain === 'string' && plain) {
+    var err;
+    if (plain && typeof plain === 'string' && plain.length <= MAX_PASSWORD_LENGTH) {
       return true;
     }
-    var err =  new Error(g.f('Invalid password: %s', plain));
+    if (plain.length > MAX_PASSWORD_LENGTH) {
+      err = new Error (g.f('Password too long: %s', plain));
+      err.code = 'PASSWORD_TOO_LONG';
+    } else {
+      err =  new Error(g.f('Invalid password: %s', plain));
+      err.code = 'INVALID_PASSWORD';
+    }
     err.statusCode = 422;
     throw err;
   };
-
   /*!
    * Setup an extended user model.
    */
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 var SG = require('strong-globalize');
-SG.SetRootDir(__dirname, {autonomousMsgLoading: 'all'});
+SG.SetRootDir(__dirname, { autonomousMsgLoading: 'all' });
 
 /**
  * loopback ~ public api
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -363,6 +363,68 @@ describe('User', function() {
     });
   });
 
+  describe('Password length validation', function() {
+    var pass72Char = new Array(70).join('a') + '012';
+    var pass73Char = pass72Char + '3';
+    var passTooLong = pass72Char + 'WXYZ1234';
+
+    it('rejects passwords longer than 72 characters', function(done) {
+      try {
+        User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
+          if (err) return done (err);
+          done(new Error('User.create() should have thrown an error.'));
+        });
+      } catch (e) {
+        expect(e).to.match(/Password too long/);
+        done();
+      }
+    });
+
+    it('rejects a new user with password longer than 72 characters', function(done) {
+      try {
+        var u = new User({ username: 'foo', password: pass73Char });
+        assert(false, 'Error should have been thrown');
+      } catch (e) {
+        expect(e).to.match(/Password too long/);
+        done();
+      }
+    });
+
+    it('accepts passwords that are exactly 72 characters long', function(done) {
+      User.create({ email: 'b@c.com', password: pass72Char }, function(err, user) {
+        if (err) return done(err);
+        User.findById(user.id, function(err, userFound)  {
+          if (err) return done (err);
+          assert(userFound);
+          done();
+        });
+      });
+    });
+
+    it('allows login with password exactly 72 characters long', function(done) {
+      User.create({ email: 'b@c.com', password: pass72Char }, function(err) {
+        if (err) return done(err);
+        User.login({ email: 'b@c.com', password: pass72Char }, function(err, accessToken) {
+          if (err) return done(err);
+          assertGoodToken(accessToken);
+          assert(accessToken.id);
+          done();
+        });
+      });
+    });
+
+    it('rejects password reset when password is more than 72 chars', function(done) {
+      User.create({ email: 'b@c.com', password: pass72Char }, function(err) {
+        if (err) return done (err);
+        User.resetPassword({ email: 'b@c.com', password: pass73Char }, function(err) {
+          assert(err);
+          expect(err).to.match(/Password too long/);
+          done();
+        });
+      });
+    });
+  });
+
   describe('Access-hook for queries with email NOT case-sensitive', function() {
     it('Should not throw an error if the query does not contain {where: }', function(done) {
       User.find({}, function(err) {
@@ -677,6 +739,23 @@ describe('User', function() {
 
           done();
         });
+    });
+
+    it('allows login with password too long but created in old LB version',
+    function(done) {
+      var bcrypt = require('bcryptjs');
+      var longPassword = new Array(80).join('a');
+      var oldHash = bcrypt.hashSync(longPassword, bcrypt.genSaltSync(1));
+
+      User.create({ email: 'b@c.com', password: oldHash }, function(err) {
+        if (err) return done(err);
+        User.login({ email: 'b@c.com', password: longPassword }, function(err, accessToken) {
+          if (err) return done(err);
+          assert(accessToken.id);
+          // we are logged in, the test passed
+          done();
+        });
+      });
     });
   });
 


loay:
@raymondfeng @bajtos PTAL

bajtos:
@loay let's start with a failing unit test please.

common/models/user.js
```diff
@@ -584,7 +584,11 @@ module.exports = function(User) {
   User.hashPassword = function(plain) {
     this.validatePassword(plain);
     var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
-    return bcrypt.hashSync(plain, salt);
+    if (salt.length < 73) {
+      return bcrypt.hashSync(plain, salt);
+    } else {
+      throw new Error('Validation Error');
```
bajtos: Notice that `validatePassword` provides a helpful error message and HTTP status code 422. We should do the same here.

common/models/user.js
```diff
@@ -584,7 +584,11 @@ module.exports = function(User) {
   User.hashPassword = function(plain) {
     this.validatePassword(plain);
     var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
-    return bcrypt.hashSync(plain, salt);
+    if (salt.length < 73) {
```
bajtos: `salt.length`? I think you should check `plain.length` instead, right?

bajtos:
> I couldn't add the fix within the validation function as the hashPassword function uses the validation, so we would get a stack size error as they will be calling each other. 

I don't understand. My expectation is that it should be enough to extend the condition in `validatePassword` to check the password length is less than 72 chars.

Having said that, the error should make it clear to the user that they have entered a password that's too long and also what's the maximum allowed length.

bajtos:
As for test coverage, I would like to see the following test cases (at least):
 - [ ] what happens when the user enters password that's MAXLEN+1 long - a helpful error is returned
 - [ ] what happens when the user enters password that's MAXLEN long - the operation succeeds
 - [ ] when logging in with a correct password that's MAXLEN long, login succeeds
 - [ ] when logging in with a  password that's MAXLEN long and has only the last character incorrect, login fails

loay:
@bajtos Fixed. PTAL

common/models/user.js
```diff
@@ -596,9 +596,15 @@ module.exports = function(User) {
   };
 
   User.validatePassword = function(plain) {
-    if (typeof plain === 'string' && plain) {
+    var MAXLENGTH = 72;
+    if (typeof plain === 'string' && plain && plain.length <= MAXLENGTH) {
       return true;
     }
+    if (plain.length > MAXLENGTH) {
+      err = new Error (g.f('Invalid password exceeds maximum length'));
```
bajtos: Please include MAXLEN value in the message.

```js
g.f('Password too long - maximum length is %s', MAXLEN);
```

index.js
```diff
@@ -4,7 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 var SG = require('strong-globalize');
-SG.SetRootDir(__dirname, {autonomousMsgLoading: 'all'});
+SG.SetRootDir(__dirname, { autonomousMsgLoading: 'all' });
```
bajtos: I don't think this will pass our linter, please check.

test/user.test.js
```diff
@@ -205,6 +205,60 @@ describe('User', function() {
...
+    describe('Password validation with Maxlength', function() {
+      var pass72 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt';
+      var pass73 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt1';
+      var badPass72 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIZ';
+
+      it('invalidates created account when password is greater than Maxlength', function(done) {
+        try {
+          User.create({ email: 'b@c.com', password: pass73 }, function(err) {
+            if (err) return done(err);
+            assert(err);
```
bajtos: This will never pass, because `if(err) return done(err)` on the previous line.

test/user.test.js
```diff
@@ -205,6 +205,60 @@ describe('User', function() {
       });
     });
 
+    describe('Password validation with Maxlength', function() {
+      var pass72 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt';
+      var pass73 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt1';
+      var badPass72 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIZ';
+
+      it('invalidates created account when password is greater than Maxlength', function(done) {
```
bajtos: I am proposing to improve the test names:

```
rejects passwords longer than 72 chars
allows passwords exactly 72 chars long
```

test/user.test.js
```diff
@@ -205,6 +205,60 @@ describe('User', function() {
...
+          if (err) return done(err);
+          assert(user.id);
+          assert(!err);
+
+          done();
+        });
+      });
+
+      it('validates login when password is at Maxlength', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err) {
```
bajtos: unhandled `err`, add `if (err) return done(err)` on the next line.

test/user.test.js
```diff
@@ -205,6 +205,60 @@ describe('User', function() {
...
+        User.create({ email: 'b@c.com', password: pass72 }, function(err, user) {
+          if (err) return done(err);
+          assert(user.id);
+          assert(!err);
+
+          done();
+        });
+      });
+
+      it('validates login when password is at Maxlength', function(done) {
```
```
it('allows login with password exactly 72 chars long')
```

test/user.test.js
```diff
@@ -205,6 +205,60 @@ describe('User', function() {
...
+            assertGoodToken(accessToken);
+            assert(accessToken.id);
+            assert(!err);
+
+            done();
+          });
+        });
+      });
+
+      it('invalidates login when password is at Maxlength but wrong', function(done) {
```
bajtos: I think in this test, you need to take password MAXLEN long and add few more characters to it.

As this test is written now, it passes even without the changes you made in `common/models/user.js`. Have you seen this test fail?

index.js
loay: yeah that space was actually added when I ran a linting test. It is still at `1.0.0`.
I will fix that once I am done with this PR.

test/user.test.js
loay: login failure error is still thrown without my implementation. 
If I use the proper password, then the test fails `Uncaught TypeError: Cannot read property 'message' of null` as it was expecting an error and login failure

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
       });
     });
 
+    describe('Password validation with Maxlength', function() {
+      var pass72 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt';
+      var pass73 = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfIt1';
+      var badPass = 'ThePasswordShouldNotExceedtheMaximumNumberOfcharactersAndItisInvalidIfItXYZ1';
```
superkhau: Might be more accurate to use `Array(72).join('a')` or something to represent N chars.

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+
+      it('rejects passwords longer than 72 characters', function(done) {
+        try {
+          User.create({ email: 'b@c.com', password: pass73 }, function(err) {
+            if (err) return done(err);
+            assert.equal(err.statusCode, 422);
```
superkhau: The callback should not be called at all so I don't think this test for statusCode is necessary here.

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+            assert(false, 'Error should have been thrown');
+          });
+        } catch (e) {
+          done();
+        }
```
superkhau: Something is weird here, are you expecting the User.create to throw before any execution or expect to handle the error in the callback?

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+      });
+
+      it('accepts passwords that are exactly 72 characters long', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err, user) {
+          if (err) return done(err);
+          assert(user.id);
+          assert(!err);
+
```
superkhau: remove empty line

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+        } catch (e) {
+          done();
+        }
+      });
+
+      it('accepts passwords that are exactly 72 characters long', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err, user) {
+          if (err) return done(err);
+          assert(user.id);
+          assert(!err);
```
superkhau: you already check for `err` on 227 remove

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+          });
+        } catch (e) {
+          done();
+        }
+      });
+
+      it('accepts passwords that are exactly 72 characters long', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err, user) {
+          if (err) return done(err);
+          assert(user.id);
```
superkhau: you should probably do a find to see if the user was created properly

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+      });
+
+      it('allows login with password exactly 72 characters long', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err) {
+          User.login({ email: 'b@c.com', password: pass72 }, function(err, accessToken) {
+            if (err) return done(err);
+            assertGoodToken(accessToken);
+            assert(accessToken.id);
+            assert(!err);
+
```
superkhau: remove empty line

common/models/user.js
```diff
@@ -548,7 +554,7 @@ module.exports = function(User) {
     cb = cb || utils.createPromiseCallback();
     var UserModel = this;
     var ttl = UserModel.settings.resetPasswordTokenTTL || DEFAULT_RESET_PW_TTL;
-
+    var MAXLENGTH = 72;
```
superkhau: `var MAX_PASSWORD_LENGTH` is more clear

common/models/user.js
```diff
@@ -247,6 +247,12 @@ module.exports = function(User) {
             debug('An error is reported from User.hasPassword: %j', err);
             fn(defaultError);
           } else if (isMatch) {
+            if ((credentials.password).length > 72) {
```
superkhau:
```
var passwordTooLong = credentials.password.length > MAX_PASSWORD_LENGTH;
if (passwordTooLong) {
  ...
}
```

common/models/user.js
```diff
@@ -558,6 +564,12 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    if ((options.password) && (options.password).length > MAXLENGTH) {
```
superkhau: `if (options.password && (options.password.length > MAX_PASSWORD_LENGTH)) {`

common/models/user.js
```diff
@@ -558,6 +564,12 @@ module.exports = function(User) {
...
+      var err = new Error(g.f('Password cannot exceed %j characters', MAXLENGTH));
```
superkhau: %j -> %d (assuming the globalization API is the same as util.format -- please verify with @Setogit)

common/models/user.js
```diff
@@ -596,14 +608,19 @@ module.exports = function(User) {
   };
 
   User.validatePassword = function(plain) {
-    if (typeof plain === 'string' && plain) {
+    var MAXLENGTH = 72;
```
superkhau: MAX_PASSWORD_LENGTH

common/models/user.js
```diff
@@ -596,14 +608,19 @@ module.exports = function(User) {
...
+    if (typeof plain === 'string' && plain && plain.length <= MAXLENGTH) {
```
superkhau: `if (plain && typeof plain === 'string' && (plain.length <= MAX_PASSWORD_LENGTH)) {`

common/models/user.js
```diff
@@ -596,14 +608,19 @@ module.exports = function(User) {
...
       return true;
     }
+    if (plain.length > MAXLENGTH) {
+      err = new Error (g.f('Password too long - maximum length is %s', MAXLEN));
+      err.statusCode = 422;
+      throw err;
+    }
```
superkhau: You should validate the other things first and reject earlier. Instead of testing the valid case above on line 612 first, you should check all the negative cases first. ie:

``
// if undefined reject
// if not a string reject
// if invalid length reject

now all valid, process accordingly

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
       });
     });
 
+    describe('Password validation with Maxlength', function() {
```
superkhau: `context('password length validation', function() {`

test/user.test.js
```diff
@@ -205,6 +205,73 @@ describe('User', function() {
...
+            assertGoodToken(accessToken);
+            assert(accessToken.id);
+            assert(!err);
+
+            done();
+          });
+        });
+      });
+
+      it('rejects login when password is proper pass of 72 chars + extra added chars',
```
superkhau: what does the + extra add chars mean? why are we adding extra chars

test/user.test.js
loay: pass73 is a good password with 73 chars, so the expected behaviour that it cannot be used to neither create nor reset, while badpass is a completely bas password that cannot be used to login after the user is already created.

loay:
Hi @bajtos PTAL.
I addressed your comments and Simon's as well.
Also, I added one more test case scenario:
-[ ] When user resets password that is greater than MAXLENGTH

common/models/user.js
bajtos: this is even better:

```js
if (options.password && options.password.length > MAX_PASSWORD_LENGTH) {
```

common/models/user.js
bajtos:
```js
if (plain && typeof plain === 'string' && plain.length <= MAX_PASSWORD_LENGTH) {
```

(no need for the extra `()`)

common/models/user.js
```diff
@@ -247,6 +247,12 @@ module.exports = function(User) {
             debug('An error is reported from User.hasPassword: %j', err);
             fn(defaultError);
           } else if (isMatch) {
+            if ((credentials.password).length > MAX_PASSWORD_LENGTH) {
```
bajtos: I think this validation should be called much earlier, just before calling `self.findOne`.

common/models/user.js
```diff
@@ -247,6 +247,12 @@ module.exports = function(User) {
...
+              err = new Error(g.f('login failed as the password entered is too long'));
```
bajtos: Please use `defaultError` here, so that we don't disclose any extra information that could be potentially used by an attacker.

common/models/user.js
bajtos: Actually, I take that back. Let's keep the validation here, to keep the response time roughly the same regardless of whether the password is less or more than MAX_PASSWORD_LENGTH long.

However, please add a comment explaining why it's important to keep the check here. For example:

```js
// Only check password length after we run the same code we would execute
// for a password with a valid length. This should prevent attackers from guessing
// the maximum allowed password length by measuring the response time.
```

common/models/user.js
```diff
@@ -558,6 +563,12 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    if (options.password && (options.password.length > MAX_PASSWORD_LENGTH)) {
```
bajtos: nitpick:

```js
if (options.password && options.password.length > MAX_PASSWORD_LENGTH) {
```

common/models/user.js
```diff
@@ -558,6 +563,12 @@ module.exports = function(User) {
...
+      var err = new Error(g.f('Password cannot exceed %j characters', MAX_PASSWORD_LENGTH));
+      err.statusCode = 400;
```
bajtos: please add `err.code`, e.g. `err.code = 'PASSWORD_TOO_LONG'`.

common/models/user.js
```diff
@@ -596,14 +607,18 @@ module.exports = function(User) {
   };
 
   User.validatePassword = function(plain) {
-    if (typeof plain === 'string' && plain) {
+    if (typeof plain === 'string' && plain && plain.length <= MAX_PASSWORD_LENGTH) {
       return true;
     }
+    if (plain.length > MAX_PASSWORD_LENGTH) {
+      err = new Error (g.f('Password too long - maximum length is %s', MAX_PASSWORD_LENGTH));
```
bajtos: Don't leak this information. Since the system does not allow passwords to be longer than 72 characters, any password longer than 72 characters is invalid.

test/user.test.js
bajtos: +1

If you want to assert that `User.create` throws, which I find rather weird, to be honest, because `User.create` should report errors via callback, then use [should#throws](https://shouldjs.github.io/#should-throws). Also verify that an expected error was thrown, and not something like `TypeError: undefined is not a function`.

test/user.test.js
bajtos: This is a problem of `bcrypt` we are trying to fix/workaround in this patch. The library looks at the first 72 chars only and discards the rest.

bajtos:
I am concerned about the backwards-incompatibility introduced in this patch. Because we refuse passwords longer than 72 characters at login time, if there is any existing user with such a long password, they won't be able to login any more after the server is upgraded to the newer LoopBack version.

Here are few options we have:

 - Make the breaking change, assume there are no users with so long passwords.

 - Add some sort of a flag to distinguish between password that were created after 72 char validation was added (and thus are 72 chars at most) and the legacy passwords that we created before. Then at login time, we can reject too long values for accounts with the flag set, but allow them for legacy accounts. When a user resets their password, the new entry will be validated for maxlen and created with the flag, thus their account will be "upgraded" to the new, more secure mode.

@loay @superkhau @raymondfeng Thoughts? Can you see any other option?

loay:
@bajtos @raymondfeng @superkhau 
I believe a breaking change would be the best option. Even if we add a flag, the users with longer passwords before introducing the validation are still subject to vulnerabilities.

common/models/user.js
loay: That was a mistake. The error is supposed to print ```('Password too long - maximum length is %s', plain)``` .. not sure why I put the length there !

test/user.test.js
superkhau: I meant the sentence for the description needs improvement to clarify "72 chars + extra added chars" means to someone who doesn't know the situation. Either add a comment or improve the description, I vote for the latter.

test/user.test.js
loay: Yes, I want to ensure that `User.Create` will throw the error, not `User.validatePassword` because the user should be blocked from even creating a user when the password is not valid.

superkhau:
@bajtos I agree with your persective on the breaking chagne.

>Make the breaking change, assume there are no users with so long passwords.

IMO, this is unacceptable from the end user perspective.

>Add some sort of a flag to distinguish between password that were created after 72 char validation was added (and thus are 72 chars at most) and the legacy passwords that we created before. Then at login time, we can reject too long values for accounts with the flag set, but allow them for legacy accounts. When a user resets their password, the new entry will be validated for maxlen and created with the flag, thus their account will be "upgraded" to the new, more secure mode.

This way provides a much smoother transition path, so I'm voting for this one.

That said, the only other option I can think of is adding a flag that turns on the security feature and let the users decide when to turn the flag on. Then allow users to implement their own mechanism to warn users to reset their passwords. This part will probably need to be announced via blog/mailing list though. So it's similar to what you suggest above, don't don't do it automatically for them -- ie. you gotta get your users to reset their password during login or reset them after a certain date before they choose to turn the flag on.

test/user.test.js
loay: I have tried few things here without using `try/catch`. I had to dig a bit more in `dao.js` Create function to see where it is not going but still couldn't find it. 
If I try anything other than try/catch, then the error is thrown: 
```
Error: Password too long:
```
The test works fine with try/catch, as when I put the proper password `pass72`, it fails.

test/user.test.js
```diff
@@ -205,6 +205,79 @@ describe('User', function() {
...
+          done();
+        }
+      });
+
+      it('rejects a new user with password longer than 72 characters', function(done) {
+        try {
+          var u = new User({ username: 'foo', password: pass73 });
+          assert(false, 'Error should have been thrown');
+        } catch (e) {
+          done();
```
bajtos: Please verify that `e` is the error you are expecting, and not something like `TypeError: undefined is not a function`.

test/user.test.js
```diff
@@ -205,6 +205,79 @@ describe('User', function() {
...
+          User.findById(user.id, function(err, userFound)  {
+            if (err) return done (err);
+            assert(userFound);
+            done();
+          });
+        });
+      });
+
+      it('allows login with password exactly 72 characters long', function(done) {
+        User.create({ email: 'b@c.com', password: pass72 }, function(err) {
```
bajtos: unhandled `err`

test/user.test.js
```diff
@@ -205,6 +205,79 @@ describe('User', function() {
...
+          if (err) return done(err);
+          User.findById(user.id, function(err, userFound)  {
+            if (err) return done (err);
+            assert(userFound);
+            done();
+          });
+        });
+      });
+
+      it('allows login with password exactly 72 characters long', function(done) {
```
bajtos: This test is nested under `describe('User.create')` right now. Please move it to `describe('User.login')` or perhaps move `describe('Password length validation')` to higher in the hierarchy, right under the top-level `describe('User')`.

test/user.test.js
bajtos: On an unrelated note: This test is nested under `describe('User.create')` right now. Please move it to `describe('User.login')` or perhaps move `describe('Password length validation')` to higher in the hierarchy, right under the top-level `describe('User')`.

test/user.test.js
```diff
@@ -205,6 +205,79 @@ describe('User', function() {
...
+          User.login({ email: 'b@c.com', password: badPass }, function(err) {
+            assert(err && !/verified/.test(err.message),
+                'expecting "login failed" error message, received: "' + err.message + '"');
+            assert.equal(err.code, 'LOGIN_FAILED');
+            done();
+          });
+        });
+      });
+
+      it('rejects password reset when password is more than 72 chars', function(done) {
```
bajtos: This test is nested under `describe('User.create')` right now. Please move it to `describe('User.resetPassword')` or perhaps move `describe('Password length validation')` to higher in the hierarchy, right under the top-level `describe('User')`.

bajtos:
> > Add some sort of a flag to distinguish between password that were created after 72 char validation was added (and thus are 72 chars at most) and the legacy passwords that we created before. Then at login time, we can reject too long values for accounts with the flag set, but allow them for legacy accounts. When a user resets their password, the new entry will be validated for maxlen and created with the flag, thus their account will be "upgraded" to the new, more secure mode.

> This way provides a much smoother transition path, so I'm voting for this one.

I went ahead and made the change myself, see 8f269a0

@loay please `git pull` in your feature branch before continuing your work.

common/models/user.js
```diff
@@ -336,8 +330,24 @@ module.exports = function(User) {
...
+      bcrypt.compare(plain, hashed, function(err, isMatch) {
         if (err) return fn(err);
+
+        // Only check password length after we run the same code we would execute
+        // for a password with a valid length. This should prevent attackers from guessing
+        // the maximum allowed password length by measuring the response time.
+        if (plain.length > MAX_PASSWORD_LENGTH && lengthWasValidated) {
+          // Reject too long password only if we are sure that the real
+          // password is at most 72 chars long.
+          fn(null, false);
```
bajtos: Ha, this will call `fn` twice, should be `return fn(null, false)`.

@loay I'll leave it up to you to fix this.

test/user.test.js
```diff
@@ -257,8 +257,9 @@ describe('User', function() {
         User.create({ email: 'b@c.com', password: pass72 }, function(err) {
           if (err) return done (err);
           User.login({ email: 'b@c.com', password: badPass }, function(err) {
-            assert(err && !/verified/.test(err.message),
```
bajtos: FWIW, `err.message` was evaluated before lazy evaluation of `&&` kicked in. When `err` was null, the test failed with a wrong error.

test/user.test.js
loay: Done.
FWIW, the error thrown: `Error: Password too long: password`

loay:
I moved describe('Password length validation') to higher in the hierarchy, right under the top-level describe('User').

loay:
@bajtos @superkhau Updated code. PTAL. Thanks.

common/models/user.js
bajtos: @loay ^^^ PTAL

common/models/user.js
```diff
@@ -558,6 +573,13 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    if (options.password && options.password.length > MAX_PASSWORD_LENGTH) {
```
bajtos: I think we should call `User.validatePassword`, so that we don't duplicate the validation logic.

test/user.test.js
```diff
@@ -363,6 +363,82 @@ describe('User', function() {
...
+        User.login({ email: 'b@c.com', password: pass72Char }, function(err, accessToken) {
+          if (err) return done(err);
+          assertGoodToken(accessToken);
+          assert(accessToken.id);
+          done();
+        });
+      });
+    });
+
+    it('rejects login when password is proper pass of 72 chars + extra added chars',
```
bajtos: Please add a comment explaining why are we testing those extra added chars, as discussed before.

bajtos:
@loay two more commits to address. Please check why CI is failing on Windows machines.

@superkhau PTAL too, to make sure I did not miss anything important.

common/models/user.js
loay: Sorry I didn't see this comment. I will have a look.

loay:
@slnode test please

common/models/user.js
loay: Done

test/user.test.js

loay:
@slnode test please

common/models/user.js
```diff
@@ -592,18 +609,29 @@ module.exports = function(User) {
...
+    return hash;
   };
 
   User.validatePassword = function(plain) {
-    if (typeof plain === 'string' && plain) {
+    if (plain && typeof plain === 'string' && plain.length <= MAX_PASSWORD_LENGTH) {
       return true;
     }
+    if (plain.length > MAX_PASSWORD_LENGTH) {
+      err = new Error (g.f('Password too long: %s', plain));
```
superkhau: var err

superkhau: Actually we use the same statusCode for both, we should move it both out.

```
var err;
if (plain.length > MAX_PASSWORD_LENGTH) {
  err = new Error (g.f('Password too long: %s', plain));
  err.code = 'PASSWORD_TOO_LONG';
} else {
  err = new Error(g.f('Invalid password: %s', plain));
  err.code = 'INVALID_PASSWORD'; // maybe add this
}
err.statusCode = 422;
throw err;
```

test/user.test.js
```diff
@@ -363,6 +363,88 @@ describe('User', function() {
...
+
+    it('rejects passwords longer than 72 characters', function(done) {
+      try {
+        User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
+          if (err) return done (err);
+          expect(err).to.have.property('message', /Password too long/);
+        });
+      } catch (e) {
+        done();
+      }
```
superkhau: Something is still weird here. If we expect error in the callback API, why do we need try catch. We talked about this over hangout last time @loay. IMO, we should be try/catching before we return the response to the user (they should only have to handle the `err` returned in the callback. If you decide the try/catch is valid, then we should at least have an assertion in 378 to ensure when the error is throw that we assert it's the right type of error being caught -- `assert(e, undefined, 'some-fail-message', '=')`.

@bajtos Thoughts?

superkhau: Actually forget the second part of my answer, I completely disagree with having to use try/catch at all with `User.create` whatsoever.

test/user.test.js
```diff
@@ -363,6 +363,88 @@ describe('User', function() {
...
+    });
+
+    it('rejects a new user with password longer than 72 characters', function(done) {
+      try {
+        var u = new User({ username: 'foo', password: pass73Char });
+        assert(false, 'Error should have been thrown');
+      } catch (e) {
+        expect(e).to.match(/Error: Password too long/);
```
superkhau: Just need `/Password too long/`

test/user.test.js
```diff
@@ -363,6 +363,88 @@ describe('User', function() {
...
+        if (err) return done (err);
+        try {
+          User.resetPassword({ email: 'b@c.com', password: pass73Char }, function(err) {
+            assert(err);
+            assert.equal(err.statusCode, 422);
+          });
+        } catch (e) {
+          expect(e).to.match(/Error: Password too long/);
+          done();
+        }
```
superkhau: ditto on the try/catch

test/user.test.js
```diff
@@ -363,6 +363,88 @@ describe('User', function() {
     });
   });
 
+  describe('Password length validation', function() {
+    var pass72Char = new Array(73).join('a');
+    var pass73Char = new Array(74).join('a');
+    var passTooLong = pass72Char + 'WXYZ1234';
```
superkhau: should be consistent with the other 2. `new Array(80).join('a');` or whatever length you prefer.

test/user.test.js
```diff
@@ -678,6 +760,22 @@ describe('User', function() {
           done();
         });
     });
+
+    it('allows login with password too long but created in old LB version',
+    function(done) {
+      var bcrypt = require('bcryptjs');
+      var longPassword = new Array(80).join('x');
```
superkhau: nit: be consistent with repeating 'a' or 'x', but not both.

test/user.test.js
```diff
@@ -678,6 +760,22 @@ describe('User', function() {
...
+      var oldHash = bcrypt.hashSync(longPassword, bcrypt.genSaltSync(1));
+
+      User.create({ email: 'b@c.com', password: oldHash }, function(err) {
+        if (err) return done(err);
+        User.login({ email: 'b@c.com', password: longPassword }, function(err) {
+          if (err) return done(err);
+          // we are logged in, the test passed
```
superkhau: Please assert the user is logged in instead of adding a comment (ie. verify the token, etc).

superkhau:
@loay Added more comments and a few nits.

@bajtos Why do we allow try/catching the callback APIs? I find it really strange how we don't just return the error in the callback accordingly.

common/models/user.js
loay: yeah I should have refactored that !

test/user.test.js
loay: As we tried together, nothing works except the `try/catch` .. Also, the only way this function is tested elsewhere is also using `try/catch`. If I had to remove that, I would have to dig more in `dao.js` and disect the function `Create`, which is not the goal of this pr because focus would be lost. The test case works and the proper error message is obtained. 

@bajtos @raymondfeng Any thoughts?

test/user.test.js
loay: basically we need the first 72 chars to be identical with passTooLong, and `Array(80)` would still fulfill that because the word is all `a` but I wanted something different starting at the 73rd index so we can show how the test fails.

test/user.test.js
superkhau: @loay I walked through it a bit with you and said dig deeper, not give up and use try/catch as the solution. I find it hard to believe we must throw try/catch for a function that has a callback API. It is probably ok to move forward since you need this merged in ASAP and the previous behaviour also threw for the validation. Anyways, let's get @bajtos to chime in here.

bajtos: 1) IIRC, the password validation throws an error. As a result, calling `User.create` may throw an error. I agree that's weird and it would be better to fix it. However, I think it's ok to leave this out of scope of this patch.

2) The test is still weird though. The callback function is non-sense, it short-circuits if `err` is set and then asserts `err` has some properties. The `catch` block is wrong too, because it accepts any error as test passing, even `TypeError: undefined is not a function`.

loay: I added ``` expect(e).to.match(/Password too long/); ``` within the catch block and it works fine. I will update the next commit

raymondfeng:
> Make the breaking change, assume there are no users with so long passwords.

IMO, this assumption is reasonable.

test/user.test.js
```diff
@@ -363,6 +363,89 @@ describe('User', function() {
...
+    var passTooLong = pass72Char + 'WXYZ1234';
+
+    it('rejects passwords longer than 72 characters', function(done) {
+      try {
+        User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
+          if (err) return done (err);
+          expect(err).to.have.property('message', /Password too long/);
+        });
+      } catch (e) {
+        expect(e).to.match(/Password too long/);
```
raymondfeng: Is `Password too long` reported via callback or `throw`? There should be exclusive.

common/models/user.js
```diff
@@ -557,6 +571,9 @@ module.exports = function(User) {
       cb(err);
       return cb.promise;
     }
+    if (options.password && options.password.length > MAX_PASSWORD_LENGTH) {
+      this.validatePassword(options.password);
```
raymondfeng: We should always call `validatePasspord`. Please note the extended user model can override `validatePassword` to enforce more rules.

raymondfeng: I think we should try to catch the validation error and report it via callback.

superkhau: This is exactly what I'm talking about to prevent what weird try/catch on the User.create.

test/user.test.js
superkhau: The issue is validatePassword is throwing and it's bubbling up. However, we should be try/catching internally and returning the error via callback so it's transparent to the end user.

common/models/user.js
```diff
@@ -592,18 +609,30 @@ module.exports = function(User) {
   User.hashPassword = function(plain) {
     this.validatePassword(plain);
     var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
-    return bcrypt.hashSync(plain, salt);
+    var hash = bcrypt.hashSync(plain, salt);
+    if (/^\$2a\$/.test(hash)) {
+      // change the prefix to indicate that we have verified the plain password
+      // is at most 72 chars long
+      hash = '$72$'.concat(hash.slice(4));
```
raymondfeng: I don't understand why we want to change the hash. It's invasive.

common/models/user.js
bajtos: +1 for always calling `validatePassword` and wrapping it in try/catch

common/models/user.js
bajtos: @raymondfeng see https://github.com/strongloop/loopback/pull/2580#issuecomment-238186402:

> I am concerned about the backwards-incompatibility introduced in this patch. Because we refuse passwords longer than 72 characters at login time, if there is any existing user with such a long password, they won't be able to login any more after the server is upgraded to the newer LoopBack version.

In https://github.com/strongloop/loopback/pull/2580/commits/bfe674ce2097592fb30cce94f3a87eb84f0ff936, I implemented my proposal:

> Add some sort of a flag to distinguish between password that were created after 72 char validation was added (and thus are 72 chars at most) and the legacy passwords that we created before. Then at login time, we can reject too long values for accounts with the flag set, but allow them for legacy accounts. When a user resets their password, the new entry will be validated for maxlen and created with the flag, thus their account will be "upgraded" to the new, more secure mode.

There is one important constraint here: we cannot add new properties (fields) to the User model, because that would break existing applications upgrading to a newer LoopBack version - the database schema of their servers will not contain the newly added field.

In which case we have to encode the flag in the existing user fields. Encoding it in the password field makes most sense to me.

I decided to modify the hash instead of prepending/appending values to the hash to prevent another possible source of a breaking changes: what if users created their database columns as at most 60 chars long, because they new bcrypt hashes are always 60 chars long?

Having wrote that, I am open to better suggestions on how to solve this problem, as long as they preserve backwards compatibility.

bajtos: @raymondfeng I see you commented

> >  Make the breaking change, assume there are no users with so long passwords.

> IMO, this assumption is reasonable.

I personally prefer to be careful about assumptions when making breaking changes, but I don't have any data points to objectively decide here.

Considering that both you @raymondfeng and @superkhau think it's ok to make the change breaking, I'll bow to the majority, so to speak :)

In that case we can revert  bfe674c completely (remove it from this patch) and call this discussion closed.

test/user.test.js
bajtos: +1 for *something different starting at the 73rd index so we can show how the test fails.*

```js
var pass72Char = new Array(70).join('a') + '012';
var pass73Char = pass72Char + '3';
var passTooLong = pass73Char + '456789';
```

test/user.test.js
```diff
@@ -363,6 +363,89 @@ describe('User', function() {
...
 
+  describe('Password length validation', function() {
+    var pass72Char = new Array(73).join('a');
+    var pass73Char = new Array(74).join('a');
+    var passTooLong = pass72Char + 'WXYZ1234';
+
+    it('rejects passwords longer than 72 characters', function(done) {
+      try {
+        User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
+          if (err) return done (err);
```
bajtos: If we decide to keep the current implementation, where invalid passwords throw an error, then in this test you should call

```js
done(new Error('User.create() should have thrown an error.'));
```

test/user.test.js
```diff
@@ -363,6 +363,89 @@ describe('User', function() {
...
+        });
+      });
+    });
+
+    it('rejects password reset when password is more than 72 chars', function(done) {
+      User.create({ email: 'b@c.com', password: pass72Char }, function(err) {
+        if (err) return done (err);
+        try {
+          User.resetPassword({ email: 'b@c.com', password: pass73Char }, function(err) {
+            assert(err);
```
bajtos: If we decide to keep the current implementation, where invalid passwords throw an error, then in this test you should call

```js
done(new Error('User.resetPassword() should have thrown an error.'));
```

bajtos:
@loay IMO, here are the steps remaining steps needed:

0) rebase on top of the current master

1) https://github.com/strongloop/loopback/pull/2580/files#r74517648
> We should always call validatePasspord. Please note the extended user model can override validatePassword to enforce more rules.

2) https://github.com/strongloop/loopback/pull/2580/files#r74518751
> In that case we can revert bfe674c completely (remove it from this patch)

Just make sure the final commit message mentions the breaking behaviour for password too long.

3) Now depending on whether we want to fix error handling in this pull request or not, you need to either:

3a) Fix user methods to catch errors thrown by `validatePassword'
or
3b) Fix the tests, as I commented in https://github.com/strongloop/loopback/pull/2580/files#r74598530 and https://github.com/strongloop/loopback/pull/2580/files#r74598777

@raymondfeng @superkhau is there anything else?

common/models/user.js
bajtos: On the second thought, this patch is already taking way too long to complete. Let's move the try/catch wrapping out of scope of this work, so that we can get this fix landed soon.

common/models/user.js
raymondfeng: Does the proposed fix reject passwords longer than 72 chars during login? I thought we only validate passwords for `User.create` and `User.resetPassword`.

raymondfeng:
I had a conversation with @bajtos and agreed to simplify the PR as follows:

1. We don’t change the behavior for login (even the user already has a long password than 72 chars). The risk of accepting a longer password over 72 chars is very low, even lower than guessing a regular password

2. The PR just have to validatePassword to make sure no future passwords can go beyond 72 chars. The enforcement should be added for create/reset only.

common/models/user.js
```diff
@@ -558,6 +557,10 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    if (options.password) {
```
raymondfeng: What is the `options.password` for?

common/models/user.js
```diff
@@ -558,6 +557,10 @@ module.exports = function(User) {
...
+      UserModel.validatePassword(options.password);
```
raymondfeng: We should try/catch and callback with err if it happens.

common/models/user.js
```diff
@@ -558,6 +557,13 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    try {
+      if (options.password) {
+        UserModel.validatePassword(options.password);
```
loay: I have to use `options.password` because otherwise `length` will not be defined and will break few other test cases and cause errors. I mean `length` in main `validatePassword` function.

raymondfeng: Got it. The new password is passed in as `options.password`.

raymondfeng:
LGTM

loay:
@slnode  test please

loay:
@slnode test please

common/models/user.js
bajtos: I don't think that's true. `resetMethod` does not touch the password, it only creates an access token. See also the API docs.

>```
    * Create a short lived acess token for temporary login. Allows users
    * to change passwords if forgotten.
    *
    * @options {Object} options
    * @prop {String} email The user's email address
    * @callback {Function} callback
    * @param {Error} err
    * @promise
```

Let's not worry about that right now and fix it as part of https://github.com/strongloop/loopback/issues/382.

Comment (bajtos) [inline common/models/user.js]: Here is the documentation explaining how to implement password reset: https://github.com/strongloop/loopback-example-user-management#how-do-you-perform-a-password-reset-for-a-registered-user

Gist - the password is updated via `user.updateAttribute`, which calls `user.updateAttributes` under the hood IIRC.

```js
    User.findById(req.accessToken.userId, function(err, user) {
      if (err) return res.sendStatus(404);
      user.updateAttribute('password', req.body.password, function(err, user) {
```

PR #2577, opened 2016-08-02
Update dependencies to their latest versions

bajtos:
cc @strongloop/loopback-dev FYI, I think no review is needed

PR #2585, opened 2016-08-05
Update dependencies to their latest versions and fix CI failures [2.x]

bajtos:
Back-port of #2577

cc @jannyHou


bajtos:
All LoopBack jobs are green, I am going to land this patch.

PR #2604, opened 2016-08-10
Fix token middleware to not trigger CLS init

bajtos:
Rework the token middleware to access current context via `req.loopbackContext` instead of `loopback.getCurentContext()`. That way the CLS/AsyncListener machinery is configured only in applications that are using current context.

This is in line with what we are doing in master, see https://github.com/strongloop/loopback/blob/32ecd2fc5c93325e41ca41a3f232db4a3f40bd10/server/middleware/token.js#L127

Connect to strongloop-internal/scrum-loopback#989

@raymondfeng @ritch PTAL

### DIFF
--- a/server/middleware/token.js
+++ b/server/middleware/token.js
@@ -124,7 +124,7 @@ function token(options) {
     TokenModel.findForRequest(req, options, function(err, token) {
       req.accessToken = token || null;
       rewriteUserLiteral(req, currentUserLiteral);
-      var ctx = loopback.getCurrentContext();
+      var ctx = req.loopbackContext;
       if (ctx) ctx.set('accessToken', token);
       next(err);
     });


ritch:
Can you link to the PR that implements `req.loopbackContext`? This PR makes it sound like that is already in master.

If so, this LGTM.

bajtos:
> Can you link to the PR that implements req.loopbackContext? This PR makes it sound like that is already in master.

`req.loopbackContext` is set by loopback-context, see https://github.com/strongloop/loopback-context/blob/268716f69eec246b4ae0f7e402147b6fcbc924cf/server/middleware/per-request.js#L42

Here is the pull request against master that contains the change I am proposing in this patch for 2.x:

https://github.com/strongloop/loopback/pull/2564

pulkitsinghal:
Hey @bajtos

1. I really like the fact that you've started `https://github.com/strongloop/loopback-context` repo to make everythign easier/better.

1. Also appreciate the fact that you:

  > changed session/memory to use process.nextTick instead of setImmediate

  even though:
  > There may be something else in play, because node-continuation-local-storage claims to support setImmediate according to their docs 
1. but I don't understand why issues like this one are working so hard to wire `loopback-context` into the heart of `loopback` while this huge warning sign is still up on the README:

  > We recommend AGAINST using the loopback-context module until there is a stable solution to the issue below! ... see e.g. issue #59.
1. and op top of that, this is going on too: `Deprecate getCurrentContext()` #1676

bajtos:
@pulkitsinghal 

> I really like the fact that you've started `https://github.com/strongloop/loopback-context` repo to make everything easier/better.

Thank you for the appreciation :)

> Also appreciate the fact that you:

> >    changed session/memory to use process.nextTick instead of setImmediate

> even though:

> >  There may be something else in play, because node-continuation-local-storage claims to support setImmediate according to their docs

I made that change only locally, to verify my assumptions. I didn't submit any patch to make that change permanent.

> Please help me understand the vision.

Backwards compatibility. We are in a difficult situation - at one hand, we know that continuation-local-storage (and thus loopback-context too) is broken and cannot be reliably fixed in general. At the same time, there may be many users that are depending on loopback-context and for whom it may work, perhaps because they use case does not trigger any of the bugs.

My vision is to make it easy for users that wish to continue using loopback-context to do so, while also making it clear that using loopback-context is not recommended, to prevent new users from using this dangerous concept. Eventually, I would like to stop supporting loopback-context completely, but I am afraid we are not there yet.

PR #2621, opened 2016-08-15
Revert globalization of Swagger descriptions [2.x]

bajtos:
As discussed in SlackChat, Swagger descriptions should not be globalized. Translating Swagger is an open problem, see e.g. https://github.com/OAI/OpenAPI-Specification/issues/274

@0candy please review

PR #2622, opened 2016-08-15
Revert globalization of Swagger descriptions

bajtos:
This is a forward-port of #2621:

> As discussed in SlackChat, Swagger descriptions should not be globalized. Translating Swagger is an open problem, see e.g. OAI/OpenAPI-Specification#274

@0candy please review


0candy:
LGTM. There is one more, if you can revert line too. Thanks!
https://github.com/strongloop/loopback/blob/master/example/mobile-models/app.js#L46

PR #2628, opened 2016-08-15
resolve related models from correct registry

### LINKED ISSUE #2627: resolvers for dynamic roles not found when using app option `localRegistry`
related to #2551 

I'm using my app with the `localRegistry` and `loadBuiltinModels` options set to `true`. This is required when using `loopback#token` middleware and want to provide a custom `AccessToken` modelName to it's params.

```
const app =  loopback({ localRegistry: true, loadBuiltinModels: true }); 
```

However, I started applying ACLs (with **dynamic** roles) to my models and calling the REST API.
As it seemed like my ACLs are not being processed, I started digging and discovered that ACLs defined in a `model-definition.json` are not found by `ACL.getStaticACLs()` since it would lookup the `modelName` in the global registry `loopback.findModel(modelName)`. Luckily, @bajtos already solved that in #2551 (which hasn't been published in a 2.x release yet and I'm applying this patch manually).

This leads me to my actual problem: 
I'm using `app` with `local Registry` and `builtInModels` (no custom `Role` model).

Then I define a dynamic role named 'admin':

```
const { registry } = app;
const Role = registry.getModel('Role');
Role.registerResolver('admin', (role, context, cb) => {
    process.nextTick(() => {
      console.log('in admin role resolver');
      cb(null, false);
    });
  });
```

In one of my models, I define the following ACLs:

```
"acls": [{
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    }, {
      "accessType": "READ",
      "principalType": "ROLE",
      "principalId": "admin",
      "permission": "ALLOW",
      "property": "find"
    }],
```

When calling the REST API, somewhere down the line `ACL.checkAccessForContext()` gets called. 
In `common/models/acl.js` you can find this:

```
var Role = loopback.Role;
...
    ...
    var roleModel = registry.getModelByType(Role);
    ...
    // this happens for each previously resolved alc
    roleModel.isInRole(...);
    ...
}
...
```

Due to the scenario described above (no custom models with `Role` as baseModel), `registry.getModelByTpe(Role)` returns the provided argument as-is, which is the global `loopback.Role`

When now calling `roleModel.isInRole('admin')`, the 'admin' resolver is not recognized since it was registered on the local registry of our app instance. `isInRole()` proceeds to calling `this.findOne()` in order to try to resolve a role named 'admin' from the attached dataSource and ultimately the application *crashes* since this global `loopback.Role` has never been attached to a dataSouce

I found that, when modifying `ACL.checkAccessForContext()` to read the built-in model from it's registry and use that with `registry.getModelByType()`, my ACL works as expected.

```
    var registry = this.registry;
    var builtInRoleModel = registry.getModel('Role');
    ...
    var roleModel = registry.getModelByType(builtInRoleModel);
    ...
}
```

Does anyone have any objections to providing this in a PR? Am I using the localRegistry wrongly?

benkroeger:
Fixes #2627 

@bajtos 
I reused a helper-method `resolveRelatedModels()` that had been added by someone else before. Helped me figure out that there are other related models that would be resolved from the wrong registry (`RoleMapping`, `User`, `Application`).

Had to modify setup of test servers when ACL was used; forcing to `loadBuiltinModels` with localRegistry. Alternative would be to list all related Models (e.g. `Application`) in the server's `model-config.json`

That said, it is required that related models are added  to the same registry. For built-in models, that can happen via the `loadBuiltinModels` option; others should be added to `model-config.json`

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -506,19 +506,12 @@ module.exports = function(ACL) {
   };
 
   ACL.resolveRelatedModels = function() {
-    var self = this;
-    if (!self.roleModel) {
-      var reg = self.registry;
-      var builtInModels = {
-        roleModel: reg.getModel('Role'),
-        roleMappingModel: reg.getModel('RoleMapping'),
-        userModel: reg.getModel('User'),
-        applicationModel: reg.getModel('Application'),
-      };
-
-      for (var m in builtInModels) {
-        self[m] = reg.getModelByType(builtInModels[m]);
-      }
+    if (!this.roleModel) {
+      var reg = this.registry;
+      this.roleModel = reg.getModelByType('Role');
+      this.roleMappingModel = reg.getModelByType('RoleMapping');
+      this.userModel = reg.getModelByType('User');
+      this.applicationModel = reg.getModelByType('Application');
     }
   };
 
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -395,8 +395,9 @@ module.exports = function(ACL) {
    */
 
   ACL.checkAccessForContext = function(context, callback) {
-    this.resolveRelatedModels();
-    var roleModel = this.roleModel;
+    var self = this;
+    self.resolveRelatedModels();
+    var roleModel = self.roleModel;
 
     if (!(context instanceof AccessContext)) {
       context = new AccessContext(context);
@@ -419,10 +420,9 @@ module.exports = function(ACL) {
     var req = new AccessRequest(modelName, property, accessType, ACL.DEFAULT, methodNames);
 
     var effectiveACLs = [];
-    var staticACLs = this.getStaticACLs(model.modelName, property);
+    var staticACLs = self.getStaticACLs(model.modelName, property);
 
-    var self = this;
-    this.find({ where: { model: model.modelName, property: propertyQuery,
+    self.find({ where: { model: model.modelName, property: propertyQuery,
       accessType: accessTypeQuery }}, function(err, acls) {
       if (err) {
         if (callback) callback(err);
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -395,7 +395,8 @@ module.exports = function(ACL) {
    */
 
   ACL.checkAccessForContext = function(context, callback) {
-    var registry = this.registry;
+    this.resolveRelatedModels();
+    var roleModel = this.roleModel;
 
     if (!(context instanceof AccessContext)) {
       context = new AccessContext(context);
@@ -421,7 +422,6 @@ module.exports = function(ACL) {
     var staticACLs = this.getStaticACLs(model.modelName, property);
 
     var self = this;
-    var roleModel = registry.getModelByType(Role);
     this.find({ where: { model: model.modelName, property: propertyQuery,
       accessType: accessTypeQuery }}, function(err, acls) {
       if (err) {
@@ -506,12 +506,19 @@ module.exports = function(ACL) {
   };
 
   ACL.resolveRelatedModels = function() {
-    if (!this.roleModel) {
-      var reg = this.registry;
-      this.roleModel = reg.getModelByType(loopback.Role);
-      this.roleMappingModel = reg.getModelByType(loopback.RoleMapping);
-      this.userModel = reg.getModelByType(loopback.User);
-      this.applicationModel = reg.getModelByType(loopback.Application);
+    var self = this;
+    if (!self.roleModel) {
+      var reg = self.registry;
+      var builtInModels = {
+        roleModel: reg.getModel('Role'),
+        roleMappingModel: reg.getModel('RoleMapping'),
+        userModel: reg.getModel('User'),
+        applicationModel: reg.getModel('Application'),
+      };
+
+      for (var m in builtInModels) {
+        self[m] = reg.getModelByType(builtInModels[m]);
+      }
     }
   };
 
--- a/test/fixtures/access-control/server/server.js
+++ b/test/fixtures/access-control/server/server.js
@@ -5,7 +5,7 @@
 
 var loopback = require('../../../..');
 var boot = require('loopback-boot');
-var app = module.exports = loopback({ localRegistry: true });
+var app = module.exports = loopback({ localRegistry: true, loadBuiltinModels: true });
 var errorHandler = require('strong-error-handler');
 boot(app, __dirname);
 
--- a/test/fixtures/user-integration-app/server/server.js
+++ b/test/fixtures/user-integration-app/server/server.js
@@ -5,7 +5,7 @@
 
 var loopback = require('../../../../index');
 var boot = require('loopback-boot');
-var app = module.exports = loopback({ localRegistry: true });
+var app = module.exports = loopback({ localRegistry: true, loadBuiltinModels: true });
 var errorHandler = require('strong-error-handler');
 app.enableAuth();
 boot(app, __dirname);
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "express": "^4.14.0",
     "inflection": "^1.6.0",
     "loopback-connector-remote": "^1.0.3",
-    "loopback-datasource-juggler": "^3.0.0-alpha.1",
+    "loopback-datasource-juggler": "strongloop/loopback-datasource-juggler#b3907caad21f82eba1e82f005e84c68116eef052",
     "loopback-phase": "^1.2.0",
     "nodemailer": "^2.5.0",
     "nodemailer-stub-transport": "^1.0.0",
--- a/test/access-control.integration.js
+++ b/test/access-control.integration.js
@@ -142,6 +142,34 @@ describe('access control - integration', function() {
   });
 
   describe('/banks', function() {
+    var SPECIAL_USER = { email: 'special@test.test', password: 'test' };
+
+    // define dynamic role that would only grant access when the authenticated user's email is equal to
+    // SPECIAL_USER's email
+
+    before(function() {
+      var roleModel = app.registry.getModel('Role');
+      var userModel = app.registry.getModel('user');
+
+      roleModel.registerResolver('$dynamic-role', function(role, context, callback) {
+        if (!(context && context.accessToken && context.accessToken.userId)) {
+          return process.nextTick(function() {
+            callback && callback(null, false);
+          });
+        }
+        var accessToken = context.accessToken;
+        userModel.findById(accessToken.userId, function(err, user) {
+          if (err) {
+            return callback(err, false);
+          }
+          if (user && user.email === SPECIAL_USER.email) {
+            return callback(null, true);
+          }
+          return callback(null, false);
+        });
+      });
+    });
+
     lt.beforeEach.givenModel('bank');
 
     lt.it.shouldBeAllowedWhenCalledAnonymously('GET', '/api/banks');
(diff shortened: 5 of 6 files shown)


slnode:

slnode:

slnode:

slnode:

benkroeger:
did any of the dependant tests fail due to my changes? Looks like those are mostly errors within themselves.

test/fixtures/access-control/server/server.js
```diff
@@ -5,7 +5,7 @@
 
 var loopback = require('../../../..');
 var boot = require('loopback-boot');
-var app = module.exports = loopback({ localRegistry: true });
+var app = module.exports = loopback({ localRegistry: true, loadBuiltinModels: true });
```
bajtos: FWIW, this should not be necessary, loopback-boot loads builtin models as needed.

test/fixtures/user-integration-app/server/server.js
```diff
@@ -5,7 +5,7 @@
 
 var loopback = require('../../../../index');
 var boot = require('loopback-boot');
-var app = module.exports = loopback({ localRegistry: true });
+var app = module.exports = loopback({ localRegistry: true, loadBuiltinModels: true });
```
bajtos: ditto - this should not be necessary, loopback-boot loads builtin models as needed.

bajtos:
Hi @benkroeger, thank you for the pull request. The changes are much more involved than I expected :( Could you please try to keep the amount of changed lines as small as possible? For example, using `self` instead of `this` is in most places not necessary.

common/models/acl.js
```diff
@@ -506,12 +506,19 @@ module.exports = function(ACL) {
...
+      var reg = self.registry;
+      var builtInModels = {
+        roleModel: reg.getModel('Role'),
+        roleMappingModel: reg.getModel('RoleMapping'),
+        userModel: reg.getModel('User'),
+        applicationModel: reg.getModel('Application'),
+      };
+
+      for (var m in builtInModels) {
+        self[m] = reg.getModelByType(builtInModels[m]);
```
bajtos: `getModelByType` works for string names too, no need to call `reg.getModel` first.

See https://github.com/strongloop/loopback/blob/21ce174939843aef15dab28dad353a931e086c4f/lib/registry.js#L324-L326

common/models/acl.js
```diff
@@ -506,12 +506,19 @@ module.exports = function(ACL) {
   };
 
   ACL.resolveRelatedModels = function() {
-    if (!this.roleModel) {
-      var reg = this.registry;
-      this.roleModel = reg.getModelByType(loopback.Role);
-      this.roleMappingModel = reg.getModelByType(loopback.RoleMapping);
-      this.userModel = reg.getModelByType(loopback.User);
-      this.applicationModel = reg.getModelByType(loopback.Application);
```
bajtos: As I see the problem, all you need to fix in this patch is to change `loopback.MODEL` to `'MODEL'` in the lines above. Unless I am missing something?

test/fixtures/access-control/server/server.js
benkroeger: and why are tests failing without this... complaining that the builtin model `Application` is unknown?

common/models/acl.js
benkroeger: I know `getModelByType` works with strings, too... but we actually want to get those models that have the builtins as `base` model, hence resolve builtin model by `const builtInModel = getName('MODEL')` and then resolve potential custom model by `getModelByType(builtInModel)` .
The previous version used `getModelByType()` as well.

benkroeger:
@bajtos `self` was used in the 'original' function as well. All I did was to move the declaration of `self` to the top and then use it consistently (rather than mixing `this` and `self`)

test/fixtures/access-control/server/server.js
bajtos: > and why are tests failing without this... complaining that the builtin model Application is unknown?

Ah, interesting, my understanding of how this part of LoopBack works was wrong then. I assume the tests are failing because ACLs are resolving `Application` model, but they don't declare this dependency anywhere.

Let's go with what you are proposing here then. I'll delete the other "ditto" comment.

common/models/acl.js
bajtos: > ```js
const AccessToken = registry.getModelByType('AccessToken');
```
> it internally calls registry.getModel('AccessToken'), which would return the builtin AccessToken model.

That does not match my mental model of how this part works. `registry.getModelByType` calls `this.getModel` under the hood, which should be the very same `getModel` function you are calling here.

What am I missing?

**Comment by bajtos** [comment]
> @benkroeger  `self` was used in the 'original' function as well. All I did was to move the declaration of `self` to the top and then use it consistently (rather than mixing `this` and `self`)

I see your point. My problem is that changing `this` to `self` clutters the patch a lot and it's not obvious which changes are fixing the bug and which are just an unrelated cleanup.

That's why I am asking you to keep the amount of changes as small as possible.

Alternatively, move all `this->self` cleanup to a standalone commit.

File: common/models/acl.js (hunk 3 above)
Comment (benkroeger) [inline common/models/acl.js]: [the first call](https://github.com/strongloop/loopback/pull/2628#discussion-diff-74943161R513) is to resolve the builtin models (e.g. like the old code did by `loopback.Role`) but from the registry that `ACL` is registered to.
[the second call](https://github.com/strongloop/loopback/pull/2628#discussion-diff-74943161R520) is to resolve any custom model that might extend the builtins.
`getModelByType()` calls `getModel()` internally only when called with string name argument. Otherwise it compares registered model's prototypes with the model provided as argument

File: test/fixtures/access-control/server/server.js (hunk 1 above)
Comment (benkroeger) [inline test/fixtures/access-control/server/server.js]: thanks

**Comment by benkroeger** [comment]
rebased my PR and split the `self` related changes into a separate commit

File: common/models/acl.js (hunk 3 above)
Comment (bajtos) [inline common/models/acl.js]: In my understanding, calling `registry.getModelByType('AccessToken')` is the same as calling `registry.getModelByType(registry.getModel('AccessToken'))`. That's why I am asking you to use the former, as it's more concise.

**Comment by bajtos** [comment]
@benkroeger I am wondering, did you try to write a unit-test that would demonstrate the issue you are fixing here and prevent regressions in the future? I am generally reluctant to land changes not covered by tests, unless the test is way too difficult to write/maintain.

File: common/models/acl.js (hunk 3 above)
Comment (benkroeger) [inline common/models/acl.js]: sorry to say that, but your understanding is not correct.
`registry.getModel('AccessToken')` is the same as `registry.getModelByType('AccessToken')`, but it's not the same as `registry.getModelByType(AccessToken)` (where `AccessToken` is the return value of `registry.getModel('AccessToken')`

**Comment by benkroeger** [comment]
Wrote a unit-test just now and appended it to the access-control.integration `/banks` tests.

When running it without my modifications from this PR, 6 tests are failing with this:
```
1) access control - integration /banks when called by logged in user GET /api/banks "before each" hook:
     Uncaught Error: Cannot call Role.findOne(). The findOne method has not been setup. The PersistedModel has not been correctly attached to a DataSource!
    at throwNotAttached (lib/persisted-model.js:68:11)
    at Function.findOne (lib/persisted-model.js:264:5)
    at Function.Role.isInRole (common/models/role.js:332:10)
    at common/models/acl.js:450:23
    at node_modules/async/dist/async.js:3671:13
    at eachOfArrayLike (node_modules/async/dist/async.js:1002:13)
    at eachOf (node_modules/async/dist/async.js:1052:9)
    at _parallel (node_modules/async/dist/async.js:3670:9)
    at Object.parallelLimit [as parallel] (node_modules/async/dist/async.js:3750:7)
    at common/models/acl.js:461:13
    at node_modules/loopback-datasource-juggler/lib/dao.js:1769:9
    at node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:396:17
    at async.each (node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:153:20)
    at _asyncMap (node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:390:13)
    at Object.map (node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:361:23)
    at allCb (node_modules/loopback-datasource-juggler/lib/dao.js:1688:13)
    at node_modules/loopback-datasource-juggler/lib/connectors/memory.js:441:7
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

that is because `loopback.Role` has not been attached to a datasource... and the dynamic role is registered in `app.registry.getModel('Role')`

File: common/models/acl.js (hunk 3 above)
Comment (bajtos) [inline common/models/acl.js]: @benkroeger I am still confused. Here is the code in registry that makes me think that those two calls should return the same result:

```js
Registry.prototype.getModelByType = function(modelType) {
  var type = typeof modelType;
  var accepted = ['function', 'string'];

  assert(accepted.indexOf(type) > -1,
    'The model type must be a constructor or model name');

  if (type === 'string') {
    modelType = this.getModel(modelType);
  }

  var models = this.modelBuilder.models;
  for (var m in models) {
    if (models[m].prototype instanceof modelType) {
      return models[m];
    }
  }
  return modelType;
};
```

As you can see, if you call `getModelByType('AccessToken')`, the function calls `getModel('AccessToken')` first before searching for descendant models.

I have also modified your patch and run the test suite to verify my assumptions, all of the tests passed.

```diff
diff --git a/common/models/acl.js b/common/models/acl.js
index 6d413a1..cdf931b 100644
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -509,16 +509,10 @@ module.exports = function(ACL) {
     var self = this;
     if (!self.roleModel) {
       var reg = self.registry;
-      var builtInModels = {
-        roleModel: reg.getModel('Role'),
-        roleMappingModel: reg.getModel('RoleMapping'),
-        userModel: reg.getModel('User'),
-        applicationModel: reg.getModel('Application'),
-      };
-
-      for (var m in builtInModels) {
-        self[m] = reg.getModelByType(builtInModels[m]);
-      }
+      self.roleModel = reg.getModelByType('Role');
+      self.roleMappingModel = reg.getModelByType('RoleMapping');
+      self.userModel = reg.getModel('User');
+      self.applicationModel = reg.getModel('Application');
     }
   };
```

So my question is, what am I missing here?

bajtos:
> Wrote a unit-test just now and appended it to the access-control.integration /banks tests.

Great!

The only remaining item is our/my confusion around `registry.getModelByType`, see https://github.com/strongloop/loopback/pull/2628#discussion_r76037199

common/models/acl.js
benkroeger: my bad. I always read it as 
```
if (type === 'string') {
    return this.getModel(modelType);
}
```
and also mixed my assumptions with the previous implementation that used 
```
reg.getModelByType(loopback.Role);
```
which would return `loopback.Role` instead of `reg.getModel('Role')` model from because `loopback.Role` is neither of type `string` nor is there any model in `reg` that would have `loopback.Role` for it's prototype.

benkroeger:
@bajtos : I had to rebase my fork, caused me to include the latest merge for the `KeyValueModel`. That one does not finish `npm test` successfully --> causing my PR to fail as well. Figured out that the tests rely on a [feature](https://github.com/strongloop/loopback-datasource-juggler/pull/1049) in `loopback-datasource-juggler` that was added after the [latest v3.0.0-alpha.6 release](https://github.com/strongloop/loopback-datasource-juggler/releases/tag/v3.0.0-alpha.6).
I updated the dependency to a commit hash directly (since none of the released versions include the required feature) --> you should update that dependency once `loopback-datasource-juggler` releases a version incl. the required features

bajtos:
> I had to rebase my fork, caused me to include the latest merge for the KeyValueModel. That one does not finish npm test successfully --> causing my PR to fail as well. Figured out that the tests rely on a feature in loopback-datasource-juggler that was added after the latest v3.0.0-alpha.6 release.
> I updated the dependency to a commit hash directly (since none of the released versions include the required feature) --> you should update that dependency once loopback-datasource-juggler releases a version incl. the required features

That's not needed. When we run tests on our Jenkins CI, we are installing dependencies from our staging registry that contains the latest master/2.x versions. It's only the local machine and Travis CI that fails because of un-released features in dependencies.

I started with back-porting your patch to 2.x to see CI results for loopback & dependencies. I'll merge this patch to `master` after 2.x is ok and landed.

bajtos:

PR #2632, opened 2016-08-15
Update eslint infrastructure

loay:


package.json
```diff
@@ -62,7 +62,7 @@
     "chai": "^3.5.0",
     "cookie-parser": "^1.3.4",
     "es5-shim": "^4.1.0",
-    "eslint-config-loopback": "^1.0.0",
+    "eslint-config-loopback": "^4.0.0",
```
superkhau: We should add:

```
"lint": "eslint .",
"posttest": "npm run lint"
```

But please do this in a separate PR, this one is huge as is already.

superkhau:
LGTM, please check CI and make sure they are unrelated to your PR before merging.

package.json
loay: I thought we do not do that if the Gruntfile is taking care of it?

superkhau: For now it doesn't matter but I think we should move away from having a Gruntfile at all and doing it all via NPM scripts instead. As for your current PR, you don't need to do anything.

loay:
@bajtos @superkhau 
PTAL. There are some code changes in the actual code, in order to be conform with the upgrade, as well as the use strict. 

The main files that included code change and not just style:
 - common/models/application.js
 - common/models/role.js
 - test/key-value-model.test.js  --> replaced `connector` by `adapter`
 - Globalized variables in test/support.js

An interesting test case here: 

in loopback.test.js:

- expected properties are different in v6 and v4 than v0.10 and 0.12 !!!

bajtos:
Hey @loay, great stuff! Thank you for going through the pain of fixing linting. Could you please use recently released `eslint-config-loopback@5.0.0` (https://github.com/strongloop/eslint-config-loopback/pull/12) while you are at this?

common/models/application.js
```diff
@@ -71,13 +72,11 @@ function generateKey(hmacKey, algorithm, encoding) {
 
 module.exports = function(Application) {
   // Workaround for https://github.com/strongloop/loopback/issues/292
-  Application.definition.rawProperties.created.default =
```
bajtos: This change looks suspicious to me, could you please explain which `eslint` rule is violated by this line?

common/models/application.js
```diff
@@ -71,13 +72,11 @@ function generateKey(hmacKey, algorithm, encoding) {
...
   Application.definition.properties.created.default = function() {
     return new Date();
   };
 
   // Workaround for https://github.com/strongloop/loopback/issues/292
-  Application.definition.rawProperties.modified.default =
```
bajtos: This change looks suspicious to me, could you please explain which `eslint` rule is violated by this line?

common/models/email.js
```diff
@@ -15,9 +15,10 @@
  * @inherits {Model}
  */
 
-var g = require('../../lib/globalize');
+ 'use strict';
+ var g = require('../../lib/globalize');
 
-module.exports = function(Email) {
+ module.exports = function(Email) {
```
bajtos: The indentation changes in this file are suspicious. Why do we need to indent lines by one (!) more character? AFAIK, we are not doing that in any other file. PTAL.

common/models/key-value-model.js
```diff
@@ -1,13 +1,13 @@
...
  * Data model for key-value databases.
  *
  * @class KeyValueModel
  * @inherits {Model}
  */
+ 'use strict';
+ var g = require('../../lib/globalize');
 
-module.exports = function(KeyValueModel) {
+ module.exports = function(KeyValueModel) {
```
bajtos: The indentation changes in this file are suspicious. Why do we need to indent lines by one (!) more character? AFAIK, we are not doing that in any other file. PTAL.

common/models/email.js
```diff
@@ -15,9 +15,10 @@
  * @inherits {Model}
  */
 
-var g = require('../../lib/globalize');
+ 'use strict';
+ var g = require('../../lib/globalize');
```
bajtos: Hmm, I think the order of jsdoc, use-strict, require() and module.exports is wrong here. The correct order is:

```js
'use strict';

var g = require('../../lib/globalize');

/** 
 * Email model. etc.
 */

module.exports = function(Email) {
  // ...
};
```

common/models/key-value-model.js
```diff
@@ -1,13 +1,13 @@
-var g = require('../../lib/globalize');
```
bajtos: Ditto, the correct order is use-strict, require(), jsdoc, module.exports.

bajtos: Also while you are fixing this, please add a license header.

common/models/role.js
```diff
@@ -21,16 +22,14 @@ assert(RoleMapping, 'RoleMapping model must be defined before Role model');
  */
 module.exports = function(Role) {
   // Workaround for https://github.com/strongloop/loopback/issues/292
-  Role.definition.rawProperties.created.default =
```
bajtos: Again, why are we dropping `rawProperties` line?

common/models/role.js
```diff
@@ -21,16 +22,14 @@ assert(RoleMapping, 'RoleMapping model must be defined before Role model');
...
-    Role.definition.properties.created.default = function() {
-      return new Date();
-    };
+  Role.definition.properties.created.default = function() {
+    return new Date();
+  };
 
   // Workaround for https://github.com/strongloop/loopback/issues/292
-  Role.definition.rawProperties.modified.default =
```
bajtos: Ditto, why are we dropping `rawProperties` line?

common/models/role.js
```diff
@@ -97,7 +96,7 @@ module.exports = function(Role) {
       }
 
       roleModel.roleMappingModel.find({
-        where: { roleId: this.id, principalType: principalType },
+        where: {roleId: roleModel.roleMappingModel.id, principalType: principalType},
```
bajtos: I am afraid this is not a correct fix. I am landing https://github.com/strongloop/loopback/pull/2742 to fix this issue, please rebase on top of the new master after the change is merged.

package.json
```diff
@@ -64,7 +64,8 @@
     "chai": "^3.5.0",
     "cookie-parser": "^1.3.4",
     "es5-shim": "^4.1.0",
-    "eslint-config-loopback": "^1.0.0",
+    "eslint": "^2.13.1",
```
bajtos: Please remove, we are using `grunt-eslint` instead.

server/middleware/context.js
```diff
@@ -3,8 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+'use strict';
 var g = require('../../lib/globalize');
-
```
bajtos: Please keep the empty line serving as a visual delimiter.

test/remoting.integration.js
```diff
@@ -301,6 +303,7 @@ function findClass(name) {
 }
 
 function getFormattedMethodsExcludingRelations(methods) {
+  var result;
```
bajtos: Just remove `result = ` from the next line.

```js
return methods.filter(function(m) {
  // ...
});
```

test/remoting.integration.js
```diff
@@ -313,6 +316,7 @@ function getFormattedMethodsExcludingRelations(methods) {
 }
 
 function getFormattedScopeMethods(methods) {
+  var result;
```
bajtos: Ditto, `result` is not used and should be removed.

test/remoting.integration.js
```diff
@@ -325,6 +329,7 @@ function getFormattedScopeMethods(methods) {
 }
 
 function getFormattedPrototypeMethods(methods) {
+  var result;
```
bajtos: Ditto, `result` is not used and should be removed.

test/key-value-model.test.js
```diff
@@ -138,16 +139,16 @@ describe('KeyValueModel', function() {
...
     app.use(loopback.rest());
 
     CacheItem = app.registry.createModel({
       name: 'CacheItem',
       base: 'KeyValueModel',
     });
 
-    app.dataSource('kv', { connector: 'kv-memory' });
-    app.model(CacheItem, { dataSource: 'kv' });
+    app.dataSource('kv', {adapter: 'kv-memory'});
```
bajtos: I don't understand why is this change necessary, could you please explain?

The `connector` option is the canonical way of configuring data-source connector, I am not sure if `adapter` is even documented.

common/models/application.js
loay: it was giving this error:
```
>> Mocha exploded!
>> TypeError: Cannot create property 'default' on string 'date'
>>   at module.exports (/Users/loay/Desktop/NodeEE/loopback/common/models/application.js:75:56)
```

As you see, I removed the first line in `application.js` and `role.js` but I think the proper solution is in `application.json` and `role.json`

```
    "created": {
      "default": "date"
    },
    "modified": {
      "default": "date"
    }
```
I will update that in the next commit.

common/models/application.js
loay: Same answer as above https://github.com/strongloop/loopback/pull/2632#discussion_r88112866

common/models/role.js
loay: Same answer as above https://github.com/strongloop/loopback/pull/2632#discussion_r88112866

common/models/role.js
loay: Same answer as above https://github.com/strongloop/loopback/pull/2632#discussion_r88112866

common/models/role.js
loay: Thanks for landing this .. I will rebase and hopefully it will fix it in the next commit

test/remoting.integration.js
loay: Done

test/key-value-model.test.js
loay: I would get the following error:
```
  3) KeyValueModel "before each" hook: setupAppAndCacheItem for "provides "get(key)" at "GET /key"":
     TypeError: Cannot create data source "kv": Cannot create property 'name' on string 'kv-memory'
    at dataSourcesFromConfig (lib/application.js:427:29)
    at EventEmitter.app.dataSource (lib/application.js:226:14)
    at Context.setupAppAndCacheItem (test/key-value-model.test.js:150:9)
```

I did some research and it seems to be `adapter` is the equivalent and it passes

common/models/application.js
bajtos: > ```
>> Mocha exploded!
>> TypeError: Cannot create property 'default' on string 'date'
>>   at module.exports (/Users/loay/Desktop/NodeEE/loopback/common/models/application.js:75:56)
```

Ah, makes sense now.

> ```
 "created": {
      "default": "date"
    },
    "modified": {
      "default": "date"
    }
```

The correct solution is to use `"defaultFn": "now"`, see https://github.com/strongloop/loopback-datasource-juggler/pull/458

common/models/role.js
bajtos: I see. Please use the approach described in that thread then.

common/models/role.js
bajtos: I merged the master into your feature branch for you, see a75b28f

bajtos:
> Could you please use recently released eslint-config-loopback@5.0.0 (strongloop/eslint-config-loopback#12) while you are at this?

It turns out that we are using a lot of undeclared global variables in our tests, therefore upgrading to config@5 requires quite a lot of work.

In that light, let's stick with v4 of our eslint config, so that we can land this pull request soon and avoid painful merge conflicts.

loay:
@bajtos That was too late. I used already eslint@5 on Tuesday night but I didn't commit as I had few other changes to do.  
- I declared all required variables. 
- removed test/support.js since it is not needed anymore. global declaration was not working. 
- Some of my changes got merged with yours. Sorry about that !!
- Your changes are still there as far as I can tell. Can you double check? ..  just in case, but I made sure they are already there few times. 
- There was a node version conflict in `loopback.test.js`. I used semver. 
- test cases finally pass

superkhau:
@slnode test please

common/models/application.js
```diff
@@ -72,12 +72,13 @@ function generateKey(hmacKey, algorithm, encoding) {
 
 module.exports = function(Application) {
   // Workaround for https://github.com/strongloop/loopback/issues/292
+  Application.definition.rawProperties.created.default =
   Application.definition.properties.created.default = function() {
     return new Date();
   };
```
bajtos: These five lines (starting with the "workaround" comment) should be removed, since the default property values is already configured via JSON, right?

common/models/application.js
```diff
@@ -72,12 +72,13 @@ function generateKey(hmacKey, algorithm, encoding) {
...
 
   // Workaround for https://github.com/strongloop/loopback/issues/292
-  Application.definition.rawProperties.created.default =
+  Application.definition.rawProperties.modified.default =
   Application.definition.properties.modified.default = function() {
     return new Date();
   };
```
bajtos: Ditto - These five lines (starting with the "workaround" comment) should be removed, since the default property values is already configured via JSON, right?

test/access-token.test.js
```diff
@@ -280,7 +280,7 @@ describe('loopback.token(options)', function() {
       ' when enableDoubkecheck is true',
     function(done) {
       var token = this.token;
-
+      var app = loopback({localRegistry: true, loadBuiltinModels: true});
```
bajtos: This looks rather confusing to me, as this line is creating a local model registry but the test itself uses token in the global registry.

The correct solution is to rework the infrastructure in this test file to use a local per-app registry for everything and drop global `AccessToken` (see lines 18-21).

As a short-term fix to unblock your eslint work, I am proposing to change this line to `var app = loopback();` and if that does not work, then `var app = express();`.

test/access-token.test.js
```diff
@@ -315,7 +315,7 @@ describe('loopback.token(options)', function() {
     function(done) {
       var token = this.token;
       var tokenStub = {id: 'stub id'};
-
+      var app = loopback({localRegistry: true, loadBuiltinModels: true});
```
bajtos: Ditto - `var app = loopback();` or `var app = express();`

test/loopback.test.js
```diff
@@ -99,9 +100,72 @@ describe('loopback', function() {
         'version',
       ];
 
+      var EXPECTED_GTNODE4 = [
```
bajtos: This makes it very difficult to understand what's the difference between `EXPECTED_GTNODE4` and `EXPECTED`. Please rework the definition of these two arrays in such way, that one contains the shared items and the second adds/removes only what's different.

I find it also very surprising that this test was passing before you started fixing eslint stuff. I want you to investigate this more and find out what is causing the different set of `loopback` properties. I suspect this change may have been caused by enabling "strict mode".

common/models/application.js
loay: Correct .. Removed. 
Also removed in models/role.js as well.

test/access-token.test.js
loay: `var app = loopback();` was not working but it is working now for some reason, maybe because of the globak AccessToken thing.

test/loopback.test.js
loay: correct .. I didn't realize that we shouldn't worry anymore about pre4.0 so I fixed that. 
`arguments` and `caller` properties are removed though! Possibly because of the use strict.

loay:
@bajtos I hope there will be no merge conflicts till your next review :)

bajtos:
@loay I have pushed two commits - the first one fixes remaining details (e511235), the second one brings back `saltWorkFactor:4` to speed up tests (8213148).

Unfortunately the patch is out of date again, I'll rebase it myself so that we can get this landed ASAP.

bajtos:
CI failures are caused by unreachable npm registry, let's ignore them and land this pull request to avoid further merge conflicts.

ebarault:
@batjos, @loay, @superkhau  It seems this PR breaks support for db connectors. I'm not able to run a basic app like "loopback-access-control-example" since this merge. I'm installing with 
`npm install "https://github.com/strongloop/loopback.git#06cb481c3fbf467ee11ace5c186238840699e225"` 

I'm defining a simple mongodb db with:
```
{
	"db": {
		"name": "db",
		"connector": "mongodb",
		"host": "localhost",
		"port": 27017,
		"database": "loopback"
	}
}
```
and get the following error:
````
/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback/lib/application.js:237
    throw err;
    ^

TypeError: Cannot create data source "db": Cannot create property 'name' on string 'mongodb'
    at dataSourcesFromConfig (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback/lib/application.js:427:29)
    at Function.app.dataSource (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback/lib/application.js:226:14)
    at /Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/lib/executor.js:190:9
    at /Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/lib/executor.js:281:5
    at Array.forEach (native)
    at forEachKeyedObject (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/lib/executor.js:280:20)
    at setupDataSources (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/lib/executor.js:180:3)
    at execute (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/lib/executor.js:38:3)
    at bootLoopBackApp (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/node_modules/loopback-boot/index.js:154:3)
    at Object.<anonymous> (/Users/Eric/carbip/loopback_tests/loopback-example-access-control/server/server.js:17:1)
````

everything's fine if working with versions before this merge

PR #2636, opened 2016-08-16
need to check if the context is active

azatoth:
if you for example load the explorer, you don't have any active context,
and it will fail here then with the message:
  "Error: No context available. ns.run() or ns.bind() must be called
  first."

Fix https://github.com/strongloop/loopback-context/issues/6

### DIFF
--- a/server/middleware/token.js
+++ b/server/middleware/token.js
@@ -125,7 +125,7 @@ function token(options) {
       req.accessToken = token || null;
       rewriteUserLiteral(req, currentUserLiteral);
       var ctx = req.loopbackContext;
-      if (ctx) ctx.set('accessToken', token);
+      if (ctx && ctx.active) ctx.set('accessToken', token);
       next(err);
     });
   };


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
Thank you @azatoth for the pull request. As I mentioned in Gitter, I think this is fixing a wrong place. `req.loopbackContext` should be set only when there is an active context configured for the request. The problem you are observing is probably a bug in loopback-context. See https://github.com/strongloop/loopback-context/blob/268716f69eec246b4ae0f7e402147b6fcbc924cf/server/middleware/per-request.js#L36-L59.

bajtos:
@azatoth please reproduce the problem using loopback-context only, add a failing test to https://github.com/strongloop/loopback-context/blob/268716f69eec246b4ae0f7e402147b6fcbc924cf/test/main.test.js and then try to fix the problem directly in loopback-context.

bajtos:
Reopening.

So apparently `getCurrentContext` checks for ns.active too, see https://github.com/strongloop/loopback-context/blob/268716f69eec246b4ae0f7e402147b6fcbc924cf/server/current-context.js#L92. The patch strongloop/loopback#2604 which inlined getCurrentContext did not include this check.

slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/2649

Issue #2592, opened 2016-08-08
KeyValue: get all keys


bajtos:
- [x] API & memory connector - https://github.com/strongloop/loopback-datasource-juggler/pull/1049 and https://github.com/strongloop/loopback-datasource-juggler/pull/1058
 - [x] Redis connector - https://github.com/strongloop/loopback-connector-kv-redis/pull/6
 - [x] remoting in LoopBack - https://github.com/strongloop/loopback/pull/2638 and https://github.com/strongloop/loopback/pull/2654

bajtos:
Notes for the Redis implementation from http://redis.io/commands/KEYS
> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using SCAN or sets.
http://redis.io/commands/scan

bajtos:
I am deciding between two API approaches:
 - AsyncIterator - see https://github.com/tc39/proposal-async-iteration and https://www.npmjs.com/package/async-iterators
 - ReadableStream in object mode

bajtos:
To specify a filter for the keys to return, we can use glob-style matching. It is supported by Redis, [minimatch](https://www.npmjs.com/package/minimatch) can be used for our in-memory connector.

bajtos:
This has been already released to npmjs.org.

PR #2638, opened 2016-08-16
KeyValueModel: add API for listing keys

bajtos:
- Expose "keys()" at "GET /keys"
 - Add a dummy implementation for "iterateKeys" to serve a useful error
   message when the model is not attached correctly.

Requires https://github.com/strongloop/loopback-datasource-juggler/pull/1049
Connect to strongloop/loopback#2592

@superkhau @ritch please review

### DIFF
--- a/common/models/key-value-model.js
+++ b/common/models/key-value-model.js
@@ -16,6 +16,16 @@ module.exports = function(KeyValueModel) {
     throwNotAttached(this.modelName, 'expire');
   };
 
+  // TODO add api docs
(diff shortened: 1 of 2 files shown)


bajtos:
@superkhau @ritch Should we expose the method for listing all keys as `GET /keys` or `GET /`?

superkhau:
LGTM, I'm voting for `/keys` since the other methods should be exposed in the same way (ie. model/ttl, model/expires, etc).

PR #2665, opened 2016-08-23
Fix password expiry session

loay:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/925

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -747,8 +747,12 @@ module.exports = function(User) {
       });
       var idsToExpire = idsToExpireEmail.concat(idsToExpirePass);
       if (!idsToExpire.length) return next();
-      AccessToken.deleteAll({userId: {inq: idsToExpire}}, next);
+      UserModel.invalidateAccessTokens(AccessToken, idsToExpire, next);
     });
+
+    User.invalidateAccessTokens = function(token, id, cb) {
+      token.deleteAll({userId: {inq: id}}, cb);
+    };
 
     UserModel.remoteMethod(
       'login',
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -708,7 +708,7 @@ module.exports = function(User) {
       ctx.Model.find({where: where}, function(err, userInstances) {
         if (err) return next(err);
         ctx.hookState.originalUserData = userInstances.map(function(u) {
-          return {id: u.id, email: u.email};
+          return {id: u.id, email: u.email, password: u.password};
         });
         if (ctx.instance) {
           var emailChanged = ctx.instance.email !== ctx.hookState.originalUserData[0].email;
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -732,13 +732,20 @@ module.exports = function(User) {
       var AccessToken = ctx.Model.relations.accessTokens.modelTo;
       if (!ctx.instance && !ctx.data) return next();
       var newEmail = (ctx.instance || ctx.data).email;
-      if (!newEmail) return next();
+      var newPassword = (ctx.instance || ctx.data).password;
+      if (!newEmail && !newPassword) return next();
       if (!ctx.hookState.originalUserData) return next();
-      var idsToExpire = ctx.hookState.originalUserData.filter(function(u) {
+      var idsToExpireEmail = ctx.hookState.originalUserData.filter(function(u) {
         return u.email !== newEmail;
       }).map(function(u) {
         return u.id;
       });
+      var idsToExpirePass = ctx.hookState.originalUserData.filter(function(u) {
+        return u.password !== newPassword;
+      }).map(function(u) {
+        return u.id;
+      });
+      var idsToExpire = idsToExpireEmail.concat(idsToExpirePass);
       if (!idsToExpire.length) return next();
       AccessToken.deleteAll({userId: {inq: idsToExpire}}, next);
     });
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1998,7 +1998,7 @@ describe('User', function() {
       it('keeps sessions AS IS if firstName is added using `replaceAttributes`', function(done) {
         user.replaceAttributes({
           email: currentEmailCredentials.email,
-          password: currentEmailCredentials.password,
+          password: user.password,
           firstName: 'Candy',
         }, function(err, userInstance) {
           if (err) return done(err);
@@ -2011,7 +2011,7 @@ describe('User', function() {
           id: user.id,
           firstName: 'Loay',
           email: currentEmailCredentials.email,
-          password: currentEmailCredentials.password,
+          password: user.password,
         }, function(err, userInstance) {
           if (err) return done(err);
           assertUntouchedTokens(done);
@@ -2024,7 +2024,7 @@ describe('User', function() {
           {
             firstName: 'Miroslav',
             email: currentEmailCredentials.email,
-            password: currentEmailCredentials.password,
+            password: user.password,
           }, function(err, userInstance) {
             if (err) return done(err);
             assertUntouchedTokens(done);
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -2364,21 +2364,21 @@ describe('User', function() {
 
       async.series([
         function createOriginalAccount(next) {
-          User.create({ email: email, password: password }, function(err, userInstance) {
+          User.create({email: email, password: password}, function(err, userInstance) {
             if (err) return next(err);
             user = userInstance;
             next();
           });
         },
         function(next) {
-          User.login({ email: email, password: password }, function(err, accessToken1) {
+          User.login({email: email, password: password}, function(err, accessToken1) {
             if (err) return next(err);
             assert(accessToken1);
             next();
           });
         },
         function(next) {
-          User.login({ email: email, password: password }, function(err, accessToken2) {
+          User.login({email: email, password: password}, function(err, accessToken2) {
             if (err) return next(err);
             assert(accessToken2);
             next();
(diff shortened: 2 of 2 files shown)


common/models/user.js
```diff
@@ -577,6 +578,15 @@ module.exports = function(User) {
         return cb(err);
       }
 
+      var AccessToken = UserModel.relations.accessTokens.modelTo;
+      AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+        tokens.forEach(function(token) {
+          AccessToken.destroyById(token.id, function(err, info) {
+            if (err) return cb(err);
```
bajtos: This will can potentially call `cb` multiple times, once for each time an error happens.

common/models/user.js
```diff
@@ -577,6 +578,15 @@ module.exports = function(User) {
         return cb(err);
       }
 
+      var AccessToken = UserModel.relations.accessTokens.modelTo;
+      AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+        tokens.forEach(function(token) {
```
bajtos: Querying a list of tokens and then deleting them one by one is rather suboptimal. Please use bulk-delete instead.

```js
AccessToken.deleteAll({ userId: user.id }, function(err) {
  // etc.
});
```

bajtos:
@loay I am afraid you are invalidating access token in a wrong method. In your current implementation, you will invalidate tokens as soon as somebody clicks on "reset password" link. That's not correct, the tokens should be invalidated only when the password is actually going to change.

I think you should to use "before save" hook, detect when the data contains password field, and fail the whole operation if existing tokens cannot be deleted.

Please update the test case accordingly.

loay:
I thought of `before save` but I don't think `deleteById` or `DeleteAll` is supported there !

bajtos:
> I thought of before save but I don't think deleteById or DeleteAll is supported there !

@loay Show me the code that does not work for you please.

Here is what I am thinking of:

```js
User.observe('before save', function(ctx, next) {
  // detect password change, return next() if password is not changed
  var userId = // figure out how to get the id from ctx
  var AccessToken = ctx.Model.relations.accessTokens.modelTo;
  AccessToken.deleteAll({ userId: userId }, next);
});
```

Few more things:
 - Please add a unit-test to verify that access tokens of other users are not changed
 - Token removal should be configurable, i.e. the user should be able to decide whether they want to log out all other sessions or not
 - The currently active token should not be deleted. This means the code snippet above won't work OOTB.

Because we don't have a first-class API methods for changing password, it will be difficult to expose the token-removal flag via REST API and also to get hold of the current access token.  I am proposing to solve this in two steps:
 
1) In the hook, use `ctx.options` as the channel for passing the token-removal flag and the current access token.

2) Implement a new API for changing the password. I am thinking about something like this:

```js
User.prototype.changePassword = function(oldPassword, newPassword, expireOtherSessions, currentAccessToken) {
  // verify that oldPassword is valid
  // call updateAttributes, pass currentAccessToken and expireOtherSessions via options
};

// when defining remoting metadata
User.remoteMethod('prototype.changePassword', {
  accepts: [
    // oldPassword, newPassword, expireOtherSessions
    { arg: 'currentAccessToken', type: 'object', http: function(req) {
        // this may not work OOTB, see server/middleware/token.js
        // the idea is to inject the token argument from remoting context
        return req.accessToken;
      }
    }
  ]
});
```

loay:
@bajtos I took a different approach here:
I thought the usage of `resetPassword` is useless in the test cases, as it only creates a temp token, where the user may never use and keep the old password. We needed to test the effect of the new implementation once the password is actually changed.

common/models/user.js
```diff
@@ -668,6 +668,27 @@ module.exports = function(User) {
...
     });
 
+
+    UserModel.observe('before save', function(ctx, next) {
+      var userEmail, oldPassword;
+      if(ctx.isNewInstance) {
+        userEmail = ctx.instance.email;
+        oldPassword = ctx.instance.password;
+      } else {
+        return next();
```
bajtos: Huh? AFAICT, your code is expiring sessions only for new users.

common/models/user.js
```diff
@@ -668,6 +668,27 @@ module.exports = function(User) {
       next();
     });
 
+
+    UserModel.observe('before save', function(ctx, next) {
+      var userEmail, oldPassword;
+      if(ctx.isNewInstance) {
+        userEmail = ctx.instance.email;
+        oldPassword = ctx.instance.password;
```
bajtos: This will contain the new password, not the old one.

test/user.test.js
```diff
@@ -1800,6 +1800,105 @@ describe('User', function() {
...
+        },
+        function(next) {
+          User.login({ email: email, password: password }, function(err, accessToken2) {
+            if (err) return next(err);
+            assert(accessToken2);
+            next();
+          });
+        },
+        function(next) {
+          user.replaceAttributes({'email': email, password: 'newPass'}, function(err, user2) {
```
bajtos: How is `email` related to password reset? It should be enough to call `user.replaceAttribute('password', 'newPass', cb)`.

The same applies for other tests below.

bajtos:
@loay see my comments above and please make sure the CI builds are (mostly) green before asking for another round of code review.

bajtos:
>  I thought the usage of `resetPassword` is useless in the test cases, as it only creates a temp token, where the user may never use and keep the old password.

Yes, makes sense.

> We needed to test the effect of the new implementation once the password is actually changed. 

Yes please. This patch is going in the right direction, but you are not there yet. I think you should finish #2693 first, the implementation in this patch can reuse some parts from that other PR.

loay:
Hi @bajtos 
That's the implementation for deleting the sessions if the password is changed. Even if the unit tests pass, I am getting a regression here concerning a previous implementation:

3 test cases fail: 
  1)  keeps sessions AS IS if firstName is added using `replaceAttributes`
  2)  keeps sessions AS IS if firstName is added using `updateOrCreate`
  3)  keeps sessions AS IS if firstName is added using `replaceById`


The goal of the test was to only update the first name and consequently, the session stays AS IS, since the email has not been touched and it was successful till I added this PR's implementation. 

Using any of the three operations, we need to insert the password or it will give a validation error, and even if it is the same password, it is being detected as a different password after bcrypt, and that's why it is considered as an update, and therefore the test fails. 

Any thoughts? @bajtos  @raymondfeng 
Thanks

bajtos:
Apparently, `User.hashPassword` returns a different value for each invocation.

```
HASH PASSWORD bar -> $2a$04$yWVcghDkA.4tND0Sgmexye4VDiqwD1.ZQE7aLOa/2UL6m9IECMsEy
HASH PASSWORD bar -> $2a$04$1lk/nr8J/mk9hhJFG13BqOlymkpYveHpMLtZXKR5yY1Mob0l1Af.G
HASH PASSWORD bar -> $2a$04$jDAxn7X9kp9ggzN0zD4K.Ori6x7A6ZUWfJUo0tSVfEKaoj8lKg3JC
```

IMO, the probability of a "change password" request using the same password value is very low, and therefore we can afford to treat any password value as a password change.

 - If `ctx.data` contains a password field, then the request is changing the password.

 - If `ctx.instances` is set, then we consider the request as a password change.

(Note that the situation is more tricky with `ctx.instance`. Instance data returned in REST API calls excludes the hashed password, and since REST clients typically don't store the password locally, there is no easy way to make a REPLACE call from via REST. Therefore I am proposing to treat all REPLACE calls as changing password. We can always improve this part in the future if needed.)

@raymondfeng thoughts?

BTW (some) of the failing tests can be easily fixed by replacing `currentEmailCredentials.password` with `user.password` in the request data:

```js
        user.replaceAttributes({
          email: user.email,
          password: user.password,
          firstName: 'Candy',
        }
```

loay:
@bajtos Thanks. I thought of using `user.password` but I thought to keep it the way it is, so I can gather feedback from you and @raymondfeng 

>IMO, the probability of a "change password" request using the same password value is very low, and therefore we can afford to treat any password value as a password change.

I agree

loay:
@slnode test please

raymondfeng:
If the purpose is to invalidate all access tokens for a given user if the password is changed. I have implemented a different way before:

- Keep a hashcode (generated using the user password) with any generated access token
- During token validation, make sure the hashcode still equals to the calculated one from the current user password

bajtos:
@raymondfeng that's an interesting alternative. What we would like to (eventually) allow, is to offer the user an option whether they want to invalidate all other sessions on a password change. How do you envision implementing that option with your approach?

loay:
@slnode test please

raymondfeng:
@bajtos If that's the case, we can add a separate field such as `tokenSecrect` to the user model and use it to generate hash codes for tokens. This way, when the user password is changed, we have the option to decide if `tokenSecret` should be reset.

loay:
@raymondfeng 
Does the current solution in this pr work as a temp solution ? i.e. where we delete the token when the password is changed .. till we rework the password other issues in the above described manner in the future: https://github.com/strongloop/loopback/issues/382

raymondfeng:
@loay I'm fine to go with your solution now. You probably should clean up the code as follows:

1. Add a method such as `User.invalidateAccessTokens(...)`
2. Trigger the method when user password/email is changed

loay:
@raymondfeng Thanks

bajtos:
@loay thank you for the update. There are still many subtle details to fix, I think it will be fastest for me to take over this pull request. As I was playing with the code, I discovered that because the hooks are registered from `setup()` method, sub-classed models (e.g. `TestUser` based on `User`) invoke the hook callback multiple times. Let's fix this problem first - see https://github.com/strongloop/loopback/pull/3014

bajtos:
At the end it turned out that the cleanup/refactoring is larger than I original thought, I had to clean up the test cases too, as it no longer makes sense to have different "describe" blocks for invalidation triggered by email and by password change.

I am closing this patch in favour of https://github.com/strongloop/loopback/pull/3018

PR #2669, opened 2016-08-24
Fixes 'this' binding for role connection to users/applications/roles 

codyolsen:
`listByPrincipalType()` loses `this` context of the `Role` and as a result the `where: this.id` in the filter is undefined and ignored. As a result the function returns any models with any assigned role rather than the role specified by `roleId`.

This commit simply adds `listByPrincipalType.call(this, ...)` to bind the `Role` instance as it's context.

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -80,7 +80,7 @@ module.exports = function(Role) {
         };
 
         var model = relsToModels[rel];
-        listByPrincipalType(model, relsToTypes[rel], query, callback);
+        listByPrincipalType.call(this, model, relsToTypes[rel], query, callback);
       };
     });
 
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -80,26 +80,27 @@ module.exports = function(Role) {
         };
 
         var model = relsToModels[rel];
-        listByPrincipalType.call(this, model, relsToTypes[rel], query, callback);
+        listByPrincipalType(this, model, relsToTypes[rel], query, callback);
       };
     });
 
     /**
      * Fetch all models assigned to this role
      * @private
+     * @param {object} model context
      * @param {*} model model type to fetch
      * @param {String} [principalType] principalType used in the rolemapping for model
      * @param {object} [query] query object passed to model find call
      * @param  {Function} [callback] callback function called with `(err, models)` arguments.
      */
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@codyolsen thank you for the pull request. Could you please add a unit-test that is failing with the current implementation and starts to pass with your fix in place?

Also `Function.prototype.call` is rather expensive performance-wise, it's better to pass the role instance as a regular function argument.

bajtos:
@slnode ok to test

codyolsen:
@bajtos Sounds like a plan!

codyolsen:
@bajtos I've updated the PR. Let me know if there is anything else needed to complete the merge.

bajtos:
@codyolsen thank you for the update.

First of all, make sure `npm test` is passing on your machine. Your code is not passing our linter checks, as can be seen e.g. [here](https://travis-ci.org/strongloop/loopback/jobs/158616081).

Secondly, please rebase your patch on top of the current (latest) `master` and edit the commit message to follow the [50/72 style](https://github.com/strongloop/loopback-contributor-docs/blob/master/git-commit-messages.md).

bajtos:
>  Could you please add a unit-test that is failing with the current implementation and starts to pass with your fix in place?

@codyolsen I see that you have refactored an existing test, but I don't see any new tests being added. What am I missing?

In general, we try to avoid making changes in code that works, as it makes code archaeology more difficult. Could you please keep the original test intact and add a new one for your fix?

codyolsen:
@bajtos I should have taken a closer look at the contributing code wiki. My apologies.

codyolsen:
Closing to submit a new PR following code submission wiki.

bajtos:
> Closing to submit a new PR following code submission wiki.

For the future: it's better to keep the work in the same pull request. Git history can be cleaned up (rewritten) before landing the patch, so there is no need to open a new pull request. When the work stays in the same pull request, then also all comments and discussions stay in the same place, which makes it much much easier for (future) readers to build a full understanding.

PR #2676, opened 2016-08-26
Add remoting for kv model ttl

superkhau:
@bajtos PTAL

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "mBaaS"
   ],
   "scripts": {
+    "lint": "grunt eslint",
     "test": "grunt mocha-and-karma"
   },
   "dependencies": {
--- a/common/models/key-value-model.js
+++ b/common/models/key-value-model.js
@@ -17,6 +17,11 @@ module.exports = function(KeyValueModel) {
   };
 
   // TODO add api docs
+  KeyValueModel.ttl = function(key, options, callback) {
+    throwNotAttached(this.modelName, 'ttl');
+  };
+
+  // TODO add api docs
   KeyValueModel.keys = function(filter, options, callback) {
     throwNotAttached(this.modelName, 'keys');
   };
@@ -60,6 +65,15 @@ module.exports = function(KeyValueModel) {
           http: { source: 'form' }},
       ],
       http: { path: '/:key/expire', verb: 'put' },
+    });
+
+    this.remoteMethod('ttl', {
+      accepts: {
+        arg: 'key', type: 'string', required: true,
+        http: { source: 'path' },
+      },
+      returns: { arg: 'value', type: 'any', root: true },
+      http: { path: '/:key/ttl', verb: 'get' },
     });
 
     this.remoteMethod('keys', {
--- a/test/key-value-model.test.js
+++ b/test/key-value-model.test.js
@@ -81,6 +81,47 @@ describe('KeyValueModel', function() {
         .expect(404, done);
     });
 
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "mBaaS"
   ],
   "scripts": {
+    "lint": "eslint .",
```
bajtos: I am afraid this is not correct. You should run `grunt eslint` here.

bajtos: For long term, I think we should remove `eslint` config from `Gruntfile` and use the usual posttest + lint scripts, as we do in other projects. However, that's a more complex change that should have its own pull request.

I am ok to add `"lint": "grunt eslint"` as a quick fix though.

package.json
```diff
@@ -46,7 +47,7 @@
     "express": "^4.14.0",
     "inflection": "^1.6.0",
     "loopback-connector-remote": "^1.0.3",
-    "loopback-datasource-juggler": "^3.0.0-alpha.1",
+    "loopback-datasource-juggler": "^3.0.0-alpha.6",
```
bajtos: This should be `^3.0.0-alpha.7`, right?

package.json
superkhau: I was thinking this would be the end result and skip the intermediate changes, but if this is the preferred method, NP. I will use your second proposal for the quick fix for now.

However, I would like to know your thought process -- the intermediate changes between seem subjective (adding more work when we know the end result we want is the one I proposed). Is the idea you want consistency here and change everything at once when we actually decide to deprecate Grunt in favour of NPM scripts?

package.json
superkhau: Haha now it should be (there was no ^3.0.0-alpha.7 when I submitted this PR). ;)

package.json
```diff
@@ -32,6 +32,7 @@
     "mBaaS"
   ],
   "scripts": {
+    "lint": "grunt eslint",
```
superkhau: @bajtos Fixed up.

superkhau:
@bajtos PTAL again

superkhau:
@slnode test please

superkhau:
@slnode test please

rmg:
The failing `dependant` builds appear to be mostly transient failures.

More importantly, though, they are all using `loopback@2` (with the exception of `loopback-api-definition` which is using `loopback@1.7`). In other words, none of the failing dependant builds were actually testing this PR anyway.

In this case, I went through them all to confirm this, but in the general case this is indicated by the lack of `[cis-jenkins] downstream ...` builds which are version aware in order to prevent this.

package.json
```diff
@@ -46,14 +47,14 @@
     "express": "^4.14.0",
     "inflection": "^1.6.0",
     "loopback-connector-remote": "^1.0.3",
-    "loopback-datasource-juggler": "^3.0.0-alpha.1",
+    "loopback-datasource-juggler": "^3.0.0-alpha.7",
```
bajtos: This is clashing with https://github.com/strongloop/loopback/pull/2680 Let's land my PR first and then rebase your on top of the new master.

bajtos:
@superkhau the commit history got messed up, the fix for "npm run lint" ended up in the KV-related commit. I am proposing to either move that fix to the first commit (that would be ideal), or alternatively squash all commits into a single one.

Other than this and my comment above, the patch LGTM. No further review is necessary.

rmg:
@slnode test please

rmg:
@slnode test please

PR #2680, opened 2016-08-26
Update pre-release dependencies

bajtos:
Apparently `^3.0.0-alpha.1` does not cover `3.0.0-alpha.7`, we need to upgrade pre-release dependencies explicitly.


superkhau:
LGTM

I'm also curious as to whether we should remove the ^ since its "sort" of directly installing the version we're specifying (not sure if ^ picks up ^4.0.0-alpha.x -- but it probably not important and less confusing to say we're "explicitly" using X alpha version).

bajtos:
> I'm also curious as to whether we should remove the ^ since it's sort of directly installing the version we're specifying (not sure if ^ picks up ^4.0.0-alpha.x -- but it probably not important and less confusing to say we're "explicitly" using X alpha version). Thoughts?

Well, `^` will pick regular release versions. I don't know if that matters or not.

I agree not using `^` will be less confusing.

Let's do as you proposed, @superkhau

bajtos:
> I'm also curious as to whether we should remove the ^ since it's sort of directly installing the version we're specifying (not sure if `^` picks up `^4.0.0-alpha.x` -- but it probably not important and less confusing to say we're "explicitly" using X alpha version). Thoughts?


This is weird. When running on Node v6.4.0, `^3.0.0-alpha.1` actually does pick `3.0.0-alpha.2`. From @rmg comments, I though it would not pick it up. This may be node/npm version specific.

What's more important, when I remove the `^` from the version range, a wrong (an older) version is installed from our staging registry, because staging registry automatically increase the number after the last dot. For example, the latest strong-remoting alpha is `strong-remoting@3.0.0-alpha.98`. This caused all jenkins-based builds to fail, even though Travis CI passed.

I am reverting back to using `^3.0.0-alpha.X`.

cc @rmg @strongloop/loopback-dev

superkhau:
>This is weird. When running on Node v6.4.0, ^3.0.0-alpha.1 actually does pick 3.0.0-alpha.2. From @rmg comments, I though it would not pick it up. This may be node/npm version specific

This is interesting. Didn't think removing the ^ would cause that many issues. I agree with staying the course until we get an answer on why Jenkins is acting differently than Travis.

PR #2693, opened 2016-08-30
Invalidate sessions after email change

loay:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/924

 - [x] Delete all tokens of the current user after an email was changed.
 - ~~The currently active token should not be deleted.~~

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -673,6 +673,34 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    UserModel.observe('before save', function beforeEmailUpdate(ctx, next) {
+      if (ctx.isNewInstance) return next();
+      if (!ctx.where && !ctx.instance) return next();
+      var where = ctx.where || { id: ctx.instance.id };
+      ctx.Model.find({ where: where }, function(err, userInstances) {
+        if (err) return next(err);
+        ctx.hookState.originalUserData = userInstances.map(function(u) {
+          return { id: u.id, email: u.email };
+        });
+        next();
+      });
+    });
+
+    UserModel.observe('after save', function afterEmailUpdate(ctx, next) {
+      if (!ctx.Model.relations.accessTokens) return next();
+      var AccessToken = ctx.Model.relations.accessTokens.modelTo;
+      var newEmail = (ctx.instance || ctx.data).email;
+      if (!ctx.hookState.originalUserData) return next();
+      var idsToExpire = ctx.hookState.originalUserData.filter(function(u) {
+        return u.email !== newEmail;
+      }).map(function(u) {
+        return u.id;
+      });
+      if (!idsToExpire.length) return next();
+      AccessToken.deleteAll({ userId: { inq: idsToExpire }}, next);
+    });
+
     UserModel.remoteMethod(
       'login',
       {
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -45,9 +45,12 @@ describe('User', function() {
     app.model(Email, { dataSource: 'email' });
 
     // attach User and related models
+    // forceId is set to false for the purpose of updating the same affected user within the
+    // `Email Update` test cases.
     User = app.registry.createModel('TestUser', {}, {
       base: 'User',
       http: { path: 'test-users' },
+      forceId: false,
     });
     app.model(User, { dataSource: 'db' });
 
@@ -1801,6 +1804,316 @@ describe('User', function() {
             done();
           });
       });
+    });
+  });
+
+  describe('Email Update', function() {
+    describe('User changing email property', function() {
+      var user, originalUserToken1, originalUserToken2, newUserCreated;
+      var currentEmailCredentials = { email: 'original@example.com', password: 'bar' };
+      var updatedEmailCredentials = { email: 'updated@example.com', password: 'bar' };
+      var newUserCred = { email: 'newuser@example.com', password: 'newpass' };
+
+      beforeEach('create user then login', function createAndLogin(done) {
+        async.series([
+          function createUserWithOriginalEmail(next) {
+            User.create(currentEmailCredentials, function(err, userCreated) {
+              if (err) return next(err);
+              user = userCreated;
+              next();
+            });
+          },
+          function firstLoginWithOriginalEmail(next) {
+            User.login(currentEmailCredentials, function(err, accessToken1) {
+              if (err) return next(err);
+              assert(accessToken1.userId);
+              originalUserToken1 = accessToken1.id;
+              next();
+            });
+          },
+          function secondLoginWithOriginalEmail(next) {
+            User.login(currentEmailCredentials, function(err, accessToken2) {
+              if (err) return next(err);
+              assert(accessToken2.userId);
+              originalUserToken2 = accessToken2.id;
+              next();
+            });
+          },
+        ], done);
+      });
+
+      it('invalidates sessions when email is changed using `updateAttributes`', function(done) {
+        user.updateAttributes(
+          { email: updatedEmailCredentials.email },
+          function(err, userInstance) {
+            if (err) return done(err);
+            assertNoAccessTokens(done);
+          });
+      });
+
+      it('invalidates sessions when email is changed using `replaceAttributes`', function(done) {
+        user.replaceAttributes(updatedEmailCredentials, function(err, userInstance) {
+          if (err) return done(err);
+          assertNoAccessTokens(done);
+        });
+      });
+
+      it('invalidates sessions when email is changed using `updateOrCreate`', function(done) {
+        User.updateOrCreate({
+          id: user.id,
+          email: updatedEmailCredentials.email,
+          password: updatedEmailCredentials.password,
+        }, function(err, userInstance) {
+          if (err) return done(err);
+          assertNoAccessTokens(done);
+        });
+      });
+
+      it('invalidates sessions when the email is changed using `replaceById`', function(done) {
+        User.replaceById(user.id, updatedEmailCredentials, function(err, userInstance) {
+          if (err) return done(err);
+          assertNoAccessTokens(done);
+        });
+      });
+
+      it('invalidates sessions when the email is changed using `replaceOrCreate`', function(done) {
+        User.replaceOrCreate({
+          id: user.id,
+          email: updatedEmailCredentials.email,
+          password: updatedEmailCredentials.password,
+        }, function(err, userInstance) {
+          if (err) return done(err);
+          assertNoAccessTokens(done);
+        });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `updateAttributes`', function(done) {
+        user.updateAttributes({ 'firstName': 'Janny' }, function(err, userInstance) {
+          if (err) return done(err);
+          assertUntouchedTokens(done);
+        });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `replaceAttributes`', function(done) {
+        user.replaceAttributes({
+          email: currentEmailCredentials.email,
+          password: currentEmailCredentials.password,
+          firstName: 'Candy',
+        }, function(err, userInstance) {
+          if (err) return done(err);
+          assertUntouchedTokens(done);
+        });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `updateOrCreate`', function(done) {
+        User.updateOrCreate({
+          id: user.id,
+          firstName: 'Loay',
+          email: currentEmailCredentials.email,
+          password: currentEmailCredentials.password,
+        }, function(err, userInstance) {
+          if (err) return done(err);
+          assertUntouchedTokens(done);
+        });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `replaceById`', function(done) {
+        User.replaceById(
+          user.id,
+          {
+            firstName: 'Miroslav',
+            email: currentEmailCredentials.email,
+            password: currentEmailCredentials.password,
+          }, function(err, userInstance) {
+            if (err) return done(err);
+            assertUntouchedTokens(done);
+          });
+      });
+
+      it('keeps sessions AS IS if a new user is created using `create`', function(done) {
+        async.series([
+          function(next) {
+            User.create(newUserCred, function(err, newUserInstance) {
+              if (err) return done(err);
+              newUserCreated = newUserInstance;
+              next();
+            });
+          },
+          function(next) {
+            User.login(newUserCred, function(err, newAccessToken) {
+              if (err) return done(err);
+              assert(newAccessToken.id);
+              assertPreservedToken(next);
+            });
+          },
+        ], done);
+      });
+
+      it('keeps sessions AS IS if a new user is created using `updateOrCreate`', function(done) {
+        async.series([
+          function(next) {
+            User.create(newUserCred, function(err, newUserInstance2) {
+              if (err) return done(err);
+              newUserCreated = newUserInstance2;
+              next();
+            });
+          },
+          function(next) {
+            User.login(newUserCred, function(err, newAccessToken2) {
+              if (err) return done(err);
+              assert(newAccessToken2.id);
+              assertPreservedToken(next);
+            });
+          },
+        ], done);
+      });
+
+      function assertPreservedToken(done) {
+        AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+          if (err) return done(err);
+          expect(tokens.length).to.equal(2);
+          expect([tokens[0].id, tokens[1].id]).to.have.members([originalUserToken1,
+            originalUserToken2]);
+          done();
+        });
+      };
+
+      function assertNoAccessTokens(done) {
+        AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+          if (err) return done(err);
+          expect(tokens.length).to.equal(0);
+          done();
+        });
+      }
+
+      function assertUntouchedTokens(done) {
+        AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+          if (err) return done(err);
+          expect(tokens.length).to.equal(2);
+          done();
+        });
+      }
+    });
+
+    describe('User not changing email property', function() {
+      var user1, user2, user3;
+      it('preserves other users\' sessions if their email is  untouched', function(done) {
+        async.series([
+          function(next) {
+            User.create({ email: 'user1@example.com', password: 'u1pass' }, function(err, u1) {
+              if (err) return done(err);
+              User.create({ email: 'user2@example.com', password: 'u2pass' }, function(err, u2) {
+                if (err) return done(err);
+                User.create({ email: 'user3@example.com', password: 'u3pass' }, function(err, u3) {
+                  if (err) return done(err);
+                  user1 = u1;
+                  user2 = u2;
+                  user3 = u3;
+                  next();
+                });
+              });
+            });
+          },
+          function(next) {
+            User.login(
+              { email: 'user1@example.com', password: 'u1pass' },
+              function(err, accessToken1) {
+                if (err) return next(err);
+                User.login(
+                  { email: 'user2@example.com', password: 'u2pass' },
+                  function(err, accessToken2) {
+                    if (err) return next(err);
+                    User.login({ email: 'user3@example.com', password: 'u3pass' },
+                    function(err, accessToken3) {
+                      if (err) return next(err);
+                      next();
+                    });
+                  });
+              });
+          },
+          function(next) {
+            user2.updateAttribute('email', 'user2Update@b.com', function(err, userInstance) {
+              if (err) return next(err);
+              assert.equal(userInstance.email, 'user2Update@b.com');
+              next();
+            });
+          },
+          function(next) {
+            AccessToken.find({ where: { userId: user1.id }}, function(err, tokens1) {
+              if (err) return next(err);
+              AccessToken.find({ where: { userId: user2.id }}, function(err, tokens2) {
+                if (err) return next(err);
+                AccessToken.find({ where: { userId: user3.id }}, function(err, tokens3) {
+                  if (err) return next(err);
+
+                  expect(tokens1.length).to.equal(1);
+                  expect(tokens2.length).to.equal(0);
+                  expect(tokens3.length).to.equal(1);
+                  next();
+                });
+              });
+            });
+          },
+        ], done);
+      });
+    });
+
+    it('invalidates sessions after using updateAll', function(done) {
+      var userSpecial, userNormal;
+      async.series([
+        function createSpecialUser(next) {
+          User.create(
+              { email: 'special@example.com', password: 'pass1', name: 'Special' },
+              function(err, specialInstance) {
+                if (err) return next (err);
+                userSpecial = specialInstance;
+                next();
+              });
+        },
+        function createNormaluser(next) {
+          User.create(
+              { email: 'normal@example.com', password: 'pass2' },
+              function(err, normalInstance) {
+                if (err) return next (err);
+                userNormal = normalInstance;
+                next();
+              });
+        },
+        function loginSpecialUser(next) {
+          User.login({ email: 'special@example.com', password: 'pass1' }, function(err, ats) {
+            if (err) return next (err);
+            next();
+          });
+        },
+        function loginNormalUser(next) {
+          User.login({ email: 'normal@example.com', password: 'pass2' }, function(err, atn) {
+            if (err) return next (err);
+            next();
+          });
+        },
+        function updateSpecialUser(next) {
+          User.updateAll(
+                { name: 'Special' },
+                { email: 'superspecial@example.com' }, function(err, info) {
+                  if (err) return next (err);
+                  next();
+                });
+        },
+        function verifyTokensOfSpecialUser(next) {
+          AccessToken.find({ where: { userId: userSpecial.id }}, function(err, tokens1) {
+            if (err) return done (err);
+            expect(tokens1.length).to.equal(0);
+            next();
+          });
+        },
+        function verifyTokensOfNormalUser(next) {
+          AccessToken.find({ userId: userNormal.userId }, function(err, tokens2) {
+            if (err) return done (err);
+            expect(tokens2.length).to.equal(1);
+            next();
+          });
+        },
+      ], done);
     });
   });
 


common/models/user.js
```diff
@@ -668,6 +668,26 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    UserModel.observe('before save', function(ctx, next) {
+      if (ctx.instance) {
+        var oldEmail = ctx.instance.email;
+      }
+      UserModel.observe('after save', function(ctx, next) {
```
bajtos: This will add a new observer after whenever a save operation is called. That won't work.

The correct solution is to store `oldEmail` in `ctx.hookState` instead of creating a local closure.

```
// in before save
ctx.hookState.oldEmail = ctx.instance.email

// in after save
var oldEmail = ctx.hookState.oldEmail
```

Having said that, I am afraid we don't have any easy way how to detect when a field was changed.

The only reliable solution is to query the database for the old version and compare it with the data being persisted.

common/models/user.js
```diff
@@ -668,6 +668,26 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    UserModel.observe('before save', function(ctx, next) {
+      if (ctx.instance) {
```
bajtos: `updateAttributes` and `upsert` don't fill `ctx.instance`, they provide `ctx.data` instead.

bajtos:
Please test also that sessions are not invalidated when email was not changed.

Both test cases should be run for all operations changing data (create, updateAttributes, updateOrCreate, replaceAttributes, replaceOrCreate).

bajtos:
Most of the concerns from https://github.com/strongloop/loopback/pull/2665#issuecomment-243781927 apply here too:

> - Please add a unit-test to verify that access tokens of other users are not changed
> - Token removal should be configurable, i.e. the user should be able to decide whether they want to log out all other sessions or not
> - The currently active token should not be deleted.

loay:
@bajtos PTAL.

loay:
@bajtos PTAL. Thanks.

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    var oldEmail, newEmail, oldId, newId;
```
bajtos: These variables are shared among all invocations of your hook. What will happen when there are two "save" operations in progress at the same time?

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+    UserModel.observe('before save', function(ctx, next) {
+      if (ctx.instance) {
```
bajtos: Please add a check for `ctx.isNewInstance` and skip the extra database query.

```js
if (ctx.isNewInstance) return next();
```

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    var oldEmail, newEmail, oldId, newId;
+    UserModel.observe('before save', function(ctx, next) {
```
bajtos: Use a named function to make the intent of this hook clearer and also to produce more helpful stack traces.

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+      if (ctx.instance) {
+        if (ctx.instance.id !== undefined) {
```
bajtos: If `ctx.instance.id` is `undefined`, then there is no way how to find the affected user model. IIRC this happens only when a new model is going to be created, which means you can bail out via `return next()`.

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+          ctx.hookState.id = ctx.instance.id;
+          oldId = ctx.hookState.id;
+        }
+      }
+      if (ctx.data) {
+        ctx.hookState.id = ctx.where.id;
```
bajtos: `ctx.where.id` may not be always present, it also may not be a simple value like `{ where: { id: 123 }}`, but a condition like `{ where: { id: { in: [123, 456]}}}`.

This typically happens when calling `updateAll`, i.e. performing a bulk update.

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+        oldId = ctx.hookState.id;
+      }
+      ctx.Model.find({ where: { id: oldId }}, function(err, userInstances) {
```
bajtos: You should use `ctx.where` if it's present.

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+        userInstances.forEach(function(userInstance) {
+          ctx.hookState.email = userInstance.email;
```
bajtos: Since `userInstances` may have any number of items - zero, one or more - you should store a list of affected users, not only a single instance. For each item, store both `id` and `email`.

```js
ctx.hookState.originalUserData = userInstances.map(function(u) {
  return { id: u.id, email: u.email };
});
```

common/models/user.js
```diff
@@ -668,6 +668,45 @@ module.exports = function(User) {
...
+      });
+      next();
+    });
+
+    UserModel.observe('after save', function(ctx, next) {
+      var AccessToken = ctx.Model.relations.accessTokens.modelTo;
+      newEmail = ctx.instance.email;
+      newId = ctx.instance.id;
+      if (!ctx.isNewInstance && oldEmail !== newEmail && oldId === newId) {
+        AccessToken.deleteAll({ userId: ctx.instance.id }, function(err, info) {
```
bajtos: As I commented earlier, there may be more than a single user affected by a "save" operation. You should invalidate access tokens for all affected users, preferably via a single query.

```js
var newEmail = (ctx.instance || ctx.data).email;
var idsToExpire = ctx.hookState.originalUserData.filter(function(u) {
  return u.email !== newEmail;
}).map(fuction(u) {
  return u.id;
});
if (!idsToExpire.length)
  return next()l

AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
  // ...
});
```

common/models/user.js
bajtos: Also when performing a partial update, the email may not change at all. For example:

```js
user.updateAttributes({ age: user.age + 1 }, cb);
```

You should skip the session-invalidation logic in such case too.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
     });
   });
 
+  describe('Email Update', function() {
+    describe('User changing email property', function() {
+      var usersId;
+      var currentEmailCredentials = { email: 'm@c.com', password: 'bar' };
+      var updatedEmailCredentials = { email: 'n@d.com', password: 'bar' };
```
bajtos: Please use a more descriptive email addresses, for example `old@c.com` and `new@c.com`.

Also we usually use `@example.com` as the domain name. `example.com` domain was designed exactly for that.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+
+      beforeEach(function(done) {
```
bajtos: The hook should be named, see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#hooks. Consider also https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#layout-of-test-files

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+        async.series([
+          function(next) {
```
bajtos: The function should be named, e.g. `createUserWithOriginalEmail`.

test/user.test.js
bajtos: Perhaps `original@example.com` and `updated@example.com` will be best.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+            User.create({ email: currentEmailCredentials.email,
+            password: currentEmailCredentials.password }, function(err, user) {
```
bajtos: Wrong formatting/indentation, see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

Please apply this and the comment above to the rest of the newly added code in this file.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+              usersId = user.id;
```
bajtos: The name `usersId` looks strange to me. We usually use `userId` instead.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+      });
+
+      afterEach(function(done) {
+        User.deleteById(usersId, function(err, user) {
+          if (err) return done (err);
+          done();
+        });
+      });
+
+      it('invalidates sessions once the email is changed using `updateAttribute`', function(done) {
```
bajtos: Use `when` instead of `once`:

`invalidates sessions when the email...`

Please apply to all other new tests too.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+              next();
+            });
+            });
+          },
+          function(next) {
+            AccessToken.find({ where: { userId: usersId }}, function(err, tokens) {
+              if (err) return next(err);
+              expect(tokens.length).to.equal(0);
+              next();
+            });
```
bajtos: This function is shared by many tests, please refactor and extract a shared function.

```js
function assertNoAccessTokens(cb) {
  AccessToken.find(...);
}
```

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+          },
+        ], function(err) {
+          if (err) return done(err);
+          done();
```
bajtos: Just pass `done` as the callback argument of `async.series`. Apply in all new code.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+        });
+      });
+
+      it('invalidates sessions once the email is changed using `updateAttribute`', function(done) {
+        async.series([
+          function(next) {
+            User.findById(usersId, function(err, userFound)  {
+              if (err) return next(err);
+              userFound.updateAttribute('email', updatedEmailCredentials.email,
+            function(err, userInstance) {
```
bajtos: Wrong indentation, please fix.

Should be:

```js
userFound.updateAttribute('email', updatedEmailCredentials.email,
  function(err, userInstance) {
```

See https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#exception

bajtos: Apply to other places in your patch too.

test/user.test.js
bajtos: Also to avoid frequent `User.findById` in the tests below, it's better to store the full user `instance`.

```js
User.create(..., function(err, u) {
  user = u;
  userId = u.id; // probably not needed
  next();
});
```

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+              if (err) return next(err);
+              assert.equal(userInstance.email, updatedEmailCredentials.email);
```
bajtos: No need to check that `updateAttribute` does it job, that's already tested elsewhere. I hink you can pass `next` directly as `updateAttribute` callback.

Once you apply https://github.com/strongloop/loopback/pull/2693/files#r77507957, it should be possible to simplify the tests so that `async.series` is no longer necessary.

```js
it(..., function(done) {
  user.updateAttribute(..., function(err, updated) {
    if (err) return done(err);
    assertNoAccessTokens(done);
  });
});
```

I would personally use Promises instead and simplify the code even more.

```js
it(..., function() {
  return user.updateAttributes(...)
    .then(assertNoAccessTokens);
});
```

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+        ], function(err) {
+          if (err) return done(err);
+          done();
+        });
+      });
+    });
+
+    describe('User not chaning email property', function() {
+      var user1Id, user2Id, user3Id;
+      it('keeps sessions untouched if email property is not changed', function(done) {
```
bajtos: To me, this looks the same as "keeps sessions AS IS if firstName is added using {method name}".  What's the difference?

From looking at the code, it looks like you are checking that the we invalidate only sessions of the current user, right? The test name should reflect that.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+        async.series([
+          function(next) {
+            User.create({ email: 'user1@a.com', password: 'u1pass' }, function(err, user1) {
```
bajtos: Unhandled `err`. Same applies to other places below. Consider using 

```
async.each(
  [
    { email, password },
    //...
  ],
  function(data, next) { User.create(data, next) },
  function(err, users) {
    // ...
  });
});
```

I would personally use promises instead:

```js
var userData = [...];
Promise.all(userData.map(function(data) { return User.create(data) }))
  .then(function(users) {
    // ...
  });
```

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+              User.create({ email: 'user2@b.com', password: 'u2pass' }, function(err, user2) {
+                User.create({ email: 'user3@c.com', password: 'u3pass' }, function(err, user3) {
+                  if (err) return done(err);
+                  user1Id = user1.id;
+                  user2Id = user2.id;
+                  user3Id = user3.id;
```
bajtos: Since there are only two kinds of users (a user changing the email, a user that should not be affected), it's enough to have only two user instances in the test too.

test/user.test.js
```diff
@@ -1763,6 +1763,367 @@ describe('User', function() {
...
+                  assert(accessToken1.userId);
+                  assert(accessToken2.userId);
+                  assert(accessToken3.userId);
+                  next();
+                });
+              });
+            });
+          },
+          function(next) {
+            User.findById(user2Id, function(err, userFound)  {
```
bajtos: Store the instance returned by `User.create` so that you don't have to look it up again.

bajtos:
@loay the implementation looks better now. I think your tests are not covering two important cases: `User.create` and `User.updateOrCreate` when a new user is created.

Also the part `The currently active token should not be deleted.` is not implemented yet, but let's leave that as the second step in this pull request, do it only after the rest is already done and reviewed.

test/user.test.js
loay: Done. used `@example.com`

test/user.test.js

test/user.test.js

test/user.test.js
loay: Fixed

test/user.test.js
loay: Refactored and used where it belongs.

test/user.test.js
loay: I would rather test with 3 for now, since I discovered an error in my function because of that. I will remove the 3rd and use ONLY 2 just before merging.

loay:
@bajtos Refactored test cases. It reduced 74 lines even when I added 2 extra test cases.

common/models/user.js
loay: @bajtos I am not sure if that's what you meant : 
replacing the whole `if (ctx.instance)` block by this if statement??
```if (ctx.isNewInstance) return next();```

**Comment by bajtos** [comment]
@loay let's start with getting the CI builds (mostly) green.

File: common/models/user.js [hunk 25]
```diff
@@ -669,6 +669,42 @@ module.exports = function(User) {
...
 
+    // Delete old sessions once email is updated
+    UserModel.observe('before save', function beforeEmailUpdate(ctx, next) {
+      var oldEmail, newEmail, oldId, newId;
+      if (ctx.isNewInstance) return next();
+
+      if (ctx.where) {
+        ctx.hookState.id = ctx.where.id;
+        oldId = ctx.hookState.id;
+      }
```
bajtos: How about the case when `ctx.where` is not set and there is `ctx.instance` instead?

common/models/user.js
```diff
@@ -669,6 +669,42 @@ module.exports = function(User) {
...
+
+      ctx.Model.find({ where: { id: oldId }}, function(err, userInstances) {
```
bajtos: unhandled `err`

common/models/user.js
```diff
@@ -669,6 +669,42 @@ module.exports = function(User) {
       next();
     });
 
+    // Delete old sessions once email is updated
+    UserModel.observe('before save', function beforeEmailUpdate(ctx, next) {
+      var oldEmail, newEmail, oldId, newId;
+      if (ctx.isNewInstance) return next();
+
+      if (ctx.where) {
+        ctx.hookState.id = ctx.where.id;
```
bajtos: Also `ctx.where.id` may not be set, consider the case of 

```js
User.updateAll(
  { where: { name: { 'Miroslav Bajtos' } } },
  { email: 'mbajtos@cz.ibm.com' });
```

You should use `ctx.where` as-is whent it's provided. After you get back results from the database, you need to change `ctx.where` to match only the `ids` found, to prevent a race condition where somebody could add/delete another record matching the original `ctx.where` but not included in your list of users.

common/models/user.js
```diff
@@ -669,6 +669,42 @@ module.exports = function(User) {
...
+      }
+
+      ctx.Model.find({ where: { id: oldId }}, function(err, userInstances) {
+        ctx.hookState.originalUserData = userInstances.map(function(u) {
+          return { id: u.id, email: u.email };
+        });
+      });
+      next();
+
+      UserModel.observe('after save', function afterEmailUpdate(ctx, next) {
```
bajtos: Actually, I think it's better to expire the tokens _before_ performing the email change. That way, if deleting access tokens fail, the email is not changed either.

@rmg @gunjpan any opinions on this?

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
     });
   });
 
+  describe('Email Update', function() {
+    describe('User changing email property', function() {
+      var user;
+      var currentEmailCredentials = { email: 'original@example.com', password: 'bar' };
+      var updatedEmailCredentials = { email: 'updated@example.com', password: 'bar' };
+
+      function assertNoAccessTokens(done) {
```
bajtos: Move the test helpers at the end of the describe block.

```js
describe('...', function() {
  var...;
   
  it(...);
  it(...);
  
  function assertNoAccessTokens(done) {
  }
}
```

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
...
+        AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+          if (err) return done(err);
+          expect(tokens.length).to.equal(0);
+          done();
+        });
+      }
+
+      function assertUntouchedTokens(done) {
```
bajtos: ditto

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
...
+            function(err, userInstance) {
+              if (err) return done(err);
+              assertUntouchedTokens(done);
+            });
+      });
+    });
+
+    describe('sessions valid after creating new users', function() {
+      var newUserCreated;
+      function assertPreservedToken(done) {
```
bajtos: move to the bottom

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
...
+          { id: user.id,
+            email: updatedEmailCredentials.email,
+            password: updatedEmailCredentials.password },
+            function(err, userInstance) {
+              if (err) return done(err);
+              assertNoAccessTokens(done);
+            });
+      });
+
+      it('invalidates sessions when the email is changed using `replaceById`', function(done) {
```
bajtos: please add test for `replaceOrCreate` too.

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
...
+        User.replaceById(user.id,
+          { email: updatedEmailCredentials.email,
+            password: updatedEmailCredentials.password },
+            function(err, userInstance) {
+              if (err) return done(err);
+              assertNoAccessTokens(done);
+            });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `updateAttribute`', function(done) {
```
bajtos: Since updateAttribute calls updateAttributes under the hood, I think this test is redundant and can be removed

test/user.test.js
```diff
@@ -1770,6 +1770,293 @@ describe('User', function() {
...
+                  user2 = u2;
+                  user3 = u3;
+                  next();
+                });
+              });
+            });
+          },
+          function(next) {
+            User.login({ email: 'user1@example.com', password: 'u1pass' }, function(err, accessToken1) {
+              User.login({ email: 'user2@example.com', password: 'u2pass' }, function(err, accessToken2) {
```
bajtos: Unhandled `err` here and on the previous line too.

Why do we need so many users in this test? I mean why two users are not enough?

common/models/user.js
rmg: Explicit ordering here sounds like a potential source of confusing bugs. I'm imagining a scenario where the tokens are deleted in one hook and then a later hook tries to validate one of the deleted tokens and fails, resulting in the remainder of the operation failing and leaving things in a state where here is no valid token left to retry with.

I'm not intimately familiar with loopback, so the above is from a generic security token handling perspective.

loay:
`replaceOrCreate` doesn't really work with the same id. It has to create a new id, which defeats the purpose of the test. If I enforce the same id, I get a validation error:
` ValidationError: The `TestUser` instance is not valid. Details: `id` can't be set (value: 3).`

It did work though for session validation after creating new users

loay:
@bajtos PTAL. Thanks

bajtos:
> Explicit ordering here sounds like a potential source of confusing bugs. I'm imagining a scenario where the tokens are deleted in one hook and then a later hook tries to validate one of the deleted tokens and fails, resulting in the remainder of the operation failing and leaving things in a state where here is no valid token left to retry with.

> I'm not intimately familiar with loopback, so the above is from a generic security token handling perspective.

@rmg Thank you for weighting in. Before we land this patch, we need to extend the implementation to not delete the current access token (do not log out the session which initiated email change). Would that address your concern?

My concern is how to handle the case when the tokens cannot be deleted but the email was already changed? If we mark the operation as "success", then the security vulnerability remains and the user is not even aware that some of the existing sessions were not logged out. If we mark the operation as "failure", then what to do with the user record that was already modified with the new email? We can try to revert the email back to the old value, but what if this operation fail too?

Thoughts?

common/models/user.js
```diff
@@ -673,6 +673,48 @@ module.exports = function(User) {
...
+        var idsToExpire = ctx.hookState.originalUserData.filter(function(u) {
+          return u.email !== newEmail;
+        }).map(function(u) {
+          return u.id;
+        });
+        if (!idsToExpire.length) return next();
+        AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
+          if (err) return next(err);
+        });
+        next();
```
bajtos: `next` must be called only once, from `AccessToken.deleteAll` callback.

common/models/user.js
```diff
@@ -673,6 +673,48 @@ module.exports = function(User) {
...
+        });
+      }
+
+      function getOriginalUserData(userInstances) {
+        ctx.hookState.originalUserData = userInstances.map(function(u) {
+          return { id: u.id, email: u.email };
+        });
+      }
+
+      UserModel.observe('after save', function afterEmailUpdate(ctx, next) {
```
bajtos: The "after save" observer must be defined only once. Your current implementation defines new observer every time a "before save" hook is triggered.

common/models/user.js
```diff
@@ -673,6 +673,48 @@ module.exports = function(User) {
...
+          return u.id;
+        });
+        if (!idsToExpire.length) return next();
+        AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
+          if (err) return next(err);
+        });
+        next();
+      });
+
+      next();
```
bajtos: `next()` in "before save" hook must be called only after all async operations finished and we have all required data ready.

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+      var user;
+      var currentEmailCredentials = { email: 'original@example.com', password: 'bar' };
+      var updatedEmailCredentials = { email: 'updated@example.com', password: 'bar' };
+
+      beforeEach('create user then login', function createAndLogin(done) {
+        async.series([
+          function createUserWithOriginalEmail(next) {
+            User.create(
+              { email: currentEmailCredentials.email,
+                password: currentEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+                function(err, userCreated) {
+                  if (err) return next(err);
+                  user = userCreated;
+                  next();
+                });
+          },
+          function firstLoginWithOriginalEmail(next) {
+            User.login(
+              { email: currentEmailCredentials.email,
+                password: currentEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+                function(err, accessToken1) {
+                  if (err) return next(err);
+                  assert(accessToken1.userId);
+                  next();
+                });
+          },
+          function secondLoginWithOriginalEmail(next) {
+            User.login({
+              email: currentEmailCredentials.email,
+              password: currentEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+        function(err, userInstance) {
+          if (err) return done(err);
+          assertNoAccessTokens(done);
+        });
+      });
+
+      it('invalidates sessions when email is changed using `replaceAttributes`', function(done) {
+        user.replaceAttributes(
+          { 'email': updatedEmailCredentials.email,
+            'password': updatedEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

This particular place is not even consistent with the style you are using above.

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+                if (err) return done(err);
+                assertNoAccessTokens(done);
+              });
+      });
+
+      it('invalidates sessions when email is changed using `updateOrCreate`', function(done) {
+        User.updateOrCreate(
+          { id: user.id,
+            email: updatedEmailCredentials.email,
+            password: updatedEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+            function(err, userInstance) {
+              if (err) return done(err);
+              assertNoAccessTokens(done);
+            });
+      });
+
+      it('invalidates sessions when the email is changed using `replaceById`', function(done) {
+        User.replaceById(user.id,
+          { email: updatedEmailCredentials.email,
+            password: updatedEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+              if (err) return done(err);
+              assertNoAccessTokens(done);
+            });
+      });
+
+      it('invalidates sessions when the email is changed using `replaceOrCreate`', function(done) {
+        User.replaceOrCreate(
+          { id: user.id,
+            email: updatedEmailCredentials.email,
+            password: updatedEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+          if (err) return done(err);
+          assertUntouchedTokens(done);
+        });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `replaceAttributes`', function(done) {
+        user.replaceAttributes(
+          { 'email': currentEmailCredentials.email,
+            'password': currentEmailCredentials.password,
+            'firstName': 'Candy' },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+              assertUntouchedTokens(done);
+            });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `updateOrCreate`', function(done) {
+        User.updateOrCreate(
+          { id: user.id,
+            'firstName': 'Amir',
+           'email': currentEmailCredentials.email,
+           'password': currentEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

test/user.test.js
```diff
@@ -1804,6 +1805,360 @@ describe('User', function() {
...
+             if (err) return done(err);
+             assertUntouchedTokens(done);
+           });
+      });
+
+      it('keeps sessions AS IS if firstName is added using `replaceById`', function(done) {
+        User.replaceById(user.id,
+          { 'firstName': 'Miroslav',
+            'email': currentEmailCredentials.email,
+            'password': currentEmailCredentials.password },
```
bajtos: coding style: one arg per line. see https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#one-arg-per-line

I see the pair `email, password` is duplicated in many of your tests. Please consider creating a helper method and DRY.

```js
User.replaceById(
  user.id,
  userBasedOnCurrent({ firstName: 'Miroslav' }),
  function(err, userInstance) {
    // ...
  });

function userBasedOnCurrent(data) {
  return extend({
    email: currentEmailCredentials.email,
    password: currentEmailCredentials.password,
  }, data);
}
```

Also no need to quote `email` and `password` keys when building a data object.

bajtos: There are more places below where your coding style does not follow ours one-arg-per-line convention, please fix them too.

loay:
@bajtos @rmg 
About keeping the currently active token, I am not sure if I agree about that. Some apps usually give the option for passwords change, but for email, which is the username as well, I think all sessions including the current active one should be deleted. It is not too annoying to be done, and the user will psychologically feel safer when they get to log back in.

loay:
@bajtos PTAL. Thanks
Actually I removed `extend` function refactoring, as it was failing on several modules, so removed it but did other refactoring.

loay:
@slnode test please

bajtos:
> About keeping the currently active token, I am not sure if I agree about that. Some apps usually give the option for passwords change, but for email, which is the username as well, I think all sessions including the current active one should be deleted. It is not too annoying to be done, and the user will psychologically feel safer when they get to log back in.

Fair enough. We can always implement that later, if/when there is user demand.

common/models/user.js
```diff
@@ -673,6 +673,49 @@ module.exports = function(User) {
...
+    UserModel.observe('before save', function beforeEmailUpdate(ctx, next) {
+      if (ctx.isNewInstance) return next();
+      if (ctx.where) {
+        ctx.Model.find({ where: ctx.where }, function(err, userInstances) {
+          if (err) return next(err);
+          getOriginalUserData(userInstances);
+        });
+      } else if (ctx.instance) {
+        ctx.hookState.id = ctx.instance.id;
+        var oldId = ctx.hookState.id;
```
bajtos: AFAICT, `ctx.hookState.id` is not used anywhere else, please remove.

Also the name `oldId` is a bit misleading. The id is (almost) never changed, therefore there is no such thing as "old" and "new" id.

I am proposing to use `id` or `userId` instead.

common/models/user.js
```diff
@@ -673,6 +673,49 @@ module.exports = function(User) {
...
+          getOriginalUserData(userInstances);
+        });
+      }
+
+      function getOriginalUserData(userInstances) {
+        ctx.hookState.originalUserData = userInstances.map(function(u) {
+          return { id: u.id, email: u.email };
+        });
+      }
+      next();
```
bajtos: Once again. This `next` is misplaced, you need to call it from inside `Model.find` callback.

Considering that both branches use the same callback, I am suggesting the following implementation:

```js
var where = ctx.where || { id: ctx.instance.id }; // feel free to add more safety checks
ctx.Model.find({ where: where }, function(err, userInstances) {
  if (err) return next(err);
  ctx.hookState.originalUserData = userInstances.map(function(u) {
    return { id: u.id, email: u.email };
  });
  next();
});
```

common/models/user.js
```diff
@@ -673,6 +673,49 @@ module.exports = function(User) {
...
+    });
+    UserModel.observe('after save', function afterEmailUpdate(ctx, next) {
+      if (!ctx.Model.relations.accessTokens) return next();
+      var AccessToken = ctx.Model.relations.accessTokens.modelTo;
+      var  newEmail = (ctx.instance || ctx.data).email;
```
bajtos: two spaces after `var`, use one only.

common/models/user.js
```diff
@@ -673,6 +673,49 @@ module.exports = function(User) {
...
+      if (!ctx.hookState.originalUserData) return next();
+      var idsToExpire = ctx.hookState.originalUserData.filter(function(u) {
+        return u.email !== newEmail;
```
bajtos: What if we are making a partial update and email is not changed at all?

```js
user.updateAttributes({ age: user.age+1 });
```

`newEmail` is `undefined`, `u.email` is the existing email, which means you will incorrectly invalidate all access tokens. What I find surprising is that you already have a test covering this path ([here](https://github.com/strongloop/loopback/pull/2693/files#diff-a202d1e6aac601eca4c15ad2418135fdR1894)). How comes it's passing? What am I missing?

Implementation wise, I think we should detect this situation early in "before save" hook to avoid unnecessary query.

common/models/user.js
```diff
@@ -673,6 +673,49 @@ module.exports = function(User) {
...
+      }).map(function(u) {
+        return u.id;
+      });
+      if (!idsToExpire.length) return next();
+      AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
+        if (err) return next(err);
```
bajtos: No need to be so verbose, just pass `next` as the callback of `deleteAll`.

test/user.test.js
```diff
@@ -48,6 +48,7 @@ describe('User', function() {
     User = app.registry.createModel('TestUser', {}, {
       base: 'User',
       http: { path: 'test-users' },
+      forceId: false,
```
bajtos: Please add a comment explaining why we need `forceId:false`.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+              assert(accessToken2.userId);
+              next();
+            });
+          },
+        ], function(err) {
+          done();
+        });
+      });
+
+      afterEach('Deleting original user', function deleteOriginalAccounts(done) {
```
bajtos: `delete current user`

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+          },
+          function secondLoginWithOriginalEmail(next) {
+            User.login(currentEmailCredentials, function(err, accessToken2) {
+              if (err) return next(err);
+              assert(accessToken2.userId);
+              next();
+            });
+          },
+        ], function(err) {
+          done();
```
bajtos: Unhandled `err`, you need `done(err)`. Even better, simply pass `done` as the callback argument of `async.series`.

test/user.test.js
bajtos: TBH, I am not sure if it's necessary to delete these test user instances. If you think it is important, then please add another hook to delete any remaining access tokens too.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+          },
+          function(next) {
+            User.login(newUserCred, function(err, newAccessToken2) {
+              if (err) return done(err);
+              assert(newAccessToken2.id);
+              assertPreservedToken(next);
+            });
+          },
+        ], function(err) {
+          done();
```
bajtos: Unhandled `err`, pass `done` as the callback of `async.series`.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+          },
+          function(next) {
+            AccessToken.find({ userId: userNormal.userId }, function(err, tokens2) {
+              if (err) return done (err);
+              expect(tokens2.length).to.equal(1);
+              next();
+            });
+          },
+        ], function(err) {
+          done();
```
bajtos: Unhandled `err`, pass `done` as the callback of `async.series`.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+              if (err) return done(err);
+              newUserCreated = newUserInstance;
+              next();
+            });
+          },
+          function(next) {
+            User.login(newUserCred, function(err, newAccessToken) {
+              if (err) return done(err);
+              assert(newAccessToken.id);
+              assertPreservedToken(next);
```
bajtos: This check does not make sense to me. Of course the new access token is preserved, we didn't make any changes in User data.

I think you need to verify that access tokens of the first (original) user were not changed by `User.create`.

1) Create a new User
2) Verify that existing access tokens were not deleted

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+              if (err) return done(err);
+              newUserCreated = newUserInstance2;
+              next();
+            });
+          },
+          function(next) {
+            User.login(newUserCred, function(err, newAccessToken2) {
+              if (err) return done(err);
+              assert(newAccessToken2.id);
+              assertPreservedToken(next);
```
bajtos: ditto, this check does not make sense to me.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+                  user3 = u3;
+                  next();
+                });
+              });
+            });
+          },
+          function(next) {
+            User.login(
+              { email: 'user1@example.com', password: 'u1pass' },
+              function(err, accessToken1) {
```
bajtos: unhandled `err`

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+                User.login(
+                  { email: 'user2@example.com', password: 'u2pass' },
+                  function(err, accessToken2) {
```
bajtos: ditto

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+              });
+            });
+          },
+        ], function(err) {
+          done();
+        });
+      });
+    });
+
+    describe('when where object contains only the name', function() {
```
bajtos: Why is `name` so special? I think this `describe` block is extraneous, just use `it('invalidates sessions after using updateAll')` without nesting it it another `describe` block.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+          },
+          function updateSpecialUser(next) {
+            User.updateAll(
+                { name: 'Special' },
+                { email: 'superspecial@example.com' }, function(err, info) {
+                  if (err) return next (err);
+                  next();
+                });
+          },
+          function(next) {
```
bajtos: please give a name to this function, e.g. `verifyTokensOfSpecialUser`.

test/user.test.js
```diff
@@ -1804,6 +1805,332 @@ describe('User', function() {
...
+            AccessToken.find({ where: { userId: userSpecial.id }}, function(err, tokens1) {
+              if (err) return done (err);
+              expect(tokens1.length).to.equal(0);
+              next();
+            });
+          },
+          function(next) {
```
bajtos: ditto

common/models/user.js
loay: The 4 unit tests of `keeps sessions AS IS` are testing this case with different ways of update. `newEmail` is not undefined. It is the current/existing email or (original@example.com). A previous implementation was generating an `undefined` value but that was fixed.

common/models/user.js
```diff
@@ -709,7 +698,7 @@ module.exports = function(User) {
       });
       if (!idsToExpire.length) return next();
       AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
-        if (err) return next(err);
+        if (err) return next();
```
bajtos: This is dropping (ignoring) any `deleteAll` errors. Is that intentional?

I would personally use the following implementation:

```js
AccessToken.deleteAll({ userId: { inq: idsToExpire }}, next);
```

test/user.test.js
```diff
@@ -1824,25 +1827,20 @@ describe('User', function() {
...
         ], function(err) {
-          done();
-        });
-      });
-
-      afterEach('Deleting original user', function deleteOriginalAccounts(done) {
-        User.deleteById(user.id, function(err, user) {
-          if (err) return done (err);
-          done();
+          done(err);
```
bajtos: I think there is no need to create a dummy function forwarding `err` to `done`. You can write:

```js
], done);
```

Unless I am missing something?

This applies to many other places below too.

test/user.test.js
```diff
@@ -1807,9 +1809,10 @@ describe('User', function() {
 
   describe('Email Update', function() {
     describe('User changing email property', function() {
-      var user;
+      var user, originalUserToken1, originalUserToken2, newUserCreated, userSpecial, userNormal;
```
bajtos: AFAICT, `userSpecial` and `userNormal` are used only in a single test only. Could you please move the declaration of these two variables down to the test case?

test/user.test.js
```diff
@@ -1804,6 +1807,325 @@ describe('User', function() {
...
+        ], function(err) {
+          done(err);
+        });
+      });
+
+      function assertPreservedToken(done) {
+        AccessToken.find({ where: { userId: user.id }}, function(err, tokens) {
+          if (err) return done(err);
+          expect(tokens.length).to.equal(2);
+          assert(originalUserToken1);
```
bajtos: What is the intention of this `assert`? The contents of `originalUserToken1` should not change between the time it was initialized and the time you run the `assert` here. Did you perhaps want to assert that `tokens` contains both `originalUserToken1` and `originalUserToken2`?

common/models/user.js
```diff
@@ -673,6 +673,38 @@ module.exports = function(User) {
...
+      var where = ctx.where || { id: ctx.instance.id };
+      ctx.Model.find({ where: where }, function(err, userInstances) {
+        if (err) return next(err);
+        ctx.hookState.originalUserData = userInstances.map(function(u) {
+          return { id: u.id, email: u.email };
+        });
+        next();
+      });
+    });
+    UserModel.observe('after save', function afterEmailUpdate(ctx, next) {
```
bajtos: Please add an empty line between `UserModel.observe` blocks to visually delimit them.

```js
UserModel.observe('before save', .
});

UserModel.observe('after save', ...
});

```

common/models/user.js
```diff
@@ -673,6 +673,38 @@ module.exports = function(User) {
...
+        return u.id;
+      });
+      if (!idsToExpire.length) return next();
+      AccessToken.deleteAll({ userId: { inq: idsToExpire }}, function(err, info) {
+        if (err) return next();
+        next();
+      });
+    });
+
+
```
bajtos: Too many empty lines, keep one empty line only.

test/user.test.js
loay: Yes absolutely. I didn't just want to make sure that the 2 tokens are still there, but also the exact same tokens.

loay: I am tweaking the code a bit to ensure it has both tokens.

loay:
test please

doublemarked:
We've discovered this new behavior after bumping Loopback versions. Unfortunately it was not caught by our automated tests (though in the future it will be). 

I realize this was in response to a security issue, but I feel like somehow there could be a better process to introducing breaking changes. Maybe at least clear announcements/notifications when things like this are landing? How do we as developers guard against changes like this breaking our software?

doublemarked:
I guess one more thing... @bajtos you wrote,

> Also the part `The currently active token should not be deleted`. is not implemented yet, but let's leave that as the second step in this pull request, do it only after the rest is already done and reviewed.

This is precisely the part that is breaking existing code, because it requires us to rework our frontend to accommodate tokens that suddenly get invalidated. Will such a fix be anticipated in the future? We're trying to figure out whether we should be reworking our frontend to anticipate/explain why the user suddenly needs to reauthenticate, or whether we can hold out on 2.34.1 until such a fix lands.

bajtos:
Hi @doublemarked, I apologise for the troubles we have caused.

> I realize this was in response to a security issue, but I feel like somehow there could be a better process to introducing breaking changes. Maybe at least clear announcements/notifications when things like this are landing? How do we as developers guard against changes like this breaking our software?

I believe we are announcing security-related fixes, @loay @rmg @gunjpan could one of you please provide more information on the channels we are using? Did we announce this patch?

@doublemarked do you have any suggestion/advise on what channel would work best for developers like you? I think it may be best to discuss this with broader community, would you mind starting the discussion in a new github issue?

> This is precisely the part that is breaking existing code, because it requires us to rework our frontend to accommodate tokens that suddenly get invalidated. Will such a fix be anticipated in the future? We're trying to figure out whether we should be reworking our frontend to anticipate/explain why the user suddenly needs to reauthenticate after changing their email address, or whether we can hold out on 2.34.1 until such a fix lands.

I agree the current situation, where a change of email or password (see #3021) logs out the user, creates a rather poor experience and we should fix this soon. One of the reasons why this was not made in the original pull requests, is that we don't have a good solution for getting the current access token. The easiest way forward is to wait for https://github.com/strongloop/loopback/pull/3023 to land, after which we can use `ctx.options.accessToken` to exclude the current access token from the `delete` query.

Let's create a new issue so that this is not forgotten: https://github.com/strongloop/loopback/issues/3034

doublemarked:
Thank you for the reply, @bajtos. I do see/understand how #3023 is a barrier for the correct behavior. I need to catch up on the communication there, but I'm happy to see you prioritizing it. We will probably hang on at v2.34.1 for now and monitor how #3023 and #3034 progress.

Regarding communication channels - sure, great idea. I've opened #3042 on this topic. 

Thank you!

PR #2696, opened 2016-08-31
Use strong-remoting's new TypeRegistry

bajtos:
The first two commits fix issues discovered by strong-remoting's duplicate checker.

The last commit reworks model registration to use the new strong-remoting API.

Requires https://github.com/strongloop/strong-remoting/pull/343

to: @gunjpan @richardpringle @deepakrkris Please review.
cc: @ritch @STRML @0candy @davidcheung

Connect to strongloop-internal/scrum-loopback#885

**TODO before merging**
 - [x] wait for https://github.com/strongloop/strong-remoting/pull/343 and
https://github.com/strongloop/loopback-connector-remote/pull/58 to be landed
 - [x] make a new alpha release of both strong-remoting and loopback-connector-remote
 - [x] bump up minimum dependency versions for these two modules in package.json

PR #2697, opened 2016-08-31
Fix app.enableAuth and test/user [2.x]

bajtos:
This is a partial back-port of #2696.

- Fix a typo in "app.enableAuth" that caused the method to not detect
the situation when e.g. the built-in User model is already attached
to a datasource.

- don't attach User model twice in `test/user.test.js`

@gunjpan @richardpringle @deepakrkris since you are reviewing the other pull request, could one of you review this one too?


richardpringle:
LGTM

PR #2539, opened 2016-07-25
adding upsertWithWhere

mountain1234585:
Requires https://github.com/strongloop/loopback-datasource-juggler/pull/1001

Issue #1806, opened 2015-11-10
Regression: dynamic coercion of Model will clobber property values on update


fabien:
I finally went through a full update cycle with my apps, and in updating from 2.18.0 to 2.23.0 I discovered the following undesired (and BC breaking) behavior:

With a property that has some default value (either static or a fn/dynamic value):
```js
"properties": {
  "foo": {
    "type": "string",
    "default": "bar"
  }
}
```

```js
// create receives a new instance, with defaults filled in
// which we'll overwrite by our new data
this.request.post('/api/things')
  .send({ foo: 'other' }).end(function(err, res) {
  res.body.foo.should.equal('other');
  // id: 1234567890
  next();
});

// updateAttributes now receives a new instance, with defaults filled in
// but this time, we don't have any value for 'foo' set explicitly
// and thus the default will be used for the update(!)
this.request.put('/api/things/1234567890')
  .send({ biz: 'baz' }).end(function(err, res) {
  res.body.biz.should.equal('baz');
  res.body.foo.should.equal('other'); // Fail: overwritten by default: 'bar'
  next();
});
```
I think it's related to this commit: https://github.com/strongloop/strong-remoting/commit/34b0665841d763944dca3e9b4b2033366ea8a0c0

/cc @bajtos @ritch 

---

See https://github.com/strongloop/loopback/issues/1806#issuecomment-217393388 for the proposed solution.

 - [ ]  Implement support for `arg.model` in loopback-swagger
 - [ ] Rework remoting metadata in LoopBack, use `{ type: 'object', model: modelName }` everywhere. Make sure to add unit-tests reproducing the problem and then demonstrating that the problem is fixed.

The last item from the proposal is tracked by a standalone issue: https://github.com/strongloop/loopback/issues/2313

> Fix swagger metadata for updateAttributes and friends to correctly indicate all properties are optional. Check other schema parts like "default" values, we may need to remove them too.

fabien:
I was able to identify where the issue exactly happens, but the fix is unsatisfactory because it prevents dynamic (object) type conversion. 

However, it's extremely difficult to not end up with the issue here - how can we declare that the coerced values are in fact just *defaults* and that they should *not* overwrite any existing data on update?

fabien:
@raymondfeng any thoughts on this? The fix I mention over at https://github.com/strongloop/strong-remoting/issues/254 works fine, but I'm not sure if this has any actual drawbacks at the moment.

raymondfeng:
[...]

PR #2702, opened 2016-09-02
Fix remoting metadata for "data" arguments [2.x]

bajtos:
Fix the definition of "data" argument to

    { type: 'object', model: modelName, ... }

That way strong-remoting passed the request body directly to the model
method (does not create a new model instance), but the swagger will
still provide correct schema for these arguments.

This fixes a bug where upsert in relation methods was adding default
property values to request payload.

Connect to strongloop/loopback#1806
Requires also https://github.com/strongloop/loopback-swagger/pull/55 to be released around the same time

@0candy please review

This patch clashes with https://github.com/strongloop/loopback/pull/2539 which is in review too. We need to rebase the second patch to be landed after the first one is merged.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -486,7 +486,7 @@ module.exports = function(registry) {
     define('__create__' + relationName, {
       isStatic: false,
       http: {verb: 'post', path: '/' + pathName},
-      accepts: {arg: 'data', type: toModelName, http: {source: 'body'}},
+      accepts: {arg: 'data', type: 'object', model: toModelName, http: {source: 'body'}},
       description: format('Creates a new instance in %s of this model.', relationName),
       accessType: 'WRITE',
       returns: {arg: 'data', type: toModelName, root: true}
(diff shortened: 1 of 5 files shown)


0candy:
@bajtos  The code change LGTM. But does this code need to be in 3.x?

bajtos:
>  But does this code need to be in 3.x?

I'll forward-port to 3.x later. I started with a patch sent against 2.x to see how downstream dependencies are handling the change. There are very few dependents using LooBack 3.x. It has happened to us very recently that we landed a patch to 3.x because all looked well and then there was a bunch failures after back-porting to 2.x

bajtos:
I compared swagger documents produced before this patch and after this patch. The only difference is that with this patch in place, the responses for relation methods include the description from remoting metadata, which is a good thing IMO.

```diff
(...)
@@ -267,6 +270,7 @@
           description: Model instance data
           required: false
           schema:
+            description: Model instance data
             $ref: '#/definitions/Category'
       responses:
         '200':
@@ -373,6 +377,7 @@
           description: An object of model property name/value pairs
           required: false
           schema:
+            description: An object of model property name/value pairs
             $ref: '#/definitions/Category'
         - name: id
           in: path
@@ -399,6 +404,7 @@
(...)
```

Issue #2350, opened 2016-05-18
Remote methods defined in model JSON file are not inherited


bajtos:
1. `yo loopback` to create a new app
2. `yo loopback:model` to create a model
3. `yo loopback:remote-method` to create a remote method, fill in some simple implementation
4. `yo loopback:model` to create another model inheriting from the model created in step 2
5. `node .` to run the app
6. Open the explorer http://127.0.0.1:3000/explorer/
7. The model created in step 2 has a custom remote method, the model created in step 4 does not.

Related: #2270 and #157

PR #2703, opened 2016-09-02
Fix remote method inheritance

0candy:
Connect to https://github.com/strongloop/loopback/issues/2350

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -121,7 +121,7 @@ Registry.prototype.createModel = function(name, properties, options) {
   var model = BaseModel.extend(name, properties, options);
   model.registry = this;
 
-  this._defineRemoteMethods(model, options.methods);
+  this._defineRemoteMethods(model, model.settings.methods);
 
   return model;
 };
@@ -242,6 +242,15 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
       'The configuration of `%s` is missing {{`dataSource`}} property.\n' +
       'Use `null` or `false` to mark models not attached to any data source.',
       modelName);
+  }
+
+  var newMethodNames = config.methods && Object.keys(config.methods);
+  var hasNewMethods = newMethodNames && newMethodNames.length;
+  var hasDescendants = this.getModelByType(ModelCtor) !== ModelCtor;
+  if (hasNewMethods && hasDescendants) {
+    g.warn(
+      'Child models of `%s` will not inherit newly defined remote methods %s.',
+      modelName, newMethodNames);
   }
 
   // Remote methods
--- a/test/loopback.test.js
+++ b/test/loopback.test.js
@@ -7,6 +7,9 @@ var it = require('./util/it');
 var describe = require('./util/describe');
 var Domain = require('domain');
 var EventEmitter = require('events').EventEmitter;
+var loopback = require('../');
+var expect = require('chai').expect;
+var assert = require('assert');
 
 describe('loopback', function() {
   var nameCounter = 0;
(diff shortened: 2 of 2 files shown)


0candy:
@bajtos PTAL

bajtos:
@0candy I think you may be missing one important fact here. There are two model-related steps when setting up an app via loopback-boot: 1) all models are created 2) all models are configured.

Using your test code as an example, it means this is the order in which create/configure API is called:

```js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

I suspect your implementation may not work correctly in this scenario.

lib/registry.js
```diff
@@ -121,7 +121,7 @@ Registry.prototype.createModel = function(name, properties, options) {
   var model = BaseModel.extend(name, properties, options);
   model.registry = this;
 
-  this._defineRemoteMethods(model, options.methods);
+  this._defineRemoteMethods(model, model.settings.methods);
```
bajtos: Interesting. So this assumes that when a new model is created, juggler will propagate `settings.methods` from the parent model to the child model.

What will happen in this case? Could you please add a test for that?

```js
var BaseModel = loopback.createModel('BaseModel', {}, {
  methods: {
    greet: {
      http: { path: '/greet' },
    },
  },
});

var ChildModel = loopback.createModel('ChildModel', {}, {
  methods: {
    hello: {
      http: { path: '/hello' },
    },
  },
});

// ChildModel should have both "greet" and "hello"
```

lib/registry.js
```diff
@@ -223,6 +223,10 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
       'must be an object', modelName);
   }
 
+  if (typeof config.methods === 'object' && config.methods !== null) {
+    settings.methods = config.methods;
```
bajtos: IIUC, this allows developers to modify remoting metadata in `server/model-config.json`. However, this code is called after `_defineRemoteMethods`, therefore I think it won't work in this case:

```js
var BaseModel = loopback.createModel('BaseModel');
var MyCustomModel = loopback.createModel('MyCustomModel');

loopback.configureModel(BaseModel, {
  dataSource: null,
  methods: {
    greet: {
      http: { path: '/greet' },
    },
  },
});

loopback.configureModel(MyCustomModel, {
  dataSource: null,
  methods: {
    greet: {
      http: { path: '/greet' },
    },
  },
});

// I would expect MyCustomModel to have both "greet" and "hello" methods,
// but I suspect MyCustomModel won't have any remote methods
```

0candy: In the above case, wouldn't they each have their own `greet` method?

0candy:
>>>
```.js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

0candy:
>>>
```.js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

0candy:
>>>
```.js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

0candy:
>>>
```.js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

0candy:
>>>
```.js
createModel('BaseModel', ...);
createModel('MyCustomModel', ...);
configureModel(BaseModel, ...);
configureModel(MyCustomModel, ...);
```

lib/registry.js
0candy: Yes, that's correct. Added test case.

lib/registry.js
bajtos: Sorry, the `MyCustomModel` method should have been called `hello`. I forgot to change the name when copy-pasting :(

test/loopback.test.js
```diff
@@ -608,4 +611,96 @@ describe('loopback', function() {
       expect(methodNames).to.include('prototype.instanceMethod');
     });
   });
+
+  describe('Remote method inheritance', function() {
+    function getAllMethodNamesWithoutClassName(Model) {
```
bajtos: We usually put helper methods at the end of the describe block, see https://github.com/strongloop/loopback-contributor-docs/blob/671a797623450fa94f5e3da05d12c34fc8cd5aa1/style-guide.md#layout-of-test-files. It mentions before/after hooks only, but the same principle applies to helper methods. Could you send a PR to clarify that?

test/loopback.test.js
```diff
@@ -608,4 +611,96 @@ describe('loopback', function() {
...
+      return Model.sharedClass.methods().map(function(m) {
+        return m.stringName.replace(/^[^.]+\./, ''); // drop the class name
+      });
+    }
+    var app;
+
+    beforeEach(function() {
```
bajtos: before/after hooks should be named, see https://github.com/strongloop/loopback-contributor-docs/blob/671a797623450fa94f5e3da05d12c34fc8cd5aa1/style-guide.md#layout-of-test-files

test/loopback.test.js
```diff
@@ -608,4 +611,96 @@ describe('loopback', function() {
...
+          },
+        },
+      });
+      var methodNames = getAllMethodNamesWithoutClassName(MyCustomModel);
+
+      expect(methodNames).to.include('greet');
+      expect(methodNames).to.include('hello');
+    });
+
+    it('create and config model inherits method from base model', function() {
```
bajtos: I find the current test name not clear enough, it's not obvious what is important/specific in this particular test. I am proposing the following:

```js
it('inherits remote methods defined via configureModel')
```

test/loopback.test.js
```diff
@@ -608,4 +611,96 @@ describe('loopback', function() {
...
+        return m.stringName.replace(/^[^.]+\./, ''); // drop the class name
+      });
+    }
+    var app;
+
+    beforeEach(function() {
+      app = loopback({ localRegistry: true });
+    });
+
+    it('create model inherits method from base model', function() {
```
bajtos: In the light of https://github.com/strongloop/loopback/pull/2703/files#r77774974, I am proposing

```js
it('inherits remote methods defined via createModel`)
```

test/loopback.test.js
```diff
@@ -608,4 +611,96 @@ describe('loopback', function() {
...
+          },
+        },
+      });
+      var methodNames = getAllMethodNamesWithoutClassName(MyCustomModel);
+
+      expect(methodNames).to.include('greet');
+      expect(methodNames).to.include('hello');
+    });
+
+    it('configure model inherits method from base model', function() {
```
bajtos:
```js
it('inherits methods defined via configureModel after child model was created')
```

lib/registry.js
```diff
@@ -244,8 +248,12 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
       modelName);
   }
 
+  var baseName = ModelCtor.base.modelName || 'PersistedModel';
+  BaseModel = this.getModel(baseName);
+  var model = BaseModel.extend(modelName, {}, settings);
```
bajtos: You are effectively re-creating the child model here, that looks very wrong to me. I am surprised the tests are passing, considering that they store the original model constructor in  `var MyModel`. I guess this is a quirk of loopback-datasource-juggler, where it merges model configuration when a model with the same name is defined.

bajtos: I think the correct way of handling remote-method inheritance is to extend strong-remoting SharedClass to support the concept of inheritance too. I.e. `MyCustomModel.sharedClass` should be linked to `Base.sharedClass` in such way that `MyCustomModel` inherits all methods from `Base`.

bajtos:
@0candy I think we should add few more tests to cover and show what happens when the child provides different metadata for the same remote method.

```js
var BaseModel = loopback.createModel('BaseModel', {}, { methods: {
    greet: {
      http: { path: '/greet' },
    },
});

var MyCustomModel = loopback.createModel('MyCustomModel', { methods: {
    greet: {
      http: { path: '/hello' },
    },
});
```

Thoughts?

bajtos:
To move this pull request forward, I am proposing to keep the first part of the fix only, i.e. inherit remote methods at `createModel` time only. I think this is the most frequent use case which is worth fixing on its own.

Fixing remote methods at `configureModel` time can be left for later, when we get users complaining. To make the issue more obvious, perhaps we can add a bit of code to detect when `configureModel` is adding remote methods to a model that has child models, and print a warning to let the user know?

```js
...configureModel = function(...) {
  // ...
  
  var hasNewMethods = config.methods && Object.keys(config.methods).length;
  var hasDescendants = this.getModelByType(ModelCtor) !== ModelCtor;
  if (hasNewMethods && hasDescendants) {
    // print a warning including model name
  }

  this._defineRemoteMethods(ModelCtor, config.methods);
}
```

It would be even better to print this warning in `ModelCtor.remoteMethod`, but I am not sure if we have access to the registry there (if `Model.registry` is set early enough)/

@0candy thoughts? I may be making this task more complex than necessary, please let me know which parts make sense to you and which do you find as not needed.

0candy:
Agree with above 2 comments, making fixes now.

0candy:
@bajtos PTAL

lib/registry.js
```diff
@@ -244,6 +244,14 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
       modelName);
   }
 
+  var hasNewMethods = config.methods && Object.keys(config.methods).length;
+  var hasDescendants = this.getModelByType(ModelCtor) !== ModelCtor;
+  if (hasNewMethods && hasDescendants) {
+    g.warn(
+      'Child models of `%s` will not inherit newly defined remote methods %j.',
+      modelName, config.methods);
```
bajtos: I think dumping the whole method-definition object will clutter the output. I am proposing to print only method names?

```js
g.warn(
  'Child models of `%s` will not inherit newly defined remote methods %j.',
  modelName, Object.keys(config.methods));
```

To avoid calling `Object.keys` twice, it's better to store it in a variable, e.g. `newMethodNames`.

test/loopback.test.js
```diff
@@ -608,4 +611,133 @@ describe('loopback', function() {
...
+          },
+        },
+      });
+      var methodNames = getAllMethodNamesWithoutClassName(MyCustomModel);
+
+      expect(methodNames).to.not.include('greet');
+      expect(methodNames).to.include('hello');
+    });
+
+    it('warn remote methods defined via configureModel after child model was ' +
```
bajtos: The test does not verify that a warning was printed.

lib/registry.js
```diff
@@ -244,12 +244,13 @@ Registry.prototype.configureModel = function(ModelCtor, config) {
       modelName);
   }
 
-  var hasNewMethods = config.methods && Object.keys(config.methods).length;
+  var newMethodNames = Object.keys(config.methods);
+  var hasNewMethods = config.methods && newMethodNames.length;
```
bajtos:
```
$ node
> Object.keys(undefined)
TypeError: Cannot convert undefined or null to object
```

I think you need to rewrite this a bit:

```js
var newMethodNames = config.methods && Object.keys(config.methods);
var hasNewMethods = newMethodNames.length;
```

test/loopback.test.js
```diff
@@ -698,7 +698,7 @@ describe('loopback', function() {
       expect(methodNames).to.include('hello');
     });
 
-    it('warn remote methods defined via configureModel after child model was ' +
+    it('remote methods defined via configureModel after child model was ' +
```
bajtos: `it('does not inherit remote methods defined via ...')`

Remember, the test name should read like an English sentence.

bajtos:
The CI seems to be failing everywhere, PTAL.

Two more minor comment to address, the patch LGTM otherwise. No more reviews are needed.

0candy:
CI is failing on windows node 10. There is already an issue created for that (https://github.com/strongloop-internal/scrum-loopback/issues/1052). Merging this PR.

PR #2407, opened 2016-06-07
Add globalization

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/910
Connect to https://github.com/strongloop/loopback/issues/2422

PR #2704, opened 2016-09-02
Add rootDir to loopack file

0candy:
Add SetRootDir to lib/loopback.js

### DIFF
--- a/lib/loopback.js
+++ b/lib/loopback.js
@@ -7,6 +7,9 @@
  * Module dependencies.
  */
 
+var path = require('path');
+var SG = require('strong-globalize');
+SG.SetRootDir(path.join(__dirname, '..'), { autonomousMsgLoading: 'all' });
 var express = require('express');
 var loopbackExpress = require('./server-app');
 var proto = require('./application');


0candy:
PTAL

bajtos:
I think the approach where each source file calls `require('strong-globalize')` is wrong.

It is not safe to assume that `index.js` or `lib/loopback.js` will be called first. For example, loopback-boot can load `common/models/user.js` directly.

Secondly, we are duplicating `autonomousMsgLoading` setting in multiple places.  That will be difficult to maintain in the future.

We should have a shared file`lib/globalize.js` that calls both `require('strong-globalize')` and `SetRootDir`, and then require this file from all other files where we need strong-globalize instance. That's the easiest and most reliable way how to ensure `SetRootDir` is called when the first loopback file is loaded.

```js
// lib/globalize.js
var SG = require('strong-globalize');
SG.SetRootDir(path.join(__dirname, '..'), { autonomousMsgLoading: 'all' });
module.exports = SG();

// lib/loopback.js
var g = require('./globalize');

// lib/registry.js
var g = require('./globalize');

// etc.
```

0candy:
We had this discussion before.
https://github.com/strongloop/loopback/pull/2407#discussion_r70572500

I agree with you that it should be in a common file for easy maintenance. However, it breaks the `strong-globalize` string extraction.
See https://github.com/strongloop/loopback/pull/2407#discussion_r71798595

Maybe we can request this feature in `strong-globalize`?

bajtos:
> We had this discussion before.
https://github.com/strongloop/loopback/pull/2407#discussion_r70572500

> I agree with you that it should be in a common file for easy maintenance. However, it breaks the `strong-globalize` string extraction.
See https://github.com/strongloop/loopback/pull/2407#discussion_r71798595

> Maybe we can request this feature in `strong-globalize`?

@0candy oh, I did not read your reply in https://github.com/strongloop/loopback/pull/2407#discussion_r71798595 until now - I am unable to keep up with all GH notifications :( I read only those containing my GH handle :)

I agree we should request this feature from strong-globalize, could you please open a GH issue to keep track of this request? TBH I think it's very likely we will have to implement it ourselves, but that's fine with me.

I am thinking how to fix the problem with the current strong-globalize version. Would the following approach work?

```js
// lib/globalize.js
module.exports = function(SG) {
  SG.SetRootDir(path.join(__dirname, '..'), { autonomousMsgLoading: 'all' });
  return SG();
}

// lib/loopback.js
var g = require('./globalize')(require('strong-globalize'));

// lib/registry.js
var g = require('./globalize')(require('strong-globalize'));

0candy:
I created an enhancement issue in `strong-globalize`.
https://github.com/strongloop/strong-globalize/issues/87

As for your suggested approach, will need to test it out to find out. @Setogit Do you know if @bajtos suggestion will work?

Setogit:
var g = require('./globalize')(require('strong-globalize'));
generates a different syntax tree the current strong-globalize supports. we could implement that form as well.  the question is how is it possible to make sure `require('strong-globalize')` is actually reused without getting into clients source code analysis?

bajtos:
Let's move the discussion about strong-globalize enhancements to https://github.com/strongloop/strong-globalize/issues/87

bajtos:
As for this pull request, I don't fully understand what issue we are trying to fix and whether it's still an issue after https://github.com/strongloop/strong-globalize/pull/86 was landed. I'll leave it up to you @0candy and @Setogit to decide whether this patch is worth landing.

bajtos:
@0candy I think this pull request is ready to be reworked to follow the new approach that's supported by strong-globalize since  strongloop/strong-globalize#86. Thoughts?

0candy:
Closing PR. See https://github.com/strongloop/loopback/pull/2754

Issue #2452, opened 2016-06-17
Verify URL incorrect with "/" restApiRoot


jonathan-chin:
I've set my restApiRoot to "/", expecting everything to go through http://api.example.com (instead of http://www.example.com/api)

everything seems to work right except that the url generated for verification emails includes a double slash:

http://api.example.com//users/verify etc etc

It's obviously expecting not to be run at the root level. I also tried passing in an empty string "" in the options of verify but the comparison operator interpreted it as false.

how can I avoid the double slash in the generated verify url if my restapiroot is "/"?

gunjpan:
@jonathan-chin : Hi, looks like you have stumbled upon a valid bug as there seems to be no handling of `/` as restApiRoot in creating verify link ( [code](https://github.com/strongloop/loopback/blob/45e523d66c629b3a389b84b01d28f83adbd0a1fd/common/models/user.js#L394-L405)):
```javascript
options.verifyHref = options.verifyHref ||
      options.protocol +
      '://' +
      options.host +
      displayPort +
      options.restApiRoot +
      userModel.http.path +  // for User model this will be: `/users`
      userModel.sharedClass.findMethodByName('confirm').http.path +
      '?uid=' +
      options.user.id +
      '&redirect=' +
      options.redirect;
```

This would be a simple fix. Would you like to submit a patch? I will be happy to assist to get it landed.
Meanwhile, we will be planning to get it fixed, stay tuned. Thank you.

jonathan-chin:
I fixed it with a simple ternary comparison:

```
var restApiRoot = options.restApiRoot == "/" ? "" : options.restApiRoot;
options.verifyHref = options.verifyHref ||
options.protocol +
'://' +
options.host +
':' +
options.port +
restApiRoot +
userModel.http.path +
userModel.sharedClass.find('confirm', true).http.path +
'?uid=' +
options.user.id +
'&redirect=' +
options.redirect;
```

However, I'm still on version 2.22.2. I'd like to submit a patch for 3.0 but I'd have to set everything up for it.

gunjpan:
@jonathan-chin : You can follow the contribution guideline and cloning/forking from GitHub would give you 3.x. Let me know, should you need further help. Thanks.

bajtos:
Fixed & released in `loopback@2.34.1`.

PR #2712, opened 2016-09-05
Added option to set restApiRoot to "" for servers run at root.

jonathan-chin:
Previously, if you ran loopback at the root directory, verification urls wouldn't be generated correctly. For example:

`https://api.example.com//users/etc-etc`
instead of
`https://api.example.com/users/etc-etc`
(note the double slash)

This was due to empty strings ("") resolving to false in javascript. A simple if statement fixes it.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -400,8 +400,9 @@ module.exports = function(User) {
     var app = userModel.app;
     options.host = options.host || (app && app.get('host')) || 'localhost';
     options.port = options.port || (app && app.get('port')) || 3000;
-    options.restApiRoot = options.restApiRoot || (app && app.get('restApiRoot')) || '/api';
-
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/2718.

Note that you should set `restApiRoot` to `/` if you want to mount your API in the root.

jonathan-chin:
hi @bajtos 

Sorry, setting the ```restApiRoot``` to ```/``` also causes the double slash in the verifyHref. Maybe the patch should check for ```restApiRoot === "/"``` and set it to ```""```

Original post: #2452

PR #2716, opened 2016-09-06
Rework email validation to use isemail [2.x]

### LINKED ISSUE #1723: Too strict validation on User.email

bajtos:
Drop hand-crafted RegExp in favour of a 3rd-party module that supports RFC5321, RFC5322 and other relevant standards.

This pull request is based on #1782 (kudos to @vdust) and fixes #1723 (cc @daankets).

cc @Amir-61 @davidcheung @superkhau

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -9,6 +9,7 @@
 
 var g = require('strong-globalize')();
 
+var isEmail = require('isemail');
 var loopback = require('../../lib/loopback');
 var utils = require('../../lib/utils');
 var path = require('path');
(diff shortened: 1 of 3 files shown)


bajtos:
For posterity, I had to downgrade `isemail@^2.2.1` to `isemail@^1.2.0` because the 2.x series does not support ES5 runtimes (Node v0.10, PhantomJS).

PR #2718, opened 2016-09-06
Fix double-slash in confirmation URL [2.x]

bajtos:
Fix the code building the URL used in the email-verification email
to prevent double-slash in the URL when e.g. restApiRoot is '/'.

Before:

    http://example.com//users/confirm?...

Now:

    http://example.com/users/confirm?...

This patch is based on #2712 from @jonathan-chin  

Connect to #2452

@loay @richardpringle please review

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1551,6 +1551,41 @@ describe('User', function() {
 
         done();
       });
+
+      it('should squash "//" when restApiRoot is "/"', function(done) {
+        var emailBody;
+        User.afterRemote('create', function(ctx, user, next) {
+          assert(user, 'afterRemote should include result');
+
+          var options = {
+            type: 'email',
+            to: user.email,
+            from: 'noreply@myapp.org',
+            redirect: '/',
+            host: 'myapp.org',
+            port: 3000,
+            restApiRoot: '/',
+          };
+
+          user.verify(options, function(err, result) {
+            if (err) return next(err);
+            emailBody = result.email.response.toString('utf-8');
+            next();
+          });
+        });
+
+        request(app)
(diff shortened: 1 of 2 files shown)


bajtos:
Cross-posting https://github.com/strongloop/loopback/pull/2712#issuecomment-244938277

> Sorry, setting the ```restApiRoot``` to ```/``` also causes the double slash in the verifyHref. Maybe the patch should check for ```restApiRoot === "/"``` and set it to ```""```

@jonathan-chin In this patch, I am fixing the code building `verifyHref` to squash double slashes, see
https://github.com/strongloop/loopback/pull/2718/files#diff-5d81af95449574544e7317f3eda2badeL404

Is there any other place where we need to take care of this too?

jonathan-chin:
@bajtos I haven't come across any other spots with double slashes. Thanks!

test/user.test.js
```diff
@@ -1551,6 +1551,41 @@ describe('User', function() {
...
+            emailBody = result.email.response.toString('utf-8');
+            next();
+          });
+        });
+
+        request(app)
+          .post('/test-users')
+          .expect('Content-Type', /json/)
+          .expect(200)
+          .send({email: 'bar@bat.com', password: 'bar'})
```
loay: I thought we were supposed to use `@example.com` in test cases. I am not sure if that's applicable  in this case.

loay:
One minor comment. the rest LGTM

test/user.test.js
bajtos: Good point. I copied and modified an existing test, but that's not an excuse to not follow our best practices. Will fix.

PR #2729, opened 2016-09-07
Fix data argument for upsertWithWhere

Amir-61:
* Fix data argument for upsertWithWhere

Related PR #2727

CC: @bajtos


Amir-61:
Since it is a simple fix; I land it without review.

bajtos:
@Amir-61 Awesome, thank you for taking care of this 🙇

Issue #2525, opened 2016-07-20
EPIC: KeyValue models


ritch:
### Problem

I should be able to easily read and write values in cache-friendly data stores. I should have control over exactly when each value expires. I should be able to choose my store.

### Example

```js
var wait = require('timer-as-promised');
var Color = app.createModel({
  base: 'KeyValueModel',
  keyValue: {
    ttl: {
      default: 120
    },
    namespace: 'colors'
  }
});
var redis = app.createDataSource('redis', {
  host: 'localhost',
  port: 6379
});
Color.attachTo(redis);

var SECONDS_TO_LIVE = 60;
var key = Math.random().toString().split('.')[1];
var myValue = new Color({
  key: key,
  value: 'blue',
});

myValue
  .setWithTTL(SECONDS_TO_LIVE)
  .then(function() {
    return myValue.ttl();
  })
  .then(function(ttl) {
    console.log(ttl); // 60
    return myValue.expire(120);
  })
  .then(function() {
    return wait(120);
  })
  .then(function() {
    return myValue.get();
  })
  .then(function() {
    console.log(myValue.value); // => undefined
  });
```

## Tasks
- [x] The first cut
- https://github.com/strongloop/loopback-datasource-juggler/pull/1023 (master)
- https://github.com/strongloop/loopback-datasource-juggler/pull/1031 (2.x)
- [x] Globalize the first cut
- https://github.com/strongloop/loopback-datasource-juggler/pull/1056
- https://github.com/strongloop/loopback-datasource-juggler/pull/1057
- [x] prototype methods (syntactic sugar)
- [x] `Color.ttl(key)` and `color.ttl()`
- https://github.com/strongloop/loopback-datasource-juggler/pull/1030
- https://github.com/strongloop/loopback-connector-kv-redis/pull/9
- [x] `Color.keys()` - https://github.com/strongloop/loopback/issues/2592
- [x] API docs - both in juggler and loopback
- https://github.com/strongloop/loopback/pull/2743
- https://github.com/strongloop/loopback-datasource-juggler/pull/1093
- [x] REST API in loopback repo - https://github.com/strongloop/loopback/pull/2594
- [x] KV memory connector example - https://github.com/strongloop-internal/scrum-loopback/issues/1017
- [x] Redis KV connector example - https://github.com/strongloop-internal/scrum-loopback/issues/1018
- [ ] Blog post on KV connectors - https://github.com/strongloop-internal/scrum-loopback/issues/1015

[...]

PR #2743, opened 2016-09-14
Add docs for KeyValue model

superkhau:
@bajtos PTAL -- it's WIP, but I figure it's better to get some feedback from you now.

cc @crandmck 

Connect to strongloop/loopback#2525

### DIFF
--- a/common/models/key-value-model.js
+++ b/common/models/key-value-model.js
@@ -1,36 +1,163 @@
 var g = require('strong-globalize')();
 
+/**
+ * Data model for key-value databases.
+ *
+ * @class KeyValueModel
+ * @inherits {Model}
+ */
+
 module.exports = function(KeyValueModel) {
-  // TODO add api docs
+  /**
+   * Return the value associated with a given key.
+   *
+   * @param {String} key Key to use when searching the database.
+   * @options {Object} options
+   * @callback {Function} callback
+   * @param {Error} err Error object.
+   * @param {Any} result Value associated with the given key.
+   * @promise
+   *
+   * @header KeyValueModel.get(key, cb)
+   */
   KeyValueModel.get = function(key, options, callback) {
     throwNotAttached(this.modelName, 'get');
   };
 
-  // TODO add api docs
+  /**
+   * Persist a value and associate it with the given key.
+   *
+   * @param {String} key Key to associate with the given value.
+   * @param {Any} value Value to persist.
+   * @options {Number|Object} options Optional settings for the key-value
+   *   pair. If a Number is provided, it is set as the TTL (time to live) in ms
+   *   (milliseconds) for the key-value pair.
+   * @property {Number} ttl TTL for the key-value pair in ms.
+   * @callback {Function} callback
+   * @param {Error} err Error object.
+   * @promise
+   *
+   * @header KeyValueModel.set(key, value, cb)
+   */
   KeyValueModel.set = function(key, value, options, callback) {
     throwNotAttached(this.modelName, 'set');
   };
 
-  // TODO add api docs
+  /**
+   * Set the TTL (time to live) in ms (milliseconds) for a given key. TTL is the
+   * remaining time before a key-value pair is discarded from the database.
+   *
+   * @param {String} key Key to use when searching the database.
+   * @param {Number} ttl TTL in ms to set for the key.
+   * @options {Object} options
+   * @callback {Function} callback
+   * @param {Error} err Error object.
+   * @promise
+   *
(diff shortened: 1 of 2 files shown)


common/models/key-value-model.js
```diff
@@ -1,36 +1,93 @@
...
+   * @promise
+   */
   KeyValueModel.keys = function(filter, options, callback) {
     throwNotAttached(this.modelName, 'keys');
   };
 
-  // TODO add api docs
+  /*!
+   * Iterate through all keys for a given filter.
+   */
```
superkhau: @bajtos Do we need to document this? Could you add a comment with a valid description if so?

common/models/key-value-model.js
```diff
@@ -1,36 +1,93 @@
...
   KeyValueModel.iterateKeys = function(filter, options) {
     throwNotAttached(this.modelName, 'iterateKeys');
   };
 
+  /*!
+   * Set up the extended model.
+   */
```
superkhau: @bajtos Please comment on this too.

common/models/key-value-model.js
```diff
@@ -1,36 +1,93 @@
...
   };
 
-  // TODO add api docs
+  /**
+   * Persist a value using a given key.
+   *
+   * @param {String} key The key to use when searching.
+   * @param {Any} value The value to persist.
+   * @options {Object} options
+   * @property
```
bajtos: IIRC, we support `options.ttl` and it's also possible to pass `ttl` directly as the `options` argument.

```js
KVM.set('key', 'value', 100, cb);
KVM.set('key', 'value', { ttl: 100 }, cb);
```

common/models/key-value-model.js
bajtos: Yes, it allows users to iterate over a large set of keys without having to load them all in memory.

See https://github.com/mirkokiefer/async-iterators for more details about the concept of an Async Iterator.

We should also add a note (a warning?) to `.keys()` method, explaining that `keys()` method is not suitable for large key sets and users should use `iterateKeys()` instead.

common/models/key-value-model.js
bajtos: `setup` is just the way how base models define remoting metadata so that it's correctly inherited (see this bug https://github.com/strongloop/loopback/issues/2350). The method is called automatically by `Model.extend` (or `app.registry.createModel`), no need to document it.

crandmck:
LGTM.  

When we're ready to publish the API docs publicly, you just need to add an entry in https://github.com/strongloop/loopback/blob/master/docs.json for `common/models/key-value-model.js`.  Be sure to test locally first.
You can do it, or LMK and I'll try to find some time to do so.

superkhau:
@bajtos PTAL, should be good now.

common/models/key-value-model.js
```diff
@@ -1,36 +1,116 @@
...
+  /**
+   * Set the TTL (time to live) in ms (milliseconds) for a given key. TTL is the
+   * remaining time before a key-value pair is discarded from the database.
+   *
+   * @param {String} key Key to use when searching the database.
+   * @param {Number} ttl The TTL in ms to set for the key.
+   * @options {Number|Object} [options] Optional settings for the key-value
+   *   pair. If a Number is provided, it is set as the TTL in ms for the
+   *   key-value pair.
+   * @property {Number} ttl TTL in ms for the key-value pair.
```
bajtos: This looks like a mistake to me.

`@options {Number|Object}` is supported by `set`, not by `expire.

IIRC, `expire` does not have any `options` yet and always expect `ttl` in the second argument.

common/models/key-value-model.js
```diff
@@ -1,36 +1,116 @@
...
 
-  // TODO add api docs
+  /**
+   * Return all keys for the given filter. Similar to `.keys()` but instead
+   * allows for iteration over a large data set without having to load
+   * everything into memory at once.
+   *
+   * @param {Array.<String>} filter The key to use when searching the database.
+   * @options {Object} [options] Placeholder for optional settings.
+   * @callback {Function} callback The function to execute upon completion.
```
bajtos: There is not callback and no promise is returned. Instead, the method returns an instance of `AsyncIterator` as described in juggler's API.

common/models/key-value-model.js
superkhau: Yeah this is wrong, should be for `set`, was sleepy lol. Fixed

superkhau:
@bajtos PTAL again

superkhau:
test please

Issue #651, opened 2014-10-15
Disable all remote methods and enable only selected ones (white-list)


voitau:
There is a bunch of existing open/closed issues with questions how to hide the method of the model. My question is about how can all the methods be disabled first and then enabled only those which are necessary? (basically, duplicates this comment: https://github.com/strongloop/loopback/issues/465#issuecomment-54851858)

Here is the simple function which hides all the methods of the model:
```javascript
/**
 * Hides all the methods besides those in 'methods'.
 *
 * @param Model model to be updated.
 * @param methods array of methods to expose, e.g.: ['find', 'updateAttributes'].
 */
var setMethodsVisibility = function(Model, methods) {
  methods = methods || [];
  Model.sharedClass.methods().forEach(function(method) {
    method.shared = methods.indexOf(method.name) > -1;
  });
};
```

However, it doesn't hide two groups:
* remote methods, which have to be hidden like this: Model.disableRemoteMethod(name, isStatic)
* custom methods, like User.login or User.resetPassword (didn't find any way to hide them at all);

Could you please advise what would be the correct way to hide everything (without specific call for each method by its name explicitly) and allow only the methods which are necessary?

elhoyos:
I'm able to hide my remote methods from the explorer when using @voitau's function in 2.17.2.

ericprieto:
+1

gholias:
great code @voitau, thanks!
I'm new to loopback, what is the best place to put that code so I can use for all my models?

ericprieto:
To hide ALL methods, I ended up doing this

```
function disableAllMethodsBut(model, methodsToExpose)
{
	if(model && model.sharedClass)
	{
		methodsToExpose = methodsToExpose || [];

		var modelName = model.sharedClass.name;
		var methods = model.sharedClass.methods();
		var relationMethods = [];
		var hiddenMethods = [];

		try
		{
			Object.keys(model.definition.settings.relations).forEach(function(relation)
			{
				relationMethods.push({ name: '__findById__' + relation, isStatic: false });
				relationMethods.push({ name: '__destroyById__' + relation, isStatic: false });
				relationMethods.push({ name: '__updateById__' + relation, isStatic: false });
				relationMethods.push({ name: '__exists__' + relation, isStatic: false });
				relationMethods.push({ name: '__link__' + relation, isStatic: false });
				relationMethods.push({ name: '__get__' + relation, isStatic: false });
[...]

PR #2758, opened 2016-09-19
Expose new method for disabling remote methods.

richardpringle:
There is a new `SharedClass` method available in strong-remoting for disabling a method by name instead of specifying the `isStatic` boolean. 

I have exposed the method on the `Model` Class such that it can be used for a spike related to #651.

@bajtos, please review

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -15,6 +15,8 @@ var SharedClass = require('strong-remoting').SharedClass;
 var extend = require('util')._extend;
 var format = require('util').format;
 
+var deprecated = require('depd')('loopback');
+
 module.exports = function(registry) {
   /**
    * The base class for **all models**.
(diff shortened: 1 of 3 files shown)


richardpringle:
@slnode test please

richardpringle:
@bajtos @0candy PTAL

richardpringle:
We can either land this or #2767

bajtos:
+1 for adding a new API that's consistent with strong-remoting API.

Could you please add a unit-test? At least copy & modify the existing test [here](https://github.com/strongloop/loopback/blob/a6f8ec672dbffcd7a11db40c464b949b1b9a1b62/test/model.test.js#L692-L704) that verifies `remoteMethodDisabled` event. It would be nice to have a test to verify that disabled methods are not accessible via REST too, but I am ok to land this PR even without that.

As for the existing `Model.disableRemoteMethod` method, I am proposing to 
 - rework it to keep the same public API for backwards compatibility
 - call the new `disableRemoteMethodByName` under the hood
 - print a loopback-specific deprecation warning

See how `SharedClass.prototype.disableMethod` is implemented in strong-remoting ([link](https://github.com/strongloop/strong-remoting/blob/6c895128fbc42ea39620206e21ea01de8f36e727/lib/shared-class.js#L223-L228)) and use the same approach here.

richardpringle:
@bajtos changes have been made as per your request

bajtos:
All three failed Windows builds are caused by `Cannot start PhantomJS`, I think it's ok to ignore this error.

bajtos:
The loopback-component-explorer failure is unrelated, see https://github.com/strongloop/loopback-component-explorer/pull/178

bajtos:
Landed, thank you @richardpringle. Could you please back-port the new API to 2.x (without the deprecation notice)?

bajtos:
@richardpringle we have a unit-test in loopback-component-explorer that's triggering a deprecation warning, could you PTAL?

```
strong-remoting deprecated SharedClass.prototype.disableMethod is deprecated. Use SharedClass.prototype.disableMethodByName instead. node_modules/loopback/lib/model.js:439:22
    ✓ updates swagger object when a remote method is disabled
```

Because the same version of loopback-component-explorer is supporting both LB 2.x and LB 3.x, it will be best to have the new API backported to 2.x, so that the tests in explorer can rely on this new API.

richardpringle:
@bajtos:
>Could you please back-port
backport -> https://github.com/strongloop/loopback-component-explorer/pull/181

LB-component-explorer PR -> https://github.com/strongloop/loopback-component-explorer/pull/181

PR #2765, opened 2016-09-20
Require verification after email change

loay:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/1029

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -683,6 +683,19 @@ module.exports = function(User) {
         ctx.hookState.originalUserData = userInstances.map(function(u) {
           return { id: u.id, email: u.email };
         });
+        if (ctx.instance) {
+          var emailChanged = ctx.instance.email !== ctx.hookState.originalUserData[0].email;
+          if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
+            ctx.instance.emailVerified = false;
+          }
+        } else {
+          var emailChanged = ctx.hookState.originalUserData.some(function(data) {
+            return data.email != ctx.data.email;
+          });
+          if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
+            ctx.data.emailVerified = false;
+          }
+        }
         next();
       });
     });
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -2193,6 +2193,70 @@ describe('User', function() {
     });
   });
 
+  describe('Verification after updating email', function() {
+    var NEW_EMAIL = 'updated@example.com';
+    var userInstance;
+
+    beforeEach(createOriginalUser);
+
+    it('sets verification to false after email update if verification is required',
+    function(done) {
+      User.settings.emailVerificationRequired = true;
+      async.series([
+        function updateUser(next) {
+          userInstance.updateAttribute('email', NEW_EMAIL, function(err, info) {
+            if (err) return next (err);
+            assert.equal(info.email, NEW_EMAIL);
+            next();
+          });
+        },
+        function findUser(next) {
+          User.findById(userInstance.id, function(err, info) {
+            if (err) return next (err);
+            assert.equal(info.email, NEW_EMAIL);
+            assert.equal(info.emailVerified, false);
+            next();
+          });
+        },
+      ], done);
+    });
+
+    it('leaves verification as is after email update if verification is not required',
+    function(done) {
+      User.settings.emailVerificationRequired = false;
+      async.series([
+        function updateUser(next) {
+          userInstance.updateAttribute('email', NEW_EMAIL, function(err, info) {
+            if (err) return next (err);
+            assert.equal(info.email, NEW_EMAIL);
+            next();
+          });
+        },
+        function findUser(next) {
+          User.findById(userInstance.id, function(err, info) {
+            if (err) return next (err);
+            assert.equal(info.email, NEW_EMAIL);
+            assert.equal(info.emailVerified, true);
+            next();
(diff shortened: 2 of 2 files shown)


loay:
@bajtos @raymondfeng 
PTAL, this function `updateEmailVerified`
if I comment out line 714, it gets stuck and times out .. if I use this line, it gives the error `Error: Callback was already called.` and  tests fail. thoughts?

bajtos:
If you add `console.log` statements add the beginning of your two new functions `deleteSessions` and `updateEmailVerified`, then you will see they are repeatedly called in an infinite loop:

 - `ctx.Model.updateAll` triggers "before save"/"after save" hooks
 - there is a bug in the implementation where you treat missing `data.email` as an email change
 - as a result, the nested "before/after" save hooks thinks and email was changed and triggers another update of the flag

I would like to to address this complexly.

1) Create a new PR to fix the problem I described in https://github.com/strongloop/loopback/pull/2693#r77506564 and that you haven't addressed in the pull request:

> Also when performing a partial update, the email may not change at all. For example:
> 
> ```js
user.updateAttributes({ age: user.age + 1 }, cb);
```
>
> You should skip the session-invalidation logic in such case too.

2) As for this pull request, I think it will be better to move the code resetting `emailVerified` flag to the "before save" hook. You already know whether the email is going to be changed at that point, thus you can modify the database query to reset the flag too. This will save us one database call.

**Comment by loay** [comment]
Thanks @bajtos having a look

**Comment by loay** [comment]
@bajtos PTAL. Thanks

**Comment by bajtos** [comment]
>  As for this pull request, I think it will be better to move the code resetting `emailVerified` flag to the "before save" hook. You already know whether the email is going to be changed at that point, thus you can modify the database query to reset the flag too. This will save us one database call.

@loay AFAICT, you are still resetting `emailVerified` in `after save` hook. Did you try to reset it in `before save`, as I suggested? Were there any problems?

**Comment by loay** [comment]
yeah, it didn't work for me, as I would run to the same problem as before, either timeout or infinite loop, as the validations of newEmail are not in the `before` hook
```
      if (!ctx.instance && !ctx.data) return next();
      var newEmail = (ctx.instance || ctx.data).email;
      if (!newEmail) return next();
```

**Comment by bajtos** [comment]
> yeah, it didn't work for me, as I would run to the same problem as before, either timeout or infinite loop, as the validations of newEmail are not in the before hook

What is stopping you from adding the validation to the "before save" hook? It should also improve the performance a bit, because there will be one database query less when email is not changed.

**Comment by loay** [comment]
@bajtos I moved the `emailVerified` part up to the `before save` function as requested. PTAL. Thanks.

File: test/user.test.js [hunk 1]
```diff
@@ -2193,6 +2193,72 @@ describe('User', function() {
     });
   });
 
+  describe('Verification after updating email', function() {
+    var nonVerifiedUser;
```
bajtos: I find the variable name confusing. You are creating the user with `emailVerified` set to `true` and then store it in a variable called `nonVerifiedUser`.

Can you use a different name please? As far as I am concerned, `user`, `userInstance` or `testUser` are good enough. Feel free to come up with your own name though!

common/models/user.js
```diff
@@ -683,6 +683,17 @@ module.exports = function(User) {
         ctx.hookState.originalUserData = userInstances.map(function(u) {
           return { id: u.id, email: u.email };
         });
+        if (ctx.instance) {
+          var emailChanged = ctx.instance.email !== ctx.hookState.originalUserData[0].email;
+          if (!ctx.Model.settings.emailVerificationRequired) return next();
```
bajtos: How about simplifying this and the following line to the followin?

```js
if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
  ctx.instance.emailVerified = false;
}
```

common/models/user.js
```diff
@@ -683,6 +683,17 @@ module.exports = function(User) {
...
+          if (emailChanged) ctx.instance.emailVerified = false;
+        } else {
+          var emailChanged = ctx.hookState.originalUserData.some(function(data) {
+            return data.email != ctx.data.email;
+          });
+          if (!ctx.Model.settings.emailVerificationRequired) return next();
```
bajtos: Same comment applies here.

```js
if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
  ctx.data.emailVerified = false;
}
```

test/user.test.js
```diff
@@ -2193,6 +2193,72 @@ describe('User', function() {
...
+    var originalMyEmailCred = { email: 'original@example.com', password: 'bar',
+    emailVerified: true };
+    var myUpdatedCred = { email: 'updated@example.com' };
+    it('sets verification to false after email update if verification is required',
+    function(done) {
+      User.settings.emailVerificationRequired = true;
+      async.series([
+        function createUser(next) {
+          User.create(originalMyEmailCred, function(err, specialInstance) {
```
bajtos: AFAICT, this setup is shared by both tests. Please extract it into a shared `beforeEach` hook.

```js
beforeEach(setupTestUser);

it(...);

it(...);

function setupTestUser(done) {
  User.create(originalMyEmailCred, function(err, inst) {
    if (err) return done(err);
    nonVerifiedUser = inst; // use a different name than nonVerifiedUser
    done();
  });
}
```

loay:
@bajtos I have addressed the changes. PTAL. Thanks.

bajtos:
@loay landed, thank you. Please back-port to 2.x yourself.

PR #2782, opened 2016-09-22
Update tests to use registry for model creation [2.x]

gunjpan:
Current implementation of `app.model(modelName, settings)`
works as a sugar for model creation. From LB 3.0 onwards, this is
not supported. 
This backporting:
- keeps the sugar method for model creation for backward
compatibility
- updates test cases to use `app.registry.createModel()`
for model creation

Backport of #2401 

Connect to strongloop-internal/scrum-loopback#772

### DIFF
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -363,18 +363,16 @@ describe('security ACLs', function() {
 });
 
 describe('access check', function() {
-  var app;
-  before(function() {
-    app = loopback();
-    app.use(loopback.rest());
-    app.enableAuth();
-    app.dataSource('test', {connector: 'memory'});
-  });
-
   it('should occur before other remote hooks', function(done) {
-    var MyTestModel = app.model('MyTestModel', {base: 'PersistedModel', dataSource: 'test'});
+    var app = loopback();
+    var MyTestModel = app.registry.createModel('MyTestModel');
     var checkAccessCalled = false;
     var beforeHookCalled = false;
+
+    app.use(loopback.rest());
+    app.enableAuth();
+    app.dataSource('test', { connector: 'memory' });
+    app.model(MyTestModel, { dataSource: 'test' });
 
     // fake / spy on the checkAccess method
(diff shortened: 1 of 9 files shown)


gunjpan:
@bajtos : Could you please review this backport PR? Thank you.

test/acl.test.js
```diff
@@ -363,19 +363,18 @@ describe('security ACLs', function() {
...
-
   it('should occur before other remote hooks', function(done) {
-    var MyTestModel = app.model('MyTestModel', {base: 'PersistedModel', dataSource: 'test'});
+    var app = loopback();
+    var MyTestModel = app.registry.createModel('MyTestModel');
     var checkAccessCalled = false;
     var beforeHookCalled = false;
 
+    app.use(loopback.rest());
+    app.set('remoting', { errorHandler: { debug: true, log: false }});
```
bajtos: I think this errorHandler configuration was added in LB 3.x only, could you please check?

test/app.test.js
```diff
@@ -617,11 +617,12 @@ describe('app', function() {
   });
 
   describe('app.model(Model)', function() {
-    var app;
-    var db;
+    var app, db, MyTestModel;
     beforeEach(function() {
       app = loopback();
-      db = loopback.createDataSource({connector: loopback.Memory});
+      app.set('remoting', { errorHandler: { debug: true, log: false }});
```
bajtos: ditto

test/acl.test.js
gunjpan: I am not sure as I see `errorHandler`in [`package.json`](https://github.com/strongloop/loopback/blob/2.x/package.json)  for 2.x.

gunjpan: I'm removing these lines as they weren't present in 2.x

gunjpan:
@bajtos : Addressed your comments, removed use of `error handler` as it's not there in 2.x. PTAL.

superkhau:
test please

Issue #2770, opened 2016-09-21
Fix expire in explorer


superkhau:
_From @superkhau on September 20, 2016 1:53_

KV connector expire feature isn't working properly in explorer:

# Step to reproduce

1. Create a typical app with a model and browse to explorer
2. Create an object via set without TTL
3. Try to set TTL using expire

The request will fail with:

```
{
  "error": {
    "statusCode": 400,
    "name": "Error",
    "message": "ttl is a required argument"
  }
}
```

# Workaround

None at the moment, we probably need to fix this one right away.

_Copied from original issue: strongloop-internal/scrum-loopback#1072_

superkhau:
_From @bajtos on September 20, 2016 14:59_

@superkhau is there any particular reason for keeping this issue private? I would like us to be as transparent and open as possible, otherwise it's difficult to get community involvement. I would like us to default to filling issues in the public `loopback` project unless there is a good reason for keeping it private.

superkhau:
_From @bajtos on September 20, 2016 15:0_

> 1. Create a typical app with a model and browse to explorer
> 2. Create an object via set without TTL
> 3. Try to set TTL using expire

Which request failed? The second one, setting a key without expire? Or the third one? For the third one, I think it's expected that the `ttl` argument is required, right?

superkhau:
_From @bajtos on September 20, 2016 15:2_

@superkhau I am also adding missing `Milestone: #Epic KV Connectors` to this issue and the others you created recently.

superkhau:
@bajtos No reason, I should've created in `loopback` in the first place -- I actually created all the CI issues in scrum-loopback and forgot to switch over for the KV connector issues. Moving.

superkhau:
>Which request failed? The second one, setting a key without expire? Or the third one? For the third one, I think it's expected that the ttl argument is required, right?

First request. There is something wrong with explorer not passing the TTL entered in the textbox to the remote method.

richardpringle:
@superkhau, what does your input look like?

superkhau:
@richardpringle the remoting for expires is at https://github.com/strongloop/loopback/blob/master/common/models/key-value-model.js#L187-L195.

After some digging, the interesting part is API explorer with the above remoting settings doesn't submit any values in the curl request (missing -d ... contents or -F for form data). Meaning it doesn't submit any data to the backend.

[...]

PR #2806, opened 2016-09-30
Fix expire in API explorer for KV connectors

superkhau:
**Depends on https://github.com/strongloop/loopback-datasource-juggler/pull/1122 to pass tests**

Expire feature wasn't sending data as part of it's request in API
explorer. Tested manually through clicking API explorer.

More info for why at https://github.com/strongloop/loopback/issues/2770#issuecomment-250597461

Connect to #2770 

@bajtos PTAL

### DIFF
--- a/common/models/key-value-model.js
+++ b/common/models/key-value-model.js
@@ -188,8 +188,8 @@ module.exports = function(KeyValueModel) {
       accepts: [
         { arg: 'key', type: 'string', required: true,
           http: { source: 'path' }},
-        { arg: 'ttl', type: 'number', required: true,
-          http: { source: 'form' }},
+        { arg: 'ttl', type: 'object', required: true,
+          http: { source: 'body' }},
       ],
       http: { path: '/:key/expire', verb: 'put' },
     });


bajtos:
@superkhau I don't like the hack where `ttl` is passed as an object. What other, more clean solutions are available?

Here is a couple that comes to my mind:

 1. Send `ttl` in query string, i.e. `PUT /my-model/{key}/expire?ttl=123`. Downsides: it's a breaking change, also `PUT` request with an empty body looks weird to me.

 2. Expose the `expire` method on both `PUT` and `POST` verbs. Downside: we have two HTTP endpoints for the same API, but then there are already precedents like `exists` exposed on `GET /{id}/exists` and `HEAD /{id}`.

bajtos:
On the second thought, I think this needs more investigation, see https://github.com/strongloop/loopback-component-explorer/blob/cb6eafd5929170c96490d660eb483ed5e0046431/public/lib/loadSwaggerUI.js#L24

bajtos:
On the second thought, I think this needs more investigation, see https://github.com/strongloop/loopback/issues/2770#issuecomment-250690884

superkhau:
Closing in favour of https://github.com/strongloop/loopback-swagger/pull/69

PR #2795, opened 2016-09-27
Add partial update test case

loay:

PR #2814, opened 2016-10-03
Validate non-email property partial update

loay:
Add partial update test case to validate non-email properties partial-update
backport of https://github.com/strongloop/loopback/pull/2795


0candy:
LGTM

PR #2836, opened 2016-10-06
Update ja and nl translation files

0candy:
Connect to: https://github.com/strongloop-internal/scrum-loopback/issues/1118


0candy:
@superkhau PTAL

PR #1797, opened 2015-11-05
allow non expiring authentication tokens by setting ttl to -1.

jesucarr:
I guess this was the initial intent since the -1 value is allowed already

PR #2841, opened 2016-10-10
Allow tokens with eternal TTL (value -1)

bajtos:
- Add a new User setting `allowEternalTokens`
 - Enhance `AccessToken.validate()` to support eternal tokens with ttl value -1 when the user model allows it.

This patch supersedes #1797.

@ritch please review

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -147,11 +147,19 @@ module.exports = function(AccessToken) {
       assert(this.ttl, 'token.ttl must exist');
       assert(this.ttl >= -1, 'token.ttl must be >= -1');
 
+      var AccessToken = this.constructor;
+      var userRelation = AccessToken.relations.user; // may not be set up
+      var User = userRelation && userRelation.modelTo;
+
       var now = Date.now();
       var created = this.created.getTime();
       var elapsedSeconds = (now - created) / 1000;
       var secondsToLive = this.ttl;
-      var isValid = elapsedSeconds < secondsToLive;
+      var eternalTokensAllowed = !!(User && User.settings.allowEternalTokens);
+      var isEternalToken = secondsToLive === -1;
+      var isValid = isEternalToken ?
+        eternalTokensAllowed :
+        elapsedSeconds < secondsToLive;
 
       if (isValid) {
         cb(null, isValid);
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -357,11 +357,38 @@ describe('AccessToken', function() {
     assert(Object.prototype.toString.call(this.token.created), '[object Date]');
   });
 
-  it('should be validateable', function(done) {
-    this.token.validate(function(err, isValid) {
-      assert(isValid);
+  describe('.validate()', function() {
+    it('accepts valid tokens', function(done) {
+      this.token.validate(function(err, isValid) {
+        assert(isValid);
+        done();
+      });
+    });
 
-      done();
+    it('rejects eternal TTL by default', function(done) {
+      this.token.ttl = -1;
+      this.token.validate(function(err, isValid) {
+        if (err) return done(err);
+        expect(isValid, 'isValid').to.equal(false);
+        done();
+      });
+    });
+
+    it('allows eternal tokens when enabled by User.allowEternalTokens',
+    function(done) {
+      var Token = givenLocalTokenModel();
+
+      // Overwrite User settings - enable eternal tokens
+      Token.app.models.User.settings.allowEternalTokens = true;
+
+      Token.create({ userId: '123', ttl: -1 }, function(err, token) {
+        if (err) return done(err);
+        token.validate(function(err, isValid) {
+          if (err) return done(err);
+          expect(isValid, 'isValid').to.equal(true);
+          done();
+        });
+      });
     });
   });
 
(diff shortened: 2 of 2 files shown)


josephlaw:
Anyone can tell me where to setup allowEternalTokens? e.g. server/model-config.json? config.json? or server.js?

bajtos:
@josephlaw you should create a new model inheriting from the built-in User model. Then you can change the setting in `common/my-user.json` file.

Shyri:
I couldn't make it work.

Having a custom user model leads the accessToken default model not to find the user relationship at the moment of verifying the token, see how User is undefined in the debug screenshot.
screen shot 2016-11-20 at 12 36 22

Shyri:
@bajtos Any idea why User is undefined in my previous message?

bajtos:
> @Shyri Any idea why User is undefined in my previous message?

Not really ☹️ Could you please create a small app reproducing the issue (see http://loopback.io/doc/en/contrib/Reporting-issues.html#bug-report) and open a new issue?

brilvio:
It's not passing the validation for me, I have a custom JSON model inherited from built-in model User.
```json
"name": "empresa",
  "base": "User",
  "idInjection": true,
  "options": {
    "validateUpsert": true,
    "allowEternalTokens": true  
  }
```
The User is defined but `User.settings.allowEternalTokens` is always undefined :(

bajtos:
@brilvio have you updated the configuration of your `AccessToken` to `belongTo` your new user model? I think the best way is to create a custom AccessToken model too, see e.g. http://loopbackloops.blogspot.com/ (ignore the last part about overriding AccessToken methods).

@crandmck Google query [loopback custom accesstoken model](https://www.google.com/search?q=loopback+custom+accesstoken+model) returns the mentioned blog post as the first result, it looks like we may not have this use case documented on loopback.io. Could you PTAL?

brilvio:
@bajtos tried to do what is in the link, new custom model inherited of `AccessToken` set the relation of this new model to my custom user model as `belongTo` and now I having this errors `AUTHORIZATION_REQUIRED`it is like I didn't send the access token at all. Here is the relation in the new model:
```json
"relations": {
    "empresa": {
      "type": "belongsTo",
      "model": "empresa",
      "foreignKey": "userId"
    }
  }
```
Setting the same relation on the `node_modules/loopback/common/models/access-token.json`file throws the `INVALID_TOKEN` because this property is undefined `User.settings.allowEternalTokens`

bajtos:
@brilvio I am afraid I have run out of easy answers. Please open a new issue, provide a sample app reproducing the problem (see http://loopback.io/doc/en/contrib/Reporting-issues.html#bug-report) and include a link to this discussion (#2841) in issue's description.

brilvio:
Thank you for your time. I will do that.

crandmck:
Added it to the reference http://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#advanced-options but we need a more thorough explanation, too.

jdhrivas:
@Shyri In my configuration I have extended both the User and AccessToken and I have verified that this feature works.

Here is are my settings:

```
// Extending User model user.json
{
  "name": "user",
  "base": "User",
  "strict": true,
  "idInjection": true,
  "options": {
    "validateUpsert": true,
    "allowEternalTokens": true
  },
 "relations": {
    "accessTokens": {
      "type": "hasMany",
      "model": "accessToken",
      "foreignKey": "userId",
      "options": {
        "disableInclude": true
      }
    }
}
// Note: I'm skipping all my properties, ACL and such. These are the properties that matter for this feature
// Extending AccessToken model access-token.json
{
  "name": "accessToken",
  "plural": "accessTokens",
  "base": "AccessToken",
  "strict": true,
  "idInjection": true,
  "properties": {},
  "validations": [],
  "relations": {
    "user": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "userId"
    }
  },
  "acls": [],
  "methods": {}
}
```

As you can see I have extended both User and AccessToken Model, now when the function AccessToken.prototype.validate is called, User.settings.allowEternalTokens returns true. Without extending AccessToken, User.settings.allowEternalTokens returns false because the AccessToken model is not related to the extended user model where the property allowEternalTokens is set. 

@crandmck should the documentation be updated?

bajtos:
> should the documentation be updated?

Yes please. Would you mind contributing this patch yourself, @jdhrivas? See http://loopback.io/doc/en/contrib/doc-contrib.html to get started.

jdhrivas:
@bajtos sure I can. This time I'll take a look at the guidelines before any pull request.

crandmck:
Awesome!  Please make a note in https://github.com/strongloop/loopback.io/issues/200 when you open a PR... thanks.

IbrahimAmin:
You can also add this line in the boot folder if you don't want to create a custom AccessToken model, I tried this and it definitely does the trick.

User.settings.allowEternalTokens = true;

Legited:
@IbrahimAmin Thanks, this works, but I still had to create a "belongsTo" relationship for the AccessToken model : 

1. Add AccessToken model to the `model-config.json`
2. Then in a boot script:
```
module.exports = function (app) {
  const models = app.models;
  const User = models.PUser; // my custom User model
  const AccessToken = models.AccessToken;

// enable eternal tokens
  User.settings.allowEternalTokens = true;
  // Since we are using custom User model, must hack the built-in relations for the AcessToken manually
  // PAccessToken = AccessToken;
  AccessToken.belongsTo(User, {foreignKey: 'userId'});
  AccessToken.relations.user = AccessToken.relations.pUser;
  delete AccessToken.relations.pUser;
};
``` 

Keep in mind, that if you have included `"allowEternalTokens": true` in your user model, then `  User.settings.allowEternalTokens = true;` is unnecessary, since it would be already set

bajtos:
Thank you @Legited for the instructions.

FWIW, we have updated the documentation for setting up custom user/token models, see http://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#preparing-access-control-models.

There is no need to manipulate AccessToken relations in a boot script, the relation can be re-configured via `server/model-config.json` - see http://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#access-control-with-a-single-user-model

Legited:
@bajtos 

Thanks, setting up relations in the model-config worker for me

PR #2842, opened 2016-10-11
Fix description of updateAll response

bajtos:
Correctly describe the first non-error callback arg as an `info` object containing a `count` property.

Supersedes #2535
Connect to #2104

@superkhau please review

PR #2846, opened 2016-10-12
Fix description of updateAll response [2.x]

bajtos:
Correctly describe the first non-error callback arg as an `info` object containing a `count` property.

Back-port #2842
Connect to #2104

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -795,10 +795,15 @@ module.exports = function(registry) {
           description: 'An object of model property name/value pairs'},
       ],
       returns: {
-        arg: 'count',
-        description: 'The number of instances updated',
-        type: 'object',
-        root: true
+        arg: 'info',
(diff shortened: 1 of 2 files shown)


bajtos:
I'll wait with landing this patch until https://github.com/strongloop/loopback-sdk-ios-codegen/pull/12 is landed, so that I can verify the CI build of loopback-sdk-ios-codegen is green again.

bajtos:
> [cis-jenkins] downstream: loopback-sdk-ios-codegen@master — Failed! (bf5c206)

@rmg I think it does not make sense to run loopback-sdk-ios-codegen on CIS, because it requires CocoaPods and XCode, which is/are MacOS only. OTOH, we need ci.strongloop.com to test loopback-sdk-ios-codegen.

What should I do to make this happen?

cc @superkhau

bajtos:
@slnode test please

PR #2879, opened 2016-10-20
allow batch create for persisted models

davidcheung:
connect to strongloop-internal/scrum-loopback#1138

### Description
In strong-remoting 3.x, we have stricken the coercion of inputs
methods that are expecting an Object will nolonger accept an array
as input, to preserve backwards compatibility we have added flag
allowArray in remote arguments, which would accept an array of objects

#### Related issues

- None

### Checklist

- [x] New tests are added to cover all changes

List of other remoteMethods in persistedModel that receives a `data` with `type:object`
i looked at them and dont think they should support batch
- patchOrCreate (data) 
- replaceOrCreate (data)
- upsertWithWhere (where, data)
- replaceById (id, data)
- updateAll (where, data) 
- patchAttributes (data)
- updateLastChange (id, data)

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -640,7 +640,7 @@ module.exports = function(registry) {
       description: 'Create a new instance of the model and persist it into the data source.',
       accessType: 'WRITE',
       accepts: {
-        arg: 'data', type: 'object', model: typeName,
+        arg: 'data', type: 'object', model: typeName, allowArray: true,
         description: 'Model instance data',
         http: { source: 'body' },
       },
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -124,6 +124,44 @@ describe.onServer('Remote Methods', function() {
     app.use(loopback.rest());
   });
 
+  describe('Model.create(data, callback)', function() {
+    it('creates model', function(done) {
+      var anObject = { first: 'June' };
+      request(app)
+        .post('/users')
+        // sends an object
+        .send(anObject)
+        .expect('Content-Type', /json/)
+        .expect(200)
+        .end(function(err, res) {
+          if (err) return done(err);
+          expect(res.body).to.have.property('id');
+          expect(res.body).to.have.property('first', 'June');
+          done();
+        });
+    });
+    // batch create must be tested with a remote request because there are
+    // coercion being done on strong-remoting side
+    it('creates array of models', function(done) {
+      var arrayOfObjects = [
+        { first: 'John' }, { first: 'Jane' },
+      ];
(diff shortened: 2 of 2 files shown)


davidcheung:
@bajtos PTAL

lib/persisted-model.js
```diff
@@ -641,7 +641,7 @@ module.exports = function(registry) {
       accessType: 'WRITE',
       accepts: {
         arg: 'data', type: 'object', model: typeName,
-        description: 'Model instance data',
+        description: 'Model instance data', allowArray: true,
```
bajtos: It's a good practice to keep related things close together and unrelated things apart. IMO, `allowArray` and `description` are rather unrelated.

Please move `allowArray` closer to `type`/`model`, e.g. on the previous line.

test/model.test.js
```diff
@@ -742,6 +742,37 @@ describe.onServer('Remote Methods', function() {
 
       expect(callbackSpy).to.have.been.calledWith(TestModel.sharedClass, 'findOne');
     });
+
+    describe('batch operations', function() {
```
bajtos: I am proposing to move this new test into `describe.onServer('Remote Methods')` block as `describe('Model.create(data, callback')`. I think it would be worth adding two tests - one sending an object, the other sending an array. That way the tests make it clear that both variants are supported.

test/model.test.js
```diff
@@ -124,6 +124,43 @@ describe.onServer('Remote Methods', function() {
     app.use(loopback.rest());
   });
 
+  describe('Model.create(data, callback)', function() {
+    it('creates model', function(done) {
+      var anObject = { first: 'June' };
+      request(app)
+        .post('/users')
```
davidcheung: i noticed the other test cases in this `describe` is using the method itself instead of going thru strong-remoting
like `User.destroyAll(....)`, instead of `POST /users/destroyAll`, but that wouldnt be able to verify our scenario that strong-remoting rejects the array.  should i add comments to describe why its different?

bajtos: `destroyAll` is not exposed via REST.

> should i add comments to describe why its different?

yes please, that would be helpful

davidcheung:
updated comments
>  >should i add comments to describe why its different?
>
> yes please, that would be helpful


travis wouldnt pass unless strong-remoting is published
due to https://github.com/strongloop/strong-remoting/pull/369

davidcheung:
@slnode test please

davidcheung:
@slnode test please

davidcheung:
@slnode test please

PR #2888, opened 2016-10-24
make test individually runable

davidcheung:
### Description

previous when you do command such as `mocha test/model.test.js`
we get errors such as "expect is not a function", "assert is not a function"
this is to organize the test file in a slightly different way, so we can do `mocha test/model.test.js`

not sure if we need to modify [`karma.conf`](https://github.com/strongloop/loopback/blob/c28a1568b31b8fb8d75273b4940b1aa8cabc2828/test/karma.conf.js#L23) too, npm test is locally passing for me
#### Related issues
- None
### Checklist
- [ ] New tests are added to cover all changes

**Update 14/11/16**:
  - Needs re-review from @bajtos or @strongloop/squad-epitome 

**Update 14/11/16**:
  - @pthieu will finish review.

### DIFF
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -16,7 +16,11 @@ Token.attachTo(ds);
 var ACL = loopback.ACL;
 
 describe('loopback.token(options)', function() {
-  beforeEach(createTestingToken);
+  var app;
+  beforeEach(function(done) {
+    app = loopback();
+    createTestingToken.call(this, done);
+  });
 
   it('should populate req.token from the query string', function(done) {
     createTestAppAndRequest(this.token, done)
@@ -428,7 +432,9 @@ describe('AccessToken', function() {
 });
 
 describe('app.enableAuth()', function() {
+  var app;
   beforeEach(function setupAuthWithModels() {
+    app = loopback();
     app.enableAuth({ dataSource: ds });
   });
   beforeEach(createTestingToken);
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -16,11 +16,15 @@ var expect = require('chai').expect;
 var it = require('./util/it');
 
 describe('app', function() {
+  var app;
+  beforeEach(function() {
+    app = loopback();
+  });
+
   describe.onServer('.middleware(phase, handler)', function() {
-    var app, steps;
+    var steps;
 
     beforeEach(function setup() {
-      app = loopback();
       steps = [];
     });
 
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,0 +1,1 @@
+--require ./test/support.js
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -4,12 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 var async = require('async');
+var chai = require('chai');
+var describe = require('./util/describe');
 var loopback = require('../');
 var ACL = loopback.ACL;
 var defineModelTestsWithDataSource = require('./util/model-tests');
 var PersistedModel = loopback.PersistedModel;
+var sinonChai = require('sinon-chai');
 
-var describe = require('./util/describe');
+var expect = chai.expect;
+chai.use(sinonChai);
 
 describe('Model / PersistedModel', function() {
   defineModelTestsWithDataSource({
--- a/test/support.js
+++ b/test/support.js
@@ -12,7 +12,6 @@ expect = require('chai').expect;
 loopback = require('../');
 memoryConnector = loopback.Memory;
 GeoPoint = loopback.GeoPoint;
-app = null;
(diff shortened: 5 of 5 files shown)


davidcheung:
cc @superkhau

davidcheung:
@bajtos do you know if there is any negative implication with this change, i have manually tried to run the tests individually and it seems to work fine. 

cc @pthieu @bosco-li

test/access-token.test.js
```diff
@@ -15,6 +15,10 @@ var ds = loopback.createDataSource({ connector: loopback.Memory });
 Token.attachTo(ds);
 var ACL = loopback.ACL;
 
+beforeEach(function() {
+  this.app = app = loopback();
```
pthieu: How does this pass `eslint`? `app` would be a global here.

test/access-token.test.js
```diff
@@ -15,6 +15,10 @@ var ds = loopback.createDataSource({ connector: loopback.Memory });
 Token.attachTo(ds);
```
pthieu: No `use strict`; at the top.

test/access-token.test.js
superkhau: LoopBack is using eslint-config-loopback@^1.0.0, the use strict check was added later on -- we need to update this project to use ^4.0.0

davidcheung: the mocha-opts requires `support.js` which app ends up global in our test suite 
https://github.com/strongloop/loopback/blob/d85ff32abb681a127659e4ed6ab4cc0fa3fac922/test/support.js#L10-L18

bajtos: Please define a describe-local `var app` and move the hook into the top-level describe block, so that we don't pollute global namespace and don't run this hook for tests that don't need it.

```js
describe('loopback.token(options)', function() {
  var app;
  beforeEach(function() {
    this.app = app = loopback();
  });

  // ...
});
```

test/mocha.opts
```diff
@@ -0,0 +1,2 @@
+--require ./test/support.js
+--require ./test/helpers/loopback-testing-helper.js
```
bajtos: Why is this second line required? AFAICT, all modules using `loopback-testing-helper` are directly calling `require()` to get it.

test/support.js
```diff
@@ -21,10 +21,6 @@ var RemoteObjects = require('strong-remoting');
 // for tests using the built-in User model
 loopback.User.settings.saltWorkFactor = 4;
 
-beforeEach(function() {
-  this.app = app = loopback();
```
bajtos: Also remove the global `app` variable defined at line 15

test/access-token.test.js
bajtos: I agree with add `use strict` to all files in a standalone commit.

test/app.test.js
```diff
@@ -15,6 +15,10 @@ var describe = require('./util/describe');
 var expect = require('chai').expect;
 var it = require('./util/it');
 
+beforeEach(function() {
+  this.app = app = loopback();
```
bajtos: Same comment as above, move this code to `describe` block and define `app` as a describe-local variable.

test/access-token.test.js
```diff
@@ -10,13 +10,17 @@ var loopback = require('../');
 var extend = require('util')._extend;
 var session = require('express-session');
 
+var app;
```
bajtos: Nitpick: we usually put describe-scoped variables inside a `describe` block.

test/model.test.js
```diff
@@ -10,6 +10,7 @@ var defineModelTestsWithDataSource = require('./util/model-tests');
 var PersistedModel = loopback.PersistedModel;
 
 var describe = require('./util/describe');
+require('./helpers/loopback-testing-helper');
```
bajtos: As I commented before, I don't understand why is this needed, could you please explain?

davidcheung: i will change it to directly use those lines

davidcheung:
@bajtos I have addressed the feedback, I think this is pretty close anyways, will ask @phong to PTAL

test/model.test.js
```diff
@@ -10,6 +10,9 @@ var defineModelTestsWithDataSource = require('./util/model-tests');
 var PersistedModel = loopback.PersistedModel;
 
 var describe = require('./util/describe');
+var chai = require('chai');
```
pthieu: alphabetical

pthieu:
LGTM from me after alphabetical change -- @bajtos can you approve yours if you're ok with changes.

pthieu:
@bajtos  is on vacation for a week, @davidcheung, since @superkhau has approved, I think you can merge.

test/access-token.test.js
davidcheung: addressed

PR #2893, opened 2016-10-25
Add options to bulk update

kobaska:
### Description

Passing options to bulkUpdate, so we can detect saves and updates during sync. Saves during sync can be identified by adding a property on options which gets passed into to replicate.

e.g.:
When you want to update last modified date column on each save through operation hooks, but not during sync:

```
var options = {
  sync: true,
  ... other options
}

SomeModel.replicate(since, targetModel, options, callback);

SomeModel.observe('before save', function updateLastModDate(ctx, next) {
if (ctx.options.sync) {
  // do not update
} else {
  // update
}

next();
});
```

#### Related issues

- None

### Checklist

- [ ] New tests are added to cover all changes

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -67,9 +67,9 @@ module.exports = function(registry) {
   function throwNotAttached(modelName, methodName) {
     throw new Error(
       g.f('Cannot call %s.%s().' +
-      ' The %s method has not been setup.' +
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

PR #2897, opened 2016-10-25
Update test confirmation text in PR template

superkhau:
We only ask users to check when adding new tests, but
some PRs only modify existing tests. Also added
instructions to mark with an "x" for checking off items.

@bajtos PTAL

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,9 @@ Please use the following link syntaxes:
 ### Checklist
 
 <!--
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,6 +14,10 @@ Please use the following link syntaxes:
 
 ### Checklist
 
-- [ ] New tests are added to cover all changes
+<!--
+Please mark your choice with an "x" (eg. [x], not [*])
```
bajtos: Perhaps add a link to https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

PR #2957, opened 2016-11-19
propagate resolved roles in remoting context

ebarault:
### Description

For some **advanced CRUD access control in remote methods**, it is often required to **get/check roles resolved** (i.e. allowed) for a user against a method with `Role.getRoles(context)` or `Role.isInRole(role, context)`.
In addition to clutter the remote methods / hooks code, it forces the app to run again through all static roles and registered dynamic role resolvers, inducing performance loss and extra calls to db.

Meanwhile, the exact same course of action is performed right when the `Model.checkAccess()` method is called to verify if a principal is allowed to access a protected resource : 

`Model.checkAccess(context)` -> `ACL.checkAccessForContext(context)` -> `acls.forEach(acl, ...)` ->`roleModel.isInRole(acl.principalId, context)` 

This PR proposes to add a new param (`Array`) in remoting context (name proposed: `resolvedRoles`, to be discussed) which is filled during **Model Access Permission Checking** phase with the **roles** for which the user is allowed to run the remote method.

an example of `resolvedRoles` is: `['admin', '$owner']` where admin is a custom role statically mapped to a principal, and $owner is a built-in smart role

`resolvedRoles` param can then be accessed in remote method and remote method hooks as `ctx.req.remotingContext.resolvedRoles`

_To be noted: the `resolvedRoles` param is set in a way that the calling user cannot temper the value to fool the additional access control policies added by user in remote methods_ 

**Reviewers:** please review/discuss the general idea before i provide the required tests  

#### Related issues

- None

### Checklist

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -34,6 +34,7 @@ var g = require('../../lib/globalize');
 var loopback = require('../../lib/loopback');
 var utils = require('../../lib/utils');
 var async = require('async');
+var extend = require('util')._extend;
 var assert = require('assert');
 var debug = require('debug')('loopback:security:acl');
 
@@ -204,11 +205,12 @@ module.exports = function(ACL) {
   /*!
    * Resolve permission from the ACLs
    * @param {Object[]) acls The list of ACLs
-   * @param {Object} req The request
-   * @returns {AccessRequest} result The effective ACL
+   * @param {AccessRequest} req The access request
+   * @returns {AccessRequest} result The resolved access request
    */
   ACL.resolvePermission = function resolvePermission(acls, req) {
     if (!(req instanceof AccessRequest)) {
+      req.registry = this.registry;
       req = new AccessRequest(req);
     }
     // Sort by the matching score in descending order
@@ -250,9 +252,16 @@ module.exports = function(ACL) {
         debug('with score:', acl.score(req));
       });
     }
+    var res = new AccessRequest({
+      model: req.model,
+      property: req.property,
+      accessType: req.accessType,
+      permission: permission || ACL.DEFAULT,
+      registry: this.registry});
 
-    var res = new AccessRequest(req.model, req.property, req.accessType,
-        permission || ACL.DEFAULT);
+    // Elucidate permission status if DEFAULT
+    res.settleDefaultPermission();
+
     return res;
   };
 
@@ -316,8 +325,8 @@ module.exports = function(ACL) {
    * @param {String} property The property/method/relation name.
    * @param {String} accessType The access type.
    * @callback {Function} callback Callback function.
-   * @param {String|Error} err The error object
-   * @param {AccessRequest} result The access permission
+   * @param {String|Error} err The error object.
+   * @param {AccessRequest} result The resolved access request.
    */
   ACL.checkPermission = function checkPermission(principalType, principalId,
                                                  model, property, accessType,
@@ -332,10 +341,11 @@ module.exports = function(ACL) {
     var accessTypeQuery = (accessType === ACL.ALL) ? undefined :
       {inq: [accessType, ACL.ALL, ACL.EXECUTE]};
 
-    var req = new AccessRequest(model, property, accessType);
+    var req = new AccessRequest({model, property, accessType, registry: this.registry});
 
     var acls = this.getStaticACLs(model, property);
 
+    // resolved is an instance of AccessRequest
     var resolved = this.resolvePermission(acls, req);
 
     if (resolved && resolved.permission === ACL.DENY) {
@@ -355,11 +365,8 @@ module.exports = function(ACL) {
           return callback(err);
         }
         acls = acls.concat(dynACLs);
+        // resolved is an instance of AccessRequest
         resolved = self.resolvePermission(acls, req);
-        if (resolved && resolved.permission === ACL.DEFAULT) {
-          var modelClass = self.registry.findModel(model);
-          resolved.permission = (modelClass && modelClass.settings.defaultPermission) || ACL.ALLOW;
-        }
         return callback(null, resolved);
       });
     return callback.promise;
@@ -377,30 +384,63 @@ module.exports = function(ACL) {
     }
   };
 
+  // NOTE Regarding ACL.isAllowed() and ACL.prototype.isAllowed()
+  // Extending existing logic, including from ACL.checkAccessForContext() method,
+  // ACL instance with missing property `permission` are not promoted to
+  // permission = ACL.DEFAULT config. Such ACL instances will hence always be
+  // inefective
+
+  /**
+   * Test if ACL's permission is ALLOW
+   * @param {String} permission The permission to test, expects one of 'ALLOW', 'DENY', 'DEFAULT'
+   * @param {String} defaultPermission The default permission to apply if not providing a finite one in the permission parameter
+   * @returns {Boolean} true if ACL permission is ALLOW
+   */
+  ACL.isAllowed = function(permission, defaultPermission) {
+    if (permission === ACL.DEFAULT) {
+      permission = defaultPermission || ACL.ALLOW;
+    }
+    return permission !== loopback.ACL.DENY;
+  };
+
+  /**
+   * Test if ACL's permission is ALLOW
+   * @param {String} defaultPermission The default permission to apply if missing in ACL instance
+   * @returns {Boolean} true if ACL permission is ALLOW
+   */
+  ACL.prototype.isAllowed = function(defaultPermission) {
+    return this.constructor.isAllowed(this.permission, defaultPermission);
+  };
+
   /**
    * Check if the request has the permission to access.
-   * @options {Object} context See below.
+   * @options {AccessContext|Object} context
+   * An AccessContext instance or a plain object with the following properties.
    * @property {Object[]} principals An array of principals.
    * @property {String|Model} model The model name or model class.
-   * @property {*} id The model instance ID.
+   * @property {*} modelId The model instance ID.
    * @property {String} property The property/method/relation name.
    * @property {String} accessType The access type:
-   *   READ, REPLICATE, WRITE, or EXECUTE.
-   * @param {Function} callback Callback function
+   * READ, REPLICATE, WRITE, or EXECUTE.
+   * @callback {Function} callback Callback function
+   * @param {String|Error} err The error object.
+   * @param {AccessRequest} result The resolved access request.
    */
-
   ACL.checkAccessForContext = function(context, callback) {
     if (!callback) callback = utils.createPromiseCallback();
     var self = this;
     self.resolveRelatedModels();
     var roleModel = self.roleModel;
 
-    context.registry = this.registry;
     if (!(context instanceof AccessContext)) {
+      context.registry = this.registry;
       context = new AccessContext(context);
     }
 
+    var authorizedRoles = {};
+    var remotingContext = context.remotingContext;
     var model = context.model;
+    var modelDefaultPermission = model && model.settings.defaultPermission;
     var property = context.property;
     var accessType = context.accessType;
     var modelName = context.modelName;
@@ -414,7 +454,13 @@ module.exports = function(ACL) {
         {inq: [ACL.REPLICATE, ACL.WRITE, ACL.ALL]} :
         {inq: [accessType, ACL.ALL]};
 
-    var req = new AccessRequest(modelName, property, accessType, ACL.DEFAULT, methodNames);
+    var req = new AccessRequest({
+      model: modelName,
+      property,
+      accessType,
+      permission: ACL.DEFAULT,
+      methodNames,
+      registry: this.registry});
 
     var effectiveACLs = [];
     var staticACLs = self.getStaticACLs(model.modelName, property);
@@ -445,6 +491,9 @@ module.exports = function(ACL) {
               function(err, inRole) {
                 if (!err && inRole) {
                   effectiveACLs.push(acl);
+                  // add the role to authorizedRoles if allowed
+                  if (acl.isAllowed(modelDefaultPermission))
+                    authorizedRoles[acl.principalId] = true;
                 }
                 done(err, acl);
               });
@@ -455,17 +504,30 @@ module.exports = function(ACL) {
       async.parallel(inRoleTasks, function(err, results) {
         if (err) return callback(err, null);
 
+        // resolved is an instance of AccessRequest
         var resolved = self.resolvePermission(effectiveACLs, req);
-        if (resolved && resolved.permission === ACL.DEFAULT) {
-          resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;
-        }
         debug('---Resolved---');
         resolved.debug();
+
+        // set authorizedRoles in remotingContext options argument if
+        // resolved AccessRequest permission is ALLOW, else set it to empty object
+        authorizedRoles = resolved.isAllowed() ? authorizedRoles : {};
+        saveAuthorizedRolesToRemotingContext(remotingContext, authorizedRoles);
         return callback(null, resolved);
       });
     });
     return callback.promise;
   };
+
+  function saveAuthorizedRolesToRemotingContext(remotingContext, authorizedRoles) {
+    const options = remotingContext && remotingContext.args && remotingContext.args.options;
+    // authorizedRoles key/value map is added to the options argument only if
+    // the latter exists and is an object. This means that the feature's availability
+    // will depend on the app configuration
+    if (options && typeof options === 'object') { // null is object too
+      options.authorizedRoles = authorizedRoles;
+    }
+  }
 
   /**
    * Check if the given access token can invoke the method
@@ -489,9 +551,9 @@ module.exports = function(ACL) {
       modelId: modelId,
     });
 
-    this.checkAccessForContext(context, function(err, access) {
+    this.checkAccessForContext(context, function(err, accessRequest) {
       if (err) callback(err);
-      else callback(null, access.permission !== ACL.DENY);
+      else callback(null, accessRequest.isAllowed());
     });
     return callback.promise;
   };
@@ -510,7 +572,9 @@ module.exports = function(ACL) {
    * Resolve a principal by type/id
    * @param {String} type Principal type - ROLE/APP/USER
    * @param {String|Number} id Principal id or name
-   * @param {Function} cb Callback function
+   * @callback {Function} callback Callback function
+   * @param {String|Error} err The error object
+   * @param {Object} result An instance of principal (Role, Application or User)
    */
   ACL.resolvePrincipal = function(type, id, cb) {
     cb = cb || utils.createPromiseCallback();
@@ -530,7 +594,7 @@ module.exports = function(ACL) {
           {where: {or: [{name: id}, {email: id}, {id: id}]}}, cb);
         break;
       default:
-        // try resolving a user model that matches principalType
+        // try resolving a user model with a name matching the principalType
         var userModel = this.registry.findModel(type);
         if (userModel) {
           userModel.findOne(
@@ -553,7 +617,9 @@ module.exports = function(ACL) {
    * @param {String} principalType Principal type
    * @param {String|*} principalId Principal id/name
    * @param {String|*} role Role id/name
-   * @param {Function} cb Callback function
+   * @callback {Function} callback Callback function
+   * @param {String|Error} err The error object
+   * @param {Boolean} isMapped is the ACL mapped to the role
    */
   ACL.isMappedToRole = function(principalType, principalId, role, cb) {
     cb = cb || utils.createPromiseCallback();
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -12,17 +12,29 @@ var debug = require('debug')('loopback:security:access-context');
  * Access context represents the context for a request to access protected
  * resources
  *
+ * NOTE While the method expects an array of principals in the AccessContext instance/object,
+ * it also accepts a single principal defined with the following properties:
+ * ```js
+ * {
+ *   // AccessContext instance/object
+ *   // ..
+ *   principalType: 'somePrincipalType', // APP, ROLE, USER, or custom user model name
+ *   principalId: 'somePrincipalId',
+ * }
+ * ```
+ *
  * @class
- * @options {Object} context The context object
+ * @options {AccessContext|Object} context An AccessContext instance or an object
  * @property {Principal[]} principals An array of principals
  * @property {Function} model The model class
  * @property {String} modelName The model name
- * @property {String} modelId The model id
+ * @property {*} modelId The model id
  * @property {String} property The model property/method/relation name
  * @property {String} method The model method to be invoked
- * @property {String} accessType The access type
- * @property {AccessToken} accessToken The access token
- *
+ * @property {String} accessType The access type: READ, REPLICATE, WRITE, or EXECUTE.
+ * @property {AccessToken} accessToken The access token resolved for the request
+ * @property {RemotingContext} remotingContext The request's remoting context
+ * @property {Registry} registry The application or global registry
  * @returns {AccessContext}
  * @constructor
  */
@@ -250,16 +262,23 @@ Principal.prototype.equals = function(p) {
 
 /**
  * A request to access protected resources.
- * @param {String} model The model name
- * @param {String} property
+ *
+ * The method can either be called with the following signature or with a single
+ * argument: an AccessRequest instance or an object containing all the required properties.
+ *
+ * @class
+ * @options {String|AccessRequest|Object} model|req The model name,<br>
+ *    or an AccessRequest instance/object.
+ * @param {String} property The property/method/relation name
  * @param {String} accessType The access type
  * @param {String} permission The requested permission
+ * @param {String[]} methodNames The names of involved methods
+ * @param {Registry} registry The application or global registry
  * @returns {AccessRequest}
- * @class
  */
-function AccessRequest(model, property, accessType, permission, methodNames) {
+function AccessRequest(model, property, accessType, permission, methodNames, registry) {
   if (!(this instanceof AccessRequest)) {
-    return new AccessRequest(model, property, accessType);
+    return new AccessRequest(model, property, accessType, permission, methodNames);
   }
   if (arguments.length === 1 && typeof model === 'object') {
     // The argument is an object that contains all required properties
@@ -268,14 +287,19 @@ function AccessRequest(model, property, accessType, permission, methodNames) {
     this.property = obj.property || AccessContext.ALL;
     this.accessType = obj.accessType || AccessContext.ALL;
     this.permission = obj.permission || AccessContext.DEFAULT;
-    this.methodNames = methodNames || [];
+    this.methodNames = obj.methodNames || [];
+    this.registry = obj.registry;
   } else {
     this.model = model || AccessContext.ALL;
     this.property = property || AccessContext.ALL;
     this.accessType = accessType || AccessContext.ALL;
     this.permission = permission || AccessContext.DEFAULT;
     this.methodNames = methodNames || [];
+    this.registry = registry;
   }
+  // do not create AccessRequest without a registry
+  assert(this.registry,
+    'Application registry is mandatory in AccessRequest but missing in provided argument(s)');
 }
 
 /**
@@ -306,6 +330,28 @@ AccessRequest.prototype.exactlyMatches = function(acl) {
   }
 
   return false;
+};
+
+/**
+ * Settle the accessRequest's permission if DEFAULT
+ * In most situations, the default permission can be resolved from the nested model
+ * config. An default permission can also be explicitly provided to override it or
+ * cope with AccessRequest instances without a nested model (e.g. model is '*')
+ *
+ * @param {String} defaultPermission (optional) the default permission to apply
+ */
+
+AccessRequest.prototype.settleDefaultPermission = function(defaultPermission) {
+  if (this.permission !== 'DEFAULT')
+    return;
+
+  var modelName = this.model;
+  if (!defaultPermission) {
+    var modelClass = this.registry.findModel(modelName);
+    defaultPermission = modelClass && modelClass.settings.defaultPermission;
+  }
+
+  this.permission = defaultPermission || 'ALLOW';
 };
 
 /**
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -5,10 +5,13 @@
 
 'use strict';
 var assert = require('assert');
+var expect = require('./helpers/expect');
 var loopback = require('../index');
 var Scope = loopback.Scope;
 var ACL = loopback.ACL;
 var request = require('supertest');
+var Promise = require('bluebird');
+var supertest = require('supertest');
 var Role = loopback.Role;
 var RoleMapping = loopback.RoleMapping;
 var User = loopback.User;
@@ -157,11 +160,24 @@ describe('security ACLs', function() {
     acls = acls.map(function(a) { return new ACL(a); });
 
     var perm = ACL.resolvePermission(acls, req);
+    // remove the registry from AccessRequest instance to ease asserting
+    delete perm.registry;
     assert.deepEqual(perm, {model: 'account',
       property: 'find',
       accessType: 'WRITE',
       permission: 'ALLOW',
       methodNames: []});
+
+    // NOTE: when fixed in chaijs, use this implement rather than modifying
+    // the resolved access request
+    //
+    // expect(perm).to.deep.include({
+    //   model: 'account',
+    //   property: 'find',
+    //   accessType: 'WRITE',
+    //   permission: 'ALLOW',
+    //   methodNames: [],
+    // });
   });
 
   it('should allow access to models for the given principal by wildcard', function() {
@@ -250,6 +266,7 @@ describe('security ACLs', function() {
       ],
     });
 
+    // ACL default permission is to DENY for model Customer
     Customer.settings.defaultPermission = ACL.DENY;
 
     ACL.checkPermission(ACL.USER, 'u001', 'Customer', 'name', ACL.WRITE,
@@ -471,4 +488,124 @@ describe('access check', function() {
         done();
       });
   });
+});
+
+describe('authorized roles propagation in RemotingContext', function() {
+  var app, request, accessToken;
+  var models = {};
+
+  beforeEach(setupAppAndRequest);
+
+  it('contains all authorized roles for a principal if query is allowed', function() {
+    return createACLs('MyTestModel', [
+      {permission: ACL.ALLOW, principalId: '$everyone'},
+      {permission: ACL.ALLOW, principalId: '$authenticated'},
+      {permission: ACL.ALLOW, principalId: 'myRole'},
+    ])
+    .then(makeAuthorizedHttpRequestOnMyTestModel)
+    .then(function() {
+      var ctx = models.MyTestModel.lastRemotingContext;
+      expect(ctx.args.options.authorizedRoles).to.eql(
+        {
+          $everyone: true,
+          $authenticated: true,
+          myRole: true,
+        }
+      );
+    });
+  });
+
+  it('does not contain any denied role even if query is allowed', function() {
+    return createACLs('MyTestModel', [
+      {permission: ACL.ALLOW, principalId: '$everyone'},
+      {permission: ACL.DENY,  principalId: '$authenticated'},
+      {permission: ACL.ALLOW, principalId: 'myRole'},
+    ])
+    .then(makeAuthorizedHttpRequestOnMyTestModel)
+    .then(function() {
+      var ctx = models.MyTestModel.lastRemotingContext;
+      expect(ctx.args.options.authorizedRoles).to.eql(
+        {
+          $everyone: true,
+          myRole: true,
+        }
+      );
+    });
+  });
+
+  it('honors default permission setting', function() {
+    // default permission is set to DENY for MyTestModel
+    models.MyTestModel.settings.defaultPermission = ACL.DENY;
+
+    return createACLs('MyTestModel', [
+      {permission: ACL.DEFAULT, principalId: '$everyone'},
+      {permission: ACL.DENY,    principalId: '$authenticated'},
+      {permission: ACL.ALLOW,   principalId: 'myRole'},
+    ])
+    .then(makeAuthorizedHttpRequestOnMyTestModel)
+    .then(function() {
+      var ctx = models.MyTestModel.lastRemotingContext;
+      expect(ctx.args.options.authorizedRoles).to.eql(
+        // '$everyone' is not expected as default permission is DENY
+        {myRole: true}
+      );
+    });
+  });
+
+  // helpers
+  function setupAppAndRequest() {
+    app = loopback({localRegistry: true, loadBuiltinModels: true});
+    app.use(loopback.rest());
+    app.set('remoting', {errorHandler: {debug: true, log: true}});
+    app.dataSource('db', {connector: 'memory'});
+    request = supertest(app);
+
+    app.enableAuth({dataSource: 'db'});
+    models = app.models;
+
+    // creating a custom model
+    const MyTestModel = app.registry.createModel('MyTestModel');
+    app.model(MyTestModel, {dataSource: 'db'});
+
+    // capturing the value of the last remoting context
+    models.MyTestModel.beforeRemote('find', function(ctx, unused, next) {
+      models.MyTestModel.lastRemotingContext = ctx;
+      next();
+    });
+
+    // creating a user, a role and a rolemapping binding that user with that role
+    return Promise.all([
+      models.User.create({username: 'myUser', email: 'myuser@example.com', password: 'pass'}),
+      models.Role.create({name: 'myRole'}),
+    ])
+    .spread(function(myUser, myRole) {
+      return Promise.all([
+        myRole.principals.create({principalType: 'USER', principalId: myUser.id}),
+        models.User.login({username: 'myUser', password: 'pass'}),
+      ]);
+    })
+    .spread(function(role, token) {
+      accessToken = token;
+    });
+  }
+
+  function createACLs(model, acls) {
+    acls = acls.map(function(acl) {
+      return models.ACL.create({
+        principalType: acl.principalType || ACL.ROLE,
+        principalId: acl.principalId,
+        model: acl.model || model,
+        property: acl.property || ACL.ALL,
+        accessType: acl.accessType || ACL.ALL,
+        permission: acl.permission,
+      });
+    });
+    return Promise.all(acls);
+  };
+
+  function makeAuthorizedHttpRequestOnMyTestModel() {
+    return request.get('/MyTestModels')
+      .set('X-Access-Token', accessToken.id)
+      .expect(200);
+  }
 });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -897,7 +897,7 @@ describe.onServer('Remote Methods', function() {
       request(app).get('/TestModels/saveOptions')
         .expect(204, function(err) {
           if (err) return done(err);
-          expect(actualOptions).to.eql({accessToken: null});
+          expect(actualOptions).to.include({accessToken: null});
           done();
         });
     });


slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@ebarault thank you for the pull request!

@raymondfeng I feel you are the best person to review this proposal at high level, could you PTAL?

bajtos:
@slnode ok to test

ebarault:
just rebased, but PR linter states 'PR out of date, needs to be rebased'

.eslintrc
```diff
@@ -5,6 +5,7 @@
       "ignoreComments": true,
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
-    }]
+    }],
+  "no-unused-expressions": "off",
```
bajtos: Please revert.

Gruntfile.js
```diff
@@ -4,6 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 /*global module:false*/
+'use strict';
```
bajtos: Changes in this file looks unrelated to me, could you please revert?

bajtos:
@ebarault 

> This PR proposes to add a new param (Array) in remoting context (name proposed: resolvedRoles, to be discussed) which is filled during Model Access Permission Checking phase with the roles for which the user is allowed to run the remote method.

I am little bit confused here. As far as I understand LoopBack's ACL system, there may be several different role arrays:
 - a list of roles the current user belongs to
 - a list of roles allowed to run the remote method
 - the intersection of the above two lists

What is stored in your new context property `resolvedRoles`?

ebarault:
@bajtos 
sorry about the mess with rebase, it went through at the moment of the large eslint refactor, it must have slipped in, i'll scratch and start over
about

about the pr intention: you're right, I should have made it clearer: the one i'm focusing here is the 3rd one: intersection of roles allowed to run the remote method and roles the current user belongs to

-or- rephrasing in my terms, to make sure we get each other: the roles for which the user is authorized to run a given remote method, which is in essence the exact purpose of the following course of actions:
`Model.checkAccess(context)` -> `ACL.checkAccessForContext(context)` -> `acls.forEach(acl, ...)` ->`roleModel.isInRole(acl.principalId, context)`  (i can detail the exact points of references of each call in the codebase if you will)

so when this set of tasks is achieved, for each acl covering the method the question has been solved "is the user in a role that can run the method?" : the general idea is to keep track of this result for further use in methods code or hooks to perform advanced CRUD.

A very basic example of when this can be useful: say a user is allowed to run a method by 2 roles, those 2 roles having conceptually very different access levels (custom role $organization:owner on one hand, and built-in smart role $authenticated) : inside that single method implementation you might want to treat the user differently from just an authenticated user, compared to an organization admin. Keeping track of the roleResolving status and making it accessible in the methods code greatly simplifies this process by avoiding running the role resolvers again for each role of interest for the method.

slnode:

slnode:

slnode:

slnode:

slnode:

ebarault:
@bajtos : clean rebase

bajtos:
@ebarault Thanks for rebasing the patch and providing a detailed explanation, your proposal makes sense to me.

I would like us to find a better name for `context.resolvedRoles` to avoid confusion about what exactly is stored there. How about `currentUserAllowedRoles`, `resolvedAndAllowedRoles` or `allowedRolesOfCurrentUser`?

BTW I think the `context` variable in this part of LoopBack refers to acl/security context, not the remoting context from remote hooks (strong-remoting)

ebarault:
the context variable here is rather mix, it is assembled from the call to [Model.checkAccess()](https://github.com/strongloop/loopback/blob/master/lib/model.js#L313) in model.js, including remotingContext as a nested param.
So as we have the chance to have the remotingContext here I believe it's best to attach the new variable to it directly, as it is more widely available in methods code and hooks than security context
 
As for the name, why not just `allowedRoles` ?

bajtos:
> So as we have the chance to have the remotingContext here I believe it's best to attach the new variable to it directly, as it is more widely available in methods code and hooks than security context

Ah, I see it now. Yes, I agree this makes more sense.

> As for the name, why not just `allowedRoles` ?

I would personally expect that `allowedRoles` is referring to *all* roles allowed to invoke the remote method, not only the roles of the current user.

Maybe we should use a name does not contain "resolved" (referring to current user) nor "allowed" (referring to roles configured for a method). What do you think about `effectiveRoles`?

ebarault:
`effective` as in [this](https://github.com/strongloop/loopback/blob/master/common/models/acl.js#L446)? ... this is tempting... but at this stage, the term `effective` stands both for granted and denied ACLs. the final check to allow or grant access is only done [there](https://github.com/strongloop/loopback/blob/master/lib/access-context.js#L291)

maybe should we keep the reference to `allowed` in the var name, or maybe `authorized`. So back to your proposal: `currentUserAllowedRoles` or `currentUserAuthorizedRoles`, although i find both a bit long..., maybe `authorizedRoles` is sufficiently explicit to refer to roles actively authorized to user?

bajtos:
Haha, naming is so hard!

As I was reading your comment, I think `currentAuthorizedRoles` may be the best trade-off between clarity and length. `authorizedRoles` looks good to me too, even if little bit less than `currentAuthorizedRoles`.

I'll let the final decision up to you, the author of this feature :)

ebarault:
@bajtos : finally opted for authorizedRoles :-)
just rebased.
as for updating the tests, this sounds more as an e2e test but could not find a proper test boiler test to add it (including custom role definition, associated custom role resolver, ACL definition, model access test)
tests in acl.test.js and role.test.js didn't quite match either imo
Could you advise on the best place to add the tests?

common/models/acl.js
```diff
@@ -376,6 +376,15 @@ module.exports = function(ACL) {
   };
 
   /**
+   * Test if ACL permission is not DENY
+   * @returns {Boolean} is permission type different than 'DENY' ?
+   */
+
+  ACL.prototype.isAllowed = function() {
+    return this.permission !== loopback.ACL.DENY;
```
bajtos: This is assuming that the permission value `ACL.DEFAULT` is always `ACL.ALLOW`. I think that's not always true, see L480:

```js
         if (resolved && resolved.permission === ACL.DEFAULT) {
            resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;		            resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;
          }		      
```

Unless I am missing something here?

bajtos:
> as for updating the tests, this sounds more as an e2e test but could not find a proper test boiler test to add it (including custom role definition, associated custom role resolver, ACL definition, model access test)
tests in acl.test.js and role.test.js didn't quite match either imo
Could you advise on the best place to add the tests?

I am afraid our test coverage of ACL is not as great as it should be and therefore it's very likely you will have to start from scratch here :(

I think the closest existing work is in [test/access-control.integration.js#L149-L208](https://github.com/strongloop/loopback/blob/fab5bd4fc5d055aafda9976f037f230c845c2486/test/access-control.integration.js#L149-L208), but to be honest I don't think that particular style of tests is suitable for your change.

I am proposing to start a new `describe` block in `test/acl.test.js` and create the necessary setup there.

@ebarault Thoughts?

ebarault:
oh i missed the tests in access-control.integration.js... knew there were some somewhere but could not put an hand on them. 

E2E app style app tests would be better i thing (if not required), since we need the REST layer to work for the method calls to have remotingContext, so we can check the content of authorizedRoles

bajtos:
@ebarault ping, what's the status of this pull request? Do you need any more information/help from me to move it forward?

ebarault:
@bajtos : i still need to work on the additional test cases
in the meantime you pointed out that 
> This is assuming that the permission value ACL.DEFAULT is always ACL.ALLOW. I think that's not always true

on which i replied [here](https://github.com/strongloop/loopback/pull/2957#issuecomment-265182216) -> your thoughts?

common/models/acl.js
bajtos: > I'm wondering of a case where the interim result of ACL scoring would be something like {role1: {score: 900, permission: 'ALLOW'}, role2: {score: 800, permission: 'DENY'}, role3: {score: 700, permission: 'ALLOW'}}, and what should be the content of authorizedRoles in such a case.

I think there should be a test case to show what's the expected result in this case.

I would expect that such request will be allowed, and my expectation as a LoopBack user would be that `role2` is not included in `authorizedRoles`.

If I understand/remember your code correctly, if the permission was `ACL.DEFAULT` and `model.settings.defaultPermission` was set to `DENY`, then `authorizedRoles` will include `role2`. Again, there should be a test exercising this scenario and verifying that an expected outcome is achieved.

ebarault: > I would expect that such request will be allowed, and my expectation as a LoopBack user would be that role2 is not included in authorizedRoles

agreed: this is the result one would expect. I'll check my code to make sure it is always the case and work on test cases when possible

ebarault:
@bajtos, i adjusted the PR to comply to our discussion and added the tests
+ rebased
could you please review?

common/models/acl.js
```diff
@@ -378,6 +379,18 @@ module.exports = function(ACL) {
...
+   * Test if ACL permission is ALLOW
+   * @returns {Boolean} true if ACL permission is ALLOW
+   */
+  ACL.prototype.isAllowed = function(defaultPermission) {
+    var permission = this.permission;
+    if (permission === ACL.DEFAULT) {
+      permission = defaultPermission || ACL.ALLOW;
+    }
+    return permission !== loopback.ACL.DENY;
+  };
```
ebarault: I changed the implem of this new helper following our discussion so to honour the defaultPermission setting for the given model

common/models/acl.js
```diff
@@ -378,6 +379,18 @@ module.exports = function(ACL) {
   };
 
   /**
+   * Test if ACL permission is ALLOW
+   * @returns {Boolean} true if ACL permission is ALLOW
+   */
+  ACL.prototype.isAllowed = function(defaultPermission) {
+    var permission = this.permission;
+    if (permission === ACL.DEFAULT) {
```
ebarault: I assumed we shouldn't treat the case where ACL.permission is not set as equivalent to ACL.permission = ACL.DEFAULT. 

Currently ([see here](https://github.com/ebarault/loopback/blob/0a2d3f9dc6ef2eff2fbab3d734cdb4d159e5b32b/common/models/acl.js#L488)) ACL.permission = undefined is not handled as equivalent to DEFAULT), and thus [this conditional test](https://github.com/ebarault/loopback/blob/0a2d3f9dc6ef2eff2fbab3d734cdb4d159e5b32b/common/models/acl.js#L494) would always fail if an ACL defined by code does not define the permission property to either DEFAULT, DENY or ALLOW

(did not verify for ACLs defined through the model JSON definition : how are handled missing `permission` property?)

common/models/acl.js
```diff
@@ -399,8 +411,16 @@ module.exports = function(ACL) {
...
+    var remotingContext = context && context.remotingContext;
+    if (remotingContext) {
+      var options = remotingContext.args.options || {};
+      remotingContext.args.options = extend(options, {authorizedRoles: []});
+    }
+    var authorizedRoles = [];
 
     var model = context.model;
+    var defaultPermission = model && model.settings.defaultPermission;
     var property = context.property;
```
ebarault: we get the ACL defaultPermission for the given model here

common/models/acl.js
```diff
@@ -445,6 +471,9 @@ module.exports = function(ACL) {
               function(err, inRole) {
                 if (!err && inRole) {
                   effectiveACLs.push(acl);
+                  if (acl.isAllowed(defaultPermission)) {
+                    addRole(authorizedRoles, acl.principalId);
+                  }
```
ebarault: and we use the defaultPermision here

common/models/acl.js
```diff
@@ -457,10 +486,14 @@ module.exports = function(ACL) {
 
         var resolved = self.resolvePermission(effectiveACLs, req);
         if (resolved && resolved.permission === ACL.DEFAULT) {
-          resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;
+          resolved.permission = defaultPermission || ACL.ALLOW;
         }
```
ebarault: shortened the line, reusing the defaultPermission var helper

common/models/acl.js
```diff
@@ -457,10 +486,14 @@ module.exports = function(ACL) {
...
         debug('---Resolved---');
         resolved.debug();
+        // add authorizedRoles to remotingContext if overall resolved permission status is ALLOW
+        if (resolved.permission === ACL.ALLOW && remotingContext) {
+          remotingContext.args.options.authorizedRoles = authorizedRoles;
+        };
         return callback(null, resolved);
```
ebarault: only add the authorizedRoles array to remotingContext if the overall resolved permission is ALLOW
the condition should not be necessary considering the request would return a 401 if resolved permission is DENY, but it's probably safer if we consider custom middlewares running just after the auth stage (would they run if the auth result is 401?)

test/acl.test.js
```diff
@@ -5,10 +5,13 @@
 
 'use strict';
 var assert = require('assert');
+var expect = require('./helpers/expect');
```
ebarault: reviewed some of your own new tests: should we use expect from loopback helpers or vanilla chai expect?

test/acl.test.js
```diff
@@ -5,10 +5,13 @@
...
 var loopback = require('../index');
 var Scope = loopback.Scope;
 var ACL = loopback.ACL;
 var request = require('supertest');
+var Promise = require('bluebird');
+var supertest = require('supertest-as-promised')(Promise);
```
bajtos: FWIW, the recent versions of supertest provide Promise API natively, supertest-as-promised was deprecated. Let's wait for https://github.com/strongloop/loopback/pull/3196 to be landed.

test/acl.test.js
```diff
@@ -472,3 +475,138 @@ describe('access check', function() {
...
+  function setupAppAndRequest() {
+    app = loopback({localRegistry: true, loadBuiltinModels: true});
+    app.use(loopback.rest());
+    app.set('remoting', {errorHandler: {debug: true, log: true}});
+    app.enableAuth();
+    app.dataSource('db', {connector: 'memory'});
+    request = supertest(app);
+
+    // attaching base models
+    ['User', 'Role', 'RoleMapping', 'ACL', 'AccessToken'].map(setupModel);
```
bajtos: - `app.enableAuth({ dataSource: 'db' })` automatically attaches all required models to the given datasource.
- any app model can be accessed via `app.models`, so it should be enough to set `var models = app.models;`

```js
app.enableAuthentication({dataSource: 'db'});
models = app.models;
```

This way, `setupModel` shouldn't be needed at all.

test/acl.test.js
```diff
@@ -472,3 +475,138 @@ describe('access check', function() {
...
+
+    // creating a custom model
+    models.MyTestModel = app.registry.createModel('MyTestModel');
+    app.model(models.MyTestModel, {dataSource: 'db'});
+
+    // beforeRemote hook on MyTestModel, as a promise
+    models.MyTestModel.onBeforeRemote = new Promise(function(resolve, reject) {
```
bajtos: Uff, I find this very confusing. Can you simplify to the following?

```js
models.MyTestModel.beforeRemote('find', function(ctx, notUsed,next) {
  models.MyTestModel.lastRemotingContext = ctx;
  next();
});
```

Unless I am missing something here?

test/acl.test.js
```diff
@@ -472,3 +475,138 @@ describe('access check', function() {
...
+      models.MyTestModel.beforeRemote('find', function(ctx, notUsed, next) {
+        resolve(ctx);
+        next();
+      });
+    });
+
+    // creating a user, a role and a rolemapping binding that user with that role
+    return Promise.all([
+      models.User.create({username: 'myUser', email: 'myuser@email.com', password: 'password'}),
```
bajtos: We use `@example.com` addresses, and usually set password to `pass` in the tests.

See http://loopback.io/doc/en/contrib/style-guide.html#email-examples, although it shows password set to `bar` (don't know why).

test/model.test.js
```diff
@@ -897,7 +897,7 @@ describe.onServer('Remote Methods', function() {
       request(app).get('/TestModels/saveOptions')
         .expect(204, function(err) {
           if (err) return done(err);
-          expect(actualOptions).to.eql({accessToken: null});
```
bajtos: Why is this change needed? The original code should produce a more helpful error message on failure.

test/acl.test.js
bajtos: #3196 was landed.

bajtos:
@ebarault ping - I am just wondering, what's the status of this patch?

ebarault:
i still plan to finish it in the coming days

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
...
+          permission = defaultPermission || ACL.ALLOW;
         }
         debug('---Resolved---');
         resolved.debug();
+
+        // set authorizedRoles in remotingContext options argument if
+        // checkAccessForContext resolved permission status is ALLOW,
+        // else set it to empty object
+        authorizedRoles = permission === ACL.ALLOW ? authorizedRoles : {};
+        setAuthorizedRoles(authorizedRoles);
```
ebarault: reduced down the whole setAuthorizedRoles logic to just these 2 lines in the main code block

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
...
+    function setAuthorizedRoles(authorizedRoles) {
+      var remotingContext = context.remotingContext;
+      const options = remotingContext && remotingContext.args.options;
+      // authorizedRoles key/value map is added to the options argument only if
+      // the latter exists and is an object. That means that the feature's availability
+      // will depend on the app configuration
+      if (options && typeof options === 'object') { // null is object too
+        options.authorizedRoles = authorizedRoles;
+      }
+    }
```
ebarault: the related helper to perform check ups on context.remotingContext.args.options before proceeding

test/acl.test.js
ebarault: 👍

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
         if (err) return callback(err, null);
 
         var resolved = self.resolvePermission(effectiveACLs, req);
+        var permission = resolved && resolved.permission;
         if (resolved && resolved.permission === ACL.DEFAULT) {
-          resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;
+          permission = defaultPermission || ACL.ALLOW;
         }
```
ebarault: you were proposing to share here the same logic as in ACL.isAllowed() but here were are not quite doing the same thing and not testing the same type of object: 
1. resolved is not an instance of ACL but an instance of AccessRequest. I'm reluctant in using ACL.isAllowed() on the permission part of an AccessRequest instance. Proper way would be to use the existing [AccessRequest.prototype.isAllowed](https://github.com/strongloop/loopback/blob/master/lib/access-context.js#L317) after adapting it to accept a default permission
2. we are not testing if allowed, we are changing the permission type if actually DEFAULT, because it is not done either [here](https://github.com/strongloop/loopback/blob/master/common/models/acl.js#L494) or [there](https://github.com/strongloop/loopback/blob/master/lib/access-context.js#L318) which is used in [model.js](https://github.com/strongloop/loopback/blob/master/lib/model.js#L353). This could be avoided and the code simplified/unified if were to align the implementation of `AccessRequest.prototype.isAllowed`

If you agree, i can track this as a different issue and address later

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
...
         debug('---Resolved---');
         resolved.debug();
+
+        // set authorizedRoles in remotingContext options argument if
+        // checkAccessForContext resolved permission status is ALLOW,
+        // else set it to empty object
+        authorizedRoles = permission === ACL.ALLOW ? authorizedRoles : {};
```
ebarault: then if we commit the changes discussed before, we can change this to 
`authorizedRoles = resolved.isAllowed() ? authorizedRoles : {};`
which would be better

(no need to even pass the defaultPermission since the accessRequest has already the targeted model and hence the defaultPermission)

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
         if (err) return callback(err, null);
 
         var resolved = self.resolvePermission(effectiveACLs, req);
+        var permission = resolved && resolved.permission;
         if (resolved && resolved.permission === ACL.DEFAULT) {
```
bajtos: I think this can be simplified to `if (permission === ACL.DEFAULT`, thoughts?

common/models/acl.js
bajtos: Uff, this goes too deep, beyond my current understanding. I'll leave this decision up to you to make.

Perhaps add a code comment explaining why we cannot use `ACL.isAllowed()` here?

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
...
+        // else set it to empty object
+        authorizedRoles = permission === ACL.ALLOW ? authorizedRoles : {};
+        setAuthorizedRoles(authorizedRoles);
         return callback(null, resolved);
       });
     });
     return callback.promise;
+
+    // helpers
+    function setAuthorizedRoles(authorizedRoles) {
```
bajtos: I am proposing to move this function out of `ACL.checkAccessForContext` and add `remotingContext` as a new parameter, so that the function does not require a closure from `ACL.checkAccessForContext` to access that reference.

There are two reasons for my proposal:
 - I personally consider such code easier to reason about, because it's easier to see what data is read/touched by `setAuthorizedRoles`
 - It's a also a (micro) performance optimisation. In the current implementation, V8 has to create a new function every time `ACL.checkAccessForContext` is called. I don't know what's the current status, but V8 was not able to optimise such short-living functions in the past.  By moving the function out, there is only one function that's called many times, and therefore V8 can optimise the heck out of it.

I would also like to propose  to use a different name, e.g. `saveAuthorizedRolesToRemotingContext`, and to  omit the `//helpers` comment.

common/models/acl.js
```diff
@@ -456,15 +489,34 @@ module.exports = function(ACL) {
...
+      var remotingContext = context.remotingContext;
+      const options = remotingContext && remotingContext.args.options;
```
bajtos: Should we check that `remotingContext.args` is set too, just to be 100% sure?

test/acl.test.js
```diff
@@ -472,3 +476,130 @@ describe('access check', function() {
...
+  it('contains all authorized roles for a principal if query is allowed', function() {
+    return createACLs('MyTestModel', [
+      {permission: ACL.ALLOW, principalId: '$everyone'},
+      {permission: ACL.ALLOW, principalId: '$authenticated'},
+      {permission: ACL.ALLOW, principalId: 'myRole'},
+    ])
+    .then(function() {
+      return request.get('/MyTestModels')
+        .set('X-Access-Token', accessToken.id)
+        .expect(200);
```
bajtos: This code is repeated in all tests, can you extract a shared helper function please? E.g.

```js
// usage
.then(makeAuthorizedHttpRequest)

// implementation
function makeAuthorizedHttpRequest() {
  return request.get('/MyTestModels')
    .set(...)
    .expect(200);
}
```

test/acl.test.js
```diff
@@ -472,3 +476,130 @@ describe('access check', function() {
...
+      );
+    });
+  });
+
+  // helpers
+  function setupAppAndRequest() {
+    app = loopback({localRegistry: true, loadBuiltinModels: true});
+    app.use(loopback.rest());
+    app.set('remoting', {errorHandler: {debug: true, log: true}});
+    app.enableAuth();
```
bajtos: Please remove in favour of L563.

test/acl.test.js
```diff
@@ -472,3 +476,130 @@ describe('access check', function() {
...
+    app.dataSource('db', {connector: 'memory'});
+    request = supertest(app);
+
+    app.enableAuth({dataSource: 'db'});
+    models = app.models;
+
+    // creating a custom model
+    models.MyTestModel = app.registry.createModel('MyTestModel');
+    app.model(models.MyTestModel, {dataSource: 'db'});
```
bajtos: Uff, modifying `app.models` directly is not a good practice. I am proposing the following:

```js
const MyTestModel = app.registry.createModel('MyTestModel');
app.model(MyTestModel, {dataSource: 'db'});
// models.MyTestModel was set by app.model() by now
```

test/acl.test.js
```diff
@@ -472,3 +476,130 @@ describe('access check', function() {
...
+
+    // capturing the value of the last remoting context
+    models.MyTestModel.beforeRemote('find', function(ctx, notUsed, next) {
```
bajtos: Nitpick (feel free to ignore): `unused` instead of `notUsed`?

test/model.test.js
ebarault: so... that's funny : you asked 29 days ago why this change was required, and i could not remember why. It happens that the test as-is fails since actualOptions is no longer equal to `{accessToken: null}`but to `{accessToken: null, authorizedRoles: {...}}`.
So i reverted my fix, and then you had the error you mentioned in [your comment here](https://github.com/strongloop/loopback/pull/2957#pullrequestreview-25762621)

i rolled the fix back, testing if accessToken is null is actually what we are testing here

ebarault:
@bajtos : i applied changes from last code review
I also fixed a number of related erroneous jsdoc comments
I isolated in a [dedicated commit](https://github.com/strongloop/loopback/pull/2957/commits/169e81a27ec684bac577bc30d0e893f75eb93abb) an additional change that adds a new method `settleDefaultPermission` on AccessRequest instances, that helps keeping the code tidier. The method replaces the messy code used to set the AccessRequest's permission to the correct default value when initial value is `DEFAULT`.

More comments inlined.

common/models/acl.js
```diff
@@ -204,11 +205,12 @@ module.exports = function(ACL) {
...
    * @param {Object[]) acls The list of ACLs
-   * @param {Object} req The request
-   * @returns {AccessRequest} result The effective ACL
+   * @param {AccessRequest} req The access request
+   * @returns {AccessRequest} result The resolved access request
    */
   ACL.resolvePermission = function resolvePermission(acls, req) {
     if (!(req instanceof AccessRequest)) {
+      req.registry = this.registry;
       req = new AccessRequest(req);
```
ebarault: as I did with AccessContext objects, I propose to add the Registry instance to the AccessRequest definition, so we can access models from that class, this is required by the new `settleDefaultPermission` method

common/models/acl.js
```diff
@@ -250,9 +252,12 @@ module.exports = function(ACL) {
...
-
-    var res = new AccessRequest(req.model, req.property, req.accessType,
-        permission || ACL.DEFAULT);
+    var res = new AccessRequest({
+      model: req.model,
+      property: req.property,
+      accessType: req.accessType,
+      permission: permission || ACL.DEFAULT,
+      registry: this.registry});
     return res;
```
ebarault: switching to one object param declaration when adding the registry as it reduces chances for further errors (adding things like `(param1, param2, null, null, param5)`)

Note: this should the default documented option, as for AccessContext

common/models/acl.js
```diff
@@ -316,8 +321,8 @@ module.exports = function(ACL) {
    * @param {String} property The property/method/relation name.
    * @param {String} accessType The access type.
    * @callback {Function} callback Callback function.
-   * @param {String|Error} err The error object
-   * @param {AccessRequest} result The access permission
+   * @param {String|Error} err The error object.
+   * @param {AccessRequest} result The resolved access request.
```
ebarault: the former definition is very vague

common/models/acl.js
```diff
@@ -332,10 +337,11 @@ module.exports = function(ACL) {
     var accessTypeQuery = (accessType === ACL.ALL) ? undefined :
       {inq: [accessType, ACL.ALL, ACL.EXECUTE]};
 
-    var req = new AccessRequest(model, property, accessType);
+    var req = new AccessRequest({model, property, accessType, registry: this.registry});
```
ebarault: adding registry to AccessRequest, going to one object param

common/models/acl.js
```diff
@@ -332,10 +337,11 @@ module.exports = function(ACL) {
...
 
     var acls = this.getStaticACLs(model, property);
 
+    // resolved is an instance of AccessRequest
     var resolved = this.resolvePermission(acls, req);
```
ebarault: further to commenting, i'd like to ask if ok (here or latter) to switch the name of the variable: resolved is just vague. This should be called something with accessRequest in it to make further code maintenance easier

common/models/acl.js
```diff
@@ -355,11 +361,9 @@ module.exports = function(ACL) {
           return callback(err);
         }
         acls = acls.concat(dynACLs);
+        // resolved is an instance of AccessRequest
         resolved = self.resolvePermission(acls, req);
-        if (resolved && resolved.permission === ACL.DEFAULT) {
-          var modelClass = self.registry.findModel(model);
-          resolved.permission = (modelClass && modelClass.settings.defaultPermission) || ACL.ALLOW;
-        }
+        resolved.settleDefaultPermission();
```
ebarault: see new AccessRequest instance method

common/models/acl.js
```diff
@@ -377,30 +381,62 @@ module.exports = function(ACL) {
...
+   * @property {*} modelId The model instance ID.
    * @property {String} property The property/method/relation name.
-   * @property {String} accessType The access type:
-   *   READ, REPLICATE, WRITE, or EXECUTE.
-   * @param {Function} callback Callback function
+   * @property {String} accessType The access type: READ, REPLICATE, WRITE, or EXECUTE.
+   * @callback {Function} callback Callback function
+   * @param {String|Error} err The error object.
+   * @param {AccessRequest} result The resolved access request.
    */
```
ebarault: jsdoc edit, following definition in [AccessContext](https://github.com/strongloop/loopback/blob/master/lib/access-context.js#L11)

common/models/acl.js
```diff
@@ -377,30 +381,62 @@ module.exports = function(ACL) {
...
-
   ACL.checkAccessForContext = function(context, callback) {
     if (!callback) callback = utils.createPromiseCallback();
     var self = this;
     self.resolveRelatedModels();
     var roleModel = self.roleModel;
 
-    context.registry = this.registry;
     if (!(context instanceof AccessContext)) {
+      context.registry = this.registry;
```
ebarault: in fact there's no reason why to add the registry if context is already an instance of AccessContext, as we assert the presence of registry in the class constructor

common/models/acl.js
```diff
@@ -414,7 +450,13 @@ module.exports = function(ACL) {
...
         {inq: [accessType, ACL.ALL]};
 
-    var req = new AccessRequest(modelName, property, accessType, ACL.DEFAULT, methodNames);
+    var req = new AccessRequest({
+      model: modelName,
+      property,
+      accessType,
+      permission: ACL.DEFAULT,
+      methodNames,
+      registry: this.registry});
```
ebarault: dito

common/models/acl.js
```diff
@@ -445,6 +487,9 @@ module.exports = function(ACL) {
               function(err, inRole) {
                 if (!err && inRole) {
                   effectiveACLs.push(acl);
+                  // add the role to authorizedRoles if allowed
+                  if (acl.isAllowed(modelDefaultPermission))
+                    addRoleToAuthorizedRoles(acl.principalId, authorizedRoles, context);
                 }
```
ebarault: call to helper to add Role the the interim AuthorizedRoles object

common/models/acl.js
```diff
@@ -455,18 +500,96 @@ module.exports = function(ACL) {
       async.parallel(inRoleTasks, function(err, results) {
         if (err) return callback(err, null);
 
+        // resolved is an instance of AccessRequest
         var resolved = self.resolvePermission(effectiveACLs, req);
-        if (resolved && resolved.permission === ACL.DEFAULT) {
-          resolved.permission = (model && model.settings.defaultPermission) || ACL.ALLOW;
-        }
+        resolved.settleDefaultPermission();
```
ebarault: dito

common/models/acl.js
```diff
@@ -455,18 +500,96 @@ module.exports = function(ACL) {
...
         debug('---Resolved---');
         resolved.debug();
+
+        // set authorizedRoles in remotingContext options argument if
+        // resolved AccessRequest permission is ALLOW, else set it to empty object
+        authorizedRoles = resolved.isAllowed() ? authorizedRoles : {};
+        saveAuthorizedRolesToRemotingContext(remotingContext, authorizedRoles);
         return callback(null, resolved);
```
ebarault: call to helper to save the authorizedRoles to remoting context options

common/models/acl.js
```diff
@@ -455,18 +500,96 @@ module.exports = function(ACL) {
...
     return callback.promise;
   };
 
+  function addRoleToAuthorizedRoles(role, authorizedRoles, context) {
+    let valueForAuthorizedRole = ACL.getValueForRoleInAuthorizedRoles;
+    if (typeof valueForAuthorizedRole !== 'function')
+      return;
+    authorizedRoles[role] = valueForAuthorizedRole(context);
+  }
+
```
ebarault: + this should not be a class method as we do not want users to easily modify this code
+ it allows using a custom `valueForAuthorizedRole`method in case the user wants more advanced values for authorized roles

common/models/acl.js
```diff
@@ -455,18 +500,96 @@ module.exports = function(ACL) {
...
+   * @property {*} modelId The model instance ID.
+   * @property {String} property The property/method/relation name.
+   * @property {String} accessType The access type:
+   *   READ, REPLICATE, WRITE, or EXECUTE.
+   * @returns {Object} The value to set for a role principalId in remotingContext.args.options.authorizedRoles
+   */
+  ACL.getValueForRoleInAuthorizedRoles = function(context) {
+    return true;
+  };
+
```
ebarault: this method is meant to be overriden by user if required. See jsdoc comments
maybe some of this jsdoc should go in the loopback.io doc only?

common/models/acl.js
```diff
@@ -455,18 +500,96 @@ module.exports = function(ACL) {
...
+  function saveAuthorizedRolesToRemotingContext(remotingContext, authorizedRoles) {
+    const options = remotingContext && remotingContext.args && remotingContext.args.options;
+    // authorizedRoles key/value map is added to the options argument only if
+    // the latter exists and is an object. This means that the feature's availability
+    // will depend on the app configuration
+    if (options && typeof options === 'object') { // null is object too
+      options.authorizedRoles = authorizedRoles;
+    }
+  }
+
```
ebarault: this method is not meant to be a class method, only a private helper

common/models/acl.js
```diff
@@ -489,9 +612,9 @@ module.exports = function(ACL) {
       modelId: modelId,
     });
 
-    this.checkAccessForContext(context, function(err, access) {
+    this.checkAccessForContext(context, function(err, accessRequest) {
       if (err) callback(err);
-      else callback(null, access.permission !== ACL.DENY);
+      else callback(null, accessRequest.isAllowed());
     });
```
ebarault: the existing `isAllowed` instance method of AccessRequest should be used here for more clarity instead of checking value of nested permission

common/models/acl.js
```diff
@@ -530,7 +655,7 @@ module.exports = function(ACL) {
           {where: {or: [{name: id}, {email: id}, {id: id}]}}, cb);
         break;
       default:
-        // try resolving a user model that matches principalType
+        // try resolving a user model with a name matching the principalType
         var userModel = this.registry.findModel(type);
```
ebarault: enhancing one of my own previous comment

lib/access-context.js
```diff
@@ -12,17 +12,29 @@ var debug = require('debug')('loopback:security:access-context');
...
+ * it also accepts a single principal defined with the following properties:
+ * ```js
+ * {
+ *   // AccessContext instance/object
+ *   // ..
+ *   principalType: 'somePrincipalType', // APP, ROLE, USER, or custom user model name
+ *   principalId: 'somePrincipalId',
+ * }
+ * ```
+ *
```
ebarault: adding a comment to document the ability to pass a principal definition at root of the AccessContext instance
I did not referenced the [prinicipalName](https://github.com/strongloop/loopback/blob/master/lib/access-context.js#L67) property as i can't seem to find any usage of it in the entire codebase

lib/access-context.js
```diff
@@ -12,17 +12,29 @@ var debug = require('debug')('loopback:security:access-context');
...
  * @property {String} property The model property/method/relation name
  * @property {String} method The model method to be invoked
- * @property {String} accessType The access type
- * @property {AccessToken} accessToken The access token
- *
+ * @property {String} accessType The access type: READ, REPLICATE, WRITE, or EXECUTE.
+ * @property {AccessToken} accessToken The access token resolved for the request
+ * @property {RemotingContext} remotingContext The request's remoting context
+ * @property {Registry} registry The application or global registry
  * @returns {AccessContext}
```
ebarault: fixing the jsdoc (based on AccessContext definition) + Adding registry

lib/access-context.js
```diff
@@ -250,16 +262,23 @@ Principal.prototype.equals = function(p) {
 
 /**
  * A request to access protected resources.
- * @param {String} model The model name
- * @param {String} property
+ *
+ * The method can either be called with the following signature or with a single
+ * argument: an AccessRequest instance or an object containing all the required properties.
+ *
```
ebarault: adding a comment to let know the one object param signature
(this should be the default doc, as for AccessContext)

lib/access-context.js
```diff
@@ -250,16 +262,23 @@ Principal.prototype.equals = function(p) {
...
+ * @class
+ * @options {String|AccessRequest|Object} model|req The model name,<br>
+ *    or an AccessRequest instance/object.
+ * @param {String} property The property/method/relation name
  * @param {String} accessType The access type
  * @param {String} permission The requested permission
+ * @param {String[]} methodNames The names of involved methods
+ * @param {Registry} registry The application or global registry
  * @returns {AccessRequest}
```
ebarault: jsdoc fixes

lib/access-context.js
```diff
@@ -268,14 +287,19 @@ function AccessRequest(model, property, accessType, permission, methodNames) {
     this.property = obj.property || AccessContext.ALL;
     this.accessType = obj.accessType || AccessContext.ALL;
     this.permission = obj.permission || AccessContext.DEFAULT;
-    this.methodNames = methodNames || [];
```
ebarault: this line is wrong

lib/access-context.js
```diff
@@ -268,14 +287,19 @@ function AccessRequest(model, property, accessType, permission, methodNames) {
...
+    this.methodNames = obj.methodNames || [];
+    this.registry = obj.registry;
   } else {
     this.model = model || AccessContext.ALL;
     this.property = property || AccessContext.ALL;
     this.accessType = accessType || AccessContext.ALL;
     this.permission = permission || AccessContext.DEFAULT;
     this.methodNames = methodNames || [];
+    this.registry = registry;
```
ebarault: adding registry, as in L291

lib/access-context.js
```diff
@@ -268,14 +287,19 @@ function AccessRequest(model, property, accessType, permission, methodNames) {
...
   }
+  // do not create AccessRequest without a registry
+  assert(this.registry,
+    'Application registry is mandatory in AccessRequest but missing in provided argument(s)');
 }
```
ebarault: asserting we have a registry in accessrequest instances

lib/access-context.js
```diff
@@ -309,6 +333,28 @@ AccessRequest.prototype.exactlyMatches = function(acl) {
...
+  if (!defaultPermission) {
+    var modelClass = this.registry.findModel(modelName);
+    defaultPermission = modelClass && modelClass.settings.defaultPermission;
+  }
+
+  this.permission = defaultPermission || 'ALLOW';
+};
+
+/**
  * Is the request for access allowed?
```
ebarault: new accessrequest instance method to help keeping the code tidier.
see code jsdoc comment

lib/access-context.js
```diff
@@ -250,16 +262,23 @@ Principal.prototype.equals = function(p) {
...
+ * @param {String[]} methodNames The names of involved methods
+ * @param {Registry} registry The application or global registry
  * @returns {AccessRequest}
- * @class
  */
-function AccessRequest(model, property, accessType, permission, methodNames) {
+function AccessRequest(model, property, accessType, permission, methodNames, registry) {
   if (!(this instanceof AccessRequest)) {
-    return new AccessRequest(model, property, accessType);
+    return new AccessRequest(model, property, accessType, permission, methodNames);
```
bajtos: I think this will not work, because the new argument `registry` is not forwarded.

bajtos:
> coverage/coveralls — Coverage decreased (-0.006%) to 89.295%

Looks like this decrease comes from the changes in `lib/access-context.js`, see https://coveralls.io/builds/10674265/source?filename=lib%2Faccess-context.js. I am not able to identify the cause of this decrease, it may be caused by new lines added to a branch that was not covered by any tests before (e.g. `AccessContext`/`AccessRequest` constructor).

I am proposing to ignore this tiny decrease in code coverage numbers.

test/acl.test.js
```diff
@@ -157,6 +160,8 @@ describe('security ACLs', function() {
     acls = acls.map(function(a) { return new ACL(a); });
 
     var perm = ACL.resolvePermission(acls, req);
+    // remove the registry from AccessRequest instance to ease asserting
+    delete perm.registry;
     assert.deepEqual(perm, {model: 'account',
```
ebarault: 👍  yeap this does not work ATM, i left a comment

test/model.test.js
ebarault: 👍  used this latter form

common/models/acl.js
ebarault: 👍

common/models/acl.js
```diff
@@ -377,30 +381,62 @@ module.exports = function(ACL) {
...
+  ACL.prototype.isAllowed = function(defaultPermission) {
+    return this.constructor.isAllowed(this.permission, defaultPermission);
+  };
+
   /**
    * Check if the request has the permission to access.
-   * @options {Object} context See below.
+   * @options {AccessContext} context An AccessContext instance.
    * @property {Object[]} principals An array of principals.
-   * @property {String|Model} model The model name or model class.
```
ebarault: did it and also removed modelName then
will come back to this latter, we should not be confusing one property for two different things

common/models/acl.js
```diff
@@ -377,30 +381,62 @@ module.exports = function(ACL) {
...
-   * @property {*} id The model instance ID.
+   * @property {Function} model The model class.
+   * @property {String} modelName The model name.
+   * @property {*} modelId The model instance ID.
    * @property {String} property The property/method/relation name.
-   * @property {String} accessType The access type:
-   *   READ, REPLICATE, WRITE, or EXECUTE.
-   * @param {Function} callback Callback function
+   * @property {String} accessType The access type: READ, REPLICATE, WRITE, or EXECUTE.
```
ebarault: just rolled back

lib/access-context.js
ebarault: no idea TBH

bajtos:
Landed 🎉

common/models/acl.js
ejvaitkus: When specifying an "arg": "options" in a common/models/xxx.json 'accepts' section, we ran into problems, finding 'authorizedRoles: {}' added to what was intended to be a list of options to pass to a system call, causing problems since we are simply passing the options values forward (in which case authorizedRoles becomes an invalid option). Our workaround was to rename to arg from "options" to something else (eg "lsfOptions".  It would be more convenient to not have to worry about this; without knowing the inner workings, having the authorizedRoles in some more uniquely named structure would help the situation.

ebarault: hi @ejvaitkus : i'm afraid it's not specific to the `authorizedRoles` property as this remote method argument is also used to pass remoting context properties downstream to hooks and methods. see  https://loopback.io/doc/en/lb3/Using-current-context.html

it's on purpose that the generic arg `options` is used as is in LB3.

PR #2959, opened 2016-11-21
Support multi-tenancy in change replication - allow apps to extend Change model with extra properties

kobaska:
### Description
Enable replication to scope changes by tenant. This makes querying for changes faster, as we can filter by tenant where it is provided.

#### Related issues

- None

### Checklist

### DIFF
--- a/common/models/change.js
+++ b/common/models/change.js
@@ -184,16 +184,32 @@ module.exports = function(Change) {
 
     cb = cb || utils.createPromiseCallback();
 
-    change.currentRevision(function(err, rev) {
+    const model = this.getModelCtor();
+    const id = this.getModelId();
+
+    model.findById(id, function(err, inst) {
       if (err) return cb(err);
 
+      if (inst) {
+        inst.fillCustomChangeProperties(change, function() {
+          const rev = Change.revisionForInst(inst);
+          prepareAndDoRectify(rev);
+        });
+      } else {
+        prepareAndDoRectify(null);
+      }
+    });
+
+    return cb.promise;
+
+    function prepareAndDoRectify(rev) {
       // avoid setting rev and prev to the same value
       if (currentRev === rev) {
         change.debug('rev and prev are equal (not updating anything)');
         return cb(null, change);
       }
 
-      // FIXME(@bajtos) Allo callers to pass in the checkpoint value
+      // FIXME(@bajtos) Allow callers to pass in the checkpoint value
       // (or even better - a memoized async function to get the cp value)
       // That will enable `rectifyAll` to cache the checkpoint value
       change.constructor.getCheckpointModel().current(
@@ -202,8 +218,7 @@ module.exports = function(Change) {
           doRectify(checkpoint, rev);
         }
       );
-    });
-    return cb.promise;
+    }
 
     function doRectify(checkpoint, rev) {
       if (rev) {
@@ -228,7 +243,7 @@ module.exports = function(Change) {
           if (currentRev) {
             change.prev = currentRev;
           } else if (!change.prev) {
-            change.debug('ERROR - could not determing prev');
+            change.debug('ERROR - could not determine prev');
             change.prev = Change.UNKNOWN;
           }
           change.debug('updated prev');
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -1052,6 +1052,7 @@ module.exports = function(registry) {
     var idName = this.dataSource.idName(this.modelName);
     var Change = this.getChangeModel();
     var model = this;
+    const changeFilter = this.createChangeFilter(since, filter);
 
     filter = filter || {};
     filter.fields = {};
@@ -1059,10 +1060,7 @@ module.exports = function(registry) {
     filter.fields[idName] = true;
 
     // TODO(ritch) this whole thing could be optimized a bit more
-    Change.find({where: {
-      checkpoint: {gte: since},
-      modelName: this.modelName,
-    }}, function(err, changes) {
+    Change.find(changeFilter, function(err, changes) {
       if (err) return callback(err);
       if (!Array.isArray(changes) || changes.length === 0) return callback(null, []);
       var ids = changes.map(function(change) {
@@ -1759,11 +1757,12 @@ module.exports = function(registry) {
     assert(BaseChangeModel,
       'Change model must be defined before enabling change replication');
 
+    const additionalChangeModelProperties =
+      this.settings.additionalChangeModelProperties || {};
+
     this.Change = BaseChangeModel.extend(this.modelName + '-change',
-      {},
-      {
-        trackModel: this,
-      }
+      additionalChangeModelProperties,
+      {trackModel: this}
     );
 
     if (this.dataSource) {
@@ -1926,6 +1925,77 @@ module.exports = function(registry) {
         next();
       };
     }
+  };
+
+  /**
+   * Get the filter for searching related changes.
+   *
+   * Models should override this function to copy properties
+   * from the model instance filter into the change search filter.
+   *
+   * ```js
+   * module.exports = (TargetModel, config) => {
+   *   TargetModel.createChangeFilter = function(since, modelFilter) {
+   *     const filter = this.base.createChangeFilter.apply(this, arguments);
+   *     if (modelFilter && modelFilter.where && modelFilter.where.tenantId) {
+   *       filter.where.tenantId = modelFilter.where.tenantId;
+   *     }
+   *     return filter;
+   *   };
+   * };
+   * ```
+   *
+   * @param {Number} since Return only changes since this checkpoint.
+   * @param {Object} modelFilter Filter describing which model instances to
+   * include in the list of changes.
+   * @returns {Object} The filter object to pass to `Change.find()`. Default:
+   * ```
+   * {where: {checkpoint: {gte: since}, modelName: this.modelName}}
+   * ```
+   */
+  PersistedModel.createChangeFilter = function(since, modelFilter) {
+    return {
+      where: {
+        checkpoint: {gte: since},
+        modelName: this.modelName,
+      },
+    };
+  };
+
+  /**
+   * Add custom data to the Change instance.
+   *
+   * Models should override this function to duplicate model instance properties
+   * to the Change instance properties, typically to allow the changes() method
+   * to filter the changes using these duplicated properties directly while
+   * querying the Change model.
+   *
+   * ```js
+   * module.exports = (TargetModel, config) => {
+   *   TargetModel.prototype.fillCustomChangeProperties = function(change, cb) {
+   *     var inst = this;
+   *     const base = this.constructor.base;
+   *     base.prototype.fillCustomChangeProperties.call(this, change, err => {
+   *       if (err) return cb(err);
+   *
+   *       if (inst && inst.tenantId) {
+   *         change.tenantId = inst.tenantId;
+   *       } else {
+   *         change.tenantId = null;
+   *       }
+   *
+   *       cb();
+   *     });
+   *   };
+   * };
+   * ```
+   *
+   * @callback {Function} callback
+   * @param {Error} err Error object; see [Error object](http://loopback.io/doc/en/lb3/Error-object.html).
+   */
+  PersistedModel.prototype.fillCustomChangeProperties = function(change, cb) {
+    // no-op by default
+    cb();
   };
 
   PersistedModel.setup();
--- a/test/change.test.js
+++ b/test/change.test.js
@@ -9,9 +9,9 @@ var async = require('async');
 var expect = require('./helpers/expect');
 var loopback = require('../');
 
-var Change, TestModel;
-
 describe('Change', function() {
+  let Change, TestModel;
+
   beforeEach(function() {
     var memory = loopback.createDataSource({
       connector: loopback.Memory,
@@ -321,7 +321,6 @@ describe('Change', function() {
       change.rectify()
         .then(function(ch) {
           assert.equal(ch.rev, test.revisionForModel);
-
           done();
         })
         .catch(done);
@@ -607,5 +606,70 @@ describe('Change', function() {
         done();
       });
     });
+  });
+});
+
+describe('Change with with custom properties', function() {
+  let Change, TestModel;
+
+  beforeEach(function() {
+    let memory = loopback.createDataSource({
+      connector: loopback.Memory,
+    });
+
+    TestModel = loopback.PersistedModel.extend('ChangeTestModelWithTenant',
+      {
+        id: {id: true, type: 'string', defaultFn: 'guid'},
+        tenantId: 'string',
+      },
+      {
+        trackChanges: true,
+        additionalChangeModelProperties: {tenantId: 'string'},
+      });
+    this.modelName = TestModel.modelName;
+
+    TestModel.prototype.fillCustomChangeProperties = function(change, cb) {
+      var inst = this;
+
+      if (inst && inst.tenantId) {
+        change.tenantId = inst.tenantId;
+      } else {
+        change.tenantId = null;
+      }
+
+      cb();
+    };
+
+    TestModel.attachTo(memory);
+    TestModel._defineChangeModel();
+    Change = TestModel.getChangeModel();
+  });
+
+  describe('change.rectify', function() {
+    const TENANT_ID = '123';
+    let change;
+
+    beforeEach(givenChangeInstance);
+
+    it('stores the custom property in the Change instance', function() {
+      return change.rectify().then(function(ch) {
+        expect(ch.toObject()).to.have.property('tenantId', TENANT_ID);
+      });
+    });
+
+    function givenChangeInstance() {
+      const data = {
+        foo: 'bar',
+        tenantId: TENANT_ID,
+      };
+
+      return TestModel.create(data)
+        .then(function(model) {
+          const modelName = TestModel.modelName;
+          return Change.findOrCreateChange(modelName, model.id);
+        }).then(function(ch) {
+          change = ch;
+        });
+    }
   });
 });
--- a/test/replication.test.js
+++ b/test/replication.test.js
@@ -1905,3 +1905,131 @@ describe('Replication / Change APIs', function() {
     });
   }
 });
+
+describe('Replication / Change APIs with custom change properties', function() {
+  this.timeout(10000);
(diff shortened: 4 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

kobaska:
@bajtos Can this test be rerun. It is failing on existing code.

bajtos:
Hi @kobaska, thank you for the pull request. I am rather reluctant to add this feature directly to LoopBack, as it's too specific in my opinion. I would prefer to see it in an independent mixin instead. If the current code in LoopBack does not provide enough extension points allowing to implement such mixin, then I am happy to add those extension points only.

common/models/change.js
```diff
@@ -186,27 +186,51 @@ module.exports = function(Change) {
...
+
+    model.findById(id, function(err, inst) {
       if (err) return cb(err);
 
-      // avoid setting rev and prev to the same value
-      if (currentRev === rev) {
-        change.debug('rev and prev are equal (not updating anything)');
-        return cb(null, change);
-      }
+      change.tenant = getTenant(model, inst);
```
bajtos: If I understand this part correctly, then it's adding a computed `tenant` property to each Change instance. I think it should be possible to accomplish this by using "before save" or "persist" hook.

```js
module.exports = myMixinFunction(TargetModel, config) {
  TargetModel.getChangeModel().observe('before save', function(ctx, next) {
    var tenant = getTenant(TargetModel, ctx.instance || ctx.data);
    if (tenant) {
     (ctx.instance || ctx.data).tenant = tenant;
    }
  });
});
```

common/models/change.js
```diff
@@ -254,23 +278,35 @@ module.exports = function(Change) {
...
-      if (err) return cb(err);
-      if (inst) {
-        cb(null, Change.revisionForInst(inst));
-      } else {
-        cb(null, null);
-      }
-    });
+
+    if (instance) {
+      cb(null, Change.revisionForInst(instance));
```
bajtos: This does not make sense to me. Just replace calls of `change.currentRevision(inst)` with `change.constructor.revisionForInst(inst)` and then you don't need to change this method.

lib/persisted-model.js
```diff
@@ -996,12 +997,13 @@ module.exports = function(registry) {
...
    * @param  {Number}   since    Return only changes since this checkpoint.
    * @param  {Object}   filter   Include only changes that match this filter, the same as for [#persistedmodel-find](find()).
+   * @param  {String}   tenant   Include only changes for this tenant
    * @callback {Function} callback Callback function called with `(err, changes)` arguments.  Required.
    * @param {Error} err Error object; see [Error object](http://docs.strongloop.com/display/LB/Error+object).
    * @param {Array} changes An array of [Change](#change) objects.
    */
 
-  PersistedModel.changes = function(since, filter, callback) {
+  PersistedModel.changes = function(since, filter, tenant, callback) {
```
bajtos: Ok, I think this part would be difficult to implement in a mixin with the current loopback code.

Since the "tenant" property is copied from the target PersistedModel to the Change model, one should be able to express the desire to limit synchronization to a certain tenant using existing `filter` argument only.

```js
MyModel.changes(since, { where: { tenant: tenantId } }, cb);
```

Then, inside `changes` function, we should be able to detect whether `filter` contains `tenant` clause and copy it to `changeFilter`. The only change needed from LoopBack is to allow mixins to override/customize the way how `changeFilter` is built.

```js
// in loopback
PersistedModel.createChangeFilter = function(since, modelFilter) {
  return {
    where: {
      checkpoint: {gte: since},
      modelName: this.modelName,
    },
  };
};

// in your mixin
module.exports = myMixinFunction(TargetModel, config) {
  // ...

  originalFilterFactory = TargetModel.createChangeFilter;
  TargetModel.createChangeFilter = function(since, modelFilter) {
    var tenantProperty = // find it out
    var filter = originalFilterFactory.call(this, since, modelFilter);
    if (modelFilter && modelFilter.where && modelFilter.where[tenantProperty]) {
      filter.where.tenant = modelFilter.where[tenantProperty];
    }
    return filter;
  };
};
```

bajtos:
@kobaska ping, what's the status of this patch? Are you still keen to get it finished?

common/models/change.js
kobaska: @bajtos 
An issue that has risen with this solution, is that we have to pass the Model instance instead of change instance to the getTenant function. We could do a query for the model instance as we have the model id from the ctx, but this adds another query to each save.

Adding tenant to the change object and calculating the rev, both require the model instance. Our solution gets the model once in rectify function and uses that to set the tenant and revision.

bajtos: First of all, I would like to acknowledge that I consider your use case, where you would like to copy certain model properties to each change record, as a valid one that we should support.

The question now is how to implement this feature in a way that's generic enough to be useful for other users of the framework too.

Can we perhaps build on top of my proposal in https://github.com/strongloop/loopback/pull/2959#discussion_r89064825, where I am describing a new method `PersistedModel.createChangeFilter` used when querying changes, and use a similar approach for the other part where we are creating/updating the change records? Something like `PersistedModel.createChangeData` that should supersede current [`Change.revisionForInstance`](https://github.com/strongloop/loopback/blob/f71030f8c2f61163b689f93e6086bfe1e4db9df6/common/models/change.js#L296-L299) method. The downside is that this proposal will require complex changes in several places, as we will have to stop passing a single property `revision` and start passing a set of properties instead.

Alternatively, can we perhaps keep your current design, but make it more flexible by calling a method on `PersistedModel` instead of calling a hard-coded `getTenant` method?

Here is a quick mock-up:

```js
// here in this method
model.findById(id, function(err, inst) {
  if (err) return cb(err);

  if (inst) {
    inst.fillChangeProperties(change);
  }

  change.currentRevision(inst, function(err, rev) { 
    // etc.
  });
});

// in lib/persisted-model.js

PersistedModel.prototype.fillChangeProperties = function(change) {
  // no-op by default
};
```

Can we come up with a better/more descriptive method name than `fillChangeProperties`?

Should this new method by async (accepting a callback arg) to allow the method to possibly fetch additional data?

kobaska:
@bajtos, I'm going to be away for a month on holiday. Will do the modifications when I'm back.

kobaska:
@bajtos I have done the changes using mixins instead as you suggested. Can you review this please?

test/change.test.js
```diff
@@ -601,10 +610,90 @@ describe('Change', function() {
...
+    var memory = loopback.createDataSource({
+      connector: loopback.Memory
+    });
+    TestModelWithTenant = loopback.PersistedModel.extend('ChangeTestModel',
+      {
+        id: {id: true, type: 'string', defaultFn: 'guid'}
+      },
+      {
+        trackChanges: true,
+        tenantProperty: tenantProperty
```
bajtos: This is more generic than needed in the test. We are not testing a multi-tenancy mixin, just verifying that it's possible to propagate properties from a model to the change model.

For the purpose of this test, please assume a hard-coded `tenant` property name, e.g. `tenantId`.

test/change.test.js
```diff
@@ -601,10 +610,90 @@ describe('Change', function() {
...
+          tenant: null,
         });
 
         done();
       });
     });
   });
 });
+
+describe('Change with multi tenancy', function() {
```
bajtos: Since we are testing the generic ability of `Change` model to support extra properties, the test name should not refer to "multi tenancy". I am proposing `Change with custom properties`.

test/change.test.js
```diff
@@ -601,10 +610,90 @@ describe('Change', function() {
...
+        tenant = inst[tenantProperty];
+      }
+
+      change.tenant = tenant;
+
+      cb(null);
+    };
+  });
+
+  describe('change.rectify - tenant variant', function() {
```
bajtos: No need to repeat the fact that our Change model has extra properties, that was already said in `describe()`.

test/change.test.js
```diff
@@ -601,10 +610,90 @@ describe('Change', function() {
...
+
+    beforeEach(function(done) {
+      ChangeWithTenant = TestModelWithTenant.getChangeModel();
+      var test = this;
+      test.data = {
+        foo: 'bar',
+        tenantId: tenantId
+      };
+
+      TestModelWithTenant.create(test.data, function(err, model) {
```
bajtos: Mocha supports promise-based hooks, no need to mix callbacks and promises here.

```js
return TestModelWithTenant.create(test.data)
  .then(function(model) {
    test.revisionForModel = ChangeWithTenant.revisionForInst(model);
    return ChangeWithTenant.findOrCreateChange(TestModelWithTenant.modelName, model.id);
  })
  .then(function(ch) {
    change = ch;
  });
}
```

bajtos: Also mixing describe-wide variables (`change`, `tenantId`) with `test` properties (`test.revisionForModel`) is confusing. Please use variables only and move `test.revisionForModel` to a regular variable.

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
     return getPropValue(list, 'id');
   }
 });
+
+describe('Replication / Change APIs with multi tenancy', function() {
```
bajtos: `... with custom change properties`

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+
+  beforeEach(function() {
+    tid++;
+    useSinceFilter = false;
+    var test = this;
+
+    dataSource = this.dataSource = loopback.createDataSource({
+      connector: loopback.Memory
+    });
+    SourceModelWithTenant = this.SourceModelWithTenant = PersistedModel.extend(
```
bajtos: We are moving away from global registry of models in favour of local per-app registries. That way there is no need for `tid`.

```js
app = loopback({localRegistry: true, loadBuiltinModels: true});

SourceModelWithTenant = app.registry.createModel({
  name: 'SourceModelWithTenant',
  properties: { /*...*/ },
  trackChanges: true
});
app.model(SourceModelWithTenant, { dataSource: 'db' });

// etc.
```

Also `tenantProperty` setting is not needed here.

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+
+      change.tenant = tenant;
+
+      cb(null);
+    };
+
+    test.startingCheckpoint = -1;
+  });
+
+  describe('Model.changes(since, filter, callback) - tenant variant', function() {
```
bajtos: No need to repeat `tenant variant`, please remove.

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+    var tenant = '123';
+
+    it('Get changes since the given checkpoint for the given tenant', function(done) {
```
bajtos: `queries changes using improved filter for custom properties`

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+      var test = this;
+      this.SourceModelWithTenant.create([{name: 'foo', tenantId: tenant}, {name: 'foo', tenantId: '456'}], function(err) {
+        if (err) return done(err);
+
+        setTimeout(function() {
+          test.SourceModelWithTenant.changes(test.startingCheckpoint, {where: {tenantId: tenant}}, function(err, changes) {
+            assert.equal(changes.length, 1);
+            assert.equal(changes[0].tenant, tenant);
```
bajtos: I believe this test would pass even without your changes applied. While it's ok to verify that the expected changes are returned, we also need to verify that correct queries were made against the datasource, i.e. that `tenantId` filter was applied when making the first `find` query on the `Change` model.

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+
+    it('excludes changes from older checkpoints', function(done) {
+      var FUTURE_CHECKPOINT = 999;
+
+      SourceModelWithTenant.create({ name: 'foo', tenantId: tenant }, function(err) {
+        if (err) return done(err);
+        SourceModelWithTenant.changes(FUTURE_CHECKPOINT, {where: {tenantId: tenant}}, function(err, changes) {
+          if (err) return done(err);
+
+          expect(changes).to.be.empty; //jshint ignore:line
```
bajtos:
```js
expect(changes).to.be.empty();
```

Just make sure `expect` is defined as `var expect = require('./helpers/expect')` at the top of this file.

test/replication.test.js
```diff
@@ -1804,3 +1804,105 @@ describe('Replication / Change APIs', function() {
...
+            assert.equal(changes.length, 1);
+            assert.equal(changes[0].tenant, tenant);
+
+            done();
+          });
+        }, 1);
+      });
+    });
+
+    it('excludes changes from older checkpoints', function(done) {
```
bajtos: I think this test is redundant - checkpoint filter is already tested, a custom `createChangeFilter` should not affect it as long as it preserves the original filter produced by built-in `createChangeFilter` method.

bajtos:
@kobaska ping, what's the status of this patch?

kobaska:
@bajtos I'll get on it next week. Been busy with other stuff.

test/replication.test.js
kobaska: It would be better to move all the tests to this format. I would rather do it separate to this pull request.

kobaska:
Hi @bajtos,

I have had another go, have a look and see. Regarding adding the custom property on Change model, I couldn't figure out a way to do that through mixins, as they all run after setup, which is too late. Hence I have added a new property additionalChangeModelProperties in settings to do this. Let me know if there is a better way to do this.

bajtos:
> I have added a new property additionalChangeModelProperties in settings to do this

I think that's a reasonable solution 👍 

The code changes looks mostly good to me, I'll make few final improvements myself. First of all, I am going to rebase this patch against `master`, as we are no longer landing new features to 2.x.

common/models/change.js
```diff
@@ -184,15 +184,15 @@ module.exports = function(Change) {
 
     cb = cb || utils.createPromiseCallback();
 
-    var model = this.getModelCtor();
-    var id = this.getModelId();
+    const model = this.getModelCtor();
```
bajtos: http://loopback.io/doc/en/contrib/style-guide.html

lib/persisted-model.js
```diff
@@ -1930,29 +1928,34 @@ module.exports = function(registry) {
...
    * module.exports = (TargetModel, config) => {
-   *   var originalFilterFactory = TargetModel.createChangeFilter;
-   *
-   *   TargetModel.createChangeFilter = function (since, modelFilter) {
-   *     var customProperty = TargetModel.settings.changeCustomProperty;
-   *
-   *     var filter = originalFilterFactory.call(this, since, modelFilter);
-   *     if (modelFilter && modelFilter.where && modelFilter.where[customProperty]) {
-   *       filter.where[customProperty] = modelFilter.where[customProperty];
+   *   TargetModel.createChangeFilter = function(since, modelFilter) {
```
bajtos: Simplified the example - use hard-coded property name "tenantId", get the base `createChangeFilter` from the parent model.

test/change.test.js
```diff
@@ -610,71 +610,66 @@ describe('Change', function() {
 });
 
 describe('Change with with custom properties', function() {
-  var ChangeWithCustomProperty, TestModelWithModifiedChangeModel;
-  var customProperty = 'tenantId';
```
bajtos: Similarly here, I find the tests easier to read/understand when the property name is hard-coded.

test/change.test.js
```diff
@@ -610,71 +610,66 @@ describe('Change', function() {
...
+    let memory = loopback.createDataSource({
       connector: loopback.Memory,
     });
 
-    var additionalChangeModelProperties = {};
-    additionalChangeModelProperties[customProperty] = {type: 'string'};
-
-    TestModelWithModifiedChangeModel = loopback.PersistedModel
-    .extend('ChangeTestModel',
+    TestModel = loopback.PersistedModel.extend('ChangeTestModelWithTenant',
```
bajtos: Since we are already within "Change with with custom properties" block, I feel there is no need to repeat that information in model property names.

It is important to use a unique model name in `extend()` arguments though.

test/change.test.js
```diff
@@ -610,71 +610,66 @@ describe('Change', function() {
...
       {
         id: {id: true, type: 'string', defaultFn: 'guid'},
+        tenantId: 'string',
```
bajtos: This was probably an oversight - the custom property should be defined on both Model and Change, right?

test/change.test.js
```diff
@@ -610,71 +610,66 @@ describe('Change', function() {
...
 
   describe('change.rectify', function() {
-    var change;
-    var customPropertyValue = '123';
+    const TENANT_ID = '123';
+    let change;
 
-    beforeEach(function() {
-      var test = this;
+    beforeEach(givenChangeInstance);
```
bajtos: http://loopback.io/doc/en/contrib/style-guide.html

test/replication.test.js
```diff
@@ -1921,99 +1919,105 @@ describe('Replication / Change APIs with custom change properties', function() {
...
+      {
+        id: {id: true, type: String, defaultFn: 'guid'},
+        customProperty: {type: 'string'},
+      },
+      {
+        trackChanges: true,
+        additionalChangeModelProperties: {customProperty: {type: 'string'}},
+      });
+
+    SourceModel.createChangeFilter = function(since, modelFilter) {
```
bajtos: Moved the setup of `createChangeFilter` and `fillCustomChangeProperties` to the before-each hook, as it's tied to `additionalChangeModelProperties` that's configured in this hook too.

test/change.test.js
```diff
@@ -610,71 +610,66 @@ describe('Change', function() {
...
+    it('stores the custom property in the Change instance', function() {
+      return change.rectify().then(function(ch) {
+        expect(ch.toObject()).to.have.property('tenantId', TENANT_ID);
       });
     });
 
-    it('creates a new change with the custom property', function() {
-      return change.rectify()
-      .then(function(ch) {
-        assert.equal(ch[customProperty], customPropertyValue);
```
bajtos: We prefer the [BDD style](http://chaijs.com/api/bdd/) assertions.

bajtos:
@slnode test please

bajtos:
@slnode test please

bajtos:
Landed 🎉  Thank you for your contribution! 👏

PR #2961, opened 2016-11-21
Optimize replication by sending multiple smaller requests to the server

kobaska:
### Description

Optimize replication to page and chunk requests to remote server. This allows for large amounts of data to be replicated and avoid network timeouts and payload being high. 

Add a new model-level setting "replicationChunkSize" which allows users to configure change replication algorithm to issue several smaller requests to fetch changes and upload updates.

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -15,6 +15,7 @@ var deprecated = require('depd')('loopback');
 var debug = require('debug')('loopback:persisted-model');
 var PassThrough = require('stream').PassThrough;
 var utils = require('./utils');
+var REPLICATION_CHUNK_SIZE = -1;
 
 module.exports = function(registry) {
   var Model = registry.getModel('Model');
@@ -1192,6 +1193,11 @@ module.exports = function(registry) {
     var Change = sourceModel.getChangeModel();
     var TargetChange = targetModel.getChangeModel();
     var changeTrackingEnabled = Change && TargetChange;
+    var replicationChunkSize = REPLICATION_CHUNK_SIZE;
+
+    if (sourceModel.settings && sourceModel.settings.replicationChunkSize) {
+      replicationChunkSize = sourceModel.settings.replicationChunkSize;
+    }
 
     assert(
       changeTrackingEnabled,
@@ -1211,7 +1217,13 @@ module.exports = function(registry) {
     async.waterfall(tasks, done);
 
     function getSourceChanges(cb) {
-      sourceModel.changes(since.source, options.filter, debug.enabled ? log : cb);
+      utils.downloadInChunks(
+        options.filter,
+        replicationChunkSize,
+        function(filter, pagingCallback) {
+          sourceModel.changes(since.source, filter, pagingCallback);
+        },
+        debug.enabled ? log : cb);
 
       function log(err, result) {
         if (err) return cb(err);
@@ -1222,7 +1234,13 @@ module.exports = function(registry) {
     }
 
     function getDiffFromTarget(sourceChanges, cb) {
-      targetModel.diff(since.target, sourceChanges, debug.enabled ? log : cb);
+      utils.uploadInChunks(
+        sourceChanges,
+        replicationChunkSize,
+        function(smallArray, chunkCallback) {
+          return targetModel.diff(since.target, smallArray, chunkCallback);
+        },
+        debug.enabled ? log : cb);
 
       function log(err, result) {
         if (err) return cb(err);
@@ -1241,9 +1259,16 @@ module.exports = function(registry) {
     function createSourceUpdates(_diff, cb) {
       diff = _diff;
       diff.conflicts = diff.conflicts || [];
+
       if (diff && diff.deltas && diff.deltas.length) {
         debug('\tbuilding a list of updates');
-        sourceModel.createUpdates(diff.deltas, cb);
+        utils.uploadInChunks(
+          diff.deltas,
+          replicationChunkSize,
+          function(smallArray, chunkCallback) {
+            return sourceModel.createUpdates(smallArray, chunkCallback);
+          },
+          cb);
       } else {
         // nothing to replicate
         done();
@@ -1253,20 +1278,29 @@ module.exports = function(registry) {
     function bulkUpdate(_updates, cb) {
       debug('\tstarting bulk update');
       updates = _updates;
-      targetModel.bulkUpdate(updates, options, function(err) {
-        var conflicts = err && err.details && err.details.conflicts;
-        if (conflicts && err.statusCode == 409) {
-          diff.conflicts = conflicts;
-          // filter out updates that were not applied
-          updates = updates.filter(function(u) {
-            return conflicts
-              .filter(function(d) { return d.modelId === u.change.modelId; })
-              .length === 0;
+      utils.uploadInChunks(
+        updates,
+        replicationChunkSize,
+        function(smallArray, chunkCallback) {
+          return targetModel.bulkUpdate(smallArray, options, function(err) {
+            // bulk update is a special case where we want to process all chunks and aggregate all errors
+            chunkCallback(null, err);
           });
-          return cb();
-        }
-        cb(err);
-      });
+        },
+        function(notUsed, err) {
+          var conflicts = err && err.details && err.details.conflicts;
+          if (conflicts && err.statusCode == 409) {
+            diff.conflicts = conflicts;
+            // filter out updates that were not applied
+            updates = updates.filter(function(u) {
+              return conflicts
+                .filter(function(d) { return d.modelId === u.change.modelId; })
+                .length === 0;
+            });
+            return cb();
+          }
+          cb(err);
+        });
     }
 
     function checkpoints() {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -4,8 +4,14 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
+
 exports.createPromiseCallback = createPromiseCallback;
+exports.uploadInChunks = uploadInChunks;
+exports.downloadInChunks = downloadInChunks;
+exports.concatResults = concatResults;
+
 var Promise = require('bluebird');
+var async = require('async');
 
 function createPromiseCallback() {
   var cb;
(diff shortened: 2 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

test/replication.test.js
```diff
@@ -19,1522 +19,1545 @@ describe('Replication / Change APIs', function() {
...
-    useSinceFilter = false;
-    var test = this;
-    dataSource = this.dataSource = loopback.createDataSource({
-      connector: loopback.Memory
-    });
-    SourceModel = this.SourceModel = PersistedModel.extend(
-      'SourceModel-' + tid,
-      { id: { id: true, type: String, defaultFn: 'guid' } },
-      { trackChanges: true });
+  describe('Replication without chunking', function() {
```
bajtos: This is adding too much whitespace changes, can we find a way how to preserve existing tests without changes and only add tests for the new "chunk" mode?

bajtos:
@slnode ok to test

test/util/utils.test.js
```diff
@@ -0,0 +1,110 @@
...
+  describe('uploadInChunks', function() {
+    it('should call process function with the chunked arrays', function() {
+      var largeArray = ['item1', 'item2', 'item3'];
+      var calls = [];
+
+      utils.uploadInChunks(largeArray, 1, processFunction, function() {
+        assert.deepEqual(calls, [['item1'], ['item2'], ['item3']]);
+      });
+
+      function processFunction(array, cb) {
```
bajtos: The usual convention for structuring unit tests is arrange-act-assert. I find your tests difficult to follow, because the `processFunction` ("act" phase) is defined after the assertions.

Could you please re-arrange all tests in this file to follow arrange-act-assert? For example:

```js
// option A
    it('should call process function with the chunked arrays', function() {
      var largeArray = ['item1', 'item2', 'item3'];
      var calls = [];

      function processFunction(array, cb) {
        calls.push(array);
        cb();
      }

      utils.uploadInChunks(largeArray, 1, processFunction, function() {
        assert.deepEqual(calls, [['item1'], ['item2'], ['item3']]);
      });
    });

// or even
    it('should call process function with the chunked arrays', function() {
      var largeArray = ['item1', 'item2', 'item3'];
      var calls = [];

      utils.uploadInChunks(
        largeArray, 1, 
        function processFunction(smallArray, cb) { 
          calls.push(smallArray);
          cb(); 
        },
        function onDone() {
          assert.deepEqual(calls, [['item1'], ['item2'], ['item3']]);
      });
    });
```

You should also check that `uploadInChunks` did not fail with an error. Also many (if not all) callback-based functions don't call the callback in the same tick of the event loop, therefore you should make all tests async:

```js
    it('should call process function with the chunked arrays', function(done) {
      var largeArray = ['item1', 'item2', 'item3'];
      var calls = [];

      utils.uploadInChunks(
        largeArray, 1, 
        function processFunction(smallArray, cb) { 
          calls.push(smallArray);
          cb(); 
        },
        function finished(err) {
          if (err) return done(err);
          assert.deepEqual(calls, [['item1'], ['item2'], ['item3']]);
          done();
      });
    });
```

The same comment applies to other new tests in this file too.

kobaska:
@bajtos I have made the changes you requested for

lib/persisted-model.js
```diff
@@ -1155,6 +1156,11 @@ module.exports = function(registry) {
     var Change = sourceModel.getChangeModel();
     var TargetChange = targetModel.getChangeModel();
     var changeTrackingEnabled = Change && TargetChange;
+    var chunkSize = CHUNK_SIZE;
+
+    if (this.settings && this.settings.chunkSize) {
```
bajtos: I believe `this` is `undefined` in this function. I think you should be using `sourceModel.chunkSize` instead?

It would be great to have a unit-test verifying this setting - if there was such test, then the problem with `this` would have been discovered long time ago.

I also find the name `chunkSize` as not descriptive enough. This setting will be usually configured in model JSON file, where it will "sit" among other general settings and it won't be obvious that `chunkSize` is related to change replication.

I am proposing `replicationChunkSize`, but feel free to come up with a different name that makes it clear that the chunk size refers to change replication/sync.

lib/persisted-model.js
```diff
@@ -1176,7 +1182,9 @@ module.exports = function(registry) {
     async.waterfall(tasks, done);
 
     function getSourceChanges(cb) {
-      sourceModel.changes(since.source, options.filter, debug.enabled ? log : cb);
+      utils.downloadInChunks(options.filter, chunkSize, function(filter, pagingCallback) {
```
bajtos: When function arguments don't fit on a single line, then we are applying "one arg per line" rule, see http://loopback.io/doc/en/contrib/style-guide.html#one-argument-per-line

```js
utils.downloadInChunks(
  options.filter, chunkSize, 
  function(filter, pagingCallback) {
    sourceModel.changes(since.source, filter, pagingCallback);
  },
  debug.enabled ? log : cb);
```  

This applies to all other `utils.*InChunks` calls below too.

lib/utils.js
```diff
@@ -31,3 +36,105 @@ function throwPromiseNotDefined() {
...
+          }
+
+          chunkCallback(err, previousResults);
+        });
+      };
+    });
+
+    async.waterfall(tasks, cb);
+  } else {
+    processFunction.call(self, largeArray, cb);
```
bajtos: Short "else" blocks after a long "if" block make it difficult to build a full picture. Please use reverse-logic with early-return instead:

```js
if (largeArray.length <= chunkSize) {
  return processFunction.call(self, largeArray, cb);
}

// the rest of the code is indented one level less
```

Also IIRC, `fn.call(...)` method has performance overhead compared to regular function calls `fn(...)`. Also AFAICT, you are not using `this` in any of the callbacks. Therefore I would prefer to use `processFunction(largeArray, cb)`. Unless I am missing something?

lib/utils.js
```diff
@@ -31,3 +36,105 @@ function throwPromiseNotDefined() {
...
+  filter = filter ? JSON.parse(JSON.stringify(filter)) : {};
+  filter.skip = 0;
+  filter.limit = chunkSize;
+
+  processFunction(filter, pageAndConcatResults);
+
+  function pageAndConcatResults(err, pagedResults) {
+    if (err) {
+      cb(err);
+    } else {
```
bajtos:
```js
if (err) {
  return cb(err);
}

// the rest of the code is indented one level less
```

test/replication.test.js
```diff
@@ -1627,6 +1627,50 @@ describe('Replication / Change APIs', function() {
...
+        var options = {};
+
+        SourceModel.create([{name: 'foo'}, {name: 'bar'}], function(err) {
+          if (err) return done(err);
+
+          test.SourceModel.replicate(test.startingCheckpoint, test.TargetModel,
+            options, function(err, conflicts) {
+              if (err) return done(err);
+
+              assertTargetModelEqualsSourceModel(conflicts, test.SourceModel,
```
bajtos: This isn't really verifying that correct `chunkSize` was applied, is it?

I think you should add some sort of an observer to one of the models and check how many times a replication-related method was called.

Thoughts?

kobaska:
@bajtos, I'm going to be away for a month on holiday. So will do it after, unless someone can continue on from here.

bajtos:
@kobaska No worries, enjoy your vacation 🏖

kobaska:
@bajtos I have done the changes you suggested. Can you review this please?

kobaska:
@bajtos Is there any more work needed for this feature? The PR Builder failed, but the link doesn't seem to work

bajtos:
Thank you for the update, I'll take a closer look next week.

test/replication.test.js
```diff
@@ -1699,6 +1699,111 @@ describe('Replication / Change APIs', function() {
...
+      TargetChange.Checkpoint = loopback.Checkpoint.extend('TargetCheckpoint');
+      TargetChange.Checkpoint.attachTo(dataSource);
+
+      TargetModel.attachTo(dataSource);
+
+      test.startingCheckpoint = -1;
+    });
+
+    describe('Model.replicate(since, targetModel, options, callback)', function() {
+      it('Replicate data using the source model with chunking', function(done) {
```
bajtos: The test name should read like a sentence, where `it` stand for the subject of "describe".

```
it('calls bulkUpdate multiple times')
```

test/replication.test.js
bajtos: Oh, I see, few lines below you are asserting the number of calls the `bulkUpdate` method was called 👍

test/replication.test.js
```diff
@@ -1699,6 +1699,111 @@ describe('Replication / Change APIs', function() {
...
+      TargetChange.Checkpoint = loopback.Checkpoint.extend('TargetCheckpoint');
+      TargetChange.Checkpoint.attachTo(dataSource);
+
+      TargetModel.attachTo(dataSource);
+
+      test.startingCheckpoint = -1;
+    });
+
+    describe('Model.replicate(since, targetModel, options, callback)', function() {
+      it('Replicate data using the source model without chunking', function(done) {
```
bajtos:
```
it('calls bulkUpdate only once')
```

test/replication.test.js
```diff
@@ -1803,4 +1908,36 @@ describe('Replication / Change APIs', function() {
   function getIds(list) {
     return getPropValue(list, 'id');
   }
+
+  function assertTargetModelEqualsSourceModel(conflicts, sourceModel,
```
bajtos: AFAICT, this helper is defined on L289 too ([source](https://github.com/kobaska/loopback/blob/b1fdd53dbc433c7ecadfeef1cf31f3d6fe61f3e2/test/replication.test.js#L298)). What is the difference between these two variants of the helper? Is there any reason prevent us from having only a single shared copy of this function?

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+var utils = require('../lib/utils');
+var assert = require('assert');
+
+describe('Utils', function() {
+  describe('uploadInChunks', function() {
+    it('should call process function with the chunked arrays', function(done) {
```
bajtos: We don't use `should` in our test names, see http://loopback.io/doc/en/contrib/style-guide.html#test-naming

```js
it('calls process function for each chunk', function(done) {
  // ...
});
```

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
...
+        calls.push(array);
+        cb();
+      }, function finished(err) {
+        if (err) return done(err);
+        assert.deepEqual(calls, [['item1'], ['item2'], ['item3']]);
+        done();
+      });
+    });
+
+    it('should call process function once when array less than chunk size', function(done) {
```
bajtos:
```js
it('calls process function only once when array is smaller than chunk size')
```

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
...
+        if (largeArray[skip + i]) {
+          results.push(largeArray[skip + i]);
+        }
+      }
+
+      skip += chunkSize;
+      cb(null, results);
+    }
+
+    it('should call process function with the correct filters', function(done) {
```
bajtos:
```js
it('calls process function with the correct filter')
```

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
...
+      var expectedFilters = [{ skip: 0, limit: chunkSize}, { skip: chunkSize, limit: chunkSize}];
+      utils.downloadInChunks({}, chunkSize, processFunction, function finished(err) {
+        if (err) return done(err);
+        assert.deepEqual(calls, expectedFilters);
+        done();
+      });
+    });
+
+    it('should concat the results from each call to the process function', function(done) {
```
bajtos:
```js
it('concats the results of all calls of the process function')
```

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
...
+      utils.downloadInChunks({}, chunkSize, processFunction, function finished(err, results) {
+        if (err) return done(err);
+        assert.deepEqual(results, largeArray);
+        done();
+      });
+    });
+  });
+
+  describe('concatResults', function() {
+    it('should concat arrays', function() {
```
bajtos:
```js
it('concats regular arrays')
```

test/util/utils.test.js
```diff
@@ -0,0 +1,114 @@
...
+      var array1 = ['item1', 'item2'];
+      var array2 = ['item3', 'item4'];
+
+      var concatResults = utils.concatResults(array1, array2);
+      assert.deepEqual(concatResults, ['item1', 'item2', 'item3', 'item4']);
+    });
+
+    it('should concat objects with arrays', function() {
```
bajtos:
```js
it('concats objects containing arrays')
```

kobaska:
@bajtos Thanks for the review. I have improved the coding style. Please let me know if any more changes are required.

bajtos:
@kobaska the patch LGTM to me now, thanks!

Just now I noticed that you are targeting 2.x branch. I am afraid the 2.x release line is in LTS mode now and we are no longer landing new features (semver-minor change) there.

I am happy to land this to `master` though. Would you like to open a new pull request against `master` yourself? Should I do it myself (preserving your commit authorship of course)? Let me know what works best for you.

bajtos:
Ah, never mind, I see GitHub allows us to change the target branch of a pull request. I have rebased your patch against `master` and cleaned up the commit message while I was at rewriting git history.

Note that `git pull` will not work for you in your local cloned branch, you can run e.g. `git reset --hard origin/optimise-replication` instead.

bajtos:
Landed 🎉 , thank you for the contribution!

kobaska:
Thanks @bajtos for your help :) . Was hoping to get it in 2.x as that is the version we are using. I'll start moving to latest loopback version.

PR #2970, opened 2016-11-22
Re-enable "no-unused-expressions" rule for eslint

bajtos:
As a follow-up to #2632, I am cleaning up the use of property-based assertions (see [Beware of libraries that assert on property access](https://github.com/moll/js-must#asserting-on-property-access)). The patch adds "dirty-chai" to modify chai's property checkers.

@superkhau please review
cc @loay


superkhau:
@slnode test please

PR #2972, opened 2016-11-22
Remove `example/context`

Amir-61:
Connected to https://github.com/strongloop-internal/scrum-epitome/issues/7

* Remove `example/context` since it is deprecated in LB 3.0

/to: @strongloop/squad-epitome @superkhau @bajtos PTAL.


bajtos:
Yes please, let's remove it. 

@Amir-61 could you please move the example code to https://github.com/strongloop/loopback-context?

PR #2973, opened 2016-11-22
Fix remote method example

Amir-61:
* Fix remote method example:
   - callback error and not ignore cartId

/to: @strongloop/squad-epitome @superkhau


Amir-61:
@slnode test please

Issue #292, opened 2014-05-29
Shortcut for changing property default value


bajtos:
When defining a model code-first, one can specify a dynamic default value for properties:

``` js
registrationDate: {type: Date, default: function () { return new Date() }},
```

This is not possible to achieve when using `app.boot()` and `models.json`. This is the currently recommended solution:

``` js
app.models.Todo.definition.rawProperties.created.default =
app.models.Todo.definition.properties.created.default = function() {
  return new Date();
};
```

@raymondfeng suggested there should be a helper method making this easier.

Related: https://groups.google.com/forum/#!topic/loopbackjs/tIlPMxVZXwE

seriousben:
The following code is not a workaround since it is executed everytime causing the created date to change. It is a workaround for the updatedAt timestamp.
```
app.models.Todo.definition.rawProperties.created.default =
app.models.Todo.definition.properties.created.default = function() {
  return new Date();
};
```

bajtos:
> The following code is not a workaround since it is executed everytime causing the created date to change.

Could you please provide more details, ideally a small example that we can run to reproduce the problem?

inawlaljar:
Hi, Did we reach on any conclusion (and also implemented) a way to set default value for a model? 

I want to set created and lastupdated fields. 

I am currently using hooks. Let me if that's the right way to set default value.

Ta.

bajtos:
The implementation of `$now` as a `Date` default value has landed in juggler's master: https://github.com/strongloop/loopback-datasource-juggler/pull/445

> @inawlaljar I want to set created and lastupdated fields.

You can use `$now` for the `lastupdated` field once the new juggler version is released. `created` field is tricky to get right, because `createOrUpdate` does not know in advance whether the `created` column should or should not be updated.

bajtos:
> You can use `$now` for the `lastupdated` field once the new juggler version is released. `created` field is tricky to get right, because `createOrUpdate` does not know in advance whether the created column should or should not be updated.

My last comment was rather incorrect, let me correct.

There is no out-of-the-box way how to implement `lastupdated` model property. The right approach is to use a `before save` hook, indeed. 

The `created` property can be implemented with `default: '$now'` and it will work as long as the calls of `updateOrCreate` are always passing the full object data retrieved from the database in the "update" case.

```js
// this works and preserves the value of the "created" property
Product.findById(existingId, function(err, prod) {
  prod.name = 'renamed';
  Product.updateOrCreate(prod);
});

[...]

PR #2974, opened 2016-11-23
Remove workaround for default value

loay:
### Description
Remove the work around for issue 292 concerning the default value, as it is not needed anymore since the json file is configured. 
The same fix is implemented for both `application` and `role`. 

#### Related issues

https://github.com/strongloop/loopback/issues/292

### Checklist


loay:
@bajtos PTAL. Thanks.

Issue #2420, opened 2016-06-10
Role.getRoles inconsistent entities return


mitsos1os:
I believe Role.getRoles functionality isn't working as expected. When issued for a user, it will return the names of dynamically resolved roles such as $authenticated, $everyone, etc.. and then for normal roles, the role id. **So you get an array of names and Ids together**. I do not see any reason for this and I find it incorrect.
- You can see that for resolved roles it adds the name [here](https://github.com/strongloop/loopback/blob/master/common/models/role.js#L404)
- But for normal roles it adds the id... [here](https://github.com/strongloop/loopback/blob/master/common/models/role.js#L439)

Any opinions???

davidcheung:
hi @mitsos1os, it would be really helpful if you could reproduce the issue(where the `role name` and `roleId` ends up in same array) in a [sand-box](https://github.com/strongloop/loopback/wiki/Reporting-issues#bug-report) 

please note that in [principalId](https://github.com/strongloop/loopback/blob/42c83f69a090664bb025d89f8dd5f5a7a3ffa34e/common/models/role-mapping.json#L5) are  `strings`
```js
"acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$owner",
      "permission": "ALLOW"
    }
  ]
```

mitsos1os:
Hello @davidcheung !
I have created the sandbox here: https://github.com/mitsos1os/loopback-sandbox/tree/get-roles-return-mixed
You can see for yourself that in the end of the boot scrip for `server/boot/admin.js` the roles of an admin user are printed, and you will see that dynamic roles (simple names), are mixed with static role ids....
In my first comment you can see the lines responsible for this, but I haven't issued a pull request because I wanted to discuss first this behavior....
Thanks

davidcheung:
Thanks for reproducing the issue, I do agree the behavior is unexpected, not sure whats the best approach to resolve this, perhaps changing
https://github.com/strongloop/loopback/blob/42c83f69a090664bb025d89f8dd5f5a7a3ffa34e/common/models/role.json#L8

```json
{
  "name": "Role",
  "properties": {

    "id": {
      "type": "string",
      "id": true,
      "generated": true 
    },
```
**right now this forces all ids into numbers**
to something like 
```json
{
  "name": "Role",
  "idInjection": true,
  "properties": {
    "id": {
      "type": "string",
      "id": true
    },
}
```

then user would be able to do 
```js
Role.create({
[...]

PR #2975, opened 2016-11-23
Role.getRoles() to return only role names

ebarault:
### Description

Currently the return type of `Role.getRoles()` method is inconsistent: role names are returned for  smart roles and role ids are returned for static roles (configured through user-role mapping)

This PR proposes to return only role names in `Role.getRoles()`
Tests have been updated.

#### Related issues

- None

### Checklist

@bajtos, @superkhau could you please have a look?

### DIFF
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -254,6 +254,20 @@ describe('role model', function() {
             },
             function(next) {
               Role.getRoles(
+                {principalType: RoleMapping.USER, principalId: user.id},
+                {returnOnlyRoleNames: true},
+                function(err, roles) {
+                  if (err) return next(err);
+                  expect(roles).to.eql([
+                    Role.AUTHENTICATED,
+                    Role.EVERYONE,
+                    role.name,
+                  ]);
+                  next();
+                });
+            },
+            function(next) {
+              Role.getRoles(
                 {principalType: RoleMapping.APP, principalId: user.id},
                 function(err, roles) {
                   if (err) return next(err);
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
@ebarault thank you for the pull request. Your proposal makes sense to me, however I am concerned about backwards compatibility. Can we implement your enhancement behind a flag, so that users can decide whether to use current behaviour (by default) or get role names only (when the flag is set)?

One options is to add a new optional parameter `options` to `Role.getRoles` method:

```js
Roles.getRoles(ctx, cb); // works as before
Roles.getRoles(ctx, { returnRoleNames: true }, cb); // the new way
```

ebarault:
@bajtos : makes sense. Renamed the option 'returnOnlyRoleNames'. Rollbacked original test, and added a new one with the option. please have a look

davidcheung:
related https://github.com/strongloop/loopback/issues/2420

test/role.test.js
```diff
@@ -254,6 +254,20 @@ describe('role model', function() {
             },
             function(next) {
               Role.getRoles(
+                {principalType: RoleMapping.USER, principalId: user.id},
+                {'returnOnlyRoleNames': true},
```
bajtos: I think the quotes are not needed, can you remove please?

bajtos:
Actually, let me make those changes for you, to get the patch landed sooner.

ebarault:
thanks for finalizing @bajtos

bajtos:

PR #2990, opened 2016-11-29
Update Downstream ignore List

loay:
Add loopback-connector-swagger to downstream ignore list. 
All servers pass except this one. The cause is not related to a fault in the unit tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,5 +115,10 @@
       "debug": "*,-mocha:*,-eslint:*"
     }
   },
-  "license": "MIT"
+  "license": "MIT",
+  "ci": {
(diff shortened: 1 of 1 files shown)


loay:
@strongloop/loopback-dev PTAL. Thanks.

rmg:
@loay how about strongloop/loopback-connector-swagger#19 instead?

rmg:
@slnode test please

superkhau:
@loay Rejecting this PR as swagger connector should be fixed instead. We should only use the flag sparingly for legitimate reasons.

PR #2980, opened 2016-11-23
Allow password reset request for users in realms

beeman:
### Description

Currently the `User.resetPassword` method only accepts an email address as a parameter. 

When you have a LoopBack application that partitions the users in Realms you need to `find` the user based on it's email address + realm, because it allows for 2 users with the same email address.

This PR fixes the issue by adding support for an optional `realm` property, that will enhance the filter used in the `findOne` method.

@bajtos @raymondfeng @superkhau could you please review?   

#### Related issues

- #2866 

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #3000, opened 2016-12-05
Opt-out downstream builds that are unstable

davidcheung:
### Description
opting out of https://cis-jenkins.swg-devops.com/job/ds/job/dashboard-controller~master

#### Related issues
#2980 

repos that are opting out are not a good indicator of stability of
this module, and are failing

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,11 @@
   },
   "config": {
     "ci": {
-      "debug": "*,-mocha:*,-eslint:*"
+      "debug": "*,-mocha:*,-eslint:*",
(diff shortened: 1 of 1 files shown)


superkhau:
Can you add `gateway-director-management-interface` and land? This PR won't pass, but please send another PR to test the ignore after this lands.

rmg:
This PR is incorrect. The `downstreamIgnoreList` is supposed to be under a root level `ci` property, not nested under `config`.

davidcheung:
thanks @rmg , will fix it

PR #3002, opened 2016-12-06
Upgrade eslint config and grunt-eslint to latest

bajtos:
Also fix linter errors reported after the upgrade and disable ES6 because PhantomJS does not support it yet (see #3003).


bajtos:
See also: https://github.com/strongloop/loopback/issues/3003

PR #1831, opened 2015-11-23
Remove options.template to prevent rejection by certain nodemailer transport plugins

sharkinsspatial:
Removes options.template after options.html is generated to prevent
rejection by certain nodemailer transport plugins.

PR #3004, opened 2016-12-06
Remove "options.template" from Email payload

bajtos:
Fix User.confirm to exclude "options.template" when sending the confirmation email. Certain nodemailer transport plugins are rejecting such requests.

This patch supersedes abandoned #1831.

@loay or @superkhau  please review
cc @sharkinsspatial


ritch:
LGTM

PR #3014, opened 2016-12-09
Fix registration of operation hooks in User model

bajtos:
Operation hooks are inherited by subclassed models, therefore they must be registered outside of `Model.setup()` function.

This commit fixes this problem in the built-in User model.

There are not tests verifying this change, as writing a test would be
too cumbersome and not worth the cost IMO.

Related: #2665 

cc @loay

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -700,49 +700,6 @@ module.exports = function(User) {
       next();
     });
 
-    // Delete old sessions once email is updated
-    UserModel.observe('before save', function beforeEmailUpdate(ctx, next) {
-      if (ctx.isNewInstance) return next();
-      if (!ctx.where && !ctx.instance) return next();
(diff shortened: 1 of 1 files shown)


bajtos:
@slnode test please

bajtos:
https://cis-jenkins.swg-devops.com/job/nb/job/loopback~master/label=x64%20&&%20linux%20&&%20nvm,nodeVersion=4/634/console
```
curl: (56) GnuTLS recv error (-110): The TLS connection was non-properly terminated.
Binary download from https://nodejs.org/dist/v4.7.0/node-v4.7.0.tar.xz failed, trying source.
grep: /home/jenkins/.nvm/.cache/src/node-v4.7.0/node-v4.7.0.tar.xz: No such file or directory
```

@rmg should we perhaps setup a local mirror to prevent CI failures like the one above?

PR #3018, opened 2016-12-09
Invalidate AccessTokens on password change

bajtos:
Invalidate all existing sessions (delete all access tokens) after user's password was changed.

This patch supersedes #2665

Connect to strongloop-internal/scrum-loopback#925

@loay @raymondfeng PTAL

PR #3021, opened 2016-12-12
Invalidate AccessTokens on password change

bajtos:
Invalidate all existing sessions (delete all access tokens) after user's password was changed.

This is a back-port of #3018

Connect to strongloop-internal/scrum-loopback#925

cc @loay

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -645,6 +645,19 @@ module.exports = function(User) {
     err.statusCode = 422;
     throw err;
   };
+
+  User._invalidateAccessTokensOfUsers = function(userIds, cb) {
+    if (!Array.isArray(userIds) || !userIds.length)
+      return process.nextTick(cb);
+
+    var accessTokenRelation = this.relations.accessTokens;
+    if (!accessTokenRelation)
+      return process.nextTick(cb);
+
+    var AccessToken = accessTokenRelation.modelTo;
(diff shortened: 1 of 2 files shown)


pelim:
this change results in invalidating the tokens on User.save() - please revert this change v2.36.0 works finde for me

bajtos:
@pelim could you please open a new issue with steps for reproducing the problem? See also https://github.com/strongloop/loopback/issues/3034 and https://github.com/strongloop/loopback/issues/3068

bajtos:
@pelim is your problem perhaps related to https://github.com/strongloop/loopback/pull/3053?

icy90:
@bajtos I have experienced the same issue, calling User.save() remove all current access tokens even not change user email or password.

in common/models/user.js: ctx.instance always true when you do User.save()
`  // Full replace of User instance => assume password change.
    // HashPassword returns a different value for each invocation,
    // therefore we cannot tell whether ctx.instance.password is the same
    // or not.
    **var isFullReplaceChangingPassword = !!ctx.instance;**

    ctx.hookState.isPasswordChange = isPartialUpdateChangingPassword ||
      isFullReplaceChangingPassword;`

bajtos:
@icy90 I see. Please open a new github issue for this problem, so that we can move the discussion there, and mention my handle there so that I get email notification.

PR #3035, opened 2016-12-16
Allow hash path in redirectUrl

### LINKED ISSUE #2307: Redirect url should be urlencoded

ambrt:
### Description
Allows GET variables and hash paths for redirecting user after email confirm.

#### Related issues

- #2307

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -427,7 +427,7 @@ module.exports = function(User) {
       '?uid=' +
       options.user.id +
       '&redirect=' +
-      options.redirect;
+      encodeURIComponent(options.redirect);
 
     options.templateFn = options.templateFn || createVerificationEmailBody;
 
@@ -814,7 +814,7 @@ module.exports = function(User) {
         if (!ctx.res) {
           return next(new Error(g.f('The transport does not support HTTP redirects.')));
         }
-        ctx.res.location(ctx.args.redirect);
+        ctx.res.location(decodeURIComponent(ctx.args.redirect));
         ctx.res.status(302);
       }
       next();


slnode:

slnode:

slnode:

slnode:

common/models/user.js
```diff
@@ -814,7 +814,7 @@ module.exports = function(User) {
         if (!ctx.res) {
           return next(new Error(g.f('The transport does not support HTTP redirects.')));
         }
-        ctx.res.location(ctx.args.redirect);
+        ctx.res.location(decodeURIComponent(ctx.args.redirect));
```
bajtos: I believe `ctx.args.redirect` should have been already url-decoded by strong-remoting, therefore this change is not necessary. Unless I am missing something?

ambrt: If its decoded then this change is not necessary.

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/3185.

PR #3044, opened 2016-12-20
Prepare for 3.x GA

superkhau:
cc @bajtos

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ LoopBack tools include:
 
 For more details, see [http://loopback.io/](http://loopback.io/).
 
+## Supported versions
+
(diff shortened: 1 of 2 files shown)


ritch:
To actually publish this I think we need to change to version `3.1.1`

superkhau:
@ritch PTAL again.

bajtos:
@ritch
> To actually publish this I think we need to change to version `3.1.1`

We use `slt-release` for releasing to npmjs.org. This tool updates the version in package.json, adds new entries to `CHANGES.md`, creates a `git` tag and finally published to npm. Pull requests should not be updating the version number in package.json.

PR #2960, opened 2016-11-21
Avoid periodic cleanup/rectification of changes

kobaska:
### Description

Add a value for which cleanup would not run. Running cleanup on a big database is make the machine run out of memory.

#### Related issues

- None

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #3047, opened 2016-12-21
Add option disabling periodic change rectification

bajtos:
When `Model.settings.changeCleanupInterval` is set to a negative value,
no periodic cleanup is performed at all.

This is a forward-port of #2960.


bajtos:
@slnode test please

Issue #3042, opened 2016-12-20
Discussion: Developer/Strongloop communication


doublemarked:
At the advice of @bajtos in #2693 I am opening this discussion regarding developer communication.

Context: Fixes #2693 and #3021 implemented access-token invalidation when the users changes an email or a password. However, the current code deletes all access tokens, including the token used to make the change password/email request. The effect of this is that client code must anticipate immediate token invalidation any time it submits changes to a user's email and password, which is notable behavior change to be slipped into a point release.

I'd like to propose several questions associated with this. These questions are fielded to both the Strongloop team and the developer community,

1. Under what conditions do we consider it acceptable to introduce breaking changes into a minor release?
2. What is the best way in which the Strongloop team can communicate to the developer community when such changes are upcoming?

### Tasks

 - [x] Setup a low-volume Announcement mailing list - see https://groups.google.com/forum/#!forum/loopbackjs-announcements
 - [x] Update loopback.io documentation - https://github.com/strongloop/loopback.io/pull/316
 - [x] Update [README](https://github.com/strongloop/loopback#resources) in this repository - see https://github.com/strongloop/loopback/pull/3261

doublemarked:
I'd like to share my own thoughts on this. I may have missed it, but I don't recall any communication regarding these changes, which I think took place in mid-September. Our team did not discover the problem until last week when we performed some cleanup that bumped up the Loopback version. 

I would like to offer the following suggestions, 

  1. Perhaps Strongloop should consider a low-volume announcement-only Google Group that is used for important developer alerts/notifications (security notices, breaking changes, notable releases, etc). The general Google Group has a lot of chatter and I would not be surprised if many developers do not subscribe or ignore it. 

  2. Similarly, perhaps the Github release notes for 2.35.0 should have emphasized the breaking changes. 
 
  3. Might it have been been more appropriate to include these changes only with configuration that allows it to be disabled, even if it defaults to on?

superkhau:
@doublemarked Thanks for bringing up this discussion. I've been advocating a Community Governance page on loopback.io for things like this (and transparency especially). @bajtos Has something in the backlog (probably won't get to it till the new year though since most people are going on vacation soon). 

IMO:

>Under what conditions do we consider it acceptable to introduce breaking changes into a minor release?
Never since we should be following semver and only breaking on majors.

>What is the best way in which the Strongloop team can communicate to the developer community when such changes are upcoming?
[...]

Issue #3034, opened 2016-12-16
Email/password change should not invalidate current session


bajtos:
In #2693 and #3021, we implemented access-token invalidation when the users changes an email or a password. However, the current code deletes *all* access tokens, including the token used to make the change password/email request.

IMO, this is a poor user experience, the access token (session) used to make the change should be preserved.

One of the reasons why this was not made in the original pull requests, is that we don't have a good solution for getting the current access token. The easiest way forward is to wait for https://github.com/strongloop/loopback/pull/3023 to land, after which we can use `ctx.options.accessToken` to exclude the current access token from the `delete` query.

bajtos:
Done.

PR #3048, opened 2016-12-22
Inject remoting context to options arg [2.x]

bajtos:
Backport of #3023

Connect to #1495

cc @fabien @PaddyMann @ebarault

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -428,8 +428,38 @@ module.exports = function(registry) {
     if (options.isStatic === undefined) {
       options.isStatic = true;
     }
+
+    if (options.accepts) {
+      options = extend({}, options);
+      options.accepts = setupOptionsArgs(options.accepts);
+    }
+
     this.sharedClass.defineMethod(name, options);
   };
+
+  function setupOptionsArgs(accepts) {
+    if (!Array.isArray(accepts))
+      accepts = [accepts];
+
+    return accepts.map(function(arg) {
+      if (arg.http && arg.http === 'optionsFromRequest') {
+        // deep clone to preserve the input value
+        arg = extend({}, arg);
+        arg.http = createOptionsViaModelMethod;
+      }
+      return arg;
+    });
+  }
+
+  function createOptionsViaModelMethod(ctx) {
+    var EMPTY_OPTIONS = {};
+    var ModelCtor = ctx.method && ctx.method.ctor;
+    if (!ModelCtor)
+      return EMPTY_OPTIONS;
+    if (typeof ModelCtor.createOptionsFromRemotingContext !== 'function')
+      return EMPTY_OPTIONS;
+    return ModelCtor.createOptionsFromRemotingContext(ctx);
+  }
 
   /**
    * Disable remote invocation for the method with the given name.
@@ -868,6 +898,46 @@ module.exports = function(registry) {
   };
 
   Model.ValidationError = require('loopback-datasource-juggler').ValidationError;
+
+  /**
+   * Create "options" value to use when invoking model methods
+   * via strong-remoting (e.g. REST).
+   *
+   * Example
+   *
+   * ```js
+   * MyModel.myMethod = function(options, cb) {
+   *   // by default, options contains only one property "accessToken"
+   *   var accessToken = options && options.accessToken;
+   *   var userId = accessToken && accessToken.userId;
+   *   var message = 'Hello ' + (userId ? 'user #' + userId : 'anonymous');
+   *   cb(null, message);
+   * });
+   *
+   * MyModel.remoteMethod('myMethod', {
+   *   accepts: {
+   *     arg: 'options',
+   *     type: 'object',
+   *     // "optionsFromRequest" is a loopback-specific HTTP mapping that
+   *     // calls Model's createOptionsFromRemotingContext
+   *     // to build the argument value
+   *     http: 'optionsFromRequest'
+   *    },
+   *    returns: {
+   *      arg: 'message',
+   *      type: 'string'
+   *    }
+   * });
+   * ```
+   *
+   * @param {Object} ctx A strong-remoting Context instance
+   * @returns {Object} The value to pass to "options" argument.
+   */
+  Model.createOptionsFromRemotingContext = function(ctx) {
+    return {
+      accessToken: ctx.req.accessToken,
+    };
+  };
 
   // setup the initial model
   Model.setup();
--- a/lib/model.js
+++ b/lib/model.js
@@ -10,6 +10,7 @@
 var g = require('strong-globalize')();
 
 var assert = require('assert');
+var debug = require('debug')('loopback:model');
 var RemoteObjects = require('strong-remoting');
 var SharedClass = require('strong-remoting').SharedClass;
 var extend = require('util')._extend;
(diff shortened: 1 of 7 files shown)


bajtos:
Quoting from https://github.com/strongloop/loopback/issues/1495#issuecomment-268782354
> When this update is published, is there a possibility it will break 2.x projects using the workaround that already uses options?

and https://github.com/strongloop/loopback/issues/1495#issuecomment-268784719

> The main reason for back-porting this new context propagation to 2.x is to allow the recently added access-token invalidation skip the access token of the current user making the request, which I consider as an important bug to be fixed (see #3034).
> I am happy to discuss how to minimise the impact of this back-ported change on existing LoopBack 2.x applications.

In the light of the discussion we have with @doublemarked in https://github.com/strongloop/loopback/issues/3042, it may be best to introduce a feature flag enabling/disabling the new context-propagation feature, so that we don't break existing applications using a workaround solution that was described in #1495.

On the other hand, such flag would complicate the fix for #3034, as the fix as I am envisioning it right now will rely on `options` object being set by remote calls. I suppose we could add another feature flag to disable access-token invalidation, but adding so many flags feels like over engineering to me.

To be honest, I personally consider manipulation of remote method arguments in application code as an unsupported usage of the framework. If we allowed applications to arbitrarily manipulate remoting metadata, then any change of remoting metadata in built-in LoopBack models would become a semver-major change, because we won't be able to tell if there is any app out there relying on the exact remoting metadata before our change.

Therefore I am personally inclining to land this pull request as it is now, and let users fix their context-propagation workarounds to work correctly with the new remoting metadata (ideally by using the new feature we are adding by this patch).

@tvdstaaij thoughts?

cc @ritch @raymondfeng @superkhau

fabien:
@bajtos can we please have a flag to enable/disable context propagation? For now, I'm relying on my own workaround, until I have a chance to use your implementation.

bajtos:
I'm leaving for vacation today, let's put this pull request on hold and revisit the discussion in January when I get back.

fabien:
@bajtos sure - have a great vacation and nice holidays!

tvdstaaij:
Agreed on the flag(s). I'll try to address your doubts below.

> I suppose we could add another feature flag to disable access-token invalidation, but adding so many flags feels like over engineering to me.

In general it is, but I assume this would be only for 2.0. Seeing how it is unlikely that new applications will be developed with 2.0, I don't think this is much of a problem (from a user perspective at least; I don't know how much it would complicate the framework code).

It's also good to realize that existing 2.0 apps were built and tested without builtin access token invalidation. It's not just some security patch or backwards compatible feature but rather something that changes the behavior of the framework. When developing my Loopback 2 application I noticed that there was no builtin token invalidation logic, so I defined custom password change and token invalidation logic using operation hooks (relying on the aforementioned context workaround). I would assume other developers did something similar if token invalidation matters for their applications. There's no telling what would happen if a framework-defined implementation would suddenly be mashed onto an application code implementation.

> To be honest, I personally consider manipulation of remote method arguments in application code as an unsupported usage of the framework. If we allowed applications to arbitrarily manipulate remoting metadata, then any change of remoting metadata in built-in LoopBack models would become a semver-major change, because we won't be able to tell if there is any app out there relying on the exact remoting metadata before our change.

Sure, it's technically an unsupported framework usage (a.k.a. "hack"), but in this particular case there was no *supported* way to do what was necessary. As you are aware the remote context implementation at the time was horribly broken and the docs warned not to use it. Some apps already relying on the old implementation suddenly broke, and some newer apps simply needed remote context for advanced features. Back then I wasn't eager to use the unofficial workaround at all, but when reviewed my options it was really the only viable solution.

So technically it's unsupported framework usage, but realistically developers just had to rely on this particular piece of "private" behavior.

Again, because this is an LTS branch, I expect that the majority of applications have already adapted some kind of solution that works for that application, and are not in immediate need of either feature (new context propagation method and token invalidation). Therefore I believe that pushing a change like this without being optional potentially causes more harm than good.

If a developer decides to do some maintenance on their Loopback 2 application and wants to migrate to the new context propagation, flipping a configuration switch is trivial. But as long as only minor/patch is upgraded and no configuration is changed, the framework behavior is expected not to change.

bajtos:
@tvdstaaij thank you for a thoughtful comment!

> If a developer decides to do some maintenance on their Loopback 2 application and wants to migrate to the new context propagation, flipping a configuration switch is trivial. But as long as only minor/patch is upgraded and no configuration is changed, the framework behavior is expected not to change.

I agree this is an important point to keep in mind.

> I assume this would be only for 2.0. Seeing how it is unlikely that new applications will be developed with 2.0, I don't think this is much of a problem (from a user perspective at least; I don't know how much it would complicate the framework code).

FWIW, based on the emails and GitHub notifications I am seeing, it looks like there are still users building applications on LoopBack 2.x now.

> It's also good to realize that existing 2.0 apps were built and tested without builtin access token invalidation. It's not just some security patch or backwards compatible feature but rather something that changes the behavior of the framework. When developing my Loopback 2 application I noticed that there was no builtin token invalidation logic, so I defined custom password change and token invalidation logic using operation hooks (relying on the aforementioned context workaround). I would assume other developers did something similar if token invalidation matters for their applications. 

I personally think most people don't realise the importance of token invalidation, and what's even worse, they don't know what they don't know. As framework authors, we need a way how to let these users know that their applications may be vulnerable.

I am thinking about adding a feature flag for token invalidation, and printing a deprecation warning if this flag is not set (i.e. the value is `undefined`). That way the behaviour of applications does not change after upgrade, yet there is a warning for users that an action is needed on their side. If the users decide the issue is something they have already handled themselves, then they can set this new flag to `false` to get rid of the deprecation warning.

Thoughts?

bajtos:
I have added one more commit de34d55 that will be landed only to 2.x. In this commit, I am adding a new model-level setting `injectOptionsFromRemoteContext` that acts as a feature flag controlling whether built-in methods include "options" argument in their remoting metadata. The flag is disabled by default for backwards compatibility.

Usage:

```js
// common/models/customer.json
{
  "name": "Customer",
  "base": "User",
  "injectOptionsFromRemoteContext": true,
  "properties": {
    // ...
  }
}
```
  
@fabien @PaddyMann @ebarault @tvdstaaij @doublemarked thoughts?

@superkhau @ritch PTAL at the proposed code changes

bajtos:
One more commit to implement the new feature flag for `sharedCtor` too: 23bb38d

fabien:
> I am thinking about adding a feature flag for token invalidation, and printing a deprecation warning if this flag is not set (i.e. the value is undefined).

+1 Would be great!

bajtos:
> > I am thinking about adding a feature flag for token invalidation, and printing a deprecation warning if this flag is not set (i.e. the value is undefined).

> +1 Would be great!

I created a new issue to keep track of this work, see https://github.com/strongloop/loopback/issues/3068.

Pull requests are welcome!

bajtos:
> I am thinking about adding a feature flag for token invalidation, and printing a deprecation warning if this flag is not set (i.e. the value is undefined).


See https://github.com/strongloop/loopback/pull/3109

bitmage:
I'm trying to utilize this to implement 1) multi-tenancy, 2) caching.

I have to say it's pretty confusing at this point... I've spent the last 4 days combing through issue tracker threads and trying a variety of methods including `getCurrentContext()`, middleware, `app.remotes().phases.addBefore`, and `app.remotes().before('*.*')`.

I finally was able to inject standard tenancy parameters into the `options` arg in a consistent way using the last approach.  And then I went to combine this with my caching middleware, and found [that they run in the wrong order](https://github.com/strongloop/loopback/issues/3434).

I guess what I would want is just more consistency.  It seems like there should be 1 general middleware implementation, and everything else should fall out from that.  And the ability to have some kind of introspection/printout is a god send as well, so I can see the middleware stacks that got created for my endpoints.

I'm on Loopback 2.x at the moment, as I saw some stability issues with generators in 3.x last I tried.  Was middleware cleanup a major thrust of 3.x?  Would I benefit from an upgrade?  Is there any other cleanup/standardization/documentation effort in the works?

bajtos:
> And the ability to have some kind of introspection/printout is a god send as well, so I can see the middleware stacks that got created for my endpoints.

I can see how this can be useful! I don't think we have any such tool in LoopBack core, have you found any community-maintained one? Would you like to start one yourself?

> I'm on Loopback 2.x at the moment, as I saw some stability issues with generators in 3.x last I tried. Was middleware cleanup a major thrust of 3.x? Would I benefit from an upgrade? Is there any other cleanup/standardization/documentation effort in the works?

IIRC, there was no significant changes related to middleware in LoopBack 3.0. The release was focused mostly on cleanup tasks that break backwards compatibility, see https://loopback.io/doc/en/lb3/3.0-Release-Notes.html. The subsequent 3.x releases contain mostly security-related improvements.

PR #3052, opened 2016-12-23

yumenohosi:
Yesterday, the loopback we are using in our system was upgraded via npm,
and since the upgrade, we noticed that every time the user model updates,
the emailVerified column would change to false.

I took a look and realized there might be an error in
https://github.com/strongloop/loopback/commit/eb640d8

The intent of the commit just mention is to make emailVerified false
becomes false.

This commit fixes the issue just mentioned.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -860,7 +860,7 @@ module.exports = function(User) {
         }
       } else {
         var emailChanged = ctx.hookState.originalUserData.some(function(data) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

PR #3053, opened 2016-12-23

yumenohosi:
### Description

Yesterday, the loopback we are using in our system was upgraded
via npm, and since the upgrade, we noticed that every time
the user model updates, the emailVerified column would change to false.

I took a look and realized there might be an error in https://github.com/strongloop/loopback/commit/eb640d8

The intent of the commit just mention is to make emailVerified false
becomes false.

This commit fixes the issue just mentioned.

#### Related issues

- None

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -858,9 +858,9 @@ module.exports = function(User) {
         if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
           ctx.instance.emailVerified = false;
         }
-      } else {
+      } else if (ctx.data.email) {
         var emailChanged = ctx.hookState.originalUserData.some(function(data) {
-          return ctx.data.email && data.email != ctx.data.email;
+          return data.email != ctx.data.email;
         });
         if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
           ctx.data.emailVerified = false;
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -860,7 +860,7 @@ module.exports = function(User) {
         }
       } else {
         var emailChanged = ctx.hookState.originalUserData.some(function(data) {
-          return data.email != ctx.data.email;
+          return ctx.data.email && data.email != ctx.data.email;
         });
         if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
           ctx.data.emailVerified = false;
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

common/models/user.js
```diff
@@ -860,7 +860,7 @@ module.exports = function(User) {
         }
       } else {
         var emailChanged = ctx.hookState.originalUserData.some(function(data) {
-          return data.email != ctx.data.email;
+          return ctx.data.email && data.email != ctx.data.email;
```
flowersinthesand: It seems that with the falsy `ctx.data.email`, you don't even need to iterate over `ctx.hookState.originalUserData`

yumenohosi: @flowersinthesand Thanks!

davidcheung:
@slnode ok to test

davidcheung:
the patch LGTM, @bajtos @loay can you guys PTAL

bajtos:
@yumenohosi thank you for the pull request, and thank you @davidcheung for the first round of review. I have assigned the patch to myself and I'll take a look in next few days.

benbrown:
This also threw a major monkeywrench into our application and caused 2 days of lost progress. 

I think all of @loay's recent changes to the emailVerified process should be carefully reviewed. For example, it is NOT desirable to log users out of applications just because they change their email address.

bajtos:
@benbrown thank you for chiming in, we are sorry for the difficulties created :(

> For example, it is NOT desirable to log users out of applications just because they change their email address.

Logging out users is important for security reasons. Consider the case when somebody hacks your email that you used when registering with a LoopBack-powered app. After you find this, you log into the app and change the email to a different one that wasn't hacked. Without session (access token) invalidation, the attacker would remain logged into your account and you would have no way how to log them out.

Having said that, I agree that this makes the user experience less optimal. I am proposing to allow the end user to decide whether they want to log out other sessions or not, see https://github.com/strongloop/loopback/issues/3071

bajtos:
Landed 🎉 Thank you @yumenohosi for your contribution! 👏

bajtos:
For posterity: I did not realise there were two commits in this pull request and I forgot to squash them before landing. My fault :(

PR #3064, opened 2017-01-03
Inject remote context createChangeStream method

BramKleinhout:
### Description

Implement, in addition to #3023, the injection of remote context to the options argument of the createChangeStream method.

#### Related issues

- #3023 

### Checklist

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -993,15 +993,8 @@ module.exports = function(registry) {
         {verb: 'post', path: '/change-stream'},
         {verb: 'get', path: '/change-stream'},
       ],
-      accepts: {
-        arg: 'options',
-        type: 'object',
-      },
-      returns: {
-        arg: 'changes',
-        type: 'ReadableStream',
-        json: true,
-      },
+      accepts: {arg: 'options', type: 'object', http: 'optionsFromRequest'},
+      returns: {arg: 'changes', type: 'ReadableStream', json: true},
     });
   };
 
--- a/test/context-options.test.js
+++ b/test/context-options.test.js
@@ -114,6 +114,11 @@ describe('OptionsFromRemotingContext', function() {
       return request.get('/products/count').expect(200)
         .then(expectInjectedOptions);
     });
+
+    it('injects options to createChangeStream()', function() {
+      return request.get('/products/change-stream').expect(200)
+        .then(expectInjectedOptions);
+    });
   });
 
   context('when invoking prototype methods', function() {


slnode:

slnode:

slnode:

slnode:

slnode:

davidcheung:
@BramKleinhout thanks for the contribution!
would you be able to add a test to prevent later regressions?

BramKleinhout:
@davidcheung That is no problem, I have added the test for the injection.

lib/persisted-model.js
```diff
@@ -993,15 +993,8 @@ module.exports = function(registry) {
...
-      accepts: {
-        arg: 'options',
-        type: 'object',
-      },
-      returns: {
-        arg: 'changes',
-        type: 'ReadableStream',
-        json: true,
-      },
+      accepts: {arg: 'options', type: 'object', http: 'optionsFromRequest'},
```
bajtos: I did not realise that `createChangeStream` was already exposing the `options` argument. The change proposed in this pull request can break existing applications, because right now it's possible to specify options via query string:

```
GET /api/mymodels/change-stream?options[flag]=value
-> options = {flag: 'value'}
```

We need to preserve backwards compatibility. The solution is to add a feature flag (a model-level setting) that will control whether the `options` argument should be provided by the user (current state) or constructed via `optionsFromRequest` (the new version).

Here is a quick mock-up:

```js
var streamOptionsArg = {arg: 'options', type: 'object'};
// TODO: remove this flag in 4.0
if (PersistedModel.settings.injectChangeStreamOptions) {
  streamOptionsArg.http = 'optionsFromRequest';
}

setRemoting(PersistedModel, 'createChangeStream', {
  // ...,
  accepts: streamOptionsArg,
  returns: // ...
});
```

bajtos: Actually, on the second thought, considering that Operation hooks observers now expect that `ctx.options` is always constructed by the server and cannot be injected by the client, I think the original `options` configuration would create a serious security vulnerability and therefore we should change it even though it may break existing apps.

bajtos:
Hmm, it turns out the built-in `createChangeStream` method does not use the `options` argument at all, and there are no operation hooks invoked where this `options` argument could be read.

In that light, it does not make sense to me to modify the way how the `options` object is initialised.

If you are overriding `createChangeStream` in your application, I am advising you to hide the built-in method and expose your custom one under a different name (but the same URL).

```js
MyModel.disableRemoteMethodByName('createChangeStream');
MyModel.myCreateChangeStream = function(options, cb) {
  // your custom implementation, possibly calling out to MyModelcreateChangeStream
}

MyModel.remoteMethod('myCreateChangeStream', {
      description: 'Create a change stream.',
      accessType: 'READ',
      http: [
        {verb: 'post', path: '/change-stream'},
        {verb: 'get', path: '/change-stream'},
      ],
      accepts: {arg: 'options', type: 'object', http: 'optionsFromRequest'},
      returns: {arg: 'changes', type: 'ReadableStream', json: true},
});
```

PR #3070, opened 2017-01-05
Fix construction of sharedCtor remoting metadata

bajtos:
Prevent the situation when we are configuring remoting metadata after strong-remoting has already picked up data from our parent (base) model.

This is a forward-port of fix that was made as part of #3048.


bajtos:
@slnode test please

PR #3072, opened 2017-01-05

bajtos:
Back-port #3053 (cc @yumenohosi):

> We noticed that every time the user model updates, the emailVerified
column would change to false, even though the email was not changed
at all.
>
> I took a look and realized there might be an error in
https://github.com/strongloop/loopback/commit/eb640d8
>
> The intent of the commit just mention is to make emailVerified false
becomes false.
>
> This commit fixes the issue just mentioned.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -845,7 +845,7 @@ module.exports = function(User) {
         if (emailChanged && ctx.Model.settings.emailVerificationRequired) {
           ctx.instance.emailVerified = false;
         }
-      } else {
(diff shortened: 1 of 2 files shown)


lschricke:
Hello! Thanks for this fix. 
Any chance you guys tag a new release soon? This would help a lot, as this issue is quite problematic for us.

Thanks again!

bajtos:
Released in `loopback@2.37.0`

lschricke:
Thanks! :)

PR #3074, opened 2017-01-05
Fix User model to handle custom PK name

ariskemper:
Fix issue with extended user model having different id name using pk name.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -388,6 +388,7 @@ module.exports = function(User) {
     var user = this;
     var userModel = this.constructor;
     var registry = userModel.registry;
+    var pkName = userModel.definition.idName() || 'id';
     assert(typeof options === 'object', 'options required when calling user.verify()');
     assert(options.type, 'You must supply a verification type (options.type)');
     assert(options.type === 'email', 'Unsupported verification type');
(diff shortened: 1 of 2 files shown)


Amir-61:
@ariskemper Thanks for the PR. Could you please add tests to verify your changes?

Thanks!

bajtos:
I modified `test/user.test.js` to use a custom `pk` property as the `idName`. With that change in place, the tests fail against `master` and pass with this patch applied. I have `git push -f` the changes to your feature branch, @ariskemper, you will need to e.g. `git fetch && git reset --hard origin/fix_user_id` to update your local clone.

@superkhau could you please review, since @Amir-61 is on vacation?

The reason why I am asking for another pair of 👀 is my concern about the test solution, it feels a bit like a shotgun to me. In general, I prefer to have focused unit-tests that make it clear what exactly is being tested. OTOH, by forcing _all_ tests to use a model with custom PK, we ensure that even new methods added in the future have to be aware of custom PKs, which may be a good reason why my approach is actually a good/valid one.

Thoughts?

superkhau:
I agree with your proposed solution. I'm swaying on modifying the existing tests.

Ideally, we would have two suites of tests -- one injecting a model with id set and another using pk like in this solution so you don't have to rewrite two set of tests doing the same thing. Meaning models like `UserWithIdAsPK` + `UserWithCustomPK` injected.

That said, I think what you're adding now will definitely work, but is not as "clear" in the sense of coming back to in 6 months and asking why we don't test do it like the ideal case. In any case, :shipit:.

PR #3080, opened 2017-01-09
Warn about injectOptionsFromRemoteContext

bajtos:
The option injectOptionsFromRemoteContext was added to LoopBack 2.x only and is not available in LoopBack 3.x (and newer).

When a model with this option is encountered, we are printing a warning message now, to let the user know about this change between 2.x and 3.x.

Connect to #1495, see also #3048 which implemented the flag in 2.x

@superkhau PTAL
@crandmck could you please review the message?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -248,6 +248,16 @@ module.exports = function(registry) {
       });
     };
 
+    if ('injectOptionsFromRemoteContext' in options) {
+      console.warn(g.f(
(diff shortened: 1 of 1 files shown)


bajtos:
The current message:

```
TestModel is using model setting injectOptionsFromRemoteContext which is no longer available.
Please rework your app to use the offical solution for injecting "options" argument from request context,
see http://loopback.io/doc/en/lb3/Using-current-context.html
```

PR #3098, opened 2017-01-13
Preserve current session when invalidating access tokens

bajtos:
Fix User model to preserve the current session (provided via "options.accessToken") when invalidating access tokens after a change of email or password property.

See #3034

@superkhau please review

cc @ivanschwarz @doublemarked

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -657,7 +657,12 @@ module.exports = function(User) {
     throw err;
   };
 
-  User._invalidateAccessTokensOfUsers = function(userIds, cb) {
+  User._invalidateAccessTokensOfUsers = function(userIds, options, cb) {
+    if (typeof options === 'function' && cb === undefined) {
+      cb = options;
+      options = {};
+    }
+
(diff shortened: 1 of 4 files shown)


common/models/user.js
```diff
@@ -666,7 +671,14 @@ module.exports = function(User) {
...
     var AccessToken = accessTokenRelation.modelTo;
-    AccessToken.deleteAll({userId: {inq: userIds}}, cb);
+
+    var query = {userId: {inq: userIds}};
+    var tokenPK = AccessToken.definition.idName() || 'id';
+    if (options.accessToken && tokenPK in options.accessToken) {
+      query[tokenPK] = {neq: options.accessToken[tokenPK]};
+    }
+
+    AccessToken.deleteAll(query, cb);
```
doublemarked: Shouldn't this pass `options`?

bajtos: Good catch, thank you @doublemarked for the review! I'll fix this ASAP.

bajtos:
Released in `loopback@3.2.1`

Issue #3105, opened 2017-01-16
[BUG ?] Passing tests in french


pierreclr:
Hi there, I'm trying to pass the loopback tests in order to make my PR properly.

Some of them are not passing because it `expect 'erreur inconnue' to be equal 'unknown error'` for example. Indeed, error class expectations are base on regex (and not 'instance of') which is problematic regarding to the automatic translation ..

Maybe the best way would be to set automatically the lang to `en` in the `test` environnement 

Any help or explanation will be welcomed for how to do so !

However I override the french translation file with the english one and my tests passed excepted this one : (message which doesn't appear to be on `intl/fr/message.json` (is this handled in an other level ?) 

<img width="1278" alt="capture d ecran 2017-01-16 a 13 59 39" src="https://cloud.githubusercontent.com/assets/14262431/21984084/83475af4-dbf4-11e6-9808-9bd30d34ebd2.png">

loay:
Hi @pierreclr 
Can you send a pr with the updated proper error message in French so the test passes?
Also, I am not sure if I understand the scenario of the error message. You just switched the language to French and you got one failed test? Can you please give a little bit more details? Thanks.

pierreclr:
No :  A lot of test do not pass because of the translation system.

However for my PR's needs I delete the translation file in `intl/fr/message.json` and then put the english one for replacing. Then I launch the tests and having only one (the one you can see on the screenshot) which is failing.

loay:
Hi @pierreclr 
In order to set your language to English, you have 2 options as indicated here:
https://github.com/strongloop/strong-globalize#demo

I will switch my system to French, then test and then let me know. 
Thanks.

Issue #3081, opened 2017-01-09
[BUG] Security issue when resolving $owner role


pierreclr:
### Bug 
considering the following model definition :
```
{
  "name": "myModel",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "strict": true,
  "properties": {
    "myProperty": "string"
  },
  "relations": {
    "user": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "userId"
    },
    "sender": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "senderId"
    }
  },
  "acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "admin",
      "permission": "ALLOW"
    },
    {
      "principalType": "ROLE",
      "principalId": "$owner",
      "permission": "ALLOW",
      "properties": "findById"
    }
  ]
}
```

If a user tries to access this ressource as a sender or as a user, it should be able to access it.

Problem : In that case, the $owner role will be resolved if the current logged in user (in the access context) has the same id that the one in userId BUT the senderId role will not be resolved which should not be the case.

No we can speak  about security issue : 
considering the following model : 

```
{
  "name": "myCompromisedModel",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "strict": true,
  "properties": {
    "myProperty": "string",
    "userId": "string"
  },
  "relations": {
    "sender": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "senderId"
    }
  },
  "acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "admin",
      "permission": "ALLOW"
    },
    {
      "principalType": "ROLE",
      "principalId": "$owner",
      "permission": "ALLOW",
      "properties": "findById"
    }
  ]
}
```

In that case the only real owner is (in theory) the one contained in the `senderId` property.
BUT, because of how the roles are resolved (see `loopback/common/models/role` L210), the framework first try to check the `userId` property even if the relation to the owner is not defined on this property and will resolve (and ALLOW) the role anyway.
[...]

PR #2454, opened 2016-06-20
Make things support multi-owner

upupzealot:
When a model has two or more "belongsTo"  relations, $owner will check
them all, instead of only checking the first one.

PR #3106, opened 2017-01-16
Allow multiple ownership relations

pierreclr:
### Description

This PR allows us to resolve multiple belongsTo role

#### Related issues

- #3081 mainly
- #3105 because I encountered some problems passing the tests because of my french computer 
- #2454 (closed as a duplicate)

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -215,13 +215,11 @@ module.exports = function(Role) {
       var ownerId = inst.userId || inst.owner;
       // Ensure ownerId exists and is not a function/relation
       if (ownerId && 'function' !== typeof ownerId) {
-        console.log('\n', 'callback line 219', '\n');
         if (callback && matches(ownerId, userId)) callback(null, true);
         return;
       }
 
       // Finally DENY the owner role after sending
-      console.log('\n', 'callback line 225', '\n');
       if (callback) callback(null, false);
     });
   };
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -171,13 +171,6 @@ module.exports = function(Role) {
   Role.isOwner = function isOwner(modelClass, modelId, userId, callback) {
     assert(modelClass, 'Model class is required');
     debug('isOwner(): %s %s userId: %s', modelClass && modelClass.modelName, modelId, userId);
-    // No userId is present
-    if (!userId) {
-      process.nextTick(function() {
-        callback(null, false);
-      });
-      return;
-    }
 
     // Is the modelClass User or a subclass of User?
     if (isUserClass(modelClass)) {
@@ -193,34 +186,43 @@ module.exports = function(Role) {
         if (callback) callback(err, false);
         return;
       }
-      debug('Model found: %j', inst);
-      var ownerId = inst.userId || inst.owner;
-      // Ensure ownerId exists and is not a function/relation
-      if (ownerId && 'function' !== typeof ownerId) {
-        if (callback) callback(null, matches(ownerId, userId));
-        return;
-      } else {
-        // Try to follow belongsTo
-        for (var r in modelClass.relations) {
-          var rel = modelClass.relations[r];
-          if (rel.type === 'belongsTo' && isUserClass(rel.modelTo)) {
-            debug('Checking relation %s to %s: %j', r, rel.modelTo.modelName, rel);
-            inst[r](processRelatedUser);
-            return;
-          }
+
+      // Try to follow belongsTo
+      for (var r in modelClass.relations) {
+        var rel = modelClass.relations[r];
+
+        if (rel.type === 'belongsTo' && isUserClass(rel.modelTo)) {
+          debug('Checking relation %s to %s: %j', r, rel.modelTo.modelName, rel);
+          if (inst[r](processRelatedUser)) return callback(null, true);
         }
+
         debug('No matching belongsTo relation found for model %j and user: %j', modelId, userId);
-        if (callback) callback(null, false);
       }
 
       function processRelatedUser(err, user) {
         if (!err && user) {
           debug('User found: %j', user.id);
-          if (callback) callback(null, matches(user.id, userId));
-        } else {
-          if (callback) callback(err, false);
+          if (callback && matches(user.id, userId)) {
+            return true;
+          }
+          return false;
         }
       }
+
+      // Checking the userId or owner field for resolving owner role
+      // is now done after fetching belongsTo relation to make possible
+      // to have multiple resolver role
+      var ownerId = inst.userId || inst.owner;
+      // Ensure ownerId exists and is not a function/relation
+      if (ownerId && 'function' !== typeof ownerId) {
+        console.log('\n', 'callback line 219', '\n');
+        if (callback && matches(ownerId, userId)) callback(null, true);
+        return;
+      }
+
+      // Finally DENY the owner role after sending
+      console.log('\n', 'callback line 225', '\n');
+      if (callback) callback(null, false);
     });
   };
 
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -172,6 +172,14 @@ module.exports = function(Role) {
     assert(modelClass, 'Model class is required');
     debug('isOwner(): %s %s userId: %s', modelClass && modelClass.modelName, modelId, userId);
 
+    // If no requester id, deny the resolver
+    if (!userId) {
+      process.nextTick(function() {
+        callback(null, false);
+      });
+      return;
+    }
+
     // Is the modelClass User or a subclass of User?
     if (isUserClass(modelClass)) {
       process.nextTick(function() {
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -15,7 +15,7 @@ function checkResult(err, result) {
   assert(!err);
 }
 
-describe.only('role model', function() {
+describe('role model', function() {
   this.timeout(10000);
 
   var app, Role, RoleMapping, User, Application, ACL;
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -15,7 +15,7 @@ function checkResult(err, result) {
   assert(!err);
 }
 
-describe('role model', function() {
+describe.only('role model', function() {
   this.timeout(10000);
 
   var app, Role, RoleMapping, User, Application, ACL;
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

pierreclr:
Sorry, I see that some points are not passing : One of my comment exceed the 50 characters size but needs to be removed anyway before merging (I guess). Could we integrate this check when passing test in local ?

bajtos:
Hi @pierreclr, thank you for the pull request. Don't worry about the git history now, we can fix it together later, when the code is good to be merged.

Are you trying to fix the same issue as #2454? cc @upupzealot

I need you to add unit-tests to verify that your implementation works as intended and also to prevent unintentional regressions in the future.

bajtos:
@slnode ok to test

pierreclr:
Hi @bajtos, I think you're right. I didn't see it (despite actively looking for it :/) 

That's the same issue but I choose another implementation (I think having a counter to check if we have finished is quite tricky (see https://github.com/strongloop/loopback/pull/2454/commits/8a497eb66f657973785ad812450f64a157df2001) and prefer to return the denied callback once every possibility has been tried.

However I see that I forgot to remove console.log ... Depending on the implementation you will choose I'll update my PR and write unit tests.

pierreclr:
Hi @bajtos, @upupzealot is not maintaining his PR anymore I added unit tests to avoid  unintentional regressions in the future. 

Let me know when you plan to implement this feature because I really need to have that going to production soon. Thank you in advance

common/models/role.js
```diff
@@ -171,13 +171,6 @@ module.exports = function(Role) {
   Role.isOwner = function isOwner(modelClass, modelId, userId, callback) {
     assert(modelClass, 'Model class is required');
     debug('isOwner(): %s %s userId: %s', modelClass && modelClass.modelName, modelId, userId);
-    // No userId is present
-    if (!userId) {
```
bajtos: I think this block should be preserved. If there is no `userId` provided by the request context, then there is no point in running any checks - they must all fail.

common/models/role.js
```diff
@@ -193,34 +186,41 @@ module.exports = function(Role) {
...
-            return;
-          }
+
+      // Try to follow belongsTo
+      for (var r in modelClass.relations) {
+        var rel = modelClass.relations[r];
+
+        if (rel.type === 'belongsTo' && isUserClass(rel.modelTo)) {
+          debug('Checking relation %s to %s: %j', r, rel.modelTo.modelName, rel);
+          if (inst[r](processRelatedUser)) return callback(null, true);
```
bajtos: `inst[r]` is async function that may not return immediately, I am surprised this code actually works.

I think we should use [async.some()](http://caolan.github.io/async/docs.html#some) to find whether any of the relation grants the owner role.

```js
async.some(
  Object.keys(modelClass.relations),
  function(r, next) {
    var rel = modelClass.relations[r];
    // ...
    next(null, matches(user.id, userId));
  },
  function(err, matchFound) {
    if (err) return callback(err);
    if (matchFound) return callback(null, matchFound);
    // handle userId/owner properties
  });
```

common/models/role.js
pierreclr: No that's the purpose of this PR : We want to be able to resolve multiple owners so if my two `blongsTo` relations are based on (for example) `requesterId` and `senderId` why would you want to DENY in case of no `userId` field ?

common/models/role.js
pierreclr: But I changed the `processRelateduser` to be not async

bajtos:
> Checking the userId or owner field for resolving owner role
> is now done after fetching belongsTo relation to make possible
> to have multiple resolver role

I am concerned about backwards-compatibility of such change. 

Right now, when a model instance has `userId` or `owner` property, and its value is different from current user id, then `isRole` returns `false`, ignoring any relations. In your proposal here, the relations can overwrite the result of `userId`/`owner` comparison. I am concerned that there may be existing application that will break after such change.

I am proposing to add a feature flag to control this behaviour, defaulting to the old mode.

The algorithm in `isRole` should then become:

 1. Check `userId`/`owner` properties using the current code
 2. If one of them matches, then return `true` via callback
 3. If none of them matches:
  a. if the new flag is not set, then return `false` via callback
  b. if the new flag is set, then continue to next steps
 4. Iterate through relations to find if any of them grants the ownership

Now the remaining question is where and how to configure the feature flag. Consider the use case where a model has two relations to a User model, but only one of the relations is granting the ownership. For example, a `Contract` has an owner who can make changes (WRITE), and an auditor who can only view the data (READ). To support this case, one need to tell LoopBack which relations to examine when determining the ownership.

I am proposing to add a new model-level setting, e.g. `ownerRelations`, with the following values:
 - `true` - all relations to `User`-like models are examined
 - `false` - no relations are examined
 - An array of relation names - only the relations specified in the array are examined

This flag can be read via `MyModel.settings.ownerRelations` and configured e.g. via `common/models/my-model.json`.

@raymondfeng @superkhau do you have any opinion how to best design this feature?

common/models/role.js
pierreclr: the `if (something)` is waiting to have a return. In that cas it will return `true` if the user is owner, `false` else.

common/models/role.js
```diff
@@ -193,34 +186,41 @@ module.exports = function(Role) {
...
+          return false;
         }
       }
+
+      // Checking the userId or owner field for resolving owner role
+      // is now done after fetching belongsTo relation to make possible
+      // to have multiple resolver role
+      var ownerId = inst.userId || inst.owner;
+      // Ensure ownerId exists and is not a function/relation
+      if (ownerId && 'function' !== typeof ownerId) {
```
pierreclr: @bajtos Concerning your comment on backward compatibility, the fallback is here : I choose to first check the belongsTo relations and then fallback to `userId` / `owner` fields. So no problem of backward compatibility in case userId exist without belongsTo relation. However I feel not comfortable with that because it means that you can resolve the owner role without having any `belongsTo` relation defined (which was already the case)

common/models/role.js
pierreclr: Sorry, done in de8cc15

common/models/role.js
pierreclr: Indeed, something is wrong here

pierreclr:
@bajtos Considering your comments I'll update my PR ASAP to make it clean

superkhau:
@bajtos Your proposal sounds solid, my thoughts are pretty much the same.

raymondfeng:
@bajtos I would also like to allow model itself to provide `isOwner(...)` function to take over the check.

upupzealot:
So will this PR be merged into master now? Or should we wait for a `isOwner(...)` solution mentioned above?

pierreclr:
@raymondfeng I don't really see why having an `isOwner(...)` method on a model (as prototype or static ?). I think this feature is a little out of scope of this PR and I don't see the use case of how can someone be owner of a ressource without `belongsTo` relation. In addition we already have `Role.registerResolver(...)`  for resolving custom (owner or not) roles ...

bajtos:
> I would also like to allow model itself to provide `isOwner(...)` function to take over the check.

Makes sense.

> I don't really see why having an isOwner(...) method on a model (as prototype or static ?). I think this feature is a little out of scope of this PR and I don't see the use case of how can someone be owner of a ressource without belongsTo relation. In addition we already have Role.registerResolver(...) for resolving custom (owner or not) roles ...

I agree adding model-specific `isOwner` is out of scope of this pull request, let's work in small incremental steps.

@pierreclr could you please rework your patch according to the proposal in my https://github.com/strongloop/loopback/pull/3106#issuecomment-273777370?

> So will this PR be merged into master now?

@upupzealot The usual rules apply - we will merge it when our concerns were addressed and the code looks good. Perhaps you can work together with @pierreclr to make this happen earlier?

common/models/role.js
bajtos: >  However I feel not comfortable with that because it means that you can resolve the owner role without having any belongsTo relation defined (which was already the case)

While the scenario may look wrong to us, it is supported by the current LoopBack version, thus there may be existing application relying on it, and therefore we need to preserve it for backwards compatibility.

pierreclr:
Hi @bajtos, sorry for the late reply. I can upgrade my PR according to your comment but I'm not familiar with `async` so I wondered If I can make it with `bluebird` Promises ? However, when you spoke about adding a new flag, I imagine that you want it at the application level (not model) ? Where can I find meaning example of how to handle that kind of things your way in loopback ? Thank you

bajtos:
> I'm not familiar with `async` so I wondered If I can make it with `bluebird` Promises ?

Our code base (the part which is published to npmjs) is still callback-based, we cannot use Promises in `common`, `lib` and `server`. I think you should not need to learn all `async` API, the [async.some()](http://caolan.github.io/async/docs.html#some) method I pointed out in my https://github.com/strongloop/loopback/pull/3106#discussion_r96855956 should suffice.

The tests are different, using Promises in tests is the direction we want to take, so feel free to use Promises there! Just please try to preserve existing tests unchanged and use Promises only in the newly added code.

> adding a new flag, I imagine that you want it at the application level (not model) ? Where can I find meaning example of how to handle that kind of things your way in loopback ?

See my previous comment https://github.com/strongloop/loopback/pull/3106#issuecomment-273777370

> I am proposing to add a new model-level setting, e.g. `ownerRelations`, with the following values:
> - `true` - all relations to `User`-like models are examined
> - `false` - no relations are examined
> - An array of relation names - only the relations specified in the array are examined
>
> This flag can be read via `MyModel.settings.ownerRelations` and configured e.g. via `common/models/my-model.json`.

Inside `Role.isOwner()` method, you can read the flag via `modelClass.settings.ownerRelations`.

To set the flag in your test model, add it to the third argument of `app.registry.createModel`, e.g.

```js
app.registry.createModel('Message', {
  name: String,
  // ...
}, {
  relations: {
    // ...
  },
  ownerRelations: true,
  // or
  // ownerRelations: ['user', 'sender']
});
```

pierreclr:
@bajtos working on it but I can't have `modelClass.settings.ownerRelations` in `Role.isOwner()` ...
Do you clean models before defining them to prevent any unauthorized flags ?

pierreclr:
@bajtos, I created a new PR (https://github.com/strongloop/loopback/pull/3140) because I restart from scratch and also because I don't know how to change my comparaison branch for PR across forks ..

bajtos:
>  I created a new PR (#3140) because I restart from scratch and also because I don't know how to change my comparaison branch for PR across forks ..

In general, we prefer to keep the work in the same pull request, as it preserves all comments in a single place. That becomes a bit more difficult when the pull request was submitted from the `master` branch as you did here, so don't worry about it now. I see that you started to use a feature branch in #3140, that's the right way! 👍 

You may find the following article about github and pull requests helpful: http://www.nearform.com/nodecrunch/first-time-with-open-source/

I am closing this pull request in favour of #3140

PR #3107, opened 2017-01-16
Role model: resolves related models by name

benkroeger:
Resolve models related to the `Role` model by name instead of class instance. This allows to use `localRegistry` in `app` without monkeypatching `Role` manually

### Description

When loading the `Role` model into a custom registry (e.g. by setting `localRegistry` to `true` when instantiating the `app` object), static roles can not be resolved because the `RoleMapping` model used inside static methods (e.g. `Role.isInRole()`) is loaded into a different registry (i.e. loopback) and thus not attached to any `dataSource`.
This PR changes resolving models related to the `Role` model by name instead of by prototype, which leads to them being resolved from the same registry that `Role` is loaded in as well.

A similar issue (around the `ACL` model) was previously discussed with @bajtos and resolved in strongloop/loopback#2627

#### Related issues

- strongloop/loopback#2627

### Checklist

### DIFF
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -47,9 +47,6 @@ describe('role model', function() {
     ACL.roleMappingModel = RoleMapping;
     ACL.userModel = User;
     ACL.applicationModel = Application;
-    Role.roleMappingModel = RoleMapping;
-    Role.userModel = User;
-    Role.applicationModel = Application;
   });
 
   it('should define role/role relations', function(done) {
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -24,9 +24,9 @@ module.exports = function(Role) {
   Role.resolveRelatedModels = function() {
     if (!this.userModel) {
       var reg = this.registry;
-      this.roleMappingModel = reg.getModelByType(loopback.RoleMapping);
-      this.userModel = reg.getModelByType(loopback.User);
-      this.applicationModel = reg.getModelByType(loopback.Application);
+      this.roleMappingModel = reg.getModelByType('RoleMapping');
+      this.userModel = reg.getModelByType('User');
+      this.applicationModel = reg.getModelByType('Application');
     }
   };
 


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
@benkroeger thank you for the pull request. Is there any reasonable way how to test verify your change in a unit-test?

test/role.test.js
```diff
@@ -47,9 +47,6 @@ describe('role model', function() {
     ACL.roleMappingModel = RoleMapping;
     ACL.userModel = User;
     ACL.applicationModel = Application;
-    Role.roleMappingModel = RoleMapping;
-    Role.userModel = User;
-    Role.applicationModel = Application;
```
benkroeger: @bajtos with these lines, the original tests implemented a work-around that made my original issue not happen in this module's unit-tests. By setting `Role`'s related models manually, this skipped the process of looking-up loopback's built-in models by prototype and hence prevented the error from happening. If you just remove these lines from the test without implementing my fix, all tests should fail (and hence automatically verify my change)

bajtos: Nice!

I am wondering about the four lines above that set `ACL.*` models - are they indicating a similar problem in the `ACL` model as you are fixing in this patch? Is it something to eventually fix too?

slnode:

slnode:

slnode:

slnode:

bajtos:

test/role.test.js
benkroeger: the ACL model was already fixed in one of my previous PRs strongloop/loopback#2627

bajtos: Can we remove those four lines from the test then?

benkroeger: yep. tried it out earlier. But didn't want to change scope of this PR.

Issue #3003, opened 2016-12-06
Test ES6 code in PhantomJS


bajtos:
Right now, we use PhantomJS headless browser to run (some of) loopback tests in a browser environment. Unfortunately, PhantomJS does not support ES6 yet - see https://github.com/ariya/phantomjs/issues/14506

To allow us to write ES6 code, we need to add some sort of ES6-to-ES5 transpiler to our Grunt pipeline.

ritch:
@bajtos anyone else able to take this on?

@kjdelisle @pthieu ??

bajtos:
>  anyone else able to take this on?

Ah, sure, of course.

So far, we are using `browserify` to build the browser bundle (see [test/karma.conf.js#L19](https://github.com/strongloop/loopback/blob/63beaa21fe270d6dc2f0fa5d6cf17d13048a3065/test/karma.conf.js#L19)) and adding `es5-shim` to add missing ES5 APIs (see [test/karma.conf.js#L23](https://github.com/strongloop/loopback/blob/63beaa21fe270d6dc2f0fa5d6cf17d13048a3065/test/karma.conf.js#L23)).

Unfortunately, ES6 is not only about extra APIs, we need to add ES6-to-ES5 transpiler to the mix. I think it should be enough to configure `karma-browserify` to transform the sources (transpile, perhaps using Babel?), the plugin already supports custom transformations - see https://github.com/nikku/karma-browserify. We may need to add [karma-es6-shim](https://www.npmjs.com/package/karma-es6-shim) to the mix too.

That's a quick intro. I didn't have time to look into this issue, so take my comment with a grain of salt.

pthieu:
Sounds interesting, I can take this on. Might need some hand-holding. What's the priority on this?

superkhau:
IMO, this is very important to having the loopback repo pass (I think there are windows issues related to phantom too).

bajtos:
> What's the priority on this?

I would prefer to get this done ASAP. We have already spent a lot of time and effort to allow us to use ES6 in our code, I think this is the last missing piece.

On the other hand, this is not related to PhantomJS passing/failing in the current code, therefore it's not a critical problem.

I have explicitly disabled ES6 syntax in eslint config of loopback. For upstream dependencies like strong-remoting and loopback-datasource-juggler, we are relying on downstream CI builds to detect the situation where an ES6 code would break PhantomJS tests.

bajtos:
@pthieu 
[...]

Issue #2164, opened 2016-03-23
Request: Return a Promise for PersistedModel.create() batch mode.


jbcpollak:
PersistedModel.create() doesn't return a Promise in batch-create model. See https://github.com/strongloop/loopback/issues/418#issuecomment-74691763 for details.

If you try, you'll get an error like this:

```
TypeError: XXX.app.models.YYY.create(...).then is not a function
    at /.../zzz.js:42:6
```

It would be good to find a solution so this mode can support promises as well. If it cannot be done, I would suggest breaking this mode out into a separate `batchCreate()` method so the lack of Promise support can be documented.

jannyHou:
hey @jbcpollak, thanks for your suggestion, when you test batch-create, are you trying to create multiple instances like this?:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L218-L224

jannyHou:
@jbcpollak could you confirm the scenario that you are testing? From my understanding batch create is like: 
```
module.exports = function(app) {
  var myUser = app.models.myuser;

  // sample data
  var mySamples = [{
  	"username": "me",
  	"password": "hah",
  	"email": "haha@gmail.com"
  }, {
  	"username": "mee",
  	"password": "hah",
  	"email": "hahah@gmail.com"
  }];

  // batch create
  myUser.create(mySamples)
    .then(function(sample) {
    	console.log(sample);
    });
};
```
But it works fine for me.

jbcpollak:
@jannyHou sorry for the delay, it took a while to get a good usecase.

I have something similar to what you showed, but it does not work:

```
	let jobs = [
		{
			"id": "f5b4cbd0-4639-4736-a35c-610bc76b11b3",
			"externalId": "30715796",
			"lines": [
				{
					// proprietary
				},
				{
					// proprietary
				}
			]
		}
	]

	let Job = app.models.Job;
	let jobsP = Job.create(jobs);

	jobsP.then(jobs => {
		console.log('num jobs' + jobs.length);
        };
```

And I see:

```
TypeError: jobsP.then is not a function
    at assertAllocation (test/server/jobTest.spec.js:83:8)
    at Context.<anonymous> (test/server/jobTest.spec.js:116:5)
    at wrappedPromise.propagateAslWrapper [as __asl_wrapper] (node_modules/async-listener/index.js:415:23)
    at node_modules/async-listener/index.js:450:32
```

Here are my loopback version specifications in package.json:

```
[...]

Issue #1610, opened 2015-08-20
User.create([newUserData]) does not return the final instances with id's


einarbmag:
I'm creating a bunch of fake data in a boot script on the backend for testing, and one of the things I want to do is pass an array of new user data into User.create(). Passing a single new user object behaves as expected, returning the newly created user with id property injected. But passing an array simply returns the original array, not an array of the newly created user instances (i.e. there are no id's). The call is working successfully otherwise, because when I check the database the users have been created and have id's.

Note that I'm using the promise functionality of the API, i.e. User.create(newUsersDataArray).then(function(users) {...})

superkhau:
Can you provide a link to a test project on GitHub? See https://github.com/strongloop/loopback/wiki/Issues#bug-report

einarbmag:
This is my first time raising an issue and making a test repo for it, hope this is enough to clarify:

https://github.com/einarbmag/loopback-issue-arrays

I made a boot/ script to generate users, and it shows how the response fails to give the new instance id's when passing an array of new users, but only when using promises, not when using callbacks.

superkhau:
Thanks! I will try to reproduce the issue on my end and let you know how it goes.

lucasSimonelli:
Same issue here. It's not that it doesn't return the `id`, the problem is that it doesn't return a promise or a thenable, i.e.: When calling `Model.create(array)`,  (array being a js array) that same `array`  is what gets returned.

The behaviour is not consistent with the one get when calling `Model.create(data)` where `data` is an object.

I'd expect it to return a promise or an array of promises (or thenables).

piti118:
This weird behaviour got me too. Even though it's easy to workaround using map, if it's not fixed, I think it should at least be documented.

superkhau:
Are you guys still running into issues here? I believe this should be fixed since promise support was implemented recently.

Piu130:
I'm still running into the issue.
```
User.create([
    {email: 'admin@admin.admin', password: 'admin'}
  ], function(err, user) {
    // user.id === undefined
  });
```
package.json:
```
"dependencies": {
  "compression": "^1.6.1",
  "cors": "^2.7.1",
  "loopback": "^2.28.0",
  "loopback-boot": "^2.18.1",
  "loopback-component-explorer": "^2.5.0",
  "loopback-component-storage": "^1.8.0",
  "loopback-connector-mongodb": "^1.14.0",
  "loopback-datasource-juggler": "^2.46.0",
[...]

PR #3140, opened 2017-01-26
Fix OWNER role to handle multiple relations

pierreclr:
### Description

Fix the code resolving OWNER role to correctly handle the situation
where the target model has multiple "belongsTo" relations to the User
model.

Introduce a new model setting "ownerRelations" that enables the new
behavior. When "ownerRelations" is set to true, then all "belongsTo"
relations are considered as granting ownership. Alternatively,
"ownerRelations" can be set to an array of the relations which
are granting ownership.

For example, a document can "belongTo" an `author` and a `reviewer`,
but only the author is an owner, the reviewer is not. In this case,
`ownerRelations` should be set to `['author']`.

#### Related issues

- #3081 mainly
- #3105 because I encountered some problems passing the tests because of my french computer 
- #2454 (closed as a duplicate)
- see also https://github.com/strongloop/loopback/pull/3106

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -222,6 +222,8 @@ module.exports = function(Role) {
    * @promise
    */
   Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, options, callback) {
+    var _this = this;
+
     if (!callback && typeof options === 'function') {
       callback = options;
       options = {};
@@ -238,8 +240,25 @@ module.exports = function(Role) {
     debug('isOwner(): %s %s userId: %s principalType: %s',
       modelClass && modelClass.modelName, modelId, userId, principalType);
 
-    // Return false if userId is missing
+    // Resolve isOwner false if userId is missing
     if (!userId) {
+      process.nextTick(function() {
+        callback(null, false);
+      });
+      return callback.promise;
+    }
+
+    // At this stage, principalType is valid in one of 2 following condition:
+    // 1. the app has a single user model and principalType is 'USER'
+    // 2. the app has multiple user models and principalType is not 'USER'
+    // multiple user models
+    var isMultipleUsers = _isMultipleUsers();
+    var isPrincipalTypeValid =
+      (!isMultipleUsers && principalType === Principal.USER) ||
+      (isMultipleUsers && principalType !== Principal.USER);
+
+    // Resolve isOwner false if principalType is invalid
+    if (!isPrincipalTypeValid) {
       process.nextTick(function() {
         callback(null, false);
       });
@@ -265,10 +284,25 @@ module.exports = function(Role) {
       }
       debug('Model found: %j', inst);
 
-      // Historically, for principalType USER, we were resolving isOwner()
-      // as true if the model has "userId" or "owner" property matching
-      // id of the current user (principalId), even though there was no
-      // belongsTo relation set up.
+      var ownerRelations = modelClass.settings.ownerRelations;
+      if (!ownerRelations) {
+        return legacyOwnershipCheck(inst);
+      } else {
+        return checkOwnership(inst);
+      }
+    });
+    return callback.promise;
+
+    // NOTE Historically, for principalType USER, we were resolving isOwner()
+    // as true if the model has "userId" or "owner" property matching
+    // id of the current user (principalId), even though there was no
+    // belongsTo relation set up.
+    // Additionaly, the original implementation did not support the
+    // possibility for a model to have multiple related users: when
+    // testing belongsTo relations, the first related user failing the
+    // ownership check induced the whole isOwner() to resolve as false.
+    // This behaviour will be pruned at next LoopBack major release.
+    function legacyOwnershipCheck(inst) {
       var ownerId = inst.userId || inst.owner;
       if (principalType === Principal.USER && ownerId && 'function' !== typeof ownerId) {
         return callback(null, matches(ownerId, userId));
@@ -282,9 +316,16 @@ module.exports = function(Role) {
         if (!belongsToUser) {
           continue;
         }
+
         // checking related user
-        var userModelName = rel.modelTo.modelName;
-        if (principalType === Principal.USER || principalType === userModelName) {
+        var relatedUser = rel.modelTo;
+        var userModelName = relatedUser.modelName;
+        var isMultipleUsers = _isMultipleUsers(relatedUser);
+        // a relation can be considered for isOwner resolution if:
+        // 1. the app has a single user model and principalType is 'USER'
+        // 2. the app has multiple user models and principalType is the related user model name
+        if ((!isMultipleUsers && principalType === Principal.USER) ||
+            (isMultipleUsers && principalType === userModelName)) {
           debug('Checking relation %s to %s: %j', r, userModelName, rel);
           inst[r](processRelatedUser);
           return;
@@ -295,15 +336,72 @@ module.exports = function(Role) {
       callback(null, false);
 
       function processRelatedUser(err, user) {
-        if (!err && user) {
-          debug('User found: %j', user.id);
-          callback(null, matches(user.id, userId));
-        } else {
-          callback(err, false);
+        if (err || !user) return callback(err, false);
+
+        debug('User found: %j', user.id);
+        callback(null, matches(user.id, userId));
+      }
+    }
+
+    function checkOwnership(inst) {
+      var ownerRelations = inst.constructor.settings.ownerRelations;
+      // collecting related users
+      var relWithUsers = [];
+      for (var r in modelClass.relations) {
+        var rel = modelClass.relations[r];
+        // relation should be belongsTo and target a User based class
+        if (rel.type !== 'belongsTo' || !isUserClass(rel.modelTo)) {
+          continue;
+        }
+
+        // checking related user
+        var relatedUser = rel.modelTo;
+        var userModelName = relatedUser.modelName;
+        var isMultipleUsers = _isMultipleUsers(relatedUser);
+        // a relation can be considered for isOwner resolution if:
+        // 1. the app has a single user model and principalType is 'USER'
+        // 2. the app has multiple user models and principalType is the related user model name
+        // In addition, if an array of relations if provided with the ownerRelations option,
+        // then the given relation name is further checked against this array
+        if ((!isMultipleUsers && principalType === Principal.USER) ||
+            (isMultipleUsers && principalType === userModelName)) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
+          if (ownerRelations === true) {
+            relWithUsers.push(r);
+          } else if (Array.isArray(ownerRelations) && ownerRelations.indexOf(r) !== -1) {
+            relWithUsers.push(r);
+          }
         }
       }
-    });
-    return callback.promise;
+      if (relWithUsers.length === 0) {
+        debug('No matching belongsTo relation found for model %j and user: %j principalType: %j',
+          modelId, userId, principalType);
+        return callback(null, false);
+      }
+
+      // check related users: someSeries is used to avoid spamming the db
+      async.someSeries(relWithUsers, processRelation, callback);
+
+      function processRelation(r, cb) {
+        inst[r](function processRelatedUser(err, user) {
+          if (err || !user) return cb(err, false);
+
+          debug('User found: %j (through %j)', user.id, r);
+          cb(null, matches(user.id, userId));
+        });
+      }
+    }
+
+    // A helper function to check if the app user config is multiple users or
+    // single user. It can be used with or without a reference user model.
+    // In case no user model is provided, we use the registry to get any of the
+    // user model by type. The relation with AccessToken is used to check
+    // if polymorphism is used, and thus if multiple users.
+    function _isMultipleUsers(userModel) {
+      var oneOfUserModels = userModel || _this.registry.getModelByType('User');
+      var accessTokensRel = oneOfUserModels.relations.accessTokens;
+      return !!(accessTokensRel && accessTokensRel.polymorphic);
+    }
   };
 
   Role.registerResolver(Role.AUTHENTICATED, function(role, context, callback) {
--- a/lib/application.js
+++ b/lib/application.js
@@ -458,6 +458,16 @@ app._verifyAuthModelRelations = function() {
 
   function verifyUserRelations(Model) {
     const hasManyTokens = Model.relations && Model.relations.accessTokens;
+
+    // display a temp warning message for users using multiple users config
+    if (hasManyTokens.polymorphic) {
+      console.warn(
+        'The app configuration follows the multiple user models setup ' +
+          'as described in http://ibm.biz/setup-loopback-auth',
+          'The built-in role resolver $owner is not currently compatible ' +
+          'with this configuration and should not be used in production.');
+    }
+
     if (hasManyTokens) return;
 
     const relationsConfig = Model.settings.relations || {};
--- a/test/multiple-user-principal-types.test.js
+++ b/test/multiple-user-principal-types.test.js
@@ -8,6 +8,7 @@ var expect = require('./helpers/expect');
 var request = require('supertest');
 var loopback = require('../');
 var ctx = require('../lib/access-context');
+var extend = require('util')._extend;
 var AccessContext = ctx.AccessContext;
 var Principal = ctx.Principal;
 var Promise = require('bluebird');
@@ -22,6 +23,7 @@ describe('Multiple users with custom principalType', function() {
   beforeEach(function setupAppAndModels() {
     // create a local app object that does not share state with other tests
     app = loopback({localRegistry: true, loadBuiltinModels: true});
+    app.set('_verifyAuthModelRelations', false);
     app.dataSource('db', {connector: 'memory'});
 
     var userModelOptions = {
@@ -350,37 +352,38 @@ describe('Multiple users with custom principalType', function() {
       });
     });
 
-    describe('built-in role resolver', function() {
-      it('supports AUTHENTICATED', function() {
+    describe('built-in role resolvers', function() {
+      it('supports $authenticated', function() {
         return Role.isInRole(Role.AUTHENTICATED, userOneBaseContext)
           .then(function(isInRole) {
             expect(isInRole).to.be.true();
           });
       });
 
-      it('supports UNAUTHENTICATED', function() {
+      it('supports $unauthenticated', function() {
         return Role.isInRole(Role.UNAUTHENTICATED, userOneBaseContext)
           .then(function(isInRole) {
             expect(isInRole).to.be.false();
           });
       });
 
-      it('supports OWNER', function() {
-        var Album = app.registry.createModel('Album', {
-          name: String,
-          userId: Number,
-        }, {
-          relations: {
-            user: {
-              type: 'belongsTo',
-              model: 'OneUser',
-              foreignKey: 'userId',
+      describe('$owner', function() {
+        it('supports legacy behavior with relations', function() {
+          var Album = app.registry.createModel('Album', {
+            name: String,
+            userId: Number,
+          }, {
+            relations: {
+              user: {
+                type: 'belongsTo',
+                model: 'OneUser',
+                foreignKey: 'userId',
+              },
             },
-          },
-        });
-        app.model(Album, {dataSource: 'db'});
+          });
+          app.model(Album, {dataSource: 'db'});
 
-        return Album.create({name: 'album', userId: userFromOneModel.id})
+          return Album.create({name: 'album', userId: userFromOneModel.id})
           .then(function(album) {
             var validContext = {
               principalType: OneUser.modelName,
@@ -393,37 +396,197 @@ describe('Multiple users with custom principalType', function() {
           .then(function(isOwner) {
             expect(isOwner).to.be.true();
           });
-      });
-
-      it('expects OWNER to resolve false if owner has incorrect principalType', function() {
-        var Album = app.registry.createModel('Album', {
-          name: String,
-          userId: Number,
-        }, {
-          relations: {
-            user: {
-              type: 'belongsTo',
-              model: 'OneUser',
-              foreignKey: 'userId',
-            },
-          },
         });
-        app.model(Album, {dataSource: 'db'});
 
-        return Album.create({name: 'album', userId: userFromOneModel.id})
+        // With multiple users config, we cannot resolve a user based just on
+        // his id, as many users from different models could have the same id.
+        it('legacy behavior resolves false without belongsTo relation', function() {
+          var Album = app.registry.createModel('Album', {
+            name: String,
+            userId: Number,
+            owner: Number,
+          });
+          app.model(Album, {dataSource: 'db'});
+
+          return Album.create({
+            name: 'album',
+            userId: userFromOneModel.id,
+            owner: userFromOneModel.id,
+          })
           .then(function(album) {
-            var invalidContext = {
-              principalType: AnotherUser.modelName,
+            var authContext = {
+              principalType: OneUser.modelName,
               principalId: userFromOneModel.id,
               model: Album,
               id: album.id,
             };
-            return Role.isInRole(Role.OWNER, invalidContext);
+            return Role.isInRole(Role.OWNER, authContext);
           })
           .then(function(isOwner) {
             expect(isOwner).to.be.false();
           });
+        });
+
+        it('legacy behavior resolves false if owner has incorrect principalType', function() {
+          var Album = app.registry.createModel('Album', {
+            name: String,
+            userId: Number,
+          }, {
+            relations: {
+              user: {
+                type: 'belongsTo',
+                model: 'OneUser',
+                foreignKey: 'userId',
+              },
+            },
+          });
+          app.model(Album, {dataSource: 'db'});
+
+          return Album.create({name: 'album', userId: userFromOneModel.id})
+          .then(function(album) {
+            var invalidPrincipalTypes = [
+              'invalidContextName',
+              'USER',
+              AnotherUser.modelName,
+            ];
+            var invalidContexts = invalidPrincipalTypes.map(principalType => {
+              return {
+                principalType,
+                principalId: userFromOneModel.id,
+                model: Album,
+                id: album.id,
+              };
+            });
+            return Promise.map(invalidContexts, context => {
+              return Role.isInRole(Role.OWNER, context)
+                .then(isOwner => {
+                  return {
+                    principalType: context.principalType,
+                    isOwner,
+                  };
+                });
+            });
+          })
+          .then(result => {
+            expect(result).to.eql([
+              {principalType: 'invalidContextName', isOwner: false},
+              {principalType: 'USER', isOwner: false},
+              {principalType: AnotherUser.modelName, isOwner: false},
+            ]);
+          });
+        });
+
+        it.skip('resolves the owner using the corrent belongsTo relation',
+        function() {
+          // passing {ownerRelations: true} will enable the new $owner role resolver
+          // with any belongsTo relation allowing to resolve truthy
+          var Message = createModelWithOptions(
+            'ModelWithAllRelations',
+            {ownerRelations: true}
+          );
+
+          var messages = [
+            {content: 'firstMessage', customerId: userFromOneModel.id},
+            {
+              content: 'secondMessage',
+              customerId: userFromOneModel.id,
+              shopKeeperId: userFromAnotherModel.id,
+            },
+
+            // this is the incriminated message where two foreignKeys have the
+            // same id but points towards two different user models. Although
+            // customers should come from userFromOneModel and shopKeeperIds should
+            // come from userFromAnotherModel. The inverted situation still resolves
+            // isOwner true for both the customer and the shopKeeper
+            {
+              content: 'thirdMessage',
+              customerId: userFromAnotherModel.id,
+              shopKeeperId: userFromOneModel.id,
+            },
+
+            {content: 'fourthMessage', customerId: userFromAnotherModel.id},
+            {content: 'fifthMessage'},
+          ];
+          return Promise.map(messages, msg => {
+            return Message.create(msg);
+          })
+          .then(messages => {
+            return Promise.all([
+              isOwnerForMessage(userFromOneModel, messages[0]),
+              isOwnerForMessage(userFromAnotherModel, messages[0]),
+              isOwnerForMessage(userFromOneModel, messages[1]),
+              isOwnerForMessage(userFromAnotherModel, messages[1]),
+
+              isOwnerForMessage(userFromOneModel, messages[2]),
+              isOwnerForMessage(userFromAnotherModel, messages[2]),
+
+              isOwnerForMessage(userFromAnotherModel, messages[3]),
+              isOwnerForMessage(userFromOneModel, messages[4]),
+            ]);
+          })
+          .then(result => {
+            expect(result).to.eql([
+              {userFrom: 'OneUser', msg: 'firstMessage', isOwner: true},
+              {userFrom: 'AnotherUser', msg: 'firstMessage', isOwner: false},
+              {userFrom: 'OneUser', msg: 'secondMessage', isOwner: true},
+              {userFrom: 'AnotherUser', msg: 'secondMessage', isOwner: true},
+
+              // these 2 tests fail because we cannot resolve ownership with
+              // multiple owners on a single model instance with a classic
+              // belongsTo relation, we need to use belongsTo with polymorphic
+              // discriminator to distinguish between the 2 models
+              {userFrom: 'OneUser', msg: 'thirdMessage', isOwner: false},
+              {userFrom: 'AnotherUser', msg: 'thirdMessage', isOwner: false},
+
+              {userFrom: 'AnotherUser', msg: 'fourthMessage', isOwner: false},
+              {userFrom: 'OneUser', msg: 'fifthMessage', isOwner: false},
+            ]);
+          });
+        });
       });
+
+      // helpers
+      function isOwnerForMessage(user, msg) {
+        var accessContext = {
+          principalType: user.constructor.modelName,
+          principalId: user.id,
+          model: msg.constructor,
+          id: msg.id,
+        };
+        return Role.isInRole(Role.OWNER, accessContext)
+          .then(isOwner => {
+            return {
+              userFrom: user.constructor.modelName,
+              msg: msg.content,
+              isOwner,
+            };
+          });
+      }
+
+      function createModelWithOptions(name, options) {
+        var baseOptions = {
+          relations: {
+            sender: {
+              type: 'belongsTo',
+              model: 'OneUser',
+              foreignKey: 'customerId',
+            },
+            receiver: {
+              type: 'belongsTo',
+              model: 'AnotherUser',
+              foreignKey: 'shopKeeperId',
+            },
+          },
+        };
+        options = extend(baseOptions, options);
+        var Model = app.registry.createModel(
+          name,
+          {content: String, senderType: String},
+          options
+        );
+        app.model(Model, {dataSource: 'db'});
+        return Model;
+      }
     });
 
     describe('isMappedToRole()', function() {
--- a/test/role.test.js
+++ b/test/role.test.js
@@ -8,6 +8,7 @@ var assert = require('assert');
 var sinon = require('sinon');
 var loopback = require('../index');
 var async = require('async');
+var extend = require('util')._extend;
 var expect = require('./helpers/expect');
 var Promise = require('bluebird');
 
@@ -363,7 +364,8 @@ describe('role model', function() {
     });
   });
 
-  it('should support owner role resolver', function(done) {
+  // this test should be split to address one resolver at a time
+  it('supports built-in role resolvers', function(done) {
     Role.registerResolver('returnPromise', function(role, context) {
       return new Promise(function(resolve) {
         process.nextTick(function() {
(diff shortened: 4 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

test/role.test.js
```diff
@@ -431,6 +431,179 @@ describe('role model', function() {
...
+          model: 'User',
+          foreignKey: 'senderId',
+        },
+      },
+      ownerRelations: true,
+    });
+
+    app.model(Message, {dataSource: 'db'});
+
+    const password = 'pass';
```
bajtos: I am afraid we cannot use ES6 in this repository yet because of https://github.com/strongloop/loopback/issues/3003 :(

pierreclr: Done in 719945d

common/models/role.js
```diff
@@ -187,38 +187,78 @@ module.exports = function(Role) {
...
+            if (callback) callback(err, false);
+          }
+        }
       } else {
-        // Try to follow belongsTo
+        /**
+         * Here is the new way to resolve owner relations
+         * First, fetches `belongsTo` relation (User)
+         * Then, fallback looking for userId / owner
+         */
```
bajtos: If I understand the code correctly, then you are proposing to ignore hard-coded property names like `userId` and `ownerId` when `ownerRelations` is configured. I guess it makes sense, we must describe this difference in our documentation.

Please add unit-test(s) to verify that when `ownerRelations` is set, and there is no relation associated with `userId`/`ownerId`, then these two properties are ignored.

pierreclr: Done in 199fa03

test/role.test.js
```diff
@@ -431,6 +431,261 @@ describe('role model', function() {
...
+        principalType: ACL.USER, principalId: user.id,
+        model: Message, id: messages[1].id,
+      };
+      var noOwnerRole = {
+        principalType: ACL.USER, principalId: sender.id,
+        model: Message, id: messages[2].id,
+      };
+
+      return Promise.all([
+        new Promise((resolve, reject) => {
```
bajtos: Please promise-ify `Role.isInRole` and other related APIs you call from new tests. See https://github.com/strongloop/loopback/issues/418 for background and https://github.com/strongloop/loopback/pull/1493/files for an example how to implement such change.

Would you mind to add Promise support in a new pull request first, before working on this feature? It will help us to keep the git history clean.

pierreclr: > Would you mind to add Promise support in a new pull request first, before working on this feature?

What do you mean by this ?

bajtos: We try to keep pull request small and have each request focused on a single thing. Adding Promise support AND allowing multi-owner Role resolution in the same pull request would go against that.

So I see two options how to move this pull request forward:
 - Rewrite your new tests to use callbacks instead of promises. This seems like a step back to me.
 - Start a new feature branch & a new pull request that will only add support for Promise style invocation to `Role` methods (and possibly other models you are going to invoke the tests you are adding here). When this other pull request is landed, you can simplify the code in this pull request by using the newly added Promise API.

pierreclr: Ok, you're definitely right. I find it uggly anyway ... However, do I need to add tests on the promisify support for role related things ?

bajtos: Yes, at least basic smoke tests are needed when adding support for Promises to our API.

I went ahead and implemented the improvement myself - see #3146.

pierreclr:
@bajtos i updated my tests with your promisify role lib ! However we need to merge this once ES6 polyfill and promisify role lib will be merged

ebarault:
@pierreclr @bajtos please see https://github.com/strongloop/loopback/issues/3180
While not addressing you original concern specifically, i believe my changes solve the issue you wrote this PR for.
The block of code you're referring to [here](https://github.com/strongloop/loopback/pull/3140/files#diff-733982793c3bd5e61d1a4350451fb494R226) is no longer usable when having multiple user models in a loopback application and so consequently the only valid method is to resolve belongsTo relations

pierreclr:
Hi @ebarault ! That's what I thought first. However adding the capability to have multiple user models imply breaking changes on the way to resolve owner role (expect nosql connectors since the id's are uniques cross dbs / collections). I think the best way to handle this problem is the following :
- keep the old way for loopback 2.X
- Keep the flag on loopback 3.X and add documentation about the necessity to have this flag in case of multiple user models. Or only keep the isOwner resolving which fetch the belongsTo relation without giving the choice to the developper. 

However, IMO this feature becomes crucial integrating multiple user models. (Think about an `author` and an `editor` which both are users in separate models. We definitely want a `book` owned by both of them ! On relations `author` and `editor`)

ebarault:
Hi @pierreclr : could you detail the breaking change you're referring that would not be covered with the implementation i proposed in https://github.com/strongloop/loopback/pull/3180 ? (similar to yours when there's a single user model)

all existing tests pass, and I don't have in mind a new test that could fail

pierreclr:
@ebarault  As already discuss and even if it was not very good for security, we used to resolve owner role on userId / owner property. In your PR you just (and only) fetch belongsTo relation (My initial PR did the exact same thing). But the braking change is the following. Considering a loopback app with model without any belongsTo relation BUT a `userId` property on it. Consider the actual way to resolve owner, here we are expecting to resolve the owner role on `userId` property which will not be the case fetching only belongsTo relations ...

ebarault:
@pierreclr: hmm right, but this implementation is so limited, ... for example it does not allow someone to use different foreign keys name than userId or owner.
Security : imagine a model configured with strict mode false : if no specific policy exists: someone with write access on the model instance could do an updateAttribute to add the role property referencing anybody else and provide him access

I'd advocate that loopback 3 should not support this owner resolving mechanism and follows only belongsTo relations. loopback 2 would remain unchanged. Thoughts?

bajtos:
@ebarault thank you for joining the discussion here! Please read also our previous discussion in #3106 to better understand the context.

In particular, https://github.com/strongloop/loopback/pull/3106#issuecomment-273777370 describes my concerns related to backwards compatibility.

> I'd advocate that loopback 3 should not support this owner resolving mechanism and follows only belongsTo relations. loopback 2 would remain unchanged. Thoughts?

IMO, changing the auth layer to follow only `belongsTo` relations is a breaking change, because existing applications based on LoopBack 3.x and depending on userId/owner property-based ownership will break after upgrading to a newer version of LoopBack.

Feature flags are the solution we are using in situations like this.
 - By default, the feature flag is turned off, therefore upgrading projects are not affected. 
 - We can modify our project templates in https://github.com/strongloop/loopback-workspace/tree/master/templates to automatically enable feature flags for new projects.
 - When a new major version is around, we can remove feature flags that were introduced for backwards compatibility only.

However, in this particular case, I think it's important to allow users to decide whether they want all "belongsTo" relations to be treated as ownership (`ownerRelations: true`) or only selected subset (e.g. `ownerRelations: ['user', 'owner']`).

> About security : imagine a model configured with strict mode false : if no specific policy exists, someone with write access on the model instance could do an updateAttribute to add the owner property referencing anybody else and provide him access.

Good point. Perhaps we should fix the code handling `userId`/`owner` properties (with no associated relation) to require these properties to be defined on the model? While it may still break existing applications, I think the security implications provide enough justification.

ebarault:
@bajtos @pierreclr : a shame I missed the discussion from https://github.com/strongloop/loopback/pull/3106

I am not using the $owner role because the behavior is not consistent to me, so I had to implement my own resolver to handler more complex logics. When required to refactor the isOwner method for the sake of the multiple user model PR, i understood then why I could not use this role resolver as I was expecting it to work and considered it more a bug than a feature : so I proposed that fix. But... okay, let's keep full backward compatibility. I'll keep focussed in the other PR to the fix required for multiple users. It should be merged before the present one.

About [this part](https://github.com/strongloop/loopback/pull/3140/files#diff-733982793c3bd5e61d1a4350451fb494R235): here is why i consider it a bug:
1. consider a model which is linked to two users (one being an owner (same id), the other not being an owner (different id)) by two different relations
2. consider now that the name of the relation for the second relation (with a non owner) comes before in the lexicographical order
**Result** : the check will callback false for the second relation even before checking the first relation which is the valid one, resulting in the isOwner returning inconsistent results depending on the names of the relations.

(ps. i considered here that **for (var i in object)** returns keys in lexicographical order, it's might not be the case, but anyway this does not change the nature of the problem)

About the security breach I mentioned : there are certainly others,  that one came to me just because my knowledge of the underlying logic, we should tinker this more.

pierreclr:
> consider now that the name of the relation for the second relation (with a non owner) comes before in the lexicographical order
Result : the check will callback false for the second relation even before checking the first relation which is the valid one, resulting in the isOwner returning inconsistent results depending on the names of the relations.

@ebarault  I don't understand that ... in case of a non owner relation to a user (that means relation type `hasOne`), the owner resolver will not try to fetch this relation since we only focus on `belongsTo` + `isUserModel`

raymondfeng:
I have a suggestion: we can use `ownerRelations` to cover both properties and belongsTo relations for owner checking. For example, `ownerRelations: ['ownerId', 'user']` will check `ownerId` (happens to be a property) followed by `user` (happens to be a relation). WDYT?

pierreclr:
@raymondfeng I feel not comfortable with what you propose for the following reasons : 
Considering two user models (authorId, editorId for example needs). Considering them both linked to MySQL datasource. Resolving the owner relation on a field in that case ispretty bad since two different users can have the same ids (however with mongodb it's ok). Indeed, my editor with `id = 1` can access a resource of an author (which has the ìd=1`too) and imply security issue ... IMO, we should definitely get rid of resolving on property and only resolve `belongsTo` (to user class) relations. Let's keep the old way on LB2.X and go forward by fetching `belongsTo` relations. The breaking change is quite easy to handle because developper only needs to add relation
```
"user": {
   "type": "belongsTo",
   "model": "User",
   "foreignKey": "userId" / "owner"
}
```
and everything will work as expected :) 

However, let me know what can I do to accelerate the merge :)

ebarault:
@pierreclr: a model could have any number of belongsTo relation with other users.
consider this:

[invitation](fk: senderId=id1) --belongsTo--> user(id1)
[invitation](fk: guestId=id2)   --belongsTo--> user(id2)

both users have hasMany relations with invitations, none of users them can "belongsTo" an invitation

sender has rights on invitation to display it, revoke it
guest has rights on invitation to display it, accept it

isOwner is a good starting point, complemented by logic inside the methods definitions, to enable such logic

ebarault:
@bajtos, @raymondfeng, @pierreclr 
any comment on what i depicted as [a bug](https://github.com/strongloop/loopback/pull/3140#issuecomment-278643043) in my previous posts before we move on?

ebarault:
Well that's my point: the current implementation does not behave consistently and fluctuates according to the relation name

bajtos:
@ebarault 
> resulting in the isOwner returning inconsistent results depending on the names of the relations.

I agree it's a bug. At the same time, there may be users depending on this buggy behaviour and I don't want to break their applications after upgrading to a newer LoopBack version.

I want to make it very clear that I am happy to make the ACL system better and fix unexpected/buggy behaviour. However, and this is even more important, I don't want to break existing applications after upgrading from a previous LoopBack version (2.x or 3.x), even if these apps are relying on buggy and/or undocumented behaviour of LoopBack.

That's why I am proposing to use feature flags in https://github.com/strongloop/loopback/pull/3140#issuecomment-278614137. There may be other solutions too, e.g. check for polymorphic `accessToken.user` relation to decide whether to use old (buggy) or new (fixed) behaviour (see https://github.com/strongloop/loopback/pull/3180). I am open-minded here.

@pierreclr 
>  Let's keep the old way on LB2.X and go forward by fetching belongsTo relations. The breaking change is quite easy to handle because developper only needs to add relation

I am afraid that's not how [Semantic Versioning](http://semver.org/) works. A breaking change is a breaking change and should be released as 4.0.0 in our case.

---


BTW I feel we may be running in circles here, perhaps a short chat on [gitter](https://gitter.im/strongloop/loopback) or a Hangout video call could help us to reach consensus faster? If you agree, then please drop me an email to the address shown on my GitHub profile with few proposed times that work for you best. I usually work between 9am-5pm CET.

pierreclr:
@bajtos mail send ! (pierre.clocher@gmail.com)

ebarault:
hi pierre, we had a long chat with @bajtos, please rich me in [gitter](https://gitter.im/strongloop/loopback), i'll post you the relevant excerpts

pierreclr:
I'm not familiar with gitter. However I connect through github and join the room but cannot gind you ..

ebarault:
@pierreclr i just reached you on gitter, check messages

ebarault:
@pierreclr @bajtos: so, now https://github.com/strongloop/loopback/pull/3180 is landed, shall we continue discussing the improvement of isOwner() method?
Let's sum up:
- we agreed to trigger the new isOwner behavior according to a feature flag
- @bajtos pointed out: "for the feature flag, we may want to use the variant where the flag value allows a list of related models which should be treated as owners"
- we should update starter templates so this new behavior is activated by default for all new projects
- now that https://github.com/strongloop/loopback/pull/3180 is landed, the present PR needs to evolve so to account for it
- i mentioned we could also discuss the ability to crawl iteratively along belongsTo relations to go up to an actual owner, this could be decided also according to an option in the feature flag

did i miss anything ?

pierreclr:
I have a question about your second point 

> @bajtos pointed out: "for the feature flag, we may want to use the variant where the flag value allows a list of related models which should be treated as owners"

I'm not sure about the consistency of such a feature since when resolving owner role we check if the realtion is `belongsTo` and related model extends the user class. This feature is similar than defining a `hasOne` relation for the related model you don't want to resolve `$owner`.

bajtos:
> we agreed to trigger the new isOwner behavior according to a feature flag

👍 

> @bajtos pointed out: "for the feature flag, we may want to use the variant where the flag value allows a list of related models which should be treated as owners"

👍 

As far as I am concerned, it's ok to leave the list variant out of scope of this pull request and implement only simple true/false flag.

> we should update starter templates so this new behavior is activated by default for all new projects

👍 

It's out of scope of this pull request though, as the templates are in loopback-workspace. In particular, I think the new feature flag should be added here: [common/models/model-definition.json#L37-L39](https://github.com/strongloop/loopback-workspace/blob/fc4f761ed7b635cf18e091b6ad111c31c78880a9/common/models/model-definition.json#L37-L39).

> now that https://github.com/strongloop/loopback/pull/3180 is landed, the present PR needs to evolve so to account for it

Yes please!

> i mentioned we could also discuss the ability to crawl iteratively along belongsTo relations to go up to an actual owner, this could be decided also according to an option in the feature flag

I personally feel this is out of scope of this pull request. Let's keep this pull request as small as reasonably possible.

@ebarault I think adding support for iterative crawl of belongsTo relations should be an incremental backward-compatible addition, are you envisioning any problems we may encounter?

---

As I was re-reading https://github.com/strongloop/loopback/pull/3106#issuecomment-273777370, and in the light of https://github.com/strongloop/loopback/pull/3140#issuecomment-278693313, I think my proposal went unnecessarily far in preserving backwards compatibility. The only thing that really matters is that users upgrading from older version of LoopBack should not be affected by the changes made in this pull request. 

 - when the new flag is disabled, then we must fully preserve current behaviour
 - when the new flag is enabled, we are free to choose whatever behaviour makes most sense for long term

So maybe it will be best to keep most of the current code intact for backwards compatibility, and add a brand-new implementation (possibly copying bits from the old impl) that will be enabled by the feature flag.

```js
Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, callback) {
  // ...
  modelClass.findById(modelId, function(err, inst) {
      if (err || !inst) {
        debug('Model not found for id %j', modelId);
        return callback(err, false);
      }
      debug('Model found: %j', inst);

      if (modelClass.settings.ownerRelations) {
        return checkOwnershipViaRelations(...);
      }

      // otherwise run the legacy check - no code changes should be necessary below

      // Historically, for principalType USER, we were resolving isOwner()
      // as true if the model has "userId" or "owner" property matching
      // id of the current user (principalId), even though there was no
      // belongsTo relation set up.
      var ownerId = inst.userId || inst.owner;
      if (principalType === Principal.USER && ownerId && 'function' !== typeof ownerId) {
        return callback(null, matches(ownerId, userId));
      }
      // ...
  });

  function checkOwnershipViaRelations(...) {
    // the new implementation
  }
};
```

Thoughts?

ebarault:
> @ebarault I think adding support for iterative crawl of belongsTo relations should be an incremental backward-compatible addition, **are you envisioning any problems we may encounter?**

@bajtos : no i don't think so 😄 . According to your next proposal we should then make sure it's available in both implementations (legacy, and feature-flag). But let's discuss in a further PR

> I personally feel this is out of scope of this pull request.

👍  yes of course, just meant to recall the full context

> it will be best to keep most of the current code intact for backwards compatibility, and add a brand-new implementation (possibly copying bits from the old impl) that will be enabled by the feature flag

👍  +**ok** with the proposed code snippet

bajtos:
I have low confidence in our test coverage of `isOwner` method, therefore I would prefer to keep the amount of changes in the legacy implementation as small as possible. This will also make future code archaeology easier. I pushed 5c090e0 where I restructured the code accordingly. @ebarault PTAL.

test/role.test.js
```diff
@@ -363,7 +363,7 @@ describe('role model', function() {
     });
   });
 
-  it('should support owner role resolver', function(done) {
+  it('should support legacy owner role resolver', function(done) {
```
bajtos: `it(supports legacy...)` - see http://loopback.io/doc/en/contrib/style-guide.html

test/role.test.js
```diff
@@ -489,48 +489,230 @@ describe('role model', function() {
...
+      return Promise.all([
+        Role.isInRole(Role.OWNER, userAccessContext),
+        Role.isInRole(Role.OWNER, senderAccesContext),
+        Role.isInRole(Role.OWNER, noOwnerAccessContext),
+      ]);
+    })
+    .spread((isUserOwner, isSenderOwner, isSenderAlsoOwner) => {
+      expect(isUserOwner).to.be.true();
+      expect(isSenderOwner).to.be.false();
+      expect(isSenderAlsoOwner).to.be.false();
```
bajtos: One assert per test please. Ideally, the original "resolves OWNER via ..." tests should be preserved.

If you are concerned about code duplication, then please extract shared helpers, or perhaps move these tests inside a `describe` block where the shared setup can be performed via `beforeEach` hook.

bajtos:
Few more things to test for the new `isOwner` resolver:
 - `owner` and `userId` property with no backing "belongsTo User" relation - verify that the model instance is _not_ considered as owned by the current user.
 - check that isOwner is correctly resolved/rejected when there are multiple User-based models defined in the application (see `test/multiple-user-principal-types.test.js`)

ebarault:
@bajtos : thanks for reviewing, and ok if your code restructuration
@pierreclr : do you have some bandwidth to go on with the tests (modification on existing ones + new ones)

pierreclr:
@ebarault I start to refacto the tests

common/models/role.js
```diff
@@ -295,8 +310,63 @@ module.exports = function(Role) {
...
+      function processRelation(r, cb) {
+        inst[r](processRelatedUser);
+        function processRelatedUser(err, user) {
+          // any error will resolve isOwner falsy
+          if (err) return cb(err);
+          if (user) {
+            debug('User found: %j (through %j)', user.id, r);
+            return cb(null, matches(user.id, userId));
+          }
+          cb(null, false);
```
pierreclr: I choose to handle it the same way than everywhere else. That means : 
```
if (err || !user) return cb(err, false);
debug(...);
cb(null, matches(user.id, userId));
```

ebarault:
thanks @pierreclr i'm taking back from here

common/models/role.js
```diff
@@ -231,14 +233,31 @@ module.exports = function(Role) {
...
+      (isMultipleUsers && principalType !== Principal.USER);
+
+    // Resolve isOwner false if principalType is invalid
+    if (!isPrincipalTypeValid) {
+      process.nextTick(function() {
+        callback(null, false);
+      });
+      return callback.promise;
+    }
+
```
ebarault: we should resolve false if the principalType is invalid given the user models config (single/multiple) : this way enables for example to resolve false if principalType is 'User' but the config is users:multiple
for this we use an helper which i comment later on

common/models/role.js
```diff
@@ -275,9 +309,16 @@ module.exports = function(Role) {
...
-        if (principalType === Principal.USER || principalType === userModelName) {
+        var relatedUser = rel.modelTo;
+        var userModelName = relatedUser.modelName;
+        var isMultipleUsers = _isMultipleUsers(relatedUser);
+        // a relation can be considered for isOwner resolution if:
+        // 1. the app has a single user model and principalType is 'USER'
+        // 2. the app has multiple user models and principalType is the related user model name
+        if ((!isMultipleUsers && principalType === Principal.USER) ||
+            (isMultipleUsers && principalType === userModelName)) {
           debug('Checking relation %s to %s: %j', r, userModelName, rel);
```
ebarault: although we already checked earlier if the config is user:multi/single in a general manner, we test it for each given user through relation. 
It is a failsafe in case the user config is not 100% consistent and we did not detect it earlier

common/models/role.js
```diff
@@ -288,15 +329,69 @@ module.exports = function(Role) {
...
+        var userModelName = relatedUser.modelName;
+        var isMultipleUsers = _isMultipleUsers(relatedUser);
+        // a relation can be considered for isOwner resolution if:
+        // 1. the app has a single user model and principalType is 'USER'
+        // 2. the app has multiple user models and principalType is the related user model name
+        // In addition, if an array of relations if provided with the ownerRelations option,
+        // then the given relation name is further checked against this array
+        if ((!isMultipleUsers && principalType === Principal.USER) ||
+            (isMultipleUsers && principalType === userModelName)) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
```
ebarault: same behavior than commented just before, but for the new isOwner behavior this time

common/models/role.js
```diff
@@ -288,15 +329,69 @@ module.exports = function(Role) {
...
+      }
+    }
+
+    // helpers
+    function _isMultipleUsers(userModel) {
+      var oneOfUserModels = userModel || _this.registry.getModelByType('User');
+      var accessTokensRel = oneOfUserModels.relations.accessTokens;
+      return !!(accessTokensRel && accessTokensRel.polymorphic);
+    }
   };
```
ebarault: this is the helper to check if the app user config is multiple users or single user
it can be used with or without a reference user model. 
in case no user model is provided, we use the registry to get any of the user model by type
the relation with AccessToken is used to check if polymorphism is used, and thus if multiple users

test/multiple-user-principal-types.test.js
```diff
@@ -350,37 +351,38 @@ describe('Multiple users with custom principalType', function() {
...
-        var Album = app.registry.createModel('Album', {
-          name: String,
-          userId: Number,
-        }, {
-          relations: {
-            user: {
-              type: 'belongsTo',
-              model: 'OneUser',
-              foreignKey: 'userId',
+      describe('$owner', function() {
```
ebarault: aggregated the tests for $owner resolver

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+          })
+          .then(result => {
+            expect(result).to.eql([
+              {principalType: 'invalidContextName', isOwner: false},
+              {principalType: 'USER', isOwner: false},
+              {principalType: AnotherUser.modelName, isOwner: false},
+            ]);
+          });
+        });
+
```
ebarault: refactored 3 tests in one to check for various incorrect principalType while keeping the error messages explicit

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+              // belongsTo relation, we need to use belongsTo with polymorphic
+              // discriminator to distinguish between the 2 models
+              {userFrom: 'OneUser', msg: 'thirdMessage', isOwner: false},
+              {userFrom: 'AnotherUser', msg: 'thirdMessage', isOwner: false},
+
+              {userFrom: 'AnotherUser', msg: 'fourthMessage', isOwner: false},
+              {userFrom: 'OneUser', msg: 'fifthMessage', isOwner: false},
+            ]);
+          });
+        });
```
ebarault: this test currently fail, see the inlined comments, support for polymorphic belongsTo should be added to isOwner() to solve this issue, let's discuss this through gitter @bajtos

test/role.test.js
```diff
@@ -363,7 +364,8 @@ describe('role model', function() {
     });
   });
 
-  it('should support owner role resolver', function(done) {
+  // this test should be split to address one resolver at a time
```
ebarault: will propose another PR for this later on

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
     });
   });
 
-  it('resolves OWNER via "userId" property with no relation', function() {
-    var Album = app.registry.createModel('Album', {
-      name: String,
-      userId: Number,
-    });
-    app.model(Album, {dataSource: 'db'});
+  describe('$owner role resolver', function() {
```
ebarault: added an aggregation layer for $owner resolver tests

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+        ]);
+      })
+      .then(result => {
+        expect(result).to.eql([
+          {user: 'sender', msg: 'firstMessage', isOwner: true},
+          {user: 'receiver', msg: 'secondMessage', isOwner: false},
+          {user: 'receiver', msg: 'thirdMessage', isOwner: false},
+        ]);
+      });
+    });
```
ebarault: refactored to have only one expect per test, while keeping the error message explicit

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+        ]);
+      })
+      .then(result => {
+        expect(result).to.eql([
+          {user: 'sender', msg: 'firstMessage', isOwner: true},
+          {user: 'receiver', msg: 'secondMessage', isOwner: true},
+          {user: 'receiver', msg: 'thirdMessage', isOwner: false},
+        ]);
+      });
+    });
```
ebarault: refactored to have only one expect per test, while keeping the error message explicit

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+        ]);
+      })
+      .then(result => {
+        expect(result).to.eql([
+          {user: 'sender', msg: 'firstMessage', isOwner: false},
+          {user: 'receiver', msg: 'secondMessage', isOwner: true},
+          {user: 'receiver', msg: 'thirdMessage', isOwner: false},
+        ]);
+      });
+    });
```
ebarault: dito, refactored to have only one expect per test, while keeping the error message explicit

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+      };
+      return Role.isInRole(Role.OWNER, accessContext)
+        .then(isOwner => {
+          return {
+            user: user.username,
+            msg: msg.content,
+            isOwner,
+          };
+        });
+    }
```
ebarault: this helper prepares accessContext for each user/msg pairs and format a self explicit isOwner response with the following format: 
``` js
{
   user: 'someUserName',
   msg: 'someMessageContent',
   isOwner: true/false
}
```

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+              customerId: userFromAnotherModel.id,
+              shopKeeperId: userFromOneModel.id,
+            },
+
+            {content: 'fourthMessage', customerId: userFromAnotherModel.id},
+            {content: 'fifthMessage'},
+          ];
+          return Promise.map(messages, msg => {
+            return Message.create(msg);
+          })
```
ebarault: just a quick note so we can discuss this:
in @pierreclr original implem, this was 
`return Message.create([messages]);`

but after starting refactoring, this suddenly returned undefined instead of a promise. so I had to switch all my `.create(withArray)` to `Promise.map(create(withObjects)`. Seems it's same thing as happening in this issue
https://github.com/strongloop/loopback/issues/1610

i fought against it for 30 minutes, since it once worked and thought i had broken something, then i moved one using Promise.map. But still this should be investigated

lib/application.js
```diff
@@ -455,6 +455,17 @@ app._verifyAuthModelRelations = function() {
...
+    // display a temp warning message for users using multiple users config
+    if (hasManyTokens.polymorphic) {
+      console.warn(
+        'The app configuration follows the multiple user models setup ' +
+          'as described in http://ibm.biz/setup-loopback-auth',
+          'The built-in role resolver $owner is not currently compatible ' +
+          'with this configuration and should not be used in production.',
+        Model.modelName);
+    }
+
```
ebarault: i added a generic warning msg that pops anytime a User model is configured with multiple-users support. This is detected with the existence of a polymorphic relation from AccessToken to users.

The message covers a bit large, but this was the easiest way to implement rather than also checking if a model has also ownerRelation option configured. Plus in some edge cases, the legacy behavior could be impacted too.

-> Will submit the fix in a new PR shortly.

ebarault:
@bajtos : 
-cleaned up the commit tree a bit
-skipped the faulty test
-added warning message

all should be good or so to take a final look and land

test/multiple-user-principal-types.test.js
bajtos: > but after starting refactoring, this suddenly returned undefined instead of a promise. 

See https://github.com/strongloop/loopback/issues/2164#issuecomment-215645590 - `create()` does not support batch create in promise mode

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+              {userFrom: 'OneUser', msg: 'fifthMessage', isOwner: false},
+            ]);
+          });
+        });
       });
+
+      // helpers
+      function isOwnerForMessage(user, msg) {
+        var accessContext = {
+          principalType: user.constructor.modelName,
```
bajtos: I think that for `legacy behaviour`, `context.principalType` should not be set.

It seems to me that the tests here may be too isolated to fully verify the implementation and that we should probably add few integration-like tests invoking `Model.checkAccess` to verify all the pieces involved in role resolution.

BTW I suspect the new multi-user feature may not be working because `Model.checkAccess` is not filling `accessContext.principalType` from `token.principalType`.

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+          id: album.id,
         });
       })
       .then(isInRole => expect(isInRole).to.be.true());
+    });
+
+    it('supports legacy behavior via a belongsTo relation', function() {
+      // passing no options will result calling
+      // the legacy $owner role resolver behavior
+      var Message = createModelWithOptions('ModelWithNoOptions');
```
bajtos: `createModelWithOptions` is too vague, it does not communicate the fact that the Model is set up with `belongsTo` relations via `senderId` and `receiverId`.

How about `createModelWithSenderReceiverRelations`? (Or even better, replace `create` with `given`  - see e.g. https://martinfowler.com/bliki/GivenWhenThen.html.)

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+        ];
+        return Promise.map(messages, msg => {
+          return Message.create(msg);
+        });
+      })
+      .then(messages => {
+        return Promise.all([
+          isOwnerForMessage(sender, messages[0]),
+          isOwnerForMessage(receiver, messages[1]),
+          isOwnerForMessage(receiver, messages[2]),
```
bajtos: Should we check `isOwnerForMessage(sender, messages[2])` too?

I find it difficult to understand what is the rationale for choosing exactly these three combinations. The tests would be much easier to read if each test case was in its own, well-named test.

```js
describe('legacy behavior via a belongsTo relation', () => {
  let message;
  beforeEach(givenModelWithSenderReceiverRelations);
  beforeEach(givenUsers);

  it('resolves true when the first belongsTo relation matches', () => {
    return Message.create({content: 'firstMessage', senderId: sender.id})
      .then(m => isOwnerForMessage(sender, m))
      .then(result => expect(result.isOwner).to.be.true());
    }
  });
});
```

I think the total number of source code lines may be even smaller using my approach.

test/role.test.js
bajtos: I find this test too difficult to understand this way ☹️  See https://github.com/strongloop/loopback/pull/3140/files#r104924017

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+      .then(result => {
+        expect(result).to.eql([
+          {user: 'sender', msg: 'firstMessage', isOwner: true},
+          {user: 'receiver', msg: 'secondMessage', isOwner: false},
+          {user: 'receiver', msg: 'thirdMessage', isOwner: false},
+        ]);
+      });
+    });
+
+    it('expects new behavior (3.5.0+) to resolve false without belongsTo relation', function() {
```
bajtos: "new behaviour (3.5.0+)" is difficult to decipher. Please mention `ownerRelation` in the test name instead.

I think it we should group the tests into two new `describe` blocks here:

```js
describe('ownerRelations not set (legacy behaviour)', () => {
  // ...
});

describe('ownerRelations', () => {
  // ...
});
```

test/role.test.js
```diff
@@ -489,48 +491,246 @@ describe('role model', function() {
...
+        ];
+        return Promise.map(messages, msg => {
+          return Message.create(msg);
+        });
+      })
+      .then(messages => {
+        return Promise.all([
+          isOwnerForMessage(sender, messages[0]),
+          isOwnerForMessage(receiver, messages[1]),
+          isOwnerForMessage(receiver, messages[2]),
```
bajtos: https://github.com/strongloop/loopback/pull/3140/files#r104924017 applies here too.

test/multiple-user-principal-types.test.js
bajtos: Please ignore this comment, I did not realise I was reviewing the tests for multiple user principal types.

However, the last two paragraphs may be still legit (although out of scope of this pull request):

> It seems to me that the tests here may be too isolated to fully verify the implementation and that we should probably add few integration-like tests invoking `Model.checkAccess` to verify all the pieces involved in role resolution.

> BTW I suspect the new multi-user feature may not be working because `Model.checkAccess` is not filling `accessContext.principalType` from `token.principalType`.

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
-              type: 'belongsTo',
-              model: 'OneUser',
-              foreignKey: 'userId',
-            },
-          },
         });
-        app.model(Album, {dataSource: 'db'});
 
-        return Album.create({name: 'album', userId: userFromOneModel.id})
+        it('legacy behavior resolves false without belongsTo relation', function() {
```
ebarault: for multiple users config no, it's just what it's meant to be, the test was missing
(how could we resolve a user based just on his id, as many users from different models could have the same id?)

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+
+          return Album.create({
+            name: 'album',
+            userId: userFromOneModel.id,
+            owner: userFromOneModel.id,
+          })
           .then(function(album) {
-            var invalidContext = {
-              principalType: AnotherUser.modelName,
+            var validContext = {
```
ebarault: didn't quite get you here i'm afraid

test/multiple-user-principal-types.test.js
ebarault: ok, thanks. then it was working by pure luck at first

test/multiple-user-principal-types.test.js
```diff
@@ -393,37 +395,194 @@ describe('Multiple users with custom principalType', function() {
...
+              isOwnerForMessage(userFromOneModel, messages[0]),
+              isOwnerForMessage(userFromAnotherModel, messages[0]),
+              isOwnerForMessage(userFromOneModel, messages[1]),
+              isOwnerForMessage(userFromAnotherModel, messages[1]),
+
+              isOwnerForMessage(userFromOneModel, messages[2]),
+              isOwnerForMessage(userFromAnotherModel, messages[2]),
+
+              isOwnerForMessage(userFromAnotherModel, messages[3]),
+              isOwnerForMessage(userFromOneModel, messages[4]),
```
ebarault: this is work in progress, currently isOwner does not work in all cases without support for belongsTo relationship with owner. Thanks for the refactor suggestion, i'll adopt this in the next PR

test/multiple-user-principal-types.test.js
ebarault: i will inspect this last part

about integration tests: could you tell me a bit more on your concerns? The binding between Model.checkAccess and ACL.checkAccessForContext is pretty straight forward

test/role.test.js
ebarault: will do

test/role.test.js
ebarault: ok, let's do this then. From within the context of this discussion i find it easy to read, but i get your concern, when someone gets to read this later on

test/role.test.js
ebarault: was expecting some educated advice on this, thanks 👍

common/models/role.js
```diff
@@ -238,14 +240,31 @@ module.exports = function(Role) {
...
     if (!userId) {
       process.nextTick(function() {
         callback(null, false);
       });
       return callback.promise;
     }
 
+    // At this stage, principalType is valid in one of 2 following condition:
+    // 1. the app has a single user model and principalType is 'USER'
+    // 2. the app has multiple user models and principalType is not 'USER'
```
bajtos: I agree it doesn't make sense to call `isOwner` passing `APP` or `ROLE` as a principal type.

However, are we sure that no code path can lead to such call? Is there any safe-guard check in `isOwner` to ensure we still return a sensible result in such case?

common/models/role.js
bajtos: > do you think we should instead do some sanity check during app boot so we warn or stop the app in case of incorrect config, and then get rid of some tests of the like?

I think that would be better - report problems to the user early.

Although we may still want to keep some of the sanity checks here, perhaps as `assert` calls.

common/models/role.js
```diff
@@ -295,15 +336,69 @@ module.exports = function(Role) {
...
+
+          debug('User found: %j (through %j)', user.id, r);
+          console.log(user.constructor.modelName, r, principalType);
+          cb(null, matches(user.id, userId));
+        });
+      }
+    }
+
+    // helpers
+    function _isMultipleUsers(userModel) {
```
bajtos: > ideally it should even be a much more generic helper at loopback app/registry object level

👍  ✖️ 💯

test/multiple-user-principal-types.test.js
bajtos: Makes sense, I think I was a bit confused and thought I was reviewing tests for the single-user scenario.

To prevent future confusion, could you please capture this important difference from the single-user scenario in a comment before the test?

```js
// With multiple users config, we cannot resolve a user based just on his id,
// as many users from different models could have the same id.
it('resolves false without belongsTo relation', function() {
// ...
```

test/multiple-user-principal-types.test.js
bajtos: The diff shows that you renamed `invalidContext` to `validContext`. I was wondering what was the reasoning for this rename and whether a better name (e.g. `authContext`) may be more appropriate? I thought that `invalidContext` was used for context objects that resolve `isOwner` as `false`.

test/multiple-user-principal-types.test.js
bajtos: My idea is to verify that `Model.checkAccess` constructs the correct `AccessContext` instance. As far as I understand the current code, because `Model.checkAccess` is not filling in `principalType`, the default `USER` value will be always used. 

Having said that, I may be wrong as it's entirely possible that I missed some other place where principal id/type is set.

upscreen:
any updates on when this feature will be merged to the master??

bajtos:
> any updates on when this feature will be merged to the master??

That's a good question, indeed. IIRC, this pull request is being blocked by #3265.

@ebarault you used to have a much better understanding of the work being done here and in #3265 (which you authored). What's the status of these to pull requests? Do you have any bandwidth to finish them? If not, could you perhaps advise us if there is a way how to finish this pull request without waiting for #3265?

ebarault:
Hi @bajtos,

> Do you have any bandwidth to finish them? If not, could you perhaps advise us if there is a way how to finish this pull request without waiting for #3265?

yes i still plan to have those addressed. but i'm currently short on bandwidth. I may have a little more time to cover those in august.

Actually, #3265 was blocked by the deficient merge algorithm on models inheritance. Now that it's fixed #3265 can be resumed.

while I believe #3140 could be merged as-is, It's maybe preferable to wait for #3265 to land as it will  help cleaning the code in #3140.

Lahrachtdi:
can we just do something like this in the root folder ?

```
module.exports = function definingDynamicRoles(app) {

  const Role = app.models.Role;

  Role.registerResolver('$owner', (role, context, next) => {

    const userId = context.accessToken.userId;
    const userPrincipalType = context.accessToken.$principalType;
    const tagetModelPrincipalType = context.modelName;
    const targetInstanceId = context.modelId;

    // Q: Is the user logged in? (there will be an accessToken with an ID if so)
    if (!userId) {
      return process.nextTick(() => next(null, false));
    }

    if (userPrincipalType !== tagetModelPrincipalType) {
      return process.nextTick(() => next(null, false));
    }

    if (parseInt(targetInstanceId, 10) !== userId) {
      return process.nextTick(() => next(null, false));
    }

    return process.nextTick(() => next(null, true));
  });

};
```

slnode:

slnode:

slnode:

slnode:

bajtos:
Let's move this pull request forward. I have rebased the changes on top of the latest master and addressed the most important comments that were also easy to fix - see 44db862.

I'll wait few days to allow you (@pierreclr and @ebarault) to review my changes and the patch in whole, if you like to. If there are no objections raised, then I'll go ahead and land this pull request early next week.

slnode:

slnode:

slnode:

bajtos:
Landed, thank you @pierreclr for the contribution 🎉 and sorry that it took so long to get it merged. Thanks @ebarault for your help along the way!

ebarault:
Hi @bajtos !
Thanks for landing this  commit as is.
For posterity, let's recall [here](https://github.com/strongloop/loopback/pull/3140/commits/e17132d06182e421b3da9a5544f68fc04c61ea17#r104612589) the main problem yet to be resolved.

Sorry for the lack of support here, i really lack time these days.

cc: @lehni

lehni:
Great! This fixed our problem in #3506! @ebarault for the mentioned remaining problem that is still to be solved, wouldn't it be good to create a separate issue so that it does not get forgotten, linking to https://github.com/strongloop/loopback/pull/3140/commits/e17132d06182e421b3da9a5544f68fc04c61ea17#r104612589?

bajtos:
> for the mentioned remaining problem that is still to be solved, wouldn't it be good to create a separate issue so that it does not get forgotten, linking it to this?

+1

Could you please create the issue yourself, @lehni, to speed things up?

lehni:
I don't understand anywhere near enough about this ongoing issue to know how to name and word it, unfortunately... : /

lehni:
@bajtos I've created an issue for it now, not sure it captures all that is important: #3644 
cc: @ebarault

bajtos:
> I've created an issue for it now, not sure it captures all that is important: #3644

@lehni this is great, thank you! 🙇

PR #3149, opened 2017-01-27
Enable ES6/ES2015 goodness

bajtos:
### Description

 - Remove ES5 parser exception from .eslintrc
 - Configure Karma to use Babel to transpile ES6 sources to ES5,
   because PhantomJS does not support ES6
 - Upgrade es5-shim to es6-shim

#### Related issues

- #3003

cc @pthieu


pierreclr:
@bajtos That means I can use ES6 code style in the test part ? What about in core ?

bajtos:
> That means I can use ES6 code style in the test part ? What about in core ?

We will be able to use ES6 code style everywhere, as we are already doing in other LoopBack packages. See also http://loopback.io/doc/en/contrib/style-guide.html

PR #3180, opened 2017-02-03
Fix Role.isOwner() when multiple user models

ebarault:
### Description

When merging https://github.com/strongloop/loopback/pull/2971 which i authored, @bajtos noted a missing test on the built-in role resolver `$owner` which internally calls the `Role.isOwner()` method ([see here](https://github.com/strongloop/loopback/pull/2971#issuecomment-276900888))
The added test actually failed, allowing users from different models but with the same id), because the `Role.isOwner()` method was not checking the user's principalType.
(good catch @bajtos 👍 )

This PR fixes this issue (and adds the corresponding test)
see the inlined comment for detailed code review

#### Related PR

https://github.com/strongloop/loopback/pull/2971 (merged)
https://github.com/strongloop/loopback/pull/3140 (in progress, will be impacted)
++ any PR/issue where someone would try to iterate recursively along belongTo relations up to the owning user (need to find those)

### Checklist

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -179,8 +179,10 @@ module.exports = function(Role) {
     }
     var modelClass = context.model;
     var modelId = context.modelId;
-    var userId = context.getUserId();
-    Role.isOwner(modelClass, modelId, userId, callback);
+    var user = context.getUser();
+    var userId = user && user.id;
+    var principalType = user && user.principalType;
+    Role.isOwner(modelClass, modelId, userId, principalType, callback);
   });
 
   function isUserClass(modelClass) {
@@ -210,18 +212,26 @@ module.exports = function(Role) {
    * @param {Function} modelClass The model class
    * @param {*} modelId The model ID
    * @param {*} userId The user ID
+   * @param {String} principalType The user principalType (optional)
    * @callback {Function} [callback] The callback function
    * @param {String|Error} err The error string or object
    * @param {Boolean} isOwner True if the user is an owner.
    * @promise
    */
-  Role.isOwner = function isOwner(modelClass, modelId, userId, callback) {
+  Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, callback) {
+    if (!callback && typeof principalType === 'function') {
+      callback = principalType;
+      principalType = undefined;
+    }
+    principalType = principalType || Principal.USER;
+
     assert(modelClass, 'Model class is required');
     if (!callback) callback = utils.createPromiseCallback();
 
-    debug('isOwner(): %s %s userId: %s', modelClass && modelClass.modelName, modelId, userId);
+    debug('isOwner(): %s %s userId: %s principalType: %s',
+      modelClass && modelClass.modelName, modelId, userId, principalType);
 
-    // No userId is present
+    // Return false if userId is missing
     if (!userId) {
       process.nextTick(function() {
         callback(null, false);
@@ -231,44 +241,58 @@ module.exports = function(Role) {
 
     // Is the modelClass User or a subclass of User?
     if (isUserClass(modelClass)) {
-      process.nextTick(function() {
-        callback(null, matches(modelId, userId));
-      });
+      var userModelName = modelClass.modelName;
+      // matching ids is enough if principalType is USER or matches given user model name
+      if (principalType === Principal.USER || principalType === userModelName) {
+        process.nextTick(function() {
+          callback(null, matches(modelId, userId));
+        });
+      }
       return callback.promise;
     }
 
     modelClass.findById(modelId, function(err, inst) {
       if (err || !inst) {
         debug('Model not found for id %j', modelId);
-        if (callback) callback(err, false);
-        return;
+        return callback(err, false);
       }
       debug('Model found: %j', inst);
+
+      // Historically, for principalType USER, we were resolving isOwner()
+      // as true if the model has "userId" or "owner" property matching
+      // id of the current user (principalId), even though there was no
+      // belongsTo relation set up.
       var ownerId = inst.userId || inst.owner;
-      // Ensure ownerId exists and is not a function/relation
-      if (ownerId && 'function' !== typeof ownerId) {
-        if (callback) callback(null, matches(ownerId, userId));
-        return;
-      } else {
-        // Try to follow belongsTo
-        for (var r in modelClass.relations) {
-          var rel = modelClass.relations[r];
-          if (rel.type === 'belongsTo' && isUserClass(rel.modelTo)) {
-            debug('Checking relation %s to %s: %j', r, rel.modelTo.modelName, rel);
-            inst[r](processRelatedUser);
-            return;
-          }
-        }
-        debug('No matching belongsTo relation found for model %j and user: %j', modelId, userId);
-        if (callback) callback(null, false);
+      if (principalType === Principal.USER && ownerId && 'function' !== typeof ownerId) {
+        return callback(null, matches(ownerId, userId));
       }
+
+      // Try to follow belongsTo
+      for (var r in modelClass.relations) {
+        var rel = modelClass.relations[r];
+        // relation should be belongsTo and target a User based class
+        var belongsToUser = rel.type === 'belongsTo' && isUserClass(rel.modelTo);
+        if (!belongsToUser) {
+          continue;
+        }
+        // checking related user
+        var userModelName = rel.modelTo.modelName;
+        if (principalType === Principal.USER || principalType === userModelName) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
+          inst[r](processRelatedUser);
+          return;
+        }
+      }
+      debug('No matching belongsTo relation found for model %j - user %j principalType %j',
+        modelId, userId, principalType);
+      callback(null, false);
 
       function processRelatedUser(err, user) {
         if (!err && user) {
           debug('User found: %j', user.id);
-          if (callback) callback(null, matches(user.id, userId));
+          callback(null, matches(user.id, userId));
         } else {
-          if (callback) callback(err, false);
+          callback(err, false);
         }
       }
     });
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -128,6 +128,15 @@ AccessContext.prototype.addPrincipal = function(principalType, principalId, prin
  * @returns {*}
  */
 AccessContext.prototype.getUserId = function() {
+  var user = this.getUser();
+  return user && user.id;
+};
+
+/**
+ * Get the user
+ * @returns {*}
+ */
+AccessContext.prototype.getUser = function() {
   var BaseUser = this.registry.getModel('User');
   for (var i = 0; i < this.principals.length; i++) {
     var p = this.principals[i];
@@ -138,17 +147,16 @@ AccessContext.prototype.getUserId = function() {
 
     // the principalType must either be 'USER'
     if (p.type === Principal.USER) {
-      return p.id;
+      return {id: p.id, principalType: p.type};
     }
 
     // or permit to resolve a valid user model
     var userModel = this.registry.findModel(p.type);
     if (!userModel) continue;
     if (userModel.prototype instanceof BaseUser) {
-      return p.id;
+      return {id: p.id, principalType: p.type};
     }
   }
-  return null;
 };
 
 /**
--- a/test/multiple-user-principal-types.test.js
+++ b/test/multiple-user-principal-types.test.js
@@ -204,8 +204,8 @@ describe('Multiple users with custom principalType', function() {
       accessContext = new AccessContext({registry: OneUser.registry});
     });
 
-    describe('getUserId()', function() {
-      it('returns userId although principals contain non USER principals',
+    describe('getUser()', function() {
+      it('returns user although principals contain non USER principals',
       function() {
         return Promise.try(function() {
           addToAccessContext([
@@ -214,12 +214,15 @@ describe('Multiple users with custom principalType', function() {
             {type: Principal.SCOPE},
             {type: OneUser.modelName, id: userFromOneModel.id},
           ]);
-          var userId = accessContext.getUserId();
-          expect(userId).to.equal(userFromOneModel.id);
+          var user = accessContext.getUser();
+          expect(user).to.eql({
+            id: userFromOneModel.id,
+            principalType: OneUser.modelName,
+          });
         });
       });
 
-      it('returns userId although principals contain invalid principals',
+      it('returns user although principals contain invalid principals',
       function() {
         return Promise.try(function() {
           addToAccessContext([
@@ -227,8 +230,11 @@ describe('Multiple users with custom principalType', function() {
             {type: 'invalidModelName'},
             {type: OneUser.modelName, id: userFromOneModel.id},
           ]);
-          var userId = accessContext.getUserId();
-          expect(userId).to.equal(userFromOneModel.id);
+          var user = accessContext.getUser();
+          expect(user).to.eql({
+            id: userFromOneModel.id,
+            principalType: OneUser.modelName,
+          });
         });
       });
 
@@ -238,8 +244,11 @@ describe('Multiple users with custom principalType', function() {
         return ThirdUser.create(commonCredentials)
         .then(function(userFromThirdModel) {
           accessContext.addPrincipal(ThirdUser.modelName, userFromThirdModel.id);
-          var userId = accessContext.getUserId();
-          expect(userId).to.equal(userFromThirdModel.id);
+          var user = accessContext.getUser();
+          expect(user).to.eql({
+            id: userFromThirdModel.id,
+            principalType: ThirdUser.modelName,
+          });
         });
       });
     });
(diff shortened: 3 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

common/models/role.js
```diff
@@ -179,8 +179,8 @@ module.exports = function(Role) {
     }
     var modelClass = context.model;
     var modelId = context.modelId;
-    var userId = context.getUserId();
-    Role.isOwner(modelClass, modelId, userId, callback);
+    var user = context.getUser();
+    Role.isOwner(modelClass, modelId, user.id, user.principalType, callback);
   });
```
ebarault: see below, fetching the userId from accessContext is no longer enough, consequently I had to introduce a `getUser()` method that also returns the user's principalType, and change the signature of `Role.isOwner()` to include the principalType as an optional parameter

common/models/role.js
```diff
@@ -210,16 +210,24 @@ module.exports = function(Role) {
...
    * @promise
    */
-  Role.isOwner = function isOwner(modelClass, modelId, userId, callback) {
+  Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, callback) {
+    if (!callback && typeof principalType === 'function') {
+      callback = principalType;
+      principalType = undefined;
+    }
+    principalType = principalType || Principal.USER;
+
```
ebarault: while principalType is an optional parameter, it defaults internally to USER if undefined for backward compatibility

common/models/role.js
```diff
@@ -231,9 +239,13 @@ module.exports = function(Role) {
...
-        callback(null, matches(modelId, userId));
-      });
+      var userModelName = modelClass.modelName;
+      // matching ids is enough if principalType is USER or matches given user model name
+      if (principalType === Principal.USER || principalType === userModelName) {
+        process.nextTick(function() {
+          callback(null, matches(modelId, userId));
+        });
+      }
       return callback.promise;
```
ebarault: we need now to check if the principalType is either USER (standard, single user model) or equal to user's model name (multiple user models)

common/models/role.js
```diff
@@ -244,31 +256,32 @@ module.exports = function(Role) {
         return;
       }
       debug('Model found: %j', inst);
-      var ownerId = inst.userId || inst.owner;
-      // Ensure ownerId exists and is not a function/relation
-      if (ownerId && 'function' !== typeof ownerId) {
-        if (callback) callback(null, matches(ownerId, userId));
-        return;
```
ebarault: this whole part is not valid anymore as it relies only on comparing userIds and cannot be modified to handle principalType, we now need to directly follow belongsTo relations if any

common/models/role.js
```diff
@@ -244,31 +256,32 @@ module.exports = function(Role) {
...
+        // relation should be belongsTo and target a User based class
+        if (rel.type !== 'belongsTo' && !isUserClass(rel.modelTo)) {
+          continue;
+        }
+        // checking related user
+        var userModelName = rel.modelTo.modelName;
+        if (principalType === Principal.USER || principalType === userModelName) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
+          inst[r](processRelatedUser);
+          return;
```
ebarault: again, principalType must either be equal to USER, or to related user model name

common/models/role.js
```diff
@@ -244,31 +256,32 @@ module.exports = function(Role) {
...
+      callback(null, false);
 
       function processRelatedUser(err, user) {
         if (!err && user) {
           debug('User found: %j', user.id);
-          if (callback) callback(null, matches(user.id, userId));
+          callback(null, matches(user.id, userId));
         } else {
-          if (callback) callback(err, false);
+          callback(err, false);
```
ebarault: just removing useless test on callback now it's always defined thanks to `utils.createPromiseCallback()`

lib/access-context.js
```diff
@@ -138,17 +146,17 @@ AccessContext.prototype.getUserId = function() {
...
     // or permit to resolve a valid user model
     var userModel = this.registry.findModel(p.type);
     if (!userModel) continue;
     if (userModel.prototype instanceof BaseUser) {
-      return p.id;
+      return {id: p.id, principalType: p.type};
     }
   }
-  return null;
+  return {};
```
ebarault: turned the `getUserId` method in `getUser`, now returning an object of the like `{id: p.id, principalType: p.type}`
of course the `getUserId` is kept for backward compatibility just a few lines above, in a shortened form, reusing the `getUser` method

test/multiple-user-principal-types.test.js
```diff
@@ -214,21 +214,27 @@ describe('Multiple users with custom principalType', function() {
...
           ]);
-          var userId = accessContext.getUserId();
-          expect(userId).to.equal(userFromOneModel.id);
+          var user = accessContext.getUser();
+          expect(user).to.eql({
+            id: userFromOneModel.id,
+            principalType: OneUser.modelName,
+          });
         });
       });
```
ebarault: adapted new tests to check getUser instead of getUserId

test/multiple-user-principal-types.test.js
```diff
@@ -214,21 +214,27 @@ describe('Multiple users with custom principalType', function() {
...
```
ebarault: same here

test/multiple-user-principal-types.test.js
```diff
@@ -238,8 +244,11 @@ describe('Multiple users with custom principalType', function() {
...
-          var userId = accessContext.getUserId();
-          expect(userId).to.equal(userFromThirdModel.id);
+          var user = accessContext.getUser();
+          expect(user).to.eql({
+            id: userFromThirdModel.id,
+            principalType: ThirdUser.modelName,
+          });
         });
       });
     });
```
ebarault: same here

test/multiple-user-principal-types.test.js
```diff
@@ -373,19 +383,36 @@ describe('Multiple users with custom principalType', function() {
...
       });
+
+      // helper
+      function isInRole(requests) {
+        var isInRoles =  requests.map(function(request) {
+          return Role.isInRole(request.role, request.context);
+        });
+        return Promise.all(isInRoles);
+      }
     });
```
ebarault: adjusted the test on built-in $owner role resolver to verify a user with similar id than an authorized user userOne but from another user model is denied access to album instance owned by userOne

common/models/role.js
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
+        if (rel.type !== 'belongsTo' || !isUserClass(rel.modelTo)) {
+          continue;
+        }
+
+        // collecting related users
+        var userModelName = rel.modelTo.modelName;
+        if (principalType === Principal.USER || principalType === userModelName) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
+          relWithUsers.push(r);
         }
```
ebarault: collecting related users rather than checking them right away : there might be several belongsTo relations linking the model instance to users, rejecting the role resolver at the first returned falsy is not right

common/models/role.js
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
+        inst[r](processRelatedUser);
+        function processRelatedUser(err, user) {
+          // any error will resolve isOwner falsy
+          if (err) return cb(err);
+          if (user) {
+            debug('User found: %j (through %j)', user.id, r);
+            return cb(null, matches(user.id, userId));
+          }
+          cb(null, false);
         }
```
ebarault: several things to note : 
- i chose asyncSeries so not to spam the db but it's clearly a tradeoff to make between performance and load here
- i chose to handle errors in processRelatedUser as rejecting the overall async.someSeries: we could just callback them falsy, thoughts ?

bajtos:
@slnode ok to test

common/models/role.js
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
+          continue;
         }
-        debug('No matching belongsTo relation found for model %j and user: %j', modelId, userId);
-        if (callback) callback(null, false);
+
+        // collecting related users
+        var userModelName = rel.modelTo.modelName;
+        if (principalType === Principal.USER || principalType === userModelName) {
+          debug('Checking relation %s to %s: %j', r, userModelName, rel);
+          relWithUsers.push(r);
```
bajtos: Could you please rewrite this loop using [Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)?

```js
var relWithUsers = modelClass.relations.filter(r => {
  const rel = modelClass.relations[r];
  // ...
});
``

File: common/models/role.js [hunk 15]
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
-          if (callback) callback(err, false);
+      // check related users: someSeries is used to avoid spamming the db
+      async.someSeries(relWithUsers, processRelation, function(err, isOwner) {
+        if (err) return callback(err);
+        return callback(null, isOwner);
+      });
+
+      function processRelation(r, cb) {
+        inst[r](processRelatedUser);
+        function processRelatedUser(err, user) {
```
bajtos: Nitpic:

```js
inst[r](function processRelatedUser(err, user) {
  // ...
});
```

common/models/role.js
bajtos: > i chose asyncSeries so not to spam the db but it's clearly a tradeoff to make between performance and load here

Yeah, it's a tradeoff and I don't have a strong opinion on which side is better. I am fine to use `asyncSeries` here.

> i chose to handle errors in processRelatedUser as rejecting the overall async.someSeries: we could just callback them falsy, thoughts ?

I think the question is what to do when we run let's say 4 queries, one of them fails, two returns information that the user is not owner, and one says that the user is owner. Should we fail the whole request because of the error? Should we ignore the error and process the request, because we know that the user is authorized, 
despite the error?

I think I like the current version better:
 - When there is an error, it's better to let the user and app administrators know about the problem
 - We can relax this behaviour later in the future if needed. Going the other way, throwing an error in situation that used to work, would be a semver-major change IMO.

lib/access-context.js
bajtos: I think we should return `null` when there is no user associated with this AccessContext, to make it easy for the consumers of this API to detect this situation.

test/multiple-user-principal-types.test.js
```diff
@@ -373,19 +383,36 @@ describe('Multiple users with custom principalType', function() {
...
+
+            return isInRole([
+              {role: Role.OWNER, context: validContext},
+              {role: Role.OWNER, context: invalidContext},
+            ]);
           })
-          .then(function(isInRole) {
-            expect(isInRole).to.be.true();
+          .then(function(data) {
+            expect(data).to.eql([true, false]);
```
bajtos: This assertion produces an error message that's difficult to understand without reading the source code of the test. Can you please split the test into two, where each tests asserts a single scenario (valid context, invalid context)?

common/models/role.js
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
 
-    debug('isOwner(): %s %s userId: %s', modelClass && modelClass.modelName, modelId, userId);
+    debug('isOwner(): %s %s userId: %s principalType: %s',
+      modelClass && modelClass.modelName, modelId, userId, principalType);
 
     // No userId is present
     if (!userId) {
       process.nextTick(function() {
-        callback(null, false);
+        callback(null);
```
ebarault: my bad, there's no reason why to change this

common/models/role.js
```diff
@@ -210,65 +210,91 @@ module.exports = function(Role) {
...
+          callback(null, matches(modelId, userId));
+        });
+      }
       return callback.promise;
     }
 
     modelClass.findById(modelId, function(err, inst) {
       if (err || !inst) {
         debug('Model not found for id %j', modelId);
-        if (callback) callback(err, false);
```
ebarault: hmm, too quickly assumed err wasn't ever falsy in that case (overlooked the `!inst`)

common/models/role.js
ebarault: I assume that you already looked at my [comment here](https://github.com/strongloop/loopback/pull/3140#issuecomment-277457162) and you prefer to keep this behavior in loopback 3 despite the comments.

So unless further comments i'll put back this behavior in place when `principalType === 'USER'`

common/models/role.js
ebarault: 👍

common/models/role.js
ebarault: **Please confirm**: I understand that in the following case : 
> 4 queries, one of them fails, two returns information that the user is not owner, and one says that the user is owner

1. you propose we resolve isOwner `truthy`
2. let the app admins know about the error -> log an error msg in the console ?
3. let the app users know about the error -> **how ?**

lib/access-context.js
ebarault: 👍

test/multiple-user-principal-types.test.js
ebarault: 👍

ebarault:
@bajtos : please review my comments so i can finalize the PR

bajtos:
@ebarault sorry for the delay, I read your comments only now. I'll try to review and come back to you by the end of this week.

common/models/role.js
bajtos: I have two concerns about this change, I have expressed them in #3140:

 - backwards compatibility - with the change in place, our ACL system will start accepting request that are rejected in existing LB versions
 - what if a model has a  `belongsTo` relation to a User-like model, but this relation does not create ownership?

Can we leave this change out of scope of this pull request please?

common/models/role.js
bajtos: I am open to discuss how to handle multiple `belongsTo` relations, but for the purpose of this pull request, I prefer to stay focused on fixing the issue created by the previous patch, i.e. "allowing users from different models but with the same id), because the `Role.isOwner()` method was not checking the user's principalType."

common/models/role.js
bajtos: Based on our recent discussions, I prefer to leave this change out of scope of this pull request, so that we don't mix several changes into one patch.

common/models/role.js
bajtos: Sorry for the confusion. What I meant is that your current implementation is good.

 - isOwner fails with an error
 - error-handling middleware reports this error e.g. to console.error
 - the app user sees that their request failed

Unless I am misunderstanding the code? Can you add a unit test executing this scenario and ensuring LoopBack behaves as we expect it to?

Having said that, this discussion is out of scope of this patch if we agree to keep current handling of "belongsTo" relations.

common/models/role.js
```diff
@@ -244,31 +256,39 @@ module.exports = function(Role) {
...
+          if (rel.type !== 'belongsTo' && !isUserClass(rel.modelTo)) {
+            continue;
+          }
+          // checking related user
+          var userModelName = rel.modelTo.modelName;
+          if (principalType === Principal.USER || principalType === userModelName) {
+            debug('Checking relation %s to %s: %j', r, userModelName, rel);
             inst[r](processRelatedUser);
             return;
           }
```
ebarault: we now need this additional check
`if (principalType === Principal.USER || principalType === userModelName) {...`
 to make sure we don't follow a belongTo relation towards an incorrect user model.

**ps**. in order to keep the level of indent controlled, i chose to first `continue` if `rel.type !== 'belongsTo' && !isUserClass(rel.modelTo)`

lib/access-context.js
ebarault: if we do this, then this will break the `getUserId()` method for which is seems normal to return null according to the tests. Should we then modify the getUserId method to return null when getUser() returns null? 
If so, then there's not much need to not return `{}` in the first place

common/models/role.js
```diff
@@ -179,8 +179,10 @@ module.exports = function(Role) {
     }
     var modelClass = context.model;
     var modelId = context.modelId;
-    var userId = context.getUserId();
-    Role.isOwner(modelClass, modelId, userId, callback);
+    var user = context.getUser();
+    var userId = user && user.id;
+    var principalType = user && user.principalType;
+    Role.isOwner(modelClass, modelId, userId, principalType, callback);
   });
```
ebarault: now that context.getUser() can return null, we need to test user before getting user.id and user.principalType

lib/access-context.js
```diff
@@ -138,17 +147,16 @@ AccessContext.prototype.getUserId = function() {
...
     }
 
     // or permit to resolve a valid user model
     var userModel = this.registry.findModel(p.type);
     if (!userModel) continue;
     if (userModel.prototype instanceof BaseUser) {
-      return p.id;
+      return {id: p.id, principalType: p.type};
     }
   }
```
ebarault: `context.getUser()` now returns `null` instead of `{}` to enable developers to more easily detect issues when accessContext misses the user

lib/access-context.js
```diff
@@ -128,6 +128,15 @@ AccessContext.prototype.addPrincipal = function(principalType, principalId, prin
  * @returns {*}
  */
 AccessContext.prototype.getUserId = function() {
+  var user = this.getUser();
+  return user && user.id;
+};
+
```
ebarault: shrinked version of `getUserId()`, leveraging `getUser()`

test/multiple-user-principal-types.test.js
```diff
@@ -373,17 +382,46 @@ describe('Multiple users with custom principalType', function() {
...
+              principalType: AnotherUser.modelName,
+              principalId: userFromOneModel.id,
+              model: Album,
+              id: album.id,
+            };
+            return Role.isInRole(Role.OWNER, invalidContext);
+          })
+          .then(function(isOwner) {
+            expect(isOwner).to.be.false();
           });
```
ebarault: split the OWNER test in 2 to check apart if resolver returns false when principalType is incorrect

ebarault:
@bajtos : i reverted the PR to only cover the required fix as discussed, and did the required modifications following your code review. You'll still have to squash the 2 commits to merge the tests you previously added. Thanks for carrying this over.

common/models/role.js
bajtos: The negation of `rel.type === 'belongsTo' && isUserClass(rel.modelTo)` is `rel.type !== 'belongsTo' || !isUserClass(rel.modelTo)`, I'll fix this myself.

bajtos:
I made few last formatting improvements and landed the patch. Most importantly, I added more details to the commit message to explain the changes, I would appreciate if you could do that yourself in the future, @ebarault.

Thank you for the contribution! 🙇

ebarault:
@bajtos : thanks. Yes i will take care of detailed commit messages myself in the future. See you soon.

Issue #3188, opened 2017-02-06
remote method PersistedModel.diff only receives one parameter [revisited]


mrjrdnthms:
As far as I can tell there is a bug in the way PersistedModel sets up the diff remote method. I think @jouke was corrected when he assessed that "setRemoting is defined for remoteChanges with http: { source: 'body' }, and 'since' is not. I then discovered that strong-remoting only sets one parameter in the request body when it's defined like this."

I have created a pull request to fix this issue. Its a simple fix. The original issue opened by @jouke was closed due to inactivity. I will be response. Need it fixed :)

Original Issue:
[https://github.com/strongloop/loopback/issues/2223](https://github.com/strongloop/loopback/issues/2223)

Pull RequestL
[https://github.com/strongloop/loopback/pull/3189](https://github.com/strongloop/loopback/pull/3189)

bajtos:
Hello, I am afraid I am still not able to reproduce the problem you are trying to solve. Please take a look at  https://github.com/strongloop/loopback/commit/8b8713e34b0baae333d7499828ddd81e078c013 which shows a test that invokes `diff` with `since` argument via the REST API and passes with the current implementation, therefore verifying that `diff` works correctly already.

I believe there is something else wrong in your application. 

To move this discussion forward, I need you to provide a small application reproducing the issue, see http://loopback.io/doc/en/contrib/Reporting-issues.html#bug-report

PR #3189, opened 2017-02-06
setRemoting is defined for remoteChanges with http: { source: 'body' …

mrjrdnthms:
### Description
In persitedModel diff method setRemoting is defined for remoteChanges with http: {source: 'body' }, and 'since' is not. strong-remoting only sets one parameter in the request body when it's defined like this. Fixed this bug by sending both params through form.

#### Related issues

- 3188 [https://github.com/strongloop/loopback/issues/3188](https://github.com/strongloop/loopback/issues/3188)

### Checklist

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -897,9 +897,10 @@ module.exports = function(registry) {
         description: 'Get a set of deltas and conflicts since the given checkpoint.',
         accessType: 'READ',
         accepts: [
-          {arg: 'since', type: 'number', description: 'Find deltas since this checkpoint'},
-          {arg: 'remoteChanges', type: 'array', description: 'an array of change objects',
-            http: {source: 'body'}},
+          { arg: 'since', type: 'number', description: 'Find deltas since this checkpoint',
+            http: { source: 'form' }},
+          { arg: 'remoteChanges', type: 'array', description: 'an array of change objects',
+            http: { source: 'form' }}
         ],
         returns: {arg: 'result', type: 'object', root: true},
         http: {verb: 'post', path: '/diff'},
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -908,7 +908,7 @@ module.exports = function(registry) {
       // set options.explicitSinceSource to true in model config
       // to explicitly set argument source and ensure its passed
       if (options.explicitSinceSource) {
-        diffOptions[0].accepts.http = {source: 'query'};
+        diffOptions.accepts[0].http = {source: 'query'};
       }
       setRemoting(PersistedModel, 'diff', diffOptions);
 
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -893,18 +893,24 @@ module.exports = function(registry) {
     }
 
     if (options.trackChanges || options.enableRemoteReplication) {
-      setRemoting(PersistedModel, 'diff', {
+      var diffOptions = {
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok test test

bajtos:
Hello @mrjrdnthms, thank you for the pull request. I am afraid I am still not able to reproduce the issue you are trying to fix. AFAIK, the `diff` remoting metadata is correct, we have automated tests that are relying on this feature - see https://github.com/strongloop/loopback/blob/304ecc4784c927be69caaa01a1b7344b169f0fa9/test/replication.test.js#L468.

Having said that, I may be wrong. To move this pull request forward, please:

 - sign our CLA at https://cla.strongloop.com/agreements/strongloop/loopback
 - make sure `npm test` passes locally, that should fix the failing Travis CI build too
 - add a test (preferably to [test/model.test.js](https://github.com/strongloop/loopback/blob/304ecc4784c927be69caaa01a1b7344b169f0fa9/test/model.test.js#L779)) that fails with the current implementation in `master` and passes with your fix in place

lib/persisted-model.js
```diff
@@ -897,9 +897,10 @@ module.exports = function(registry) {
         description: 'Get a set of deltas and conflicts since the given checkpoint.',
         accessType: 'READ',
         accepts: [
-          {arg: 'since', type: 'number', description: 'Find deltas since this checkpoint'},
-          {arg: 'remoteChanges', type: 'array', description: 'an array of change objects',
-            http: {source: 'body'}},
+          { arg: 'since', type: 'number', description: 'Find deltas since this checkpoint',
+            http: { source: 'form' }},
```
bajtos: This is a breaking change which we cannot land in `3.x` version.

I am proposing to keep `remoteChanges` mapped to `body`, and explicitly map `since` to a query string argument (I think that was the initial intention anyways). To prevent breaking change in the mapping of the `since` argument, there should be a model setting to control whether the old or the fixed mapping is used.

mrjrdnthms: Thanks for following up @bajtos. I'm a big fan. I've been working with loopback for a few years. After reading so much of your work and posts, actually interacting with you is like talking to a rock star.  :)

It's not clear to me why changing the source for the input parameters is a breaking change. I assumed the method that gets called was agnostic to the source of the parameters and that strong-remoting would handle mapping the parameters appropriately. But maybe thats not the case. Or maybe there is another place in the framework that depends on remoteChanges being mapped to body?

Whatever the case I'm happy to go with your suggestion. I updated the pull request with what I think you are suggesting. The code looks for an option "explicitSinceSource" I signed the CLA and made sure npm test passes. 

I did not add a test to catch the error when diff is called via rest. I am not sure how I would write a test that examines the parameters before they are sent to the server. If you can point me to an example you think demonstrates how I could do this I'd be happy to review and take a shot at it.

bajtos:
Hello, I am afraid I am still not able to reproduce the problem you are trying to solve. Please take a look at  https://github.com/strongloop/loopback/commit/8b8713e34b0baae333d7499828ddd81e078c013 which shows a test that invokes `diff` with `since` argument via the REST API and passes with the current implementation, therefore verifying that `diff` works correctly already.

I am closing this pull request as rejected. Let's move the discussion back to #3188, where we can try to figure out what exactly is the issue you are encountering.

Issue #3121, opened 2017-01-19
Rolemapping only accept principalId where with like


felipetoffolo1:
### Bug/Feature request

After i upgraded Loopback to version 3, This filter o rolemapping returns nothing
`{"where":{"principalId":"5880038427758463a1ae5a3b"}}`
i changed my filter to this
`{"where":{"principalId":{ "like": "5880038427758463a1ae5a3b"}}}`
and it worked. But is it a bug? Or it changed onn the new version? (i did not find anything about it in migration docs)

### Expected result
Expected
filter` {"where":{"principalId":"5880038427758463a1ae5a3b"}`} to return results

Actual
it returns nothing

### Additional information
linux x64 7.4.0

├── loopback@3.2.1
├── loopback-boot@2.23.0
├── loopback-component-explorer@4.0.0
├── loopback-component-push@1.7.0
├── loopback-component-storage@3.0.0
├── loopback-connector-mongodb@3.0.0

alFReD-NSH:
Try converting `"5880038427758463a1ae5a3b"` to object id.

totalard:
How i do that?

luncht1me:
Converting to ObjectID doesn't work. I'm having the same problem.

Trying to find a RoleMapping using principalId in the where query returns nothing.
If I try and find it as a regex of the Id string it works.
principalId is stored as a string in mongodb.

luncht1me:
It's kind of annoying, I had a nice function to return the roles of a user based on their mapping
```javascript
RoleMapping.find({
	where:{
		"principalType": RoleMapping.USER,
		"principalId": userId.toString()
	}
}, function (err, roleMap) {
	console.log("[getRolesForUser]:", err, roleMap);
	if (err || (roleMap && !roleMap.length)) {
		cb("no rolemapping found", false);
	}
	else if (roleMap && roleMap.length) {
		/* Map out the roleIds only. Some reason can't filter fields above */
		var roleIds = roleMap.map(function (r) {
			return r.roleId
		});
		console.log("[getRolesForUser - roleIds]: ", roleIds);
		/* Get our roles that match the id map */
		Role.find({
			where: {
				id: {
					inq: roleIds
				}
			}

		}, function (err, roles) {
			console.log("[Role.find]:", err, roles);
			if (err || !roles.length) {
				cb("no roles found", false);
			} else {
				/* Map out only the names of the roles found */
				var roleNames = roles.map(function (r) {
					return r.name;
				});
[...]

PR #3198, opened 2017-02-09
"strictObjectIDCoercion": true on role-mapping.json

luncht1me:
### Description

How the json file is, loopback developers are currently unable to query RoleMappings by principalId, since most connectors (mongodb in my case) will overwrite the type of the principalId for the query to an ObjectId, which is incorrect behavior.

principalId is a string, as per this definition. So having the query cast to an ObjectId always returns no matches.

#### Related issues

- #3121

### DIFF
--- a/common/models/role-mapping.json
+++ b/common/models/role-mapping.json
@@ -1,6 +1,7 @@
 {
   "name": "RoleMapping",
   "description": "Map principals to roles",
+  "strictObjectIDCoercion": true,
   "properties": {
     "id": {
       "type": "string",


slnode:

slnode:

slnode:

slnode:

slnode:

Casy:

Casy:
This bug it is not just about the roleMapping, all the id searchs fails with mongo

bajtos:
@slnode ok to test

bajtos:
Hello @luncht1me, thank you for the contribution.

What are the consequences of your change for existing applications? I am concerned that this may be a breaking change.

We also require most changes to be accompanied with a test that fails before the change is made and passes afterwards. Can you add one please?

bajtos:
@fabien @Amir-61 @jannyHou the flag "strictObjectIDCoercion" was added by https://github.com/strongloop/loopback-connector-mongodb/pull/297 which you authored/reviewed, could you please take a look at this related pull request too?

@fabien are you using loopback's built-in auth system? How do you fix object-id coercion for models like RoleMapping? Also, is there any reasonably easy way for LoopBack users to enable `strictObjectIDCoercion` flag in built-in models, for example via a boot script?

fabien:
@bajtos this pull request looks fine - I have the following in a boot script right now:

```
RoleMapping.settings.strictObjectIDCoercion = true;
```

AFAIK this is the only built-in model that was affected by this option.

bajtos:
@fabien great! thank you for chiming in.

My main concern is about backward compatibility. If we enable `strictObjectIDCoercion` in LoopBack, can this change break existing applications after they upgrade to a newer LoopBack version?

fabien:
@bajtos I think your concern is valid, as `strictObjectIDCoercion` will also dictate how ID's are written to mongo in the first place. In other words, it can only be enabled for 'fresh' projects, or after a db migration of these ID's in existing applications.

Perhaps a migration script can be provided, though?

bajtos:
I see. In that case, let's enable this setting only for new projects created using `lb`/`slc`/`apic` tools - see https://github.com/strongloop/loopback-workspace/pull/437.

I am closing this patch as rejected.

> Perhaps a migration script can be provided, though?

That would be nice, indeed. I think this belongs to our documentation - instructions on how to enable `strictObjectIDCoercion` in existing projects. That's is out of scope of this patch though.

PR #3214, opened 2017-02-20
TypeError: tokenGenerator is not a function

islamixo:
.../node_modules/loopback/common/models/user.js:433
        tokenGenerator(user, function(err, token) {
        ^

    TypeError: tokenGenerator is not a function
        at ModelConstructor.User.verify (../node_modules/loopback/
        common/models/user.js:433:5)

    This error occurs when you pass the generateVerificationToken
    parameter to the verify function. Assume that you never pass

    Sorry my english is not very good!!!

    EJ:

     user.generateVerificationToken(inst,
          function (err ,verificationToken) {
         ...
         let options = {
            type: 'email',
            to: inst.email,
            ...
            generateVerificationToken: verificationToken,
            ...
          };

          //Here jump the error
          inst.verify(options, function(err, Object) {
          ...

#### Related issues

- None

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -448,18 +448,20 @@ module.exports = function(User) {
     var Email =
       options.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
 
-    // Set a default token generation function if one is not provided
-    let tokenGenerator = new Promise(function(resolve, reject){
-      if (options.generateVerificationToken)
-        resolve(options.generateVerificationToken);        
+     // Set a default token generation function if one is not provided    
+    if(!options.generateVerificationToken){
       User.generateVerificationToken(user, function(err, token) {
-        if (err) { reject(err); }
-        resolve(token);
-      });
-    });
-
-    tokenGenerator.then(function(token){
-      user.verificationToken = token;
+        if (err) { return fn(err); }
+        user.verificationToken = token;
+        saveAndSend(user);
+      })   
+    } else {
+      user.verificationToken = options.generateVerificationToken;
+      saveAndSend(user);
+    }
+  
+    //Save and sendEmail
+    function saveAndSend(user){
       user.save(function(err) {
         if (err) {
           fn(err);
@@ -467,8 +469,7 @@ module.exports = function(User) {
           sendEmail(user);
         }
       });
-    })
-    .catch(function(err){fn(err);});
+    }
 
     // TODO - support more verification types
     function sendEmail(user) {
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -470,7 +470,7 @@ module.exports = function(User) {
       options.verifyHref += '&token=' + user.verificationToken;
 
       options.verificationToken = user.verificationToken;
-      
+
       options.text = options.text || g.f('Please verify your email by opening ' +
         'this link in a web browser:\n\t%s', options.verifyHref);
 
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -448,20 +448,14 @@ module.exports = function(User) {
     var Email =
       options.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
 
-     // Set a default token generation function if one is not provided    
-    if(!options.generateVerificationToken){
-      User.generateVerificationToken(user, function(err, token) {
-        if (err) { return fn(err); }
-        user.verificationToken = token;
-        saveAndSend(user);
-      })   
-    } else {
-      user.verificationToken = options.generateVerificationToken;
-      saveAndSend(user);
-    }
-  
-    //Save and sendEmail
-    function saveAndSend(user){
+    // Set a default token generation function if one is not provided
+    var tokenGenerator = options.generateVerificationToken || User.generateVerificationToken;
+    assert(typeof tokenGenerator === 'function', 'generateVerificationToken must be a function');
+
+    tokenGenerator(user, function(err, token) {
+      if (err) { return fn(err); }
+
+      user.verificationToken = token;
       user.save(function(err) {
         if (err) {
           fn(err);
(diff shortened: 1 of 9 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

common/models/user.js
```diff
@@ -449,11 +449,16 @@ module.exports = function(User) {
       options.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
 
     // Set a default token generation function if one is not provided
-    var tokenGenerator = options.generateVerificationToken || User.generateVerificationToken;
-
-    tokenGenerator(user, function(err, token) {
-      if (err) { return fn(err); }
+    let tokenGenerator = new Promise(function(resolve, reject){
```
raymondfeng: Instead of completely switching to `Promise` based implementation, I would suggest you add a check to test if tokenGenerator is a function to minimize the changes and keep it consistent.

common/models/user.js
```diff
@@ -448,21 +448,28 @@ module.exports = function(User) {
...
-      user.verificationToken = token;
+     // Set a default token generation function if one is not provided    
+    if(!options.generateVerificationToken){
+      User.generateVerificationToken(user, function(err, token) {
+        if (err) { return fn(err); }
+        user.verificationToken = token;
+        saveAndSend(user);
+      })   
+    } else {
+      user.verificationToken = options.generateVerificationToken;
```
raymondfeng: Hmm, do you expect `options.generateVerificationToken` to be the value for `verificationToken`? If so, that is wrong. `options.generateVerificationToken` should be a JS function that is capable of generating verification tokens for a user.

bajtos:
>  options.generateVerificationToken should be a JS function that is capable of generating verification tokens for a user.

Exactly.

I think we should simply add an assert to verify that `tokenGenerator` is a function.

```js
var tokenGenerator = options.generateVerificationToken || User.generateVerificationToken;
assert(typeof tokenGenerator === 'function', 'generateVerificationToken must be a function');
```

islamixo:
If I explained to @raymondfeng, because I thought it was better than verify () wait for a token and not a fuction, because although it returns the call it automatically sends the mail, when you need to build the confirmation url that goes in the email, for example for Confirm from (mobile and web).

I think if the de developer generates the token it is to manipulate it, build the url, and send it to the template, in case of not passing it loopback generates it to me that looks great.

islamixo:
@bajtos  true with this the TypeError is corrected.  

@raymondfeng for the case I need to generate a own token and build several url to send by email. I just have to generate the token and pass it on the CB, it's not very intuitive for me, but it works.

verificationToken = Token we've generated:
      
      let linkHref = CONFIG.verifyHref + '/#/confirm?uid=' + inst.id;
      let linkMovil = CONFIG.verifyHref + "/other.html?uid=" + inst.id + ";token=" + verificationToken;
      ....
      
     //Here we pass our token in fuction cb
      let tokenGenerator = function (inst, next) { 
        next(err, verificationToken);
      };

      let options = {
        ....
        //Passing tonken inside the function
        generateVerificationToken: tokenGenerator,
        verifyHref: linkHref,
        verifyMovil: linkMovil,
        ....
      };      
      
      //Enviando Mensaje de  Verificación
      inst.verify(options, function(err, Object) {
      ....


Thank you very much!!!  for the correction and for clarifying my doubt!!!

bajtos:
@islamixo thank you for the detailed explanation of your use case!

If I understand it correctly, your email template needs two URLs (`verifyHref`, `verifyMovil`) containing the generated verification token.

In that case, I am proposing:

1. Modify the User model to include the generated token in the options passed to template function:

    ```diff
      function sendEmail(user) {
        options.verifyHref += '&token=' + user.verificationToken;
    +   options.verificationToken = user.verificationToken
        // ...
      }
    ```

2. In your application, provide a custom `options.templateFn` that will add any additional template options needed:

    ```js
    user.verify({
      // ...
      templateFn: createVerificationEmail,
    });

    function createVerificationEmail(options, cb) {
      options.verifyMovil = CONFIG.verifyHref + "/other.html?uid=" + options.user.id + ";token=" + verificationToken;

      // copied from LoopBack's "createVerificationEmailBody()"
      var template = loopback.template(options.template);
      var body = template(options);
      cb(null, body);
    }

    ```


Thoughts?

islamixo:
Perfect, this way I have complete control of the process, solved for me. Thank you!!

bajtos:
@islamixo great!

Can you rework this pull request to implement "Modify the User model to include the generated token in the options passed to template function" from my proposal then?

islamixo:
Done!!

bajtos:
@slnode test please

islamixo:
@bajtos  Done!!, in the test I check that the token generated by loopback is being passed to the function. I did some local mail sending tests, and generateVerificationToken. Everything works for me correctly with npm test. 

If you think it necessary, testify in another way, you tell me. I am happy to contribute. Thank you.

test/user.test.js
```diff
@@ -1714,6 +1714,30 @@ describe('User', function() {
...
+        return User.create({email: 'test1@example.com', password: 'pass'})
+          .then(user => {
+            let newVerifyUrl;
+            return user.verify({
+              type: 'email',
+              to: user.email,
+              from: 'noreply@myapp.org',
+              redirect: '#/some-path?a=1&b=2',
+              templateFn: (options, cb) => {
+                options.verifyMovil = '/redirect.html?uid=' + user.id + ';token=' + options.verificationToken;
```
bajtos: I would prefer to simplify this test - the purpose is to verify that `options.templateFn` receives `options.verificationToken`.

See the test above for an example.

```js
// ...
templateFn: (options, cb) => {
  actualVerificationToken = options.actualVerificationToken;
  cb(null, 'dummy body');
},
// ...
.then(token => {
  expect(token).to.exist();
});
```

bajtos:
Could you please rebase your patch on top of the latest master too?

islamixo:
Please delete this pull request, I have created a much cleaner one in # 3249

Issue #3215, opened 2017-02-20
Problems extending User model with loopback v3.3.0


greaterweb:
# Description/Steps to reproduce
1. Scaffold a new loopback project using the CLI (eg. `lb`)
2. In `server/model-config.json` find the builtin model reference:
```
  "User": {
    "dataSource": "db"
  },
```

and replace with the custom model reference:
```
  "user": {
    "dataSource": "db"
  },
```

**the `user` name for the custom model as it is used as seen in other examples, I will retest with a custom model name as well**

3. Create a new model setting `User` as the `base` value (for example see [loopback-example-access-control](https://github.com/strongloop/loopback-example-access-control/blob/da503b0ee18f1b7ebc18b7e1764c73d6ca3ca6fe/common/models/user.json#L3)), an overly simplified example might look like:

``` json
{
  "name": "user",
  "base": "User",
  "hidden": ["password", "verificationToken"]
}

```
4. Create a boot script (eg. `/server/boot/setup.js`) with the following

``` js
'use strict';

module.exports = function(app) {
  var User = app.models.user;

  User.create({
    username: 'foo',
    email: 'foo@bar.com',
    password: 'changeme'
  }, (err, user) => {
    if (err) {
      throw err;
    }

    user.email = 'bar@foo.com';
    user.save((saveErr, user) => {
      if (saveErr) {
        throw saveErr;
      }
      // won't make it here, error is thrown
      // TypeError: Cannot read property 'polymorphic' of undefined
    });
  });
};
```

5. Run the loopback application

The following will produce an error such as:

``` sh
/code/loopback-sample/node_modules/loopback/common/models/user.js:695
    var isRelationPolymorphic = relatedUser.polymorphic && !relatedUser.modelTo;
                                           ^

TypeError: Cannot read property 'polymorphic' of undefined
    at Function.User._invalidateAccessTokensOfUsers (/code/loopback-sample/node_modules/loopback/common/models/user.js:695:44)
    at invalidateOtherTokens (/code/loopback-sample/node_modules/loopback/common/models/user.js:927:15)
    at notifySingleObserver (/code/loopback-sample/node_modules/loopback-datasource-juggler/lib/observer.js:104:22)
    at /code/loopback-sample/node_modules/async/dist/async.js:3025:16
    at replenish (/code/loopback-sample/node_modules/async/dist/async.js:881:17)
    at /code/loopback-sample/node_modules/async/dist/async.js:885:9
[...]

PR #3227, opened 2017-02-23
Fix access-token invalidation for missing relation

bajtos:
### Description

Fix the code invalidating access tokens on user email/password changes to correctly handle the case when the relation "AccessToken belongs to (subclassed) user" is not configured.

cc @ebarault @greaterweb 

#### Related issues

- #3215

### Checklist

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -2370,6 +2370,28 @@ describe('User', function() {
       });
     });
 
+    // See https://github.com/strongloop/loopback/issues/3215
+    it('handles subclassed user with no accessToken relation', () => {
+      // setup a new LoopBack app, we don't want to use shared models
+      app = loopback({localRegistry: true, loadBuiltinModels: true});
+      app.set('remoting', {errorHandler: {debug: true, log: false}});
+      app.dataSource('db', {connector: 'memory'});
+      const User = app.registry.createModel({
+        name: 'user',
+        base: 'User',
+      });
+      app.model(User, {dataSource: 'db'});
+      app.enableAuth({dataSource: 'db'});
+      expect(app.models.User.modelName).to.eql('user');
+
+      return User.create(validCredentials)
+        .then(u => {
+          u.email = 'updated@example.com';
+          return u.save();
+          // the test passes when save() does not throw any error
+        });
+    });
+
     function assertPreservedTokens(done) {
       AccessToken.find({where: {userId: user.pk}}, function(err, tokens) {
         if (err) return done(err);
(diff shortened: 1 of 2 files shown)


ebarault:
👍

test/user.test.js
```diff
@@ -2370,6 +2370,28 @@ describe('User', function() {
...
+      });
+      app.model(User, {dataSource: 'db'});
+      app.enableAuth({dataSource: 'db'});
+      expect(app.models.User.modelName).to.eql('user');
+
+      return User.create(validCredentials)
+        .then(u => {
+          u.email = 'updated@example.com';
+          return u.save();
+          // the test passes when save() does not throw any error
```
superkhau: I think we should use two arg style: http://loopback.io//doc/en/contrib/style-guide.html#asserting-a-rejected-promise

superkhau: You wrote that entry though so you would know better. ;) What I'm suggesting is we should have something that catches and calls `expect.fail()` instead of the comment.

greaterweb:
Good work on addressing this, @bajtos! 

As far as this PR goes, I've tested this branch in my [sample repo](https://github.com/greaterweb/loopback-issue-3215/tree/dev/lb-branch-fix) and another Loopback project of mine, in both cases the app is no longer crashing and the issue raised is resolved. I'm going to add some more discussions though to the original ticket as I want to clarify a few points on best practices.

test/user.test.js
bajtos: >  What I'm suggesting is we should have something that catches and calls expect.fail() instead of the comment.

By calling `expect.fail`, the test would hide the original error and report an unhelpful failure.

Note that we are asserting a resolved promise here, not a rejected promise as is described in the style guide link you pasted.

I suppose I could rewrite this test as follows, would you find it easier to read?

```js
.then(u => {
  u.email = 'updated@example.com';
  return u.save();
})
.then(u => {
  // the test passes when save() does not throw any error
});
```

bajtos:
@ebarault @raymondfeng @superkhau thank you for the review!

@greaterweb thank you for verifying my fix in your codebase. Let's get it landed to address the immediate problem and then continue the discussion in the original issue.

bajtos:
Released in `loopback@3.4.0`

bajtos:
@ebarault 
> not clear to me what part from the multiuser patch specifically broke the default accesstoken --> user belongsto relation inheritance from base User?

It's not your patch that break the relation, it's the application that did not correctly configure this relation. Your patch was just not prepared to handle such misconfiguration, which I fixed here.

test/user.test.js
superkhau: A bit better, but not worth the effort. LGTM. ;)

PR #3211, opened 2017-02-18
fixed context passing in role resolver with `isOwner`

DaGaMs:
### Description

With the move to explicit context passing of an `options` object, internal calls to `find` etc need to be updated to pass the necessary information along, too. In this case, `isOwner` calls `findById` on a Model, but does not pass an access token in `options`.
#### Related issues

- #3209 
- #3023 

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

Issue #3209, opened 2017-02-17
findById does not include "options", but find does?


DaGaMs:
# Description/Steps to reproduce
According to the new remoting context injection #3023 I expect to always get an `options` property in the current context of an operation hook. I have a mixin that will restrict the requested object to the current owner (and in the future other team members, so I can't use `$owner`):

```js
module.exports = function(Model, options) {
  Model.observe('access', function limitToTenant(ctx, next) {
    var userId = ctx.options.accessToken.userId;
    if (!userId) {
      console.error("No user in context ", ctx.options);
      var error = new Error("Not authorized");
      error.statusCode = 401;
      next(error);
    }
    if (_.has(ctx,'query')) {
      if (_.has(ctx.query, 'where')) {
        var old = ctx.query.where;
        ctx.query.where = {and: [old, {ownerId: userId}]};
      }
      else {
        ctx.query.where = {ownerId: userId};
      }
    }
    next();
  });
};
```

This works just fine when I do a request like this: `/api/projects?filter[where][id]=1`, but when I do a request like this: `/api/projects/1`, the `options` parameter is empty.

# Expected result
I expect invocation of `find` and `findById` to both propagate the options to operation hooks.

# Additional information
Here is the model definition file:

```json
{
  "name": "project",
  "base": "PersistedModel",
  "strict": true,
  "idInjection": true,
  "replaceOnPUT": false,
  "options": {
    "validateUpsert": true
  },
  "mixins": {
    "AccessControl": {}
  },
  "mysql": {
    "schema": "testdb",
    "table": "Project"
  },
  "properties": {
    "lastUpdated": {
      "type": "date",
      "required": true,
      "length": null,
      "dataType": "timestamp",
      "nullable": "N"
    },
    "created": {
      "type": "date",
      "required": true,
      "index": -1,
      "dataType": "datetime",
      "nullable": "N" 
    },
    "name": {
      "type": "string",
      "required": false,
      "length": 256,
      "index": 1,
      "nullable": "Y"
    }
  },
  "validations": [],
  "relations": {
    "owner": {
      "model": "Person",
      "type": "belongsTo"
    }
  },
  "acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$authenticated",
      "permission": "ALLOW",
      "property": "count"
    },
    {
      "accessType": "*",
[...]

PR #3230, opened 2017-02-24
Fix context passing in OWNER role resolver

bajtos:
### Description

With the move to explicit context passing of an `options` object, internal calls to `find` etc need to be updated to pass the necessary information along, too. In this case, `isOwner` calls `findById` on a Model, but does not pass an access token in `options`.

cc @DaGaMs @ebarault 

#### Related issues

- #3209 
- #3023 
- supersede #3211

### Checklist

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -182,7 +182,8 @@ module.exports = function(Role) {
     var user = context.getUser();
     var userId = user && user.id;
     var principalType = user && user.principalType;
-    Role.isOwner(modelClass, modelId, userId, principalType, callback);
+    var opts = {accessToken: context.accessToken};
+    Role.isOwner(modelClass, modelId, userId, principalType, opts, callback);
   });
 
   function isUserClass(modelClass) {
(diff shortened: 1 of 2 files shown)


common/models/role.js
```diff
@@ -213,15 +214,21 @@ module.exports = function(Role) {
...
    * @param {String|Error} err The error string or object
    * @param {Boolean} isOwner True if the user is an owner.
    * @promise
    */
-  Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, callback) {
-    if (!callback && typeof principalType === 'function') {
+  Role.isOwner = function isOwner(modelClass, modelId, userId, principalType, options, callback) {
+    if (!callback && typeof options === 'function') {
+      callback = options;
+      options = {};
```
bajtos: This branch is not covered by any test, coveralls reports "Coverage decreased (-0.04%) to 89.558%" as a result. I think we can ignore that, but please let me know if you disagree! 

(The solution is to add a test that explicitly calls `Role.isOwner(modelClass, modelId, userId, principalType, callback)`.)

ebarault:
sorry to wake after the battle @bajtos 
i see [here](https://github.com/strongloop/loopback/pull/3230/files#diff-733982793c3bd5e61d1a4350451fb494R185) that you fix the content of passed options to solely the accessContext (which i agree is the base configuration) when registering the $owner role resolver.
What if a user wants extra customization on top of that to get extra parameters in the passed options? Should he then override the Role model to specify his own way of registering this role resolver?

PR #3231, opened 2017-02-24
Warn on misconfigured access token user

bajtos:
### Description

Verify User and AccessToken relations at startup Modify `app.enableAuth()` to verify that (custom) User and AccessToken models have correctly configured their hasMany/belongsTo relations and print a warning otherwise.

As I was fixing the tests reporting this new warning, I discovered few other issues that I needed to fix first - see the first three commits  abf8246, c45954c and f0c9700.

The warning and the tests are fixed by the last commit f0d6416.

I think we should wait with merging this pull request until there is a doc page written where we explain users how to fix this problem, so that we can add URL of this doc page to the new warnings.

cc @ebarault @greaterweb

#### Related issues

- #3215

### Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -398,7 +398,86 @@ app.enableAuth = function(options) {
     }
   };
 
+  this._verifyAuthModelRelations();
+
   this.isAuthEnabled = true;
+};
+
+app._verifyAuthModelRelations = function() {
+  // Allow unit-tests (but also LoopBack users) to disable the warnings
+  if (this.get('_verifyAuthModelRelations') === false) return;
+
+  const AccessToken = this.registry.findModel('AccessToken');
+  const User = this.registry.findModel('User');
+  this.models().forEach(Model => {
+    if (Model === AccessToken || Model.prototype instanceof AccessToken) {
+      scheduleVerification(Model, verifyAccessTokenRelations);
+    }
+
+    if (Model === User || Model.prototype instanceof User) {
+      scheduleVerification(Model, verifyUserRelations);
+    }
+  });
+
+  function scheduleVerification(Model, verifyFn) {
+    if (Model.dataSource) {
+      verifyFn(Model);
+    } else {
+      Model.on('attached', () => verifyFn(Model));
+    }
+  }
+
+  function verifyAccessTokenRelations(Model) {
+    const belongsToUser = Model.relations && Model.relations.user;
+    if (belongsToUser) return;
+
+    const relationsConfig = Model.settings.relations || {};
+    const userName = (relationsConfig.user || {}).model;
+    if (userName) {
+      console.warn(
+        'The model %j configures "belongsTo User-like models" relation ' +
+          'with target model %j. However, the model %j is not attached to ' +
+          'the application and therefore cannot be used by this relation. ' +
+          'This typically happens when the application has a custom ' +
+          'custom User subclass, but does not fix AccessToken relations ' +
+          'to use this new model.\n' +
+          'Learn more at http://ibm.biz/setup-loopback-auth',
+        Model.modelName, userName, userName);
+      return;
+    }
+
+    console.warn(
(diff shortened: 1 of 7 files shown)


lib/builtin-models.js
```diff
@@ -52,8 +55,33 @@ module.exports = function(registry) {
...
     var Model = registry.createModel(definitionJson);
     customizeFn(Model);
     return Model;
   }
 };
+
+// Because we are cloning objects created by JSON.parse,
+// the cloning algorithm can stay much simpler than a general-purpose
+// "cloneDeep" e.g. from lodash.
+function cloneDeepJson(obj) {
```
raymondfeng: I think it should be fine to use `lodash` as it's one-time charge.

greaterweb:
Nice work on this, @bajtos!

I've tested this branch in the sample repo I was working with for #3215. I do get the warning for 2 different scenarios I tried:

 - [custom `Account` model extending the `User` model](https://github.com/greaterweb/loopback-issue-3215/tree/dev/warn-on-misconfigured-custom-account-model)
 - [custom `user` model extending the `User` model](https://github.com/greaterweb/loopback-issue-3215/tree/dev/warn-on-misconfigured-custom-user-model)

In both cases the `User` model is not included in the `model-config.json` file.

The unusual thing though is in my other personal project I've been testing against, I don't get the warning message when using the same `warn-on-misconfigured-accessToken-user` branch. The project is consistent with the second scenario above, a custom `user` model is used. I don't configure any relationships with the `AccessToken` model so it's a little confusing why it's not triggering the warning message. 

What I suspect is somewhere in my customization of my `user` model Loopback is being fooled into thinking my model is the same as the `User` model. This is consistent with the fact all of my access control related tests are passing even though a relationship with `AccessToken` is not configured. I'll investigate this further and report back.

lib/builtin-models.js
bajtos: We are trying to avoid using `lodash` in our runtime code in order to keep the size of browserified bundle small (see https://github.com/strongloop/loopback/issues/989). lodash contributes about 350kb unminified and increases the bundle size by approx 20% - see https://github.com/strongloop/loopback-datasource-juggler/pull/538.

bajtos:
@raymondfeng @superkhau @alshamiri1 thank you for the review!

@greaterweb 
> What I suspect is somewhere in my customization of my `user` model Loopback is being fooled into thinking my model is the same as the `User` model. This is consistent with the fact all of my access control related tests are passing even though a relationship with `AccessToken` is not configured. I'll investigate this further and report back.

AFAIK, we don't have any code in LoopBack that is depending on `AccessToken belongsTo User` relation (after #3227 was landed). I would be surprised if your access control tests were failing because of this missing relation, as long as you are not relying on this relation in your test code by calling `accessToken.user(function(err, u) {...})`.

> The unusual thing though is in my other personal project I've been testing against, I don't get the warning message when using the same warn-on-misconfigured-accessToken-user branch.

That's suspicious, I agree.

lib/application.js
```diff
@@ -398,9 +398,88 @@ app.enableAuth = function(options) {
...
+    const userName = (relationsConfig.user || {}).model;
+    if (userName) {
+      console.warn(
+        'The model %j configures "belongsTo User-like models" relation ' +
+          'with target model %j. However, the model %j is not attached to ' +
+          'the application and therefore cannot be used by this relation. ' +
+          'This typically happens when the application has a custom ' +
+          'custom User subclass, but does not fix AccessToken relations ' +
+          'to use this new model.\n' +
+          'Learn more at http://ibm.biz/setup-loopback-auth',
```
bajtos: @crandmck I setup `http://ibm.biz/setup-loopback-auth` shortcut to point to `http://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#preparing-access-control-models`. If/when you move the setup instructions to a different page, then we need to change the target url of the shortcut `http://ibm.biz/setup-loopback-auth` points to.

PR #3265, opened 2017-03-09
Validate User model config at app boot

ebarault:
### Description
**Add** a `hasMultipleUserModelsConfig` property on loopback application to reflect the User setup throughout the application
**Perform sanity check on the go on user model config** to detect and warn the user (and eventually stop the application in case the config is not workable)

This continues the work started by @bajtos in https://github.com/strongloop/loopback/pull/3231
First draft provided for general discussion.

#### Related issues
https://github.com/strongloop/loopback/pull/3140 (warn on misconfigured user/accessToken model relations)
https://github.com/strongloop/loopback/pull/3231 (enable multiple owner $owner resolving)
https://github.com/strongloop/loopback/pull/2971 (enable loopback to handle multiple user models)

### Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -19,6 +19,7 @@ var classify = require('underscore.string/classify');
 var camelize = require('underscore.string/camelize');
 var path = require('path');
 var util = require('util');
+var Promise = require('bluebird');
 
 /**
  * The `App` object represents a Loopback application.
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
 };
 
 app._verifyAuthModelRelations = function() {
+  let self = this;
+
   // Allow unit-tests (but also LoopBack users) to disable the warnings
-  if (this.get('_verifyAuthModelRelations') === false) return;
+  const warnOnBadSetup = self.get('_disableAuthModelRelationsWarnings') !== true;
+  // Prevent the app from being killed although the configuration is inconsistent
+  const dismissRejection = self.get('_dismissBadUserConfigRejection') !== true;
 
-  const AccessToken = this.registry.findModel('AccessToken');
-  const User = this.registry.findModel('User');
-  this.models().forEach(Model => {
-    if (Model === AccessToken || Model.prototype instanceof AccessToken) {
-      scheduleVerification(Model, verifyAccessTokenRelations);
-    }
+  const AccessToken = self.registry.findModel('AccessToken');
+  const User = self.registry.findModel('User');
+  const models = self.models();
 
-    if (Model === User || Model.prototype instanceof User) {
-      scheduleVerification(Model, verifyUserRelations);
-    }
+  verifyUserModelsSetup().then(()=> {
+    verifyRelationsSetup();
   });
 
+  function verifyUserModelsSetup() {
+    let userModels = models.filter(model => {
+      return (model === User || model.prototype instanceof User) ||
+        (model === AccessToken || model.prototype instanceof AccessToken);
+    });
+
+    // we use Promise.reflect() here to let all the tests go without throwing,
+    // this way we can get all the logs for all models we are checking
+    return Promise.map(userModels, (model) => {
+      return scheduleVerification(model, hasMultipleUserModelsConfig).reflect();
+    })
+    .then((inspections) => {
+      // proceed to next checkups if no error
+      let hasErrors = inspections.filter(inspection => {
+        return inspection.isRejected();
+      });
+      if (!hasErrors.length) {
+        return;
+      }
+
+      // else log and eventually kill the app
+      console.warn('Application setup is inconsistent: some models are set to use ' +
+        'the Multiple user models configuration while some other models are set to ' +
+        'use the Single user model configuration. The model config is listed below.');
+
+      // detailed logs
+      inspections.forEach(inspection => {
+        let modelSetup = (inspection.value() || inspection.reason()).modelSetup;
+        console.warn(
+          'Model %j of type %j is set to use the %j user model config',
+          modelSetup.model,
+          modelSetup.instanceOf,
+          modelSetup.hasMultipleUserModelsConfig ? 'Multiple' : 'Single'
+        );
+      });
+
+      delete self.hasMultipleUserModelsConfig;
+
+      if (!dismissRejection) {
+        throw new Error('Application setup is inconsistent, please see the log for more ' +
+          'information');
+      }
+    });
+  }
+
+  function verifyRelationsSetup() {
+    models.forEach(Model => {
+      if (Model === AccessToken || Model.prototype instanceof AccessToken) {
+        scheduleVerification(Model, verifyAccessTokenRelations);
+      }
+
+      if (Model === User || Model.prototype instanceof User) {
+        scheduleVerification(Model, verifyUserRelations);
+      }
+    });
+  }
+
+  // instant or scheduled Model verifications, as a promise
   function scheduleVerification(Model, verifyFn) {
-    if (Model.dataSource) {
-      verifyFn(Model);
-    } else {
-      Model.on('attached', () => verifyFn(Model));
+    return new Promise((resolve, reject) => {
+      if (Model.dataSource) {
+        try {
+          resolve(verifyFn(Model));
+        } catch (e) {
+          reject(e);
+        }
+      } else {
+        Model.on('attached', () => {
+          scheduleVerification(Model, verifyFn);
+        });
+      }
+    });
+  }
+
+  function hasMultipleUserModelsConfig(model) {
+    let hasMultipleUserModelsConfig, isInstanceOfUser;
+
+    // check is the model is set to use mutiple users config (with polymorphic relations)
+    if (model === User || model.prototype instanceof User) {
+      isInstanceOfUser = true;
+      const hasManyTokens = model.relations && model.relations.accessTokens;
+      hasMultipleUserModelsConfig = !!hasManyTokens.polymorphic;
+    } else if (model === AccessToken || model.prototype instanceof AccessToken) {
+      const belongsToUser = model.relations && model.relations.user;
+      // the test on belongsToUser is required as we allow AccessToken model not
+      // to define the relation with custom user model for backward compatibility
+      // see https://github.com/strongloop/loopback/pull/3227
+      hasMultipleUserModelsConfig = !!(belongsToUser && belongsToUser.polymorphic);
     }
+
+    let modelSetup = {
+      model: model.modelName,
+      instanceOf: isInstanceOfUser ? 'User' : 'AccesToken',
+      hasMultipleUserModelsConfig,
+    };
+
+    // check if model config is consistent with already parsed models' config
+    if (self.hasMultipleUserModelsConfig === undefined) {
+      self.hasMultipleUserModelsConfig = hasMultipleUserModelsConfig;
+    } else if (self.hasMultipleUserModelsConfig !== hasMultipleUserModelsConfig) {
+      let error = new Error('User models setup is inconsistent');
+      error.modelSetup = modelSetup;
+      throw error;
+    }
+
+    // if no error, return the modelSetup for further logging purposes
+    return {modelSetup};
   }
 
   function verifyAccessTokenRelations(Model) {
+    if (!warnOnBadSetup) return;
+
     const belongsToUser = Model.relations && Model.relations.user;
     if (belongsToUser) return;
 
@@ -454,6 +558,8 @@ app._verifyAuthModelRelations = function() {
   }
 
   function verifyUserRelations(Model) {
+    if (!warnOnBadSetup) return;
+
     const hasManyTokens = Model.relations && Model.relations.accessTokens;
     if (hasManyTokens) return;
 
--- a/lib/application.js
+++ b/lib/application.js
@@ -399,26 +399,25 @@ app.enableAuth = function(options) {
     }
   };
 
-  this._verifyAuthModelRelations();
-
-  this.isAuthEnabled = true;
+  return this._verifyAuthModelRelations(app._warnOnBadAuthModelRelations)
+    .then(() => {
+      this.isAuthEnabled = true;
+    });
 };
 
-app._verifyAuthModelRelations = function() {
+app._verifyAuthModelRelations = function(warnFn) {
   let self = this;
 
   // Allow unit-tests (but also LoopBack users) to disable the warnings
-  const warnOnBadSetup = self.get('_disableAuthModelRelationsWarnings') !== true;
+  const warnOnBadSetup = this.get('_warnOnBadAuthModelsSetup') !== false;
   // Prevent the app from being killed although the configuration is inconsistent
-  const dismissRejection = self.get('_dismissBadUserConfigRejection') !== true;
+  const abortOnBadUserSetup = this.get('_abortOnBadUserSetup') !== false;
 
-  const AccessToken = self.registry.findModel('AccessToken');
-  const User = self.registry.findModel('User');
-  const models = self.models();
+  const AccessToken = this.registry.findModel('AccessToken');
+  const User = this.registry.findModel('User');
+  const models = this.models();
 
-  verifyUserModelsSetup().then(()=> {
-    verifyRelationsSetup();
-  });
+  return verifyUserModelsSetup().then(verifyRelationsSetup);
 
   function verifyUserModelsSetup() {
     let userModels = models.filter(model => {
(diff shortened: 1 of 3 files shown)


lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
-    }
+  const AccessToken = self.registry.findModel('AccessToken');
+  const User = self.registry.findModel('User');
+  const models = self.models();
 
-    if (Model === User || Model.prototype instanceof User) {
-      scheduleVerification(Model, verifyUserRelations);
-    }
+  verifyUserModelsSetup().then(()=> {
+    verifyRelationsSetup();
```
ebarault: general idea of why not make the whole two together:
1. first we detect the user models config status
2. then we can use it in verifyRelationsSetup to even go further in the bad config detection since things get to change according to one config or the other

ebarault: general idea of why not make the whole two together:
1. first we detect the user models config status
2. then we can use it in verifyRelationsSetup to even go further in the bad config detection since things get to change according to one config or the other

lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
+        } catch (e) {
+          reject(e);
+        }
+      } else {
+        Model.on('attached', () => {
+          scheduleVerification(Model, verifyFn);
+        });
+      }
+    });
+  }
```
ebarault: i needed to change @bajtos first implem to add async support as we now expect on this function to resolve when done, so we can use it to first check user models config and then check how the relations are set up

lib/application.js
bajtos: Sounds good.

Code-wise, this line can be simplified to `verifyUserModelsSetup.then(verifyRelationsSetup)`.

lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
+  function verifyUserModelsSetup() {
+    let userModels = models.filter(model => {
+      return (model === User || model.prototype instanceof User) ||
+        (model === AccessToken || model.prototype instanceof AccessToken);
+    });
+
+    // we use Promise.reflect() here to let all the tests go without throwing,
+    // this way we can get all the logs for all models we are checking
+    return Promise.map(userModels, (model) => {
+      return scheduleVerification(model, hasMultipleUserModelsConfig).reflect();
```
bajtos: This can be simplified (although I am not sure if the line won't become too long):

```js
return Promise.map(userModels, 
  model => scheduleVerification(model, hasMultipleUserModelsConfig).reflect());

File: lib/application.js [hunk 4]
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
+    })
+    .then((inspections) => {
+      // proceed to next checkups if no error
+      let hasErrors = inspections.filter(inspection => {
+        return inspection.isRejected();
```
bajtos: ditto:

```js
inspections.filter(inspection => inspection.isRejected());
```

lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
+        let modelSetup = (inspection.value() || inspection.reason()).modelSetup;
+        console.warn(
+          'Model %j of type %j is set to use the %j user model config',
+          modelSetup.model,
+          modelSetup.instanceOf,
+          modelSetup.hasMultipleUserModelsConfig ? 'Multiple' : 'Single'
+        );
+      });
+
+      delete self.hasMultipleUserModelsConfig;
```
bajtos: Since this is in an arrow function `=>`, you should be able to use `this`. Arrow functions preserve the value of `this` reference.

lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
+
+      if (!dismissRejection) {
```
bajtos: I think this is another example of a double-negative.

lib/application.js
```diff
@@ -404,30 +405,133 @@ app.enableAuth = function(options) {
...
   function scheduleVerification(Model, verifyFn) {
-    if (Model.dataSource) {
-      verifyFn(Model);
-    } else {
-      Model.on('attached', () => verifyFn(Model));
+    return new Promise((resolve, reject) => {
+      if (Model.dataSource) {
+        try {
+          resolve(verifyFn(Model));
+        } catch (e) {
```
bajtos: I believe the Promise constructor is required to catch errors thrown by the factory function, therefore it shouldn't be needed to catch them ourselves.

ebarault: no, i checked. If i just do `resolve(veryFn(Model))`
then the verifyFn throws an error, the following error will be caught when trying to reflect the inspection result:

`TypeError: cannot get fulfillment value of a non-fulfilled promise`

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
     }
   };
 
-  this._verifyAuthModelRelations();
-
-  this.isAuthEnabled = true;
+  return this._verifyAuthModelRelations(app._warnOnBadAuthModelRelations)
+    .then(() => {
+      this.isAuthEnabled = true;
+    });
```
ebarault: as `_verifyAuthModelRelations` is now async and returns a promise, this change is required
also i made enableAuth() return a promise to, as it's de facto async too now, it's not a breaking change to me, the method can still be called in an app while not caring about the returned promise if most if not all use cases.

This is useful in the tests so we can wait for enableAuth to return before checking the  displayed warnings

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
...
 
-app._verifyAuthModelRelations = function() {
+app._verifyAuthModelRelations = function(warnFn) {
+  let self = this;
+
   // Allow unit-tests (but also LoopBack users) to disable the warnings
-  if (this.get('_verifyAuthModelRelations') === false) return;
+  const warnOnBadSetup = this.get('_warnOnBadAuthModelsSetup') !== false;
+  // Prevent the app from being killed although the configuration is inconsistent
+  const abortOnBadUserSetup = this.get('_abortOnBadUserSetup') !== false;
```
ebarault: reworded the flags to avoid double negations

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
...
-          'with target model %j. However, the model %j is not attached to ' +
-          'the application and therefore cannot be used by this relation. ' +
-          'This typically happens when the application has a custom ' +
-          'custom User subclass, but does not fix AccessToken relations ' +
-          'to use this new model.\n' +
-          'Learn more at http://ibm.biz/setup-loopback-auth',
-        Model.modelName, userName, userName);
+    const modelTo = (relationsConfig.user || {}).model;
+    if (modelTo) {
+      warnFn('CUSTOM_USER_MODEL_NOT_AVAILABLE', {modelFrom, modelTo});
```
ebarault: i suggest changing the var names to modelFrom and modelTo, rather than modelName and userName, as it makes the tests more readable.
if you'd like to keep the original names, i will map the original names so the tests are readable

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
...
+    const modelFrom = Model.modelName;
     const relationsConfig = Model.settings.relations || {};
-    const accessTokenName = (relationsConfig.accessTokens || {}).model;
-    if (accessTokenName) {
-      console.warn(
+    const modelTo = (relationsConfig.accessTokens || {}).model;
+    if (modelTo) {
+      warnFn('CUSTOM_ACCESS_TOKEN_MODEL_NOT_AVAILABLE', {modelFrom, modelTo});
+      return;
+    }
```
ebarault: dito

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
...
+
+    warnFn('MISSING_RELATION_TO_ACCESS_TOKEN', {modelFrom});
+  }
+};
+
+// function warnOnBadAuthModelRelations(code, args) {
+app._warnOnBadAuthModelRelations = function(code, args) {
```
ebarault: built-in warning function ("private")
i actually need it to be attached on the app object so i can swap it in the tests when calling `enableAuth()`

lib/application.js
bajtos: I am quite concerned about this change. We use `app.isAuthEnabled` in other places (e.g. https://github.com/strongloop/loopback/blob/c3ab9320127cb7726e83b9668f641f7cb2a55692/server/middleware/rest.js#L51-L54), but possibly in any 3rd party LoopBack code to detect whether the app has authentication enabled. While our REST middleware makes this check few ticks later (at which time this promise chain has already finished), I think it's not safe to assume that everybody else is doing the same.

Having said that, I agree that making `enableAuth` async & promise returning is a good direction.

What is your reasoning for enabling `isAuthEnabled` only after auth-model relations have been verified? What would happen if we keep only `_verifyAuthModelRelations` async and set `app.isAuthEnabled` right on the first tick, as we are doing now?

BTW in the current proposal, the app will end up in an inconsistent state when `_verifyAuthModelRelations` -models and the remoting authorization hook are all set up, but the flag is saying otherwise.

lib/application.js
```diff
@@ -398,85 +399,217 @@ app.enableAuth = function(options) {
     }
   };
 
-  this._verifyAuthModelRelations();
-
-  this.isAuthEnabled = true;
+  return this._verifyAuthModelRelations(app._warnOnBadAuthModelRelations)
```
bajtos: In existing applications that are not waiting for `enableAuth` to resolve, when the verification code throws an error, the rejected promise will end up as unhandled promise rejection. This typically produces only a console warning,  it does not crash the application as an uncaught error would.

```
(node:90355) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error
```

I am proposing to introduce an `options` flag to control the sync/async behaviour. For example:

```js
// old way - throws error
app.enableAuth({dataSource: 'db'});

// new way - returns rejected promise
app.enableAuth({dataSource: 'db', async: true})
  .then(...)
  .catch(...);

// alternatively, replace the flag with a new method name
app.enableAuthAsync({dataSource: 'db'});
```

(The "Async" prefix is already used by Bluebird's [promisify API](http://bluebirdjs.com/docs/api/promise.promisify.html))

lib/application.js
bajtos: That's weird. This is what Bluebird's docs say about the error:

https://github.com/petkaantonov/bluebird/blob/master/docs/docs/error-explanations.md#error-cannot-get-fulfillment-value-of-a-non-fulfilled-promise

> You can get this error when you're trying to call .value or .error when inspecting a promise where the promise has not been fulfilled or rejected yet.

> For example:

```js
var p = Promise.delay(1000);
p.inspect().value();
```

> Consider using .isPending() .isFulfilled() and .isRejected() in order to inspect the promise for status.

This looks like a possible bug in bluebird to me.

lib/application.js
bajtos: Have you considered passing the custom implementation via `enableAuth` options instead?

bajtos: Not a big deal though.

test/app.test.js
```diff
@@ -859,10 +861,11 @@ describe('app', function() {
   });
 
   describe.onServer('enableAuth', function() {
-    it('should set app.isAuthEnabled to true', function() {
+    it('sets app.isAuthEnabled to true', function() {
       expect(app.isAuthEnabled).to.not.equal(true);
-      app.enableAuth();
-      expect(app.isAuthEnabled).to.equal(true);
+      return app.enableAuth().then(() => {
+        expect(app.isAuthEnabled).to.equal(true);
```
bajtos: This goes back to my first comment https://github.com/strongloop/loopback/pull/3265/files#r107652656.

When we have to change a test in a significant way, as we are doing here, then I think it's a very strong sign that we are breaking backwards compatibility.

test/app.test.js
```diff
@@ -871,14 +874,14 @@ describe('app', function() {
       require('../lib/builtin-models')(app.registry);
       var db = app.dataSource('db', {connector: 'memory'});
 
-      app.enableAuth({dataSource: 'db'});
-
-      expect(Object.keys(app.models)).to.include.members(AUTH_MODELS);
+      return app.enableAuth({dataSource: 'db'}).then(() => {
```
bajtos: This change shouldn't be needed, because we are attaching the models synchronously in the current implementation. Unless I am missing something?

test/app.test.js
```diff
@@ -892,9 +895,71 @@ describe('app', function() {
       const AccessToken = app.registry.getModel('AccessToken');
       AccessToken.settings.relations.user.model = 'Customer';
 
-      app.enableAuth({dataSource: 'db'});
+      return app.enableAuth({dataSource: 'db'}).then(() => {
```
bajtos: ditto, I think this change is not needed.

test/user.test.js
```diff
@@ -2442,7 +2442,7 @@ describe('User', function() {
     it('handles subclassed user with no accessToken relation', () => {
       // setup a new LoopBack app, we don't want to use shared models
       app = loopback({localRegistry: true, loadBuiltinModels: true});
-      app.set('_verifyAuthModelRelations', false);
+      app.set('_warnOnBadAuthModelsSetup', false);
```
bajtos: Would you mind reverting this rename? I don't really care about the name, it's just to avoid code churn.

ebarault: it would mean that it's the exact same name as the app method 
`app. _verifyAuthModelRelations()`
is it problematic?

lib/application.js
ebarault: Yes, i agree this is the main point to discuss.

In loopback i think this is not a problem since the features/modules relying on this flag comes few ticks later, but i agree we should care about 3rd party apps. 
I dropped this implementation quickly to make this visible and get your feedback.

In fact i noted when implementing the tests that i could not easily test directly the _verifyAuthModelRelations method alone since it require the auth models and relations are all set up. Which is usually a benefit of using `enableAuth()`, so i needed to call `_verifyAuthModelRelations` through `enableAuth`, and then way for the latter's returned promise to resolve to properly get the warning messages. So at first i just modified `enableAuth` to return a promise. Then i considered the meaning of the flag and decided to set it after the health check has resolved.

**The rationale:** 
Leaving aside backward compat for a sec: a valid auth setup health check function would not result in setting an isAuthEnabled flag true if the auth model was incorrect is some ways. We could project for v4 that the app deliberately stops in situations that now trigger only warnings, until the user fixes the issues or activate some failover flags to bypass the situation. Thoughts?

👉  **For now:** Setting app.isAuthEnabling to `true` before the call _verifyAuthModelRelations would probably not have a big impact actually and could keep a variant where it still set it sync and let the method return async after the checks have completed:
- either the relations hare not well set for single user model and the flag will be set and the app will warn, this is the current behavior, also this does not mean that the flag should actually be set since the app is not configured properly
- or the multiple users config if not good (including relations), the flag will have been set but the app will crash, so there's not risk in setting the flag sync

So this could be 
```js
  // ...
  // still sync up to this point
  this.isAuthEnabled = true;
  return this._verifyAuthModelRelations(app._warnOnBadAuthModelRelations); // async
}
```

As the enableAuth() method currently returns nothing, there's no risk that it will conflict in existing 3rd party apps.

Of course we can additionaly already implement the target implement with a flag and set the flag only latter and decide to halt the app if warnings occur with a feature flag (wrapping your other proposal about an option to return sync or async)

bajtos: > I dropped this implementation quickly to make this visible and get your feedback.

👍 

> Leaving aside backward compat for a sec: a valid auth setup health check function would not result in setting an isAuthEnabled flag true if the auth model was incorrect is some ways. We could project for v4 that the app deliberately stops in situations that now trigger only warnings, until the user fixes the issues or activate some failover flags to bypass the situation. Thoughts?

I agree that if the backwards compatibility was not an issue (e.g. in v4), then `enableAuth` should fail when the configuration is invalid.

>
```js
// still sync up to this point
this.isAuthEnabled = true;
return this._verifyAuthModelRelations(app._warnOnBadAuthModelRelations); // async
```

👍  let's take this approach now.

test/user.test.js
bajtos: > it would mean that it's the exact same name as the app method

That shouldn't matter, because the flag is stored in a config hash-map.

> the rationale for this change is: the _verifyAuthModelRelations method, for a same name, has now a larger scope than the equally named flag

I see your point, let's make the proposed change then.

lib/application.js
ebarault: nevermind... thanks for digging, the error was on my side...

PR #3292, opened 2017-03-17
Fix user-literal rewrite for anonymous requests

### LINKED ISSUE #3291: invalid token on currentUserLiteral throws 500
When loading fixtures in our dev/staging environments, our token table gets reset and all clients that were previously logged in start receiving 500 responses at currentUserLiteral endpoints with the error: 
```
500 invalid input syntax for integer: "me"
length: 91
severity: ERROR
code: 22P02
file: numutils.c
line: 62
routine: pg_atoi
error: invalid input syntax for integer: "me"
    at Connection.parseE (/.../node_modules/pg/lib/connection.js:569:11)
    at Connection.parseMessage (/.../node_modules/pg/lib/connection.js:396:17)
    at Socket. (/.../node_modules/pg/lib/connection.js:132:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:548:20)
```

In looking into this we found that the authToken was being invalidated and unset, and the default findById method was being called with our currentUserLiteral as the id which throws a SQL error. 

*update*
The cause seems to be the if on line L25 of [token.js](https://github.com/strongloop/loopback/blob/master/server/middleware/token.js#L25-L34). If the access token is not set, then this request is processed by `Model.findById`. 

# Expected result

Expected a 401 Unauthorized response

# Additional information

We've temporarily worked around this with the following:
```
const router = server.loopback.Router();
const currentUserLiteral = 'me';
router.all(
    `/api/{model}/${currentUserLiteral}`,
    catchInvalidAuthForCurrentUserLiteral
);
server.use(router);

function catchInvalidAuthForCurrentUserLiteral(req, res, next) {
  if (req.accessToken == null) res.sendStatus(401);
  else next();
}
```

Env info:
```
$ node -e 'console.log(process.platform, process.arch, process.versions.node)'
darwin x64 6.9.5

$ npm ls --prod --depth 0 | grep loopback
├── loopback@3.4.0
├── loopback-boot@2.23.0
├── loopback-component-explorer@4.1.1
├── loopback-component-passport@3.2.0
├── loopback-connector-postgresql@2.8.0
├── loopback-fixtures@0.0.10 invalid
```

aaronbuchanan:
### Description

Currently any `currentUserLiteral` routes when accessed with a bad token throw a 500 due to a SQL error that is raised because `Model.findById` is invoked with `id={currentUserLiteral}` (`id=me` in our case) when the url rewrite fails. 

This PR will render a 401 response on currentUserLiteral routes instead. 

#### Related issues

- #3291 invalid token on currentUserLiteral throws 500

### Checklist

### DIFF
--- a/server/middleware/token.js
+++ b/server/middleware/token.js
@@ -8,6 +8,7 @@
  */
 
 'use strict';
+var g = require('../../lib/globalize');
 var loopback = require('../../lib/loopback');
 var assert = require('assert');
 var debug = require('debug')('loopback:middleware:token');
@@ -21,18 +22,33 @@ module.exports = token;
 /*
  * Rewrite the url to replace current user literal with the logged in user id
  */
-function rewriteUserLiteral(req, currentUserLiteral) {
-  if (req.accessToken && req.accessToken.userId && currentUserLiteral) {
+function rewriteUserLiteral(req, currentUserLiteral, next) {
+  if (!currentUserLiteral) return next();
+  const literalRegExp = new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g');
+
+  if (req.accessToken && req.accessToken.userId) {
     // Replace /me/ with /current-user-id/
     var urlBeforeRewrite = req.url;
-    req.url = req.url.replace(
-      new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g'),
+    req.url = req.url.replace(literalRegExp,
         '/' + req.accessToken.userId + '$1');
+
     if (req.url !== urlBeforeRewrite) {
       debug('req.url has been rewritten from %s to %s', urlBeforeRewrite,
         req.url);
     }
+  } else if (!req.accessToken && literalRegExp.test(req.url)) {
+    debug(
+      'URL %s matches current-user literal %s,' +
+        ' but no (valid) access token was provided.',
+      req.url, currentUserLiteral);
+
+    var e = new Error(g.f('Authorization Required'));
+    e.status = e.statusCode = 401;
+    e.code = 'AUTHORIZATION_REQUIRED';
+    return next(e);
   }
+
+  next();
 }
 
 function escapeRegExp(str) {
@@ -105,23 +121,24 @@ function token(options) {
       if (!enableDoublecheck) {
         // req.accessToken is defined already (might also be "null" or "false") and enableDoublecheck
         // has not been set --> skip searching for credentials
-        rewriteUserLiteral(req, currentUserLiteral);
-        return next();
+        return rewriteUserLiteral(req, currentUserLiteral, next);
       }
       if (req.accessToken && req.accessToken.id && !overwriteExistingToken) {
         // req.accessToken.id is defined, which means that some other middleware has identified a valid user.
         // when overwriteExistingToken is not set to a truthy value, skip searching for credentials.
-        rewriteUserLiteral(req, currentUserLiteral);
-        return next();
+        return rewriteUserLiteral(req, currentUserLiteral, next);
       }
       // continue normal operation (as if req.accessToken was undefined)
     }
+
     TokenModel.findForRequest(req, options, function(err, token) {
       req.accessToken = token || null;
-      rewriteUserLiteral(req, currentUserLiteral);
+
       var ctx = req.loopbackContext;
       if (ctx && ctx.active) ctx.set('accessToken', token);
-      next(err);
+
+      if (err) return next(err);
+      rewriteUserLiteral(req, currentUserLiteral, next);
     });
   };
 }
--- a/test/access-token.test.js
+++ b/test/access-token.test.js
@@ -289,6 +289,26 @@ describe('loopback.token(options)', function() {
         });
     });
 
+  it('should generate a 401 on a current user literal route without an authToken',
+    function(done) {
+      var app = createTestApp(null, done);
+      request(app)
+        .get('/users/me')
+        .set('authorization', null)
+        .expect(401)
+        .end(done);
+    });
+
+  it('should generate a 401 on a current user literal route with invalid authToken',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      request(app)
+        .get('/users/me')
+        .set('Authorization', 'invald-token-id')
+        .expect(401)
+        .end(done);
+    });
+
   it('should skip when req.token is already present', function(done) {
     var tokenStub = {id: 'stub id'};
     app.use(function(req, res, next) {


slnode:

slnode:

slnode:

slnode:

slnode:

server/middleware/token.js
```diff
@@ -32,6 +32,9 @@ function rewriteUserLiteral(req, currentUserLiteral) {
       debug('req.url has been rewritten from %s to %s', urlBeforeRewrite,
         req.url);
     }
+  } else if (!req.accessToken && currentUserLiteral &&
```
bajtos: http://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

I am proposing the following:

```js
function rewriteUserLiteral(req, currentUserLiteral) {
  if (!currentUserLiteral) return;
  
  if (req.accessToken && req.accessToken.userId) {
    // ...
  } else if (...) {
    // ...
  }
}
```

server/middleware/token.js
```diff
@@ -32,6 +32,9 @@ function rewriteUserLiteral(req, currentUserLiteral) {
...
+      new RegExp(currentUserLiteral).test(req.url)) {
```
bajtos: This will match requests that do not contain current user literal. Assuming the literal is `me`, this will match requests like `/products/watermelon` too.

My proposal: use the same regexp as on L29.

```js
function rewriteUserLiteral(req, currentUserLiteral) {
  if (!currentUserLiteral) return;
  const literalRegExp = new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g');
  // ...
}

File: server/middleware/token.js [hunk 3]
```diff
@@ -32,6 +32,9 @@ function rewriteUserLiteral(req, currentUserLiteral) {
...
+    req.res.status(401).end();
```
bajtos: This will by-pass the application error handler, which will result in an inconsistent error response.

You should create a new `Error` object and pass it to `next` function which was provided to `token` middleware handler.

test/access-token.test.js
```diff
@@ -289,6 +289,33 @@ describe('loopback.token(options)', function() {
...
     });
 
+  it('should generate a 401 on a current user literal route without an authToken',
+    function(done) {
+      var app = createTestApp(null, done);
+      request(app)
+        .get('/users/me')
+        .set('authorization', null)
+        .expect(401)
+        .end(function(err, res) {
```
bajtos:
```js
.end(done);
```

test/access-token.test.js
```diff
@@ -289,6 +289,33 @@ describe('loopback.token(options)', function() {
...
+
+  it('should generate a 401 on a current user literal route with invalid authToken',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var invalidTokenStub = {id: 'stub id'};
+      request(app)
+        .get('/users/me')
+        .set('Authorization', invalidTokenStub.id)
+        .expect(401)
+        .end(function(err, res) {
```
bajtos:
```js
.end(done)
```

test/access-token.test.js
```diff
@@ -289,6 +289,33 @@ describe('loopback.token(options)', function() {
...
+        });
+    });
+
+  it('should generate a 401 on a current user literal route with invalid authToken',
+    function(done) {
+      var app = createTestApp(this.token, done);
+      var invalidTokenStub = {id: 'stub id'};
+      request(app)
+        .get('/users/me')
+        .set('Authorization', invalidTokenStub.id)
```
bajtos: No need to create `invalidTokenStub` object, just pass a string that makes it clear it's an invalid token id.

```js
.set('Authorization', 'invalid-token-id')
```

bajtos:
@slnode test please

server/middleware/token.js
aaronbuchanan: Love it, thought about a guard clause there, but didn't want to introduce seemingly unnecessary changes. Will update!

server/middleware/token.js
aaronbuchanan: Good catch! [lol](https://gist.github.com/aaronbuchanan/cbaa38f45b6e4ae6ec00e9d5f707acdb)

server/middleware/token.js
aaronbuchanan: ok, I wasn't sure if it was better to `throw` or `return`, prob picked the wrong one but this is updated 👍

aaronbuchanan:
@bajtos I think I've addressed your points, let me know if you see anything else!

aaronbuchanan:
rebased per [pr linter](https://ci.sl.strongloop.com/job/strongloop-pr-linter-jobs/job/loopback/1268/console)

server/middleware/token.js
```diff
@@ -22,16 +23,25 @@ module.exports = token;
...
-      new RegExp('/' + currentUserLiteral + '(/|$|\\?)', 'g'),
+    req.url = req.url.replace(literalRegExp,
         '/' + req.accessToken.userId + '$1');
+
     if (req.url !== urlBeforeRewrite) {
       debug('req.url has been rewritten from %s to %s', urlBeforeRewrite,
         req.url);
     }
+  } else if (!req.accessToken && literalRegExp.test(req.url)) {
+    req.res.status(401).end();
```
bajtos: You shouldn't be touching the `res` object in this method. Let the error handler middleware decide how to handle the error.

Sorry if my original comment didn't make this clear.

server/middleware/token.js
```diff
@@ -101,24 +111,26 @@ function token(options) {
...
 
+    var err = null;
     if (req.accessToken !== undefined) {
       if (!enableDoublecheck) {
         // req.accessToken is defined already (might also be "null" or "false") and enableDoublecheck
         // has not been set --> skip searching for credentials
-        rewriteUserLiteral(req, currentUserLiteral);
-        return next();
+        err = rewriteUserLiteral(req, currentUserLiteral);
+        return next(err);
```
bajtos: While this solution definitely works, it's quite an unusual pattern.

In Node.js world, the convention for functions is to accept a callback argument and then pass any errors to that callback.

```js
rewriteUserLiteral(req, currentUserLiteral, next);
```

server/middleware/token.js
```diff
@@ -101,24 +111,26 @@ function token(options) {
...
       // continue normal operation (as if req.accessToken was undefined)
     }
+
     TokenModel.findForRequest(req, options, function(err, token) {
       req.accessToken = token || null;
-      rewriteUserLiteral(req, currentUserLiteral);
+      err = err || rewriteUserLiteral(req, currentUserLiteral);
       var ctx = req.loopbackContext;
       if (ctx && ctx.active) ctx.set('accessToken', token);
       next(err);
```
bajtos: I am proposing to clean up this block a bit while we are changing it.

```js
TokenModel.findForRequest(req, options, function(err, token) {
  if (err) {
    req.accessToken = null;
    return next(err);
  }

  var ctx = req.loopbackContext;
  if (ctx && ctx.active) ctx.set('accessToken', token);

  rewriteUserLiteral(req, currentUserLiteral, next);
}
```

bajtos:
@slnode ok to test

server/middleware/token.js
aaronbuchanan: Oh geez, that was a mistake. Should have been removed when adding the error.

server/middleware/token.js
aaronbuchanan: I considered that initially but didn't like transferring responsibility to call `next()` to rewriteUserLiteral because it wasn't always the last call. Since we've refactored the findForRequest I think it makes more sense to pass `next in now.

aaronbuchanan:
Hey @bajtos! I've made the changes, some other tests started failing and it's unclear why to me. In debugging, I see that some of the failing test scenarios are trying to access userLiteral routes with null `accessToken` values (where `(!req.accessToken && literalRegExp.test(req.url))` is true)

In looking further it appears `req.loopbackContext` is null and so `if (ctx && ctx.active) ctx.set('accessToken', token);` doesn't get called. 

Any idea?

server/middleware/token.js
```diff
@@ -101,27 +112,31 @@ function token(options) {
     assert(typeof TokenModel === 'function',
       'loopback.token() middleware requires a AccessToken model');
 
+    var err = null;
```
bajtos: I think this can be removed.

bajtos:
@aaronbuchanan sorry for responding so late!

> Any idea?

Not really :(

> some of the failing test scenarios are trying to access userLiteral routes with null `accessToken` values (where `(!req.accessToken && literalRegExp.test(req.url))` is true).

That's surprising. Which tests are attempting to do that?

> In looking further it appears `req.loopbackContext` is null and so `ctx.set('accessToken', token);` doesn't get called. 

I think `req.loopbackContext` shouldn't be related to user-literal rewriting. AFAIK, the only place where we are using `req.loopbackContext` is the token middleware. Those few lines of code are there to make it easier for `loopback-context` users to access the currently-logged-in user without having to install a new middleware.

bajtos:
Hey, it turns out my code example in https://github.com/strongloop/loopback/pull/3292#discussion_r107709032 contained a severe bug - `req.accessToken` was not set when the token was authenticated. I fixed the problem in dfb19e7

bajtos:

PR #3293, opened 2017-03-18
Fix acl.resolvePermission not working with wildcard request

### LINKED ISSUE #2153: ACL checkpermission not working properly
i have static ACL rules inside my model named 'Event'

this are the acl in my   Event.json

```
"acls": [
    {
      "model": "*",
      "property": "*",
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
```

```
{
  "model": "*",
  "property": "*",
  "accessType": "*",
  "principalType": "ROLE",
  "principalId": "admin",
  "permission": "ALLOW"
}`
```

it works properly during rest calls to that model 

but , i have custom route for that model . During that route i check with "ACL.checkpermission" to verify permission for principal manually . This is my custom route in my boot script

``` node
module.exports = function(app){
    app.get('/getEvents', function(req, res, next){
        app.models.RoleMapping.findOne({
                                           where  : {principalId: request.accessToken.userId},
                                           include: {relation: 'role', fields: ['name']}
                                       }).then(function(map){
            var principal     = map.role().name;
            var principalType = 'ROLE';
            app.models.ACL.checkPermission(principalType, principal, 'Event', '*', 'READ',
                                           function(err, result){
                                               if(!err){ console.log(JSON.stringify(result)); }
                                           });
        }
```

in console i got this , 

``` shell
  loopback:security:acl The following ACLs were searched:  +0ms
  loopback:security:acl ---Resolved--- +1ms
  loopback:security:acl The following ACLs were searched:  +8s
  loopback:security:acl ---ACL--- +1ms
  loopback:security:acl model Event +0ms
  loopback:security:acl property * +0ms
  loopback:security:acl principalType ROLE +0ms
  loopback:security:acl principalId admin +0ms
  loopback:security:acl accessType * +0ms
  loopback:security:acl permission ALLOW +0ms
  loopback:security:acl with score: +0ms 8020
  loopback:security:acl ---ACL--- +0ms
  loopback:security:acl model Event +0ms
  loopback:security:acl property * +0ms
  loopback:security:acl principalType ROLE +0ms
  loopback:security:acl principalId $everyone +0ms
  loopback:security:acl accessType * +0ms
  loopback:security:acl permission DENY +0ms
  loopback:security:acl with score: +0ms 8007
  loopback:security:acl Permission denied by statically resolved permission +0ms
  loopback:security:acl  Resolved Permission: {"model":"Event","property":"*","accessType":"READ","permission":"DENY","methodNames":[]} +0ms
{"model":"Event","property":"*","accessType":"READ","permission":"DENY","methodNames":[]}
```

Later while i check with 'function checkpermission' in acl.js.
i found that issue is with ''resolvePermission" invoked inside  checkpermission , 
the issue is missing 'break'  statement refer code below at '@issuehere' 

``` node
ACL.resolvePermission = function resolvePermission(acls, req) {
    if (!(req instanceof AccessRequest)) {
      req = new AccessRequest(req);
    }
    // Sort by the matching score in descending order
    acls = acls.sort(function(rule1, rule2) {
      return ACL.getMatchingScore(rule2, req) - ACL.getMatchingScore(rule1, req);
    });
    var permission = ACL.DEFAULT;
    var score = 0;

    for (var i = 0; i < acls.length; i++) {
      var candidate = acls[i];
      score = ACL.getMatchingScore(candidate, req);
      if (score < 0) {
        // the highest scored ACL did not match
        break;
      }
      if (!req.isWildcard()) {
        // We should stop from the first match for non-wildcard
        permission = candidate.permission;
        break;
      } else {
        if (req.exactlyMatches(candidate)) {
          permission = candidate.permission;
          break;
        }
        // For wildcard match, find the strongest permission
        var candidateOrder = AccessContext.permissionOrder[candidate.permission];
        var permissionOrder = AccessContext.permissionOrder[permission];
        if (candidateOrder > permissionOrder) {
          permission = candidate.permission;
           //@issuehere
          break; //This is the fix
        }
      }
    }

    if (debug.enabled) {
      debug('The following ACLs were searched: ');
      acls.forEach(function(acl) {
        acl.debug();
        debug('with score:', acl.score(req));
      });
    }

    var res = new AccessRequest(req.model, req.property, req.accessType,
        permission || ACL.DEFAULT);
    return res;
  };
```

alFReD-NSH:
### Description

When acl.resolvePermission was called with a request containing a wildcard, it would return the matching acl with lowest score instead of higher. 

#### Related issues

 - Fixes #2153
 - Also this was mentioned on [Stackoverflow](http://stackoverflow.com/q/42176493/774086).  

### Checklist

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -243,6 +243,7 @@ module.exports = function(ACL) {
         var permissionOrder = AccessContext.permissionOrder[permission];
         if (candidateOrder > permissionOrder) {
           permission = candidate.permission;
+          break;
         }
       }
     }
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -186,6 +186,41 @@ describe('security ACLs', function() {
     // });
   });
 
+  it('should order ACL entries based on the matching score even with wildcard req', function() {
+    var acls = [
+      {
+        'model': 'account',
+        'accessType': '*',
+        'permission': 'DENY',
+        'principalType': 'ROLE',
+        'principalId': '$everyone',
+      },
+      {
+        'model': 'account',
+        'accessType': '*',
+        'permission': 'ALLOW',
+        'principalType': 'ROLE',
+        'principalId': '$owner',
+      }];
+    var req = {
+      model: 'account',
+      property: '*',
+      accessType: 'WRITE',
+    };
+
+    acls = acls.map(function(a) { return new ACL(a); });
+
+    var perm = ACL.resolvePermission(acls, req);
+    // remove the registry from AccessRequest instance to ease asserting.
+    // Check the above test case for more info.
+    delete perm.registry;
+    assert.deepEqual(perm, {model: 'account',
+      property: '*',
+      accessType: 'WRITE',
+      permission: 'ALLOW',
+      methodNames: []});
+  });
+
   it('should allow access to models for the given principal by wildcard', function() {
     // jscs:disable validateIndentation
     ACL.create({


slnode:

slnode:

slnode:

slnode:

ebarault:
Thanks @alFReD-NSH for your pull request.
Please as a starting point see the my post [here](https://github.com/strongloop/loopback/issues/2153#issuecomment-287539017)

I'll further comment asap, but after some investigations to understand the method you're willing to change was implemented this way, i can at least say that this looks as a breaking change to me, more than a bug.

As i commented in the original issue you should not implement the access-control in lb this way. 
More on this latter.

bajtos:
> When acl.resolvePermission was called with a request containing a wildcard, it would return the matching acl with lowest score instead of higher.

I am afraid this is beyond my current understanding of our ACL system. @raymondfeng @ritch you are the authors of the original design for prioritisation of ACL entries, what's you opinion on this proposal?

bajtos:
@slnode ok to test

ebarault:
Given that comment `// For wildcard match, find the strongest permission` [here](https://github.com/strongloop/loopback/blob/master/common/models/acl.js#L237)

The implementation is correct and there's no missing break. Question remains: why this behavior? and what was the intention with wildcards in access requests? (@ritch, @raymondfeng)

I cannot find any instance in the code base where an access request does not represent exactly one specific request: targeting one model, one method, one accessType. 

I cannot find either any excerpts of former docs where the checkPermission was used directly by developers: shouldn't it be a private method instead as the standard behavior would be to use the built-in access control system?

raymondfeng:
In this particular case:
```
var acls = [
      {
        'model': 'account',
        'accessType': '*',
        'permission': 'DENY',
        'principalType': 'ROLE',
        'principalId': '$everyone',
      },
      {
        'model': 'account',
        'accessType': '*',
        'permission': 'ALLOW',
        'principalType': 'ROLE',
        'principalId': '$owner',
      }];
```

The 2nd rule is stronger than the 1st one as it's more specific about the principal. The score is calculated by (model, property, accessType, principal, permission). The current code seems to assume that all wildcard rules have the same score. As a result, it continues to search for stronger permission.

If two rules cannot produce the same score, adding `break` probably make sense.

ebarault:
thx @raymondfeng for your time.
the issue actually is not with wildcards in ACLs but with wildcards in AccessRequests.

🖐  AFAIK AccessRequests created by LB's access control system have no wildcards
So an AccessRequest with a wildcard would have been created by the developer for some purpose.

ATM if `ACL.checkPermission(principalType, principalId, model, property, accessType)` is called with a  wildcard on model, property or accessType, then whatever the ACLs' score order, the algorithm will resolve with the strongest permission, hence DENY (score 4) if one ACL of the model is DENY.

My main point is actually to argue that it has currently no meaning to call `ACL.checkPermission()` with any of those wildcard, and developers should not do it.
But if this was to evolve, i wanted to understand what was your idea with the following piece of code:

```js
      if (!req.isWildcard()) {  
        // ...
      } else {
        // ...
        // For wildcard match, find the strongest permission
        var candidateOrder = AccessContext.permissionOrder[candidate.permission];
        var permissionOrder = AccessContext.permissionOrder[permission];
        if (candidateOrder > permissionOrder) {
          permission = candidate.permission;
        }
      }
```

which is equivalent to say: 
```
if (accessRequest.isWildcard()) {
  // pick the strongest permissions, according to the following ranking:
  /*
  AccessContext.permissionOrder = {
    DEFAULT: 0,
    ALLOW: 1,
    ALARM: 2,
    AUDIT: 3,
    DENY: 4,
  };
*/
}
```

alFReD-NSH:
I appreciate the comments guys. @raymondfeng basically said in other words why I think this is bug and should be fixed.

@ebarault Basically I checkPermission to see if user has access to a certain screen and if it could match wildcard requests, it would make coding it simpler. Since there is `isWildcard` method on the request object, I believe it's appropriate for it be wildcard.

ebarault:
@alFReD-NSH : by screens do you mean MVC views served by the template engine?

alFReD-NSH:
@ebarault Technically the check is for the navigation menu. If the user doesn't have access to an item in it, it shouldn't be there.

ebarault:
I see.
Could you check using the `ACL.checkAccessForContext()` instead of checkPermission (without your fix) ? I'd like to test this before going further.

I'm not confident with the current behavior which consist in dealing with any wildcard in an AccessRequest the same way (see `AccessRequest.isWildCard()`)

alFReD-NSH:
@ebarault Same thing happens.

bajtos:
Ping, what's the status of this pull request? How to move it forward?

ebarault:
@bajtos, @alFReD-NSH  : i validate an improvement can be made here, and the overall idea in this PR makes sense to me now (especially now that the issue is confirmed in `ACL.checkAccessForContext()`) i'd just would like to take some time here to review the whole authorization sequence once more to understand the potential impact of treating any wildcard access request the same way.

chris08002:
Any news on this one?

ebarault:
IMO the rationale described in this PR is valid. The implementation remains to be reviewed.

slnode:

slnode:

slnode:

slnode:

bajtos:
Hello, I have re-read the discussion above and as I understand it, the changes are going in the right direction and there are no known problems. In that light, I am proposing to move forward and land this pull request as it is now.

Any objections?

@alFReD-NSH would you mind rebasing the patch on top of the latest master?

supermafete:
Definitely that 'break' in this PR is necessary in the code, but also is necessary some more stuff. Please, review my PR #3618 relating issue #3518.

slnode:

slnode:

slnode:

alFReD-NSH:
I have rebased this against the latest master.

bajtos:
I checked few of the downstream builds, they are failing because `nsp check` reports vulnerable dependencies. I believe that's not related to this pull request and we can ignore these failures.

bajtos:
Landed, thank you @alFReD-NSH for your contribution!

PR #3299, opened 2017-03-22
Add User.changePassword(id, old, new, cb)

bajtos:
### Description

Implement a new method for changing user passwords the secure way.
The method requires the old password to be provided before a new
password can be used.

JavaScript API:

    User.changePassword(userId, oldPassword, newPassword[, cb])

REST API:

    POST /api/users/change-password
    Authorization: your-token-id
    Content-Type: application/json

    {"oldPassword":"old-pass", "newPassword": "new-pass"}

#### Related issues

- #382

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -354,6 +354,80 @@ module.exports = function(User) {
   };
 
   /**
+   * Change this user's password.
+   *
+   * @param {*} userId Id of the user changing the password
+   * @param {string} oldPassword Current password, required in order
+   *   to strongly verify the identity of the requesting user
+   * @param {string} newPassword The new password to use.
+   * @param {object} [options]
+   * @callback {Function} callback
+   * @param {Error} err Error object
+   * @promise
+   */
+  User.changePassword = function(userId, oldPassword, newPassword, options, cb) {
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
+
+    // Make sure to use the constructor of the (sub)class
+    // where the method is invoked from (`this` instead of `User`)
+    this.findById(userId, options, (err, inst) => {
+      if (err) return cb(err);
+
+      if (!inst) {
+        const err = new Error(`User ${userId} not found`);
+        Object.assign(err, {
+          code: 'USER_NOT_FOUND',
+          statusCode: 401,
+        });
+        return cb(err);
+      }
+
+      inst.changePassword(oldPassword, newPassword, options, cb);
+    });
+
+    return cb.promise;
+  };
+
+  /**
+   * Change this user's password (prototype/instance version).
+   *
+   * @param {string} oldPassword Current password, required in order
+   *   to strongly verify the identity of the requesting user
+   * @param {string} newPassword The new password to use.
+   * @param {object} [options]
+   * @callback {Function} callback
+   * @param {Error} err Error object
+   * @promise
+   */
+  User.prototype.changePassword = function(oldPassword, newPassword, options, cb) {
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
+
+    this.hasPassword(oldPassword, (err, isMatch) => {
+      if (err) return cb(err);
+      if (!isMatch) {
+        const err = new Error('Invalid current password');
+        Object.assign(err, {
+          code: 'INVALID_PASSWORD',
+          statusCode: 400,
+        });
+        return cb(err);
+      }
+
+      const delta = {password: newPassword};
+      this.patchAttributes(delta, options, (err, updated) => cb(err));
+    });
+    return cb.promise;
+  };
+
+  /**
    * Verify a user's identity by sending them a confirmation email.
    *
    * ```js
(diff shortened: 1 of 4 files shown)


common/models/user.js
```diff
@@ -354,6 +354,60 @@ module.exports = function(User) {
...
+    const Ctor = this;
+
+    Ctor.findById(userId, options, function(err, inst) {
+      if (err) return cb(err);
+
+      if (!inst) {
+        const err = new Error(`User ${userId} not found`);
+        Object.assign(err, {
+          code: 'USER_NOT_FOUND',
+          statusCode: 400,
```
loay: statusCode should be 404 to be conform with the rest of the code example https://github.com/strongloop/loopback/blob/9c904629d180e3977622d04c80ea822d69282f6b/common/models/user.js#L621

test/user.test.js
```diff
@@ -1267,6 +1267,109 @@ describe('User', function() {
...
+      return User.changePassword('unknown-id', 'pass', 'pass').then(
+        success => { throw new Error('changePassword should have failed'); },
+        err => {
+          // workaround for chai problem
+          //   object tested must be an array, an object, or a string,
+          //    but error given
+          const props = Object.assign({}, err);
+          expect(props).to.contain({
+            code: 'USER_NOT_FOUND',
+            statusCode: 400,
```
loay: This should be 404... as per the description of the unit test: 'fails with 404 for unknown user id'

test/user.test.js
```diff
@@ -1267,6 +1267,109 @@ describe('User', function() {
...
+          });
+        });
+    });
+
+    it('forwards "options" argument', () => {
```
superkhau: forwards the

common/models/user.js
superkhau: Actually, I'm not so sure about 404 now. 401 seems more correct according to SO: http://stackoverflow.com/questions/1959947/whats-an-appropriate-http-status-code-to-return-by-a-rest-api-service-for-a-val

ebarault:
I understand the original issue as: "prevent someone that has stolen a valid access token to modify the related user's password".

So actually i'm not sure if there's a good reason not to make this new method an instance method, with an ACL restricted to `$owner`. This would enforce the double check: the user is logged in and knows the current password.

In the case the user is not logged in because he lost his password, then resetPassword should be used, with the caveat addressed that currently tokens emitted by resetPassword are valid application wide.

common/models/user.js
bajtos: Good catch. I was thinking about 404 too, but it seemed inappropriate to me.

> The server has not found anything matching the Request-URI.

The Request URI is `/api/users/change-password`, which is a valid existing path.

The problem is in the access token supplied in the Authorization header, because the user authenticated by this token no longer exists.

On the second thought, I agree with @superkhau that 401 is a more appropriate response status code.

common/models/user.js
```diff
@@ -812,6 +866,22 @@ module.exports = function(User) {
...
+        description: 'Change password of the current user.',
+        accepts: [
+          {arg: 'id', type: 'any',
+            http: ctx => ctx.req.accessToken && ctx.req.accessToken.userId,
+          },
+          {arg: 'oldPassword', type: 'string', required: true, http: {source: 'form'}},
+          {arg: 'newPassword', type: 'string', required: true, http: {source: 'form'}},
+          {arg: 'options', type: 'object', http: 'optionsFromRequest'},
+        ],
+        http: {verb: 'POST', path: '/change-password'},
```
bajtos: We typically use `POST` for non-RESTful actions, see e.g. `login`, `logout` or `resetPassword`.

bajtos:
@ebarault thank you for chiming in.

> prevent someone that has stolen a valid access token to modify the related user's password

Exactly, that's the problem we need to address IMO.

> + the extra comment: "This method should have ACLs configured to allow only the owner to execute it"

I think that's an implementation detail that can be changed as long as the root cause is fixed.

> So actually i'm not sure if there's a good reason not to make this new method an instance method, with an ACL restricted to $owner. This would enforce the double check: the user is logged in and knows the current password.

IMO, the current implementation is equally secure as using `$owner` ACL, because we are modifying the currently logged in user as provided by the access token used to authenticate the request.

My objections against making `changePassword` a prototype method stem from the fact that as a prototype method, it would be exposed on `/api/users/:id/change-password`. This means:
 1. Clients would have to build this URL dynamically. My proposal allows them to always use the same static URL.
 2. The request contains two fields that may be in conflict (accessToken may authenticate a different user from the one specified via the instance `id` argument). I prefer to design the API in  a way that prevents this situation, instead of having to handle it in our code.
 3. It's no secret that the performance of our ACL system is not great, the $owner check comes with performance cost that I would rather avoid.

bajtos:
I have changed the controversial status code to 401, see https://github.com/strongloop/loopback/pull/3299#discussion_r107646218 and addressed most of the comments.

The remaining one is PUT vs POST discussion - see https://github.com/strongloop/loopback/pull/3299#discussion_r107551212

ebarault:
> IMO, the current implementation is equally secure as using $owner ACL, because we are modifying the currently logged in user as provided by the access token used to authenticate the request.
👍 

May i suggest we add a prototype method shortcut not exposed over REST for further use for example in building different endpoints reusing this code?

bajtos:
> May i suggest we add a prototype method shortcut not exposed over REST for further use for example in building different endpoints reusing this code?

@ebarault good idea! Added in 2c0ab9e, PTAL

loay:
I created a PR to fix failing CI for `gateway-director-bluemix@develop`. I will let you guys know when it is landed.

test/user.test.js
```diff
@@ -1267,6 +1267,109 @@ describe('User', function() {
...
+        });
+    });
+
+    it('fails when current password does not match', () => {
+      return User.changePassword(userId, 'bad password', 'new password').then(
+        success => { throw new Error('changePassword should have failed'); },
+        err => {
+          // workaround for chai problem
+          //   object tested must be an array, an object, or a string,
+          //    but error given
```
bajtos: FWIW, I have fixed the indentation in 4e60e62, I don't know why github did not recognize this comment as outdated.

safwen1:
Why does this method delete current accessToken ?

bajtos:
> Why does this method delete current accessToken ?

@safwen1 please open a new github issue to discuss this problem and mention my handle (`@bajtos`) so that I get an email notification.

PR #3313, opened 2017-03-27
Add support for scoped access tokens

bajtos:
### Description

Define a new property `AccessToken.scopes` to contain the list of scopes granted to this access token.

~~Define a new remote method metadata `accessScope` to contain a single scope name required by this method.~~

Define a new remote method metadata `accessScopes` to contain a list of scope names required by this method.

Modify the authorization algorithm to ensure that at least one of the required scope is allowed by the scopes granted to the requesting access token.

~~To preserve backwards compatibility, remote methods without `accessScope` metadata can be invoked only by access tokens with undefined `scopes` property.~~

To preserve backwards compatibility:
 - a built-in scope `DEFAULT` is defined
 - remote methods with no `accessScopes` metadata are treated as `accessScopes: ['DEFAULT']`
 - access tokens with no `scopes` property are treated as `scopes: ['DEFAULT']`

Impact on existing applications:

 - Database schema must be updated after upgrading the loopback version

 - If the application was already using a custom `AccessToken.scopes`
   property with a type different from an array, then the relevant code
   must be updated to work with the new type "array of strings".

This pull request is an incremental step towards #382 where we want to scope password-reset tokens to allow only password reset and no other operation.

cc @strongloop/sq-lb-epitome 

#### Related issues

- connect to #382
- depends on https://github.com/strongloop/strong-remoting/pull/404

### Checklist

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -98,6 +98,8 @@ module.exports = function(ACL) {
   ACL.ROLE = Principal.ROLE;
   ACL.SCOPE = Principal.SCOPE;
 
+  ACL.DEFAULT_SCOPE = ctx.DEFAULT_SCOPES[0];
+
   /**
    * Calculate the matching score for the given rule and request
    * @param {ACL} rule The ACL entry
@@ -461,6 +463,16 @@ module.exports = function(ACL) {
       permission: ACL.DEFAULT,
       methodNames,
       registry: this.registry});
+
+    if (!context.isScopeAllowed()) {
+      req.permission = ACL.DENY;
+      debug('--Denied by scope config--');
+      debug('Scopes allowed:', context.accessToken.scopes || ctx.DEFAULT_SCOPES);
+      debug('Scope required:', context.getScopes());
+      context.debug();
+      callback(null, req);
+      return callback.promise;
+    }
 
     var effectiveACLs = [];
     var staticACLs = self.getStaticACLs(model.modelName, property);
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -8,6 +8,8 @@ var assert = require('assert');
 var loopback = require('./loopback');
 var debug = require('debug')('loopback:security:access-context');
 
+const DEFAULT_SCOPES = ['DEFAULT'];
+
 /**
  * Access context represents the context for a request to access protected
  * resources
@@ -70,6 +72,7 @@ function AccessContext(context) {
   }
 
   this.accessType = context.accessType || AccessContext.ALL;
+
   assert(loopback.AccessToken,
     'AccessToken model must be defined before AccessContext model');
   this.accessToken = context.accessToken || loopback.AccessToken.ANONYMOUS;
@@ -193,6 +196,45 @@ AccessContext.prototype.isAuthenticated = function() {
   return !!(this.getUserId() || this.getAppId());
 };
 
+/**
+ * Get the list of scopes required by the current access context.
+ */
+AccessContext.prototype.getScopes = function() {
+  if (!this.sharedMethod)
+    return DEFAULT_SCOPES;
+
+  // For backwards compatibility, methods with no scopes defined
+  // are assigned a single "DEFAULT" scope
+  const methodLevel = this.sharedMethod.accessScopes || DEFAULT_SCOPES;
+
+  // TODO add model-level and app-level scopes
+
+  debug('--Context scopes of %s()--', this.sharedMethod.stringName);
+  debug('  method-level: %j', methodLevel);
+
+  return methodLevel;
+};
+
+/**
+ * Check if the scope required by the remote method is allowed
+ * by the scopes granted to the requesting access token.
+ * @return {boolean}
+ */
+AccessContext.prototype.isScopeAllowed = function() {
+  if (!this.accessToken) return false;
+
+  // For backwards compatibility, tokens with no scopes are treated
+  // as if they have "DEFAULT" scope granted
+  const tokenScopes = this.accessToken.scopes || DEFAULT_SCOPES;
+
+  const resourceScopes = this.getScopes();
+
+  // Scope is allowed when at least one of token's scopes
+  // is found in method's (resource's) scopes.
+  return Array.isArray(tokenScopes) && Array.isArray(resourceScopes) &&
+    resourceScopes.some(s => tokenScopes.indexOf(s) !== -1);
+};
+
 /*!
  * Print debug info for access context.
  */
@@ -213,10 +255,12 @@ AccessContext.prototype.debug = function() {
     debug('property %s', this.property);
     debug('method %s', this.method);
     debug('accessType %s', this.accessType);
+    debug('accessScopes %j', this.getScopes());
     if (this.accessToken) {
       debug('accessToken:');
       debug('  id %j', this.accessToken.id);
       debug('  ttl %j', this.accessToken.ttl);
+      debug('  scopes %j', this.accessToken.scopes || DEFAULT_SCOPES);
     }
     debug('getUserId() %s', this.getUserId());
     debug('isAuthenticated() %s', this.isAuthenticated());
@@ -379,3 +423,4 @@ AccessRequest.prototype.debug = function() {
 module.exports.AccessContext = AccessContext;
 module.exports.Principal = Principal;
 module.exports.AccessRequest = AccessRequest;
+module.exports.DEFAULT_SCOPES = DEFAULT_SCOPES;
--- a/test/authorization-scopes.test.js
+++ b/test/authorization-scopes.test.js
@@ -1,0 +1,130 @@
+'use strict';
+
+const loopback = require('../');
+const supertest = require('supertest');
+const strongErrorHandler = require('strong-error-handler');
+
+describe('Authorization scopes', () => {
+  const CUSTOM_SCOPE = 'read:custom';
+
+  let app, request, User, testUser, regularToken, scopedToken;
+  beforeEach(givenAppAndRequest);
+  beforeEach(givenRemoteMethodWithCustomScope);
+  beforeEach(givenUser);
+  beforeEach(givenDefaultToken);
+  beforeEach(givenScopedToken);
+
+  it('denies regular token to invoke custom-scoped method', () => {
+    logServerErrorsOtherThan(401);
+    return request.get('/users/scoped')
+      .set('Authorization', regularToken.id)
+      .expect(401);
+  });
+
+  it('allows regular tokens to invoke default-scoped method', () => {
+    logAllServerErrors();
+    return request.get('/users/' + testUser.id)
+      .set('Authorization', regularToken.id)
+      .expect(200);
+  });
+
+  it('allows scoped token to invoke custom-scoped method', () => {
+    logAllServerErrors();
+    return request.get('/users/scoped')
+      .set('Authorization', scopedToken.id)
+      .expect(204);
+  });
+
+  it('denies scoped token to invoke default-scoped method', () => {
+    logServerErrorsOtherThan(401);
+    return request.get('/users/' + testUser.id)
+      .set('Authorization', scopedToken.id)
+      .expect(401);
+  });
+
+  describe('token granted both default and custom scope', () => {
+    beforeEach('given token with default and custom scope',
+      () => givenScopedToken(['DEFAULT', CUSTOM_SCOPE]));
+    beforeEach(logAllServerErrors);
+
+    it('allows invocation of default-scoped method', () => {
+      return request.get('/users/' + testUser.id)
+        .set('Authorization', scopedToken.id)
+        .expect(200);
+    });
+
+    it('allows invocation of custom-scoped method', () => {
+      return request.get('/users/scoped')
+        .set('Authorization', scopedToken.id)
+        .expect(204);
+    });
+  });
+
+  it('allows invocation when at least one method scope is matched', () => {
+    givenRemoteMethodWithCustomScope(['read', 'write']);
+    return givenScopedToken(['read', 'execute']).then(() => {
+      return request.get('/users/scoped')
+        .set('Authorization', scopedToken.id)
+        .expect(204);
+    });
+  });
+
+  function givenAppAndRequest() {
+    app = loopback({localRegistry: true, loadBuiltinModels: true});
+    app.set('remoting', {rest: {handleErrors: false}});
+    app.dataSource('db', {connector: 'memory'});
+    app.enableAuth({dataSource: 'db'});
+    request = supertest(app);
+
+    app.use(loopback.rest());
+
+    User = app.models.User;
+  }
+
+  function givenRemoteMethodWithCustomScope() {
+    const accessScopes = arguments[0] || [CUSTOM_SCOPE];
+    User.scoped = function(cb) { cb(); };
+    User.remoteMethod('scoped', {
+      accessScopes,
+      http: {verb: 'GET', path: '/scoped'},
+    });
+    User.settings.acls.push({
+      principalType: 'ROLE',
+      principalId: '$authenticated',
+      permission: 'ALLOW',
+      property: 'scoped',
+      accessType: 'EXECUTE',
+    });
+  }
+
+  function givenUser() {
+    return User.create({email: 'test@example.com', password: 'pass'})
+      .then(u => testUser = u);
+  }
+
+  function givenDefaultToken() {
+    return testUser.createAccessToken(60)
+      .then(t => regularToken = t);
+  }
+
+  function givenScopedToken() {
+    const scopes = arguments[0] || [CUSTOM_SCOPE];
+    return testUser.accessTokens.create({ttl: 60, scopes})
+      .then(t => scopedToken = t);
+  }
+
+  function logAllServerErrors() {
+    logServerErrorsOtherThan(-1);
+  }
+
+  function logServerErrorsOtherThan(statusCode) {
+    app.use((err, req, res, next) => {
+      if ((err.statusCode || 500) !== statusCode) {
+        console.log('Unhandled error for request %s %s: %s',
+          req.method, req.url, err.stack || err);
+      }
+      res.statusCode = err.statusCode || 500;
+      res.json(err);
+    });
+  }
+});
--- a/common/models/access-token.json
+++ b/common/models/access-token.json
@@ -11,6 +11,10 @@
       "default": 1209600,
       "description": "time to live in seconds (2 weeks by default)"
     },
+    "scopes": {
+      "type": ["string"],
+      "description": "Array of scopes granted to this access token."
+    },
     "created": {
       "type": "Date",
       "defaultFn": "now"
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -30,6 +30,12 @@ before(function() {
   ds = loopback.createDataSource({connector: loopback.Memory});
 });
 
+describe('ACL model', function() {
+  it('provides DEFAULT_SCOPE constant', () => {
+    expect(ACL).to.have.property('DEFAULT_SCOPE', 'DEFAULT');
+  });
+});
+
 describe('security scopes', function() {
   beforeEach(function() {
     var ds = this.ds = loopback.createDataSource({connector: loopback.Memory});
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -624,18 +624,18 @@ describe('User', function() {
         // Override createAccessToken
         User.prototype.createAccessToken = function(ttl, options, cb) {
           // Reduce the ttl by half for testing purpose
-          this.accessTokens.create({ttl: ttl / 2, scopes: options.scope}, cb);
+          this.accessTokens.create({ttl: ttl / 2, scopes: [options.scope]}, cb);
         };
         User.login(validCredentialsWithTTLAndScope, function(err, accessToken) {
           assertGoodToken(accessToken, validCredentialsUser);
           assert.equal(accessToken.ttl, 1800);
-          assert.equal(accessToken.scopes, 'all');
+          assert.deepEqual(accessToken.scopes, ['all']);
 
           User.findById(accessToken.userId, function(err, user) {
             user.createAccessToken(120, {scope: 'default'}, function(err, accessToken) {
               assertGoodToken(accessToken, validCredentialsUser);
               assert.equal(accessToken.ttl, 60);
-              assert.equal(accessToken.scopes, 'default');
+              assert.deepEqual(accessToken.scopes, ['default']);
               // Restore create access token
               User.prototype.createAccessToken = createToken;
 


bajtos:
Note: The implementation here does not take into account the built-in model `Scope`, because I was not able to find any documentation explaining how to setup such scopes to restrict which token can invoke which remote method.

bajtos:
Question to consider during review:

 - Is the impact on existing applications acceptable? If not, then I'll have to introduce a feature flag to preserve backwards compatibility. Besides the extra complexity, such feature flag will leave applications insecure until they opt-in.

 - Is the user-facing API (`accessScope` property on shared methods, `scopes` as an array of strings) flexible enough to allow us to evolve this initial implementation closer towards OAuth2 scopes, while preserving backwards compatibility?

Please note that my intention is to get a minimal solution for scopes that would allow me to create access tokens scoped to a single password-setter function. Anything beyond that is out of scope of this pull request.

bajtos:
@ritch thank you for chiming in.

> Nothing for setting scopes?

For the purpose of password reset, the scope will be set when creating the password-reset token - see [User.resetPassword](https://github.com/strongloop/loopback/blob/048110ee01a8c62a2f13f72f07d25b3caceb3202/common/models/user.js#L708) and [User.prototype.createAccessToken](https://github.com/strongloop/loopback/blob/048110ee01a8c62a2f13f72f07d25b3caceb3202/common/models/user.js#L89) that eventually calls

```js
    this.accessTokens.create({
      ttl: ttl,
    }, cb);
```

>  I'm a bit apprehensive to move forward without docs on such a big feature.

I see your point. To be honest, I didn't really consider how to use this feature outside of LoopBack core.

Do you have any use case in mind where access-token scopes would be useful? (Other than OAuth2.)

Is this perhaps an indication that we should use a different mechanism for scoping down the password-reset token, i.e. do not treat the password-reset token as a regular access token, but treat it as a new domain entity, similarly how we treat email confirmation token now?

ebarault:
> Do you have any use case in mind where access-token scopes would be useful? (Other than OAuth2.)
It's a very powerful toolbox for fine grain access control if this feature lands in user domain (by opposition to keeping in internal to loopback) as it enables to create multiple access tokens for a same USER principal with different rights without adding/revoking roles.

I wonder if scopes as an embedded array will prevent some future evolutions compared to using an hasMany relations to a scope model.

@bajtos : are you just roughly reusing the concept of scopes from OAuth2 or are you seeking to align more with the standard itself?
My point is to check in the standard if some other part of the implementation should already evolve with this PR to prepare the ground for real OAuth2 tokens.

raymondfeng:
@bajtos Your PR is a good starting point.

The enhanced access token model is already available at https://github.com/strongloop/loopback-component-oauth2/blob/master/common/models/oauth-token.json as part of `oAuth2`. If we align the base user/accessToken model with `oAuth2` password flow, let's make sure they are consistent.

For `accessScope`, usually a method might allow one of a set of scopes. Should it be `string[]` instead?

test/authorization-scopes.test.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function logAllServerErrors() {
+    logServerErrorsOtherThan(-1);
+  }
+
+  function logServerErrorsOtherThan(statusCode) {
+    app.use((err, req, res, next) => {
+      if ((err.statusCode || 500) !== statusCode) {
+        console.log('Unhandled error for request %s %s:',
```
superkhau: missing one %s (or %j)?

bajtos:
Thank you all for your comments, this is a great discussion!

Here is an excerpt from OAuth2 specification describing the usage of AccessToken scopes

https://tools.ietf.org/html/rfc6749#section-3.3
> The authorization and token endpoints allow the client to specify the
scope of the access request using the "scope" request parameter.  In
turn, the authorization server uses the "scope" response parameter to
inform the client of the scope of the access token issued.
>
> The value of the scope parameter is expressed as a list of space-
delimited, case-sensitive strings.  The strings are defined by the
authorization server.  If the value contains multiple space-delimited
strings, their order does not matter, and each string adds an
additional access range to the requested scope.

Should we use the same format (a single string containing a space-delimited list of scopes) too?

On the other hand, our `OAuth2AccessToken` model uses an array of strings, as can be seen in [loopback-component-oauth2:common/models/oauth-token.json#L29](https://github.com/strongloop/loopback-component-oauth2/blob/afb3b48c78eb85ad414972168dfc332cde254f6c/common/models/oauth-token.json#L29)

@ebarault 
> I wonder if scopes as an embedded array will prevent some future evolutions compared to using an hasMany relations to a scope model.

I am pretty sure we cannot use `hasMany scopes` relation, because that's `1:n` relation, where the scope object contains id of the AccessToken it belongs to. I suppose we could use `hasAndBelongsToMany` relation where the pairs of (accessTokenId, scopeId) are stored in a new model/table, but it seems to me like an overkill.

> have you checked in the OAuth2 standard if there are any conceptual incompatibilities with the proposed implementation we could not easily move away later on?
> If we align the base user/accessToken model with oAuth2 password flow, let's make sure they are consistent.

👍 

AFAICT, the proposal in this patch should be consistent with oAuth2 flow as I understand it. @raymondfeng you are much more familiar with oAuth2 flow, do you see any potential inconsistency?

> For accessScope, usually a method might allow one of a set of scopes. Should it be `string[]` instead?

I'll trust your judgment here.

What is the scope check algorithm when both the access token and the protected method/resource define multiple scopes?

If I understand the oAuth2 quote above correctly, then my expectation is that the request is allowed as long as there is at least one scope in the access token that is also in the list defined by the protected method/resource. For example:

```js
// method scopes
accessScopes: ['write:user', 'write:user:password']

// token scopes
scopes: ['write:user', 'read:products', 'read:categories']

// -> request is allowed
```

bajtos:
> For accessScope, usually a method might allow one of a set of scopes. Should it be string[] instead?

I think the more important question to answer is how to handle scopes that are a superset of another scope. Examples:

https://developer.github.com/v3/oauth/#scopes
> `user` - Grants read/write access to profile info only. Note that this scope includes `user:email` and `user:follow`.
> `user:email` - Grants read access to a user's email addresses.
> `user:follow` - Grants access to follow or unfollow other users.

https://brandur.org/oauth-scope
> `read`: Read access to all a user’s apps and their subresources, except for protected subresources like config vars and releases.
> `write`: Write access to apps and unprotected subresources. Superset of `read`.
> `read-protected`: Read including protected subresources. Superset of `read`.
> `write-protected`: Write including protected subresources. Superset of `read-protected` and `write`.

If we allow remote methods to specify multiple scopes, then a poor-man's solution for superset/included scopes is to always specify all scopes which are allowed for the method, i.e. duplicate the superset/include relations in every shared method. While that doesn't like a good solution to me, perhaps it's a sign that as long as methods can define multiple scopes, then an incremental backwards-compatible solution for supserset/included scopes can be implemented later, to keep this initial pull request small and focused.

```js
// An example to illustrate my poor-man's solution
acessScopes: ['user', 'user:email']
```

Thoughts?

raymondfeng:
> If I understand the oAuth2 quote above correctly, then my expectation is that the request is allowed as long as there is at least one scope in the access token that is also in the list defined by the protected method/resource.

Yes.

The only missing piece against the oAuth2 access token is the optional `belongsTo` relation to `Application` model, which represents the client application.

bajtos:
> The only missing piece against the oAuth2 access token is the optional belongsTo relation to Application model, which represents the client application.

Great! I think this relation can be implemented in the future as an incremental addition, let's leave it out of scope of this pull request.

ebarault:
> I think the more important question to answer is how to handle scopes that are a superset of another scope.

Does the standard precise anything about that or it is left to implementation details?
In the absence of any strong directive, i upvote your proposal for a simple solution.

Should we make the function for access scopes hierarchy resolving overridable from the start, so that if someone wants to have a policy like "a parent includes any parent:child" he can easily implement it?

bajtos:
> Does the standard precise anything about that or it is left to implementation details?

AFAICT, the standard does not say anything about this.

> Should we make the function for access scopes hierarchy resolving overridable from the start, so that if someone wants to have a policy like "a parent includes any parent:child" he can easily implement it?

I prefer to leave that out of scope of this initial pull request to stay focused on fixing the password-reset flow.

bajtos:
> coverage/coveralls — Coverage decreased (-0.02%) to 89.305%

I believe this is caused by the new lines added to `AccessContext.prototype.debug`, which is invoked only when debug logs are enabled via `DEBUG` env var. Nothing to worry about IMO.

bajtos:
I pushed few new commits, I think all comments should be addressed now.

While working on the documentation (see https://github.com/strongloop/loopback.io/pull/338), I discovered one scenario that I consider important to cover: how to allow models to add new scopes to built-in methods contributed e.g. by a database datasource (`PersistedModel`). In 6f9ae2a, I added code to read `Model.settings.accessScopes` and add any scopes defined in that object to scopes already defined at shared-method level.

@raymondfeng @ritch @superkhau @ebarault I consider this pull request as done and ready for your final review. I'll squash the commits after I get your LGTMs.

ebarault:
Regarding your last comment, 
> how to allow models to add new scopes to built-in methods contributed e.g. by a database datasource (PersistedModel).

and looking at the doc draft, i just realized i misunderstood the way you want it to work: why should the scopes be defined at remote method declaration level rather than in ACLs definition?

I think it makes access control composition (access type, roles, properties, + scopes) more easy to handle if managed in a single place.

Take the example where someone has a `groupAdmin` role but not the `group:read` scope defined at access token level, and the developer is using scopes to grant read access to groups to some contributors. 

Dealing this at ACL level could enable for example to allow groupAdmin, based on role, to have access to group, while also allowing read access to users with the scope defined as token level, but without the role.

In the scoring algorithm, matching roles should give a higher score than matching scopes.

bajtos:
@ebarault 
> and looking at the doc draft, i just realized i misunderstood the way you want it to work

That makes me glad I wrote that documentation proposal!

> why should the scopes be defined at remote method declaration level rather than in ACLs definition

As I understand scopes, they are orthogonal to ACLs. I admit my understanding of ACLs and Scopes is limited and thus I may be wrong.

@raymondfeng you are our resident expert on authorization and oAuth2, could you please comment?

If I understand the proposal correctly, then a single method may require different scope depending on what role the requesting user is (which ACL entry is applied). I think that may be confusing - consider the following extension of your example:

```js
    {
      "accessType": "EXECUTE"
      "principalType": "ROLE",
      "principalId": "$owner",
      "scopes": ["read"]
      "permission": "ALLOW",
      "property": "findById"
    },
    {
      "accessType": "EXECUTE"
      "principalType": "USER",
      "principalId": "admin",
      "scopes": ["admin"]
      "permission": "ALLOW",
      "property": "findById"
    },
```

If the user is admin, then they must have "admin" scope to execute this method. If the user is an owner, then they need "read" scope to execute this method.

Is that intended? How should we handle scopes defined in ACL entries with `"permission": "DENY"`?

---

Could you please show me an example ACL config implementing the following requirement for password-reset flow?

 1. I want to create a new method, let's call it `User.setPassword(value, cb)`, that can be used to reset the password.
 2. I want to create a new scope, let's call it "reset:password", that will be used exclusively by this new method.
 3. I want to configure the scopes in such way, that access tokens with default scope created by `User.login` are NOT ALLOWED to execute this new method, regardless of the role the user is in.
 4. I want to change `User.resetPassword` method to generate a token with a single scope, `reset:password`. This token can invoke only the new method `setPassword` and cannot access any other API. Regardless of the role the token/user is in, as long as the user is authenticated.

ebarault:
I will provide a detailed response tomorrow, but meanwhile regarding:

> If I understand the proposal correctly, then a single method may require different scope depending on what role the requesting user is

No, roles and scopes stays independant in my proposal. I specifically chose the role $authenticated in the exemple. Another way would be to introduce a new type of principalType "scope" for ACLs complementing user, role and app. This would prevent confusions and reuse familiar concepts.

So to cover the case you submitted, as per my example, you just need to allow only the scope "reset:password" for role $authenticated on method setPassword.

I agree this gives a special treatment to the $authenticated role. So the further proposal for a new principalType.

In a nutshell, my opinion is that roles and scoped should be kept orthogonal, but the ACL formalism should be used to define scopes perimeter, reusing a well known concept. 

Don't take my proposal right as per the example, it needs to be reworked.

ebarault:
other examples where dealing with scope config in ACLs make sense:

```js
// dealing with multiple methods allowed for a given scope
{
      // ...
      "scopes": ["user:*"]
      "accessType": "EXECUTE"
      "permission": "ALLOW",
      "property": ["findById", "create", ... "someCustomMethod"]
}

// dealing with property access control with scopes (a la facebook)
{
      // ...
      "scopes": ["user:public"]
      "accessType": "READ"
      "permission": "ALLOW",
      "property": ["username", "firstName", "lastName", ... "someOtherProperty"]
}
```

raymondfeng:
Here is my view:

1. There are a few related concepts here:
   - Scopes group protected resources (method/property access)
   - Roles group principals (user/app/role)
   - ACLs controls if a given role can access a scope. At the moment, LoopBack goes directly to specific protected resources without `scope`. LoopBack also allows specific principals to be used in ACLs. My take is to simplify ACL with `(role(s), scope(s), allow/deny)`
   - AccessToken represents a list of scopes granted by the system/user to a given client application. It represents authorization delegation.

When a request comes in with a valid access token, the following should happen in LoopBack:
- Validate the access token and establish the security context such as (appId, userId, roles, scopes)
- Resolve the request to a protected resource, the corresponding ACLs, and a list of allowed scopes
- If the access token doesn't have any scope in the list of allowed scopes, the request will be denied. Please note this step checks the authorization delegation.
- Perform the ACL check to make sure the established principal can access the protected resource by matching roles to scopes. Please note this is the final authorization.

2. To implement the above: 
  - We need to have a way to define a set of scopes which serve as grouping of protected resources This will tie a given protected method to a list of allowed scopes. It can be done via two ways:
    - scopes: [{name: 's1', resources: [{model: 'User', property: 'create', ...}]}, ...] at model/app level
    - accessScopes: ['s1'] at the remoting method/property level to tag a protected resource with allowed scopes

  - Roles will stay as a way to group principals for authorization purpose.

  - AccessToken model will be extended to capture scopes and principals (user/app).

ebarault:
@raymondfeng if i read you correctly we would end with a typical model json definition looking like this: (i applied some tiny mods here)

```js
{
  "name": "myModel",
  "base": "someBaseModel",
  "properties": {
    // ...
  },
  // ...
  "scopes": [
    {
      "name": "scope1",
      "property": [
        "findById",
        "create"
      ]
    },
    {
      "name": "scope1",
      "property": "findById"
    }
  ],
  "acls": [
    {
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "ALLOW",
      "property": [
        "create",
        "find"
      ]
    }
  ],
  // ..
}
```

raymondfeng:
@ebarault Here is my version:
```js
{
  "name": "myModel",
  "base": "someBaseModel",
  "properties": {
    // ...
  },
  // ...
  "authScopes": [ // Use `authScopes` to differentiate from `scopes`, which are used for pre-defined server side constraints
    {
      "name": "scope1",
      "property": [
        "find",
        "create",
        "count"
      ]
    },
    {
      "name": "scope2",
      "property": "findById"
    }
  ],
  "acls": [
    {
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "principalType": "ROLE",
      "principalId": "myRole",
      "permission": "ALLOW",
      "property": [
        "create",
        "findById"
      ],
      "authScopes": ["scope2"] // Allow `myRole` to access `scope2`, which is `findById`. This flavor is preferred over `property`
    }
  ],
  // ..
}
```

ebarault:
@raymondfeng : ok, this shreds light on your previous message, thx. Am I right then that the last part of your version should be this? (removed the `property` param)
```js
  acls: [
  // ..
  {
    "principalType": "ROLE",
      "principalId": "myRole",
      "permission": "ALLOW",
      "authScopes": ["scope2"] // Allow `myRole` to access `scope2`, which is `findById`. This flavor is preferred over `property`
     }
  ]
```

**Please confirm**: 
- you suggest to remove the property `param` from the ACLs definition in favor of `authScopes`

then regarding the following authorization sequence:
> - If the access token doesn't have any scope in the list of allowed scopes, the request will be denied. Please note this step checks the authorization delegation.
> - Perform the ACL check to make sure the established principal can access the protected resource by matching roles to scopes. Please note this is the final authorization.

**Please confirm**: 
- a scope resolved for a given access token does not suffice to allow access to a method or property. The principal still needs to be authorized (USER/ROLE/APP) by ACLs

Then to illustrate: say we had to our example a third scope that **also** gives access to `findById`: 
```js
  "authScopes": [
    // ..
    {
      "name": "scope3",
      "property": "findById"
    }
  ],
```

then an accessToken that would resolve the scope **scope3** but not the role **myRole** would nevertheless not have access to `findById` in the example we followed, as **myRole** is not authorized **scope3**. Right?

raymondfeng:
@ebarault 
> Am I right then that the last part of your version should be this? (removed the property param)
  acls: [
  // ..
  {
    "principalType": "ROLE",
      "principalId": "myRole",
      "permission": "ALLOW",
      "authScopes": ["scope2"] // Allow `myRole` to access `scope2`, which is `findById`. This flavor is preferred over `property`
     }
  ]
Please confirm:
you suggest to remove the property param from the ACLs definition in favor of authScopes
then regarding the following authorization sequence:

**I prefer to have `authScopes` over `property`. But we might have to support both for some time.**

> the access token doesn't have any scope in the list of allowed scopes, the request will be denied. Please note this step checks the authorization delegation.
Perform the ACL check to make sure the established principal can access the protected resource by matching roles to scopes. Please note this is the final authorization.
Please confirm:
a scope resolved for a given access token does not suffice to allow access to a method or property. The principal still needs to be authorized (USER/ROLE/APP) by ACLs
Then to illustrate: say we had to our example a third scope that also gives access to findById:
  "authScopes": [
    // ..
    {
      "name": "scope3",
      "property": "findById"
    }
  ],
then an accessToken that would resolve the scope scope3 but not the role myRole would nevertheless not have access to findById in the example we followed, as myRole is not authorized scope3. Right?

**Yes. Having allowed scopes in the access token alone doesn't bypass ACL rules. The principals will still be checked against authScopes.**

ebarault:
@raymondfeng : yes, got you. I voluntarily occulted backward compatibility in my last statement on ACL's properties

Now to ease discussions on backward compatibility, i've put up this table to project what should be the policies to enforce when some of the params are missing in either AccessContext (resolved Access token) or ACLs definition
✔️ should be read: is available
❌ should be read: is not available

Resolved Token's scope | ACL `authScope` for principal | ACL `property` for principal | Policy
:---: | :---: | :---: | ---
✔️  | ✔️  | ✔️  | **grant allowed authScopes**
✔️  | ✔️  | ❌  | **grant allowed authScopes**
✔️  | ❌  | ✔️  | **deny request**
❌  | ✔️  | ✔️   | **grant allowed properties**
❌  | ✔️  | ❌   | **deny request**
❌  | ❌  | ✔️   | **grant allowed properties**

**To sum up:**
- From the moment a scope is resolved from the access token, only ACL's authScopes are granted, ACL's properties are ignored
- If no scope is resolved from the access token, only properties are granted, authScopes are ignored (this is the current behavior, kept for compatibility)

bajtos:
> From the moment a scope is resolved from the access token, only ACL's authScopes are granted, ACL's properties are ignored
> If no scope is resolved from the access token, only properties are granted, authScopes are ignored (this is the current behavior, kept for compatibility)

This is different from my understanding and I think it also contradicts Raymond's https://github.com/strongloop/loopback/pull/3313#issuecomment-291203805 that says

> - If the access token doesn't have any scope in the list of allowed scopes, the request will be denied. Please note this step checks the authorization delegation.
> - Perform the ACL check to make sure the established principal can access the protected resource by matching roles to scopes. Please note this is the final authorization.

I'd like to step back a bit here. The use case I need to address is how to make a scoped access token that allows only invocation of `setPassword(pass, cb)` method, while denying "regular" access tokens from invoking this special method.

To solve that, I proposed to allow remote methods to specify list of scopes that are required to invoke them. IIUC the discussion above correctly, we all agree on this part and there are no open points.

In order to allow me to move on with the security improvements for #382, I am proposing to limit the scope of this pull request only to scopes defined at remote-method level and discuss model-level configuration in a follow-up issue/pull request.

As can be seen from 6f9ae2a, the solution based on method-level definition of scopes is easy to extend in backwards-compatible way.

I am going to rework this pull request in that spirit to speed up the review process.

bajtos:
Updated. @raymondfeng @ebarault @superkhau PTAL again.

ebarault:
I'll review your updated PR asap @bajtos.

> This is different from my understanding and I think it also contradicts Raymond's #3313 (comment) that says

**Please rather consider the whole table rather than the quoted synthesis that just focus on backward compat**
I think is see how you misunderstood it, but i really don't think i am saying something incompatible to what @raymondfeng said.

These are two specific situations that can happen in the transition period where we need to make existing app still compatible, and deal with erroneous configs.

---

> "From the moment a scope is resolved from the access token, only ACL's authScopes are granted, ACL's properties are ignored" : **what i mean:**
  - the access token resolves a scope
  - user has a role
  - model has an ACL that binds that role with that scope -> methods referenced in that scope are granted to user
  - **even if the ACL also references methods directly (param `property`) : they are ignored (considered as an erroneous definition): "authScopes" is favored over `property`)**

---

> "If no scope is resolved from the access token, only properties are granted, authScopes are ignored (this is the current behavior, kept for compatibility)" : **what i mean:**
- the access token does not resolve any scope
- user has a role
- model has an ACL that binds that role with methods with the param `property`-> methods referenced in that ACL are granted to user 
- **even if the ACL also references one or several scopes (param `authScopes`) they are ignored (considered as an erroneous definition)**

bajtos:
@ebarault Thank you for the clarification, it is indeed possible that I have misunderstood your comment https://github.com/strongloop/loopback/pull/3313#issuecomment-291794186. I admit I find it difficult to understand the proposed ACL-based scope definitions in full. 

How do you envision the ACL-based scope definitions to play with scope definitions defined at remote-method level (in this pull request)? Do we need to make any changes in this patch to make sure ACL-based scope definitions can be implemented as an incremental addition in the future?

The last question is what matters to me most, as I want to stay focused on remote-method level scopes to unblock my work on #382.

lib/access-context.js
```diff
@@ -32,6 +32,7 @@ var debug = require('debug')('loopback:security:access-context');
  * @property {String} property The model property/method/relation name
  * @property {String} method The model method to be invoked
  * @property {String} accessType The access type: READ, REPLICATE, WRITE, or EXECUTE.
+ * @property {String[]} accessScopes The access scope required by the invoked method.
```
ebarault: "The access scopes that can authorize access to the invoked resource"

lib/access-context.js
```diff
@@ -193,6 +196,28 @@ AccessContext.prototype.isAuthenticated = function() {
...
+  if (!this.accessToken) return false;
+
+  // For backwards compatibility, tokens with no scopes are treated
+  // as if they have "DEFAULT" scope granted
+  const scopes = this.accessToken.scopes || ['DEFAULT'];
+
+  // For backwards compatibility, methods with no scopes defined
+  // are assigned a single "DEFAULT" scope
+  const requiredScopes = this.accessScopes || ['DEFAULT'];
+
```
ebarault: authorizedScopes is more appropriate

ebarault:
So, you removed the definition of scopes at model level ? that's too bad, i believe it was very close to be ok.

I went through your PR, and i believe this can evolve in backward compatible ways towards the vision we've built in this discussion.

**My main concern is the following**
People will generally implement scopes at model json definition level, as they do for ACLs. I am not expecting them to define them at method levels, and then navigating from ACLs to scopes defined in each method.

Plus, having 2 different places to define scopes we will have to handle considerations as: "how do we merge scopes defined at method level with scopes defined model level?" and "how to explain this in a way that does not confuse developers ?"

I'd have only implemented (or at least promoted) the definition of scopes at model level to keep it as simple as possible. It will make even more sense to define them only there when we will evolve the ACLs to reference scopes.

I left 2 cosmetic comments in the PR.

raymondfeng:
@ebarault I think you misunderstood my view a bit. Essentially, there are two perspectives for authorization scopes.

1. `authorizedScopes` in access token which represents the grant of limited permissions by the resource owner (user/system) to the client application. For example, to allow a mobile app to see my Facebook posts but not albums, the access token can be `(app1, 'raymond', ['posts:read:'])`.

2. `protectedResourceScopes` at the app/model config level which represents a list of protected resources, for example, `posts:read --> ['Post.find', 'Post.findById']`. It maps named permissions to access of protected resources. **We can even think `Post.find` as a fine protected resource scope which only contains `Post.find` method access.**

`ACL` controls if a principal (user/app/role) can access a protected resource (scope/method/property). For example:
   - `('role:$owner', scopes: ['posts:read'], 'allow')`
   - `('role:$owner', property: ['Post.find', 'Post.findById'], 'allow')` (The same as above)

Now let's look at two requests for `GET /api/users/raymond/posts?access_token=...`:
The resolved method for the REST call is `Post.find`. 

1) AccessToken is ('app1', 'raymond', ['posts:read'])
  - If `Post.find` is not tagged with any resource scopes, we'll only apply the `property` based ACL. In this case, it will be allowed as the access token has `raymond` maps to `$owner` role.
  - If `Post.find` is tagged as part of `posts:read` scope, we'll first check the authorizedScopes from the access token, which is in `['posts:read']`. We then apply ACL rules. In this case, it will be allowed as the access token has `raymond` which maps to `$owner` role.

2) AccessToken is ('app1', 'raymond', ['albums:read'])
  - If `Post.find` is not tagged with any resource scopes, we'll only apply the `property` based ACL. In this case, it will be allowed as the access token has `raymond`. No scope check is involved here.
  - If `Post.find` is tagged as part of `posts:read` scope, we'll first check the authorizedScopes from the access token, which is NOT in `['posts:read']`, it will be denied by we further check ACL rules. Tagging a method to a protected resource scope not only allows ACL to define rules against the scope, but also enforce the scope matching between AccessToken and targeted protected resource scope.

In summary, there will be two step check:
  - Check if authorizedScopes from the access token is in the list of allowed protected resource scopes if tagged for the target method invocation.
  - Check if established principal has authorization to perform the target method invocation

WDYT?

ebarault:
@raymondfeng : prior i can respond, could you review the last part of you previous comment?
i think there's something wring with 
>If Post.find is tagged as part of posts:read scope, we'll first check the authorizedScopes from the access token, which is NOT in ['posts:read'], _it will be denied **by we further** check ACL rules._ 

this is the essential part actually :-)

Also, regarding this part: 
> Tagging a method to a protected resource scope not only **allows** ACL to define rules **against** the scope but ...

I'm not sure how to take this with regard to 
> there will be two step check

which i understand as cumulative (**1 then 2**, not **1 or 2**)

if the two steps are cumulative, then ACL rules can't go against the method scope tagging.

raymondfeng:
@ebarault 
> it will be denied by we further check ACL rules.

This was a typo. Now I fixed it as `it will be denied even before we further check ACL rules.`

> It's a two step process.

1 and 2. If 1 fails, we will deny without trying 2. 

In my example, let's assume the ACL allows an owner to see posts and albums, but not delete a post (i.e., 3 scopes - `posts:read`, `posts:delete`, and `albums:read`). 

  - If the access token has `posts:read` scope for `GET .../posts`, the access to `Post.find` will be allowed during step 1 and step 2.
  - If the access token has `albums:read` scope for `GET .../posts`, the access to `Post.find` will be denied during step 1.
  - If the access token has `posts:delete` scope for `DELETE .../posts`, the access to `Post.delete` will be allowed during step 1 but denied in step 2.

ebarault:
@raymondfeng : so i can confirm we now speak 99% the same language 👍 
 
Only remaining subtlety : there's still something wrong for me with 
> Tagging a method to a protected resource scope not only allows ACL to define rules against the scope, but ...

Still IMO, with all this discussed, i can't see how an ACL could go against a scope as scope checkup denies in step 1.

---
**Regarding the confusion**: My intention was to use `protectedResourceScopes` as aliases for resources groups so that we can avoid duplicating ACLs like 
- `('role:$owner', scopes: ['posts:read'], 'allow')`
- `('role:$owner', property: ['Post.find', 'Post.findById'], 'allow')`

when `posts:read --> ['Post.find', 'Post.findById']`

A resource group alias would allow to keep just one. 

However i had a slightly different idea on how backward compatibility would work, which led me to reuse the protectedResourceScopes for this aliasing purpose.

Your proposal makes sense, **so let's go !**

ebarault:
that being said: do you have any particular comment regarding my earlier [comment](https://github.com/strongloop/loopback/pull/3313#issuecomment-291899004)?

settling this will allow finalizing this PR

raymondfeng:
@ebarault I see `resource group alias` the same as `protectedResourceScope`. To me, it is a named permission group to a set of protected resources (such as `Post.find` and `Post.findById`). Please note a protected resource is `(Model/Method/Property, READ/WRITE/EXECUTE)`. The concept is close to Java Permissions, such as:
-  https://docs.oracle.com/javase/7/docs/api/java/io/FilePermission.html
- https://docs.oracle.com/javase/7/docs/api/java/net/SocketPermission.html

To declare protectedResourceScopes, I see multiple ways:

1. At `app` level - server/config.json
2. At 'model` level - server/model-config.json
3. At `method/property` level - tagging the method/property with an array of scope names

We can start with 3 as the PR implements and introduce 1 & 2.

superkhau:
Code looks good, as for functionality, I don't have enough expertise in this area to give a good opinion. @raymondfeng's logic seems sound.

drmikecrowe:
Are there plans for loopback-swagger in place, or may I submit a PR for that project that supports these changes?

lib/access-context.js
bajtos: I see your point. I find `authorizedScopes` confusing, because it may also mean scopes the access token is authorized to access. I am going to use `resourceScopes` names, inspired by https://github.com/strongloop/loopback/pull/3313#issuecomment-291917954

bajtos:
@drmikecrowe thank you for chiming in.

> Are there plans for loopback-swagger in place, or may I submit a PR for that project that supports these changes?

I am not sure if this implementation of scopes is compatible with OAuth2 scopes enough to make sense to expose these scopes via Swagger. Could you please open an issue in loopback-swagger repository where we can discuss this proposal before you start writing code?

bajtos:
I created https://github.com/strongloop/loopback/issues/3339 for the follow-up discussion about defining scopes at app and model level.

bajtos:
@ebarault 
> i think you could have kept the code to define scopes at model level as it was close to be ok

My concern is that "close to be ok" may take too long to get into state when it can be landed and I don't want to delay this pull request much longer.

However, I reviewed the removed commit 52ef73c and cherry-picked the preparation/refactoring changes back into this pull request - see fab857d.

I have also addressed your earlier two comments, I squashed those two fixes in the first commit, sorry for that.

Does the patch still LGTY now?

common/models/acl.js
```diff
@@ -462,6 +462,16 @@ module.exports = function(ACL) {
       methodNames,
       registry: this.registry});
 
+    if (!context.isScopeAllowed()) {
+      req.permission = ACL.DENY;
+      debug('--Denied by scope config--');
+      debug('Scopes allowed:', context.accessToken.scopes || 'DEFAULT');
+      debug('Scope required:', context.getScopes() || 'DEFAULT');
+      context.debug();
```
ebarault: why `'DEFAULT'` rather than the accurate `['DEFAULT']` ?

bajtos: Good catch, this should have been using `DEFAULT_SCOPES` for the token, and no `||` clause for context scopes (`getScopes()` should be already handling the default case). I'll fix this ASAP.

bajtos:
> I suggest we add AccessToken.DEFAULT_SCOPE = 'DEFAULT' to avoid people having to manipulate directly the 'DEFAULT' string when creating access tokens

I personally see little value in such aliases, it's just one or another string to me. However, I am happy to add that such alias to make the code/API easier to use for others.

ebarault:
👏  🎉  @bajtos

Issue #3303, opened 2017-03-23
Provide a solution to finish the user's registration when the email verification token has expired


ebarault:
#### About
I'm opening this issue to isolate the discussion on the feature request in #382 referenced as
**"Provide solution for finishing the registration when the verification link was not followed (opened) before it has expired"**

My understanding, as depicted in the issue title is that we're willing provide a solution for the following situation:
- a user a completed the registration form to register
- an email is sent to the email address he provided so that he can confirm his email address
- the token provided in this email has somehow expired and the user cannot ask to regenerate a new token 

#### Related
#382

#### Discussion
As can be seen in [user.js#L528](https://github.com/strongloop/loopback/blob/master/common/models/user.js#L528) the built-in `User.generateVerificationToken()` method to generate a verification token does not generate an instance of AccessToken but just a simple random hash. This token is saved in the User instance.

Hence, the built-in `User.confirm()` method does a simple equality check on the verificationToken in [user.js#L550](https://github.com/strongloop/loopback/blob/master/common/models/user.js#L550)

👉  **the built-in implementation of the email verification feature does not manage a TTL on the generated token.**

Of course developers, when extending the User model can provide his own `User.generateVerificationToken()` and `User.confirm()` methods, dealing with a TTL check. And in that case a function to ask for a regeneration of this token makes sense.

But in the current context of LoopBack provided as-is the requested feature has not much meaning, unless we start by adding a TTL feature to email verification tokens.

Please @bajtos, @superkhau, @loay, @raymondfeng comment.

bajtos:
I'd like to invite people involved in the original discussion in #382 to chime in too: @offlinehacker @TFaga  @karlmikko @shaunbruno @johnsoftek @lazaridis-com

> Provide a solution to finish the user's registration when the email verification token has expired
>  the built-in implementation of the email verification feature does not manage a TTL on the generated token.

Good catch!

This leaves me a bit confused about what was mine reasoning for proposing this change in https://github.com/strongloop/loopback/issues/382#issuecomment-224954978.

How about the following two scenarios, is it something we should be prepared to handle?

 - A user a completed the registration form to register
 - An email is sent to the email address they provided so that they can confirm their email address
 - The email is not delivered, or deleted by spam filter, or deleted/forgotten by the user, because they change their mind about the registration
 - When they come back to the service much later, they realise they have already registered but did not verify the email.

Scenario 2:
 - The LoopBack app was initially configured to not require email verification.
[...]

PR #3314, opened 2017-03-28
Provide a solution to finish the user's registration when the email verification token has expired

ebarault:
### Description
Provide a solution to finish the user's registration when the email verification token has expired
Discussion: see related issue

#### Related issues
https://github.com/strongloop/loopback/issues/3303

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -183,9 +183,15 @@ module.exports = function(User) {
    *
    * ```js
    *    User.login({username: 'foo', password: 'bar'}, function (err, token) {
-  *      console.log(token.id);
-  *    });
+   *      console.log(token.id);
+   *    });
    * ```
+   *
+   * If the `emailVerificationRequired` flag is set for the inherited user model
+   * and the email has not yet been verified then the method will return a 401
+   * error that will contain the user's id. This id can be used to call the
+   * `api/verify` remote method to generate a new email verification token and
+   * send back the related email to the user.
    *
    * @param {Object} credentials username/password or email/password
    * @param {String[]|String} [include] Optionally set it to "user" to include
@@ -273,6 +279,9 @@ module.exports = function(User) {
               err = new Error(g.f('login failed as the email has not been verified'));
               err.statusCode = 401;
               err.code = 'LOGIN_FAILED_EMAIL_NOT_VERIFIED';
+              err.details = {
+                userId: user.id,
+              };
               fn(err);
             } else {
               if (user.createAccessToken.length === 2) {
@@ -527,42 +536,152 @@ module.exports = function(User) {
   };
 
   /**
-   * Verify a user's identity by sending them a confirmation email.
+   * Returns default verification options to use when calling User.prototype.verify()
+   * from remote method /user/:id/verify.
+   *
+   * NOTE: the User.getVerifyOptions() method can also be used to ease the
+   * building of identity verification options.
    *
    * ```js
-   *    var verifyOptions = {
-   *      type: 'email',
-   *      from: noreply@example.com,
-   *      template: 'verify.ejs',
-   *      redirect: '/',
-   *      tokenGenerator: function (user, cb) { cb("random-token"); }
-   *    };
+   * var verifyOptions = MyUser.getVerifyOptions();
+   * user.verify(verifyOptions);
+   * ```
    *
-   *    user.verify(verifyOptions, options, next);
+   * This is the full list of possible params, with example values
+   *
+   * ```js
+   * {
+   *   type: 'email',
+   *   mailer: {
+   *     send(verifyOptions, options, cb) {
+   *       // send the email
+   *       cb(err, result);
+   *     }
+   *   },
+   *   to: 'test@email.com',
+   *   from: 'noreply@email.com'
+   *   subject: 'verification email subject',
+   *   text: 'Please verify your email by opening this link in a web browser',
+   *   headers: {'Mime-Version': '1.0'},
+   *   template: 'path/to/template.ejs',
+   *   templateFn: function(verifyOptions, options, cb) {
+   *     cb(null, 'some body template');
+   *   }
+   *   redirect: '/',
+   *   verifyHref: 'http://localhost:3000/api/user/confirm',
+   *   host: 'localhost'
+   *   protocol: 'http'
+   *   port: 3000,
+   *   restApiRoot= '/api',
+   *   generateVerificationToken: function (user, options, cb) {
+   *     cb(null, 'random-token');
+   *   }
+   * }
+   * ```
+   *
+   * NOTE: param `to` internally defaults to user's email but can be overriden for
+   * test purposes or advanced customization.
+   *
+   * Static default params can be modified in your custom user model json definition
+   * using `settings.verifyOptions`. Any default param can be programmatically modified
+   * like follows:
+   *
+   * ```js
+   * customUserModel.getVerifyOptions = function() {
+   *   const base = MyUser.base.getVerifyOptions();
+   *   return Object.assign({}, base, {
+   *     // custom values
+   *   });
+   * }
+   * ```
+   *
+   * Usually you should only require to modify a subset of these params
+   * See `User.verify()` and `User.prototype.verify()` doc for params reference
+   * and their default values.
+   */
+
+  User.getVerifyOptions = function() {
+    const verifyOptions = {
+      type: 'email',
+      from: 'noreply@example.com',
+    };
+    return this.settings.verifyOptions || verifyOptions;
+  };
+
+  /**
+   * Verify a user's identity by sending them a confirmation message.
+   * NOTE: Currently only email verification is supported
+   *
+   * ```js
+   * var verifyOptions = {
+   *   type: 'email',
+   *   from: 'noreply@example.com'
+   *   template: 'verify.ejs',
+   *   redirect: '/',
+   *   generateVerificationToken: function (user, options, cb) {
+   *     cb('random-token');
+   *   }
+   * };
+   *
+   * user.verify(verifyOptions);
+   * ```
+   *
+   * NOTE: the User.getVerifyOptions() method can also be used to ease the
+   * building of identity verification options.
+   *
+   * ```js
+   * var verifyOptions = MyUser.getVerifyOptions();
+   * user.verify(verifyOptions);
    * ```
    *
    * @options {Object} verifyOptions
-   * @property {String} type Must be 'email'.
+   * @property {String} type Must be `'email'` in the current implementation.
+   * @property {Function} mailer A mailer function with a static `.send() method.
+   *  The `.send()` method must accept the verifyOptions object, the method's
+   *  remoting context options object and a callback function with `(err, email)`
+   *  as parameters.
+   *  Defaults to provided `userModel.email` function, or ultimately to LoopBack's
+   *  own mailer function.
    * @property {String} to Email address to which verification email is sent.
-   * @property {String} from Sender email addresss, for example
-   *   `'noreply@myapp.com'`.
+   *  Defaults to user's email. Can also be overriden to a static value for test
+   *  purposes.
+   * @property {String} from Sender email address
+   *  For example `'noreply@example.com'`.
    * @property {String} subject Subject line text.
+   *  Defaults to `'Thanks for Registering'` or a local equivalent.
    * @property {String} text Text of email.
-   * @property {String} template Name of template that displays verification
-   *  page, for example, `'verify.ejs'.
+   *  Defaults to `'Please verify your email by opening this link in a web browser:`
+   *  followed by the verify link.
+   * @property {Object} headers Email headers. None provided by default.
+   * @property {String} template Relative path of template that displays verification
+   *  page. Defaults to `'../../templates/verify.ejs'`.
    * @property {Function} templateFn A function generating the email HTML body
-   * from `verify()` options object and generated attributes like `options.verifyHref`.
-   * It must accept the option object and a callback function with `(err, html)`
-   * as parameters
+   *  from `verify()` options object and generated attributes like `options.verifyHref`.
+   *  It must accept the verifyOptions object, the method's remoting context options
+   *  object and a callback function with `(err, html)` as parameters.
+   *  A default templateFn function is provided, see `createVerificationEmailBody()`
+   *  for implementation details.
    * @property {String} redirect Page to which user will be redirected after
-   *  they verify their email, for example `'/'` for root URI.
+   *  they verify their email. Defaults to `'/'`.
+   * @property {String} verifyHref The link to include in the user's verify message.
+   *  Defaults to an url analog to:
+   *  `http://host:port/restApiRoot/userRestPath/confirm?uid=userId&redirect=/``
+   * @property {String} host The API host. Defaults to app's host or `localhost`.
+   * @property {String} protocol The API protocol. Defaults to `'http'`.
+   * @property {Number} port The API port. Defaults to app's port or `3000`.
+   * @property {String} restApiRoot The API root path. Defaults to app's restApiRoot
+   *  or `'/api'`
    * @property {Function} generateVerificationToken A function to be used to
-   *  generate the verification token. It must accept the user object and a
-   *  callback function. This function should NOT add the token to the user
-   *  object, instead simply execute the callback with the token! User saving
-   *  and email sending will be handled in the `verify()` method.
-   * @param {Object} options remote context options.
+   *  generate the verification token.
+   *  It must accept the verifyOptions object, the method's remoting context options
+   *  object and a callback function with `(err, hexStringBuffer)` as parameters.
+   *  This function should NOT add the token to the user object, instead simply
+   *  execute the callback with the token! User saving and email sending will be
+   *  handled in the `verify()` method.
+   *  A default token generation function is provided, see `generateVerificationToken()`
+   *  for implementation details.
    * @callback {Function} cb Callback function.
+   * @param {Object} options remote context options.
    * @param {Error} err Error object.
    * @param {Object} object Contains email, token, uid.
    * @promise
@@ -585,9 +704,11 @@ module.exports = function(User) {
 
     // Set a default template generation function if none provided
     verifyOptions.templateFn = verifyOptions.templateFn || createVerificationEmailBody;
+
     // Set a default token generation function if none provided
     verifyOptions.generateVerificationToken = verifyOptions.generateVerificationToken ||
       User.generateVerificationToken;
+
     // Set a default mailer function if none provided
     verifyOptions.mailer = verifyOptions.mailer || userModel.email ||
       registry.getModelByType(loopback.Email);
@@ -654,10 +775,8 @@ module.exports = function(User) {
     function sendEmail(user) {
       verifyOptions.verifyHref += '&token=' + user.verificationToken;
       verifyOptions.verificationToken = user.verificationToken;
-
       verifyOptions.text = verifyOptions.text || g.f('Please verify your email by opening ' +
         'this link in a web browser:\n\t%s', verifyOptions.verifyHref);
-
       verifyOptions.text = verifyOptions.text.replace(/\{href\}/g, verifyOptions.verifyHref);
 
       // argument "options" is passed depending on templateFn function requirements
@@ -1004,9 +1123,21 @@ module.exports = function(User) {
     );
 
     UserModel.remoteMethod(
+      'prototype.verify',
+      {
+        description: 'Trigger user\'s identity verification with configured verifyOptions',
+        accepts: [
+          {arg: 'verifyOptions', type: 'object', http: ctx => this.getVerifyOptions()},
+          {arg: 'options', type: 'object', http: 'optionsFromRequest'},
+        ],
+        http: {verb: 'post'},
+      }
+    );
+
+    UserModel.remoteMethod(
       'confirm',
       {
-        description: 'Confirm a user registration with email verification token.',
+        description: 'Confirm a user registration with identity verification token.',
         accepts: [
           {arg: 'uid', type: 'string', required: true},
           {arg: 'token', type: 'string', required: true},
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -854,7 +854,7 @@ describe('User', function() {
       User.settings.emailVerificationRequired = false;
     });
 
-    it('Require valid and complete credentials for email verification error', function(done) {
+    it('requires valid and complete credentials for email verification', function(done) {
       User.login({email: validCredentialsEmail}, function(err, accessToken) {
         // strongloop/loopback#931
         // error message should be "login failed"
@@ -862,12 +862,15 @@ describe('User', function() {
         assert(err && !/verified/.test(err.message),
           'expecting "login failed" error message, received: "' + err.message + '"');
         assert.equal(err.code, 'LOGIN_FAILED');
+        // as login is failing because of invalid credentials it should to return
+        // the user id in the error message
+        assert.equal(err.details, undefined);
 
         done();
       });
     });
 
-    it('Require valid and complete credentials for email verification error - promise variant',
+    it('requires valid and complete credentials for email verification - promise variant',
     function(done) {
       User.login({email: validCredentialsEmail})
         .then(function(accessToken) {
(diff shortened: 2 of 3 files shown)


ebarault:
First commit for draft review.
Needs tests.

common/models/user.js
```diff
@@ -428,6 +428,117 @@ module.exports = function(User) {
...
+   *  callback function. This function should NOT add the token to the user
+   *  object, instead simply execute the callback with the token! User saving
+   *  and email sending will be handled in the `verify()` method.
+   * @callback {Function} fn Callback function.
+   * @param {Error} err Error object.
+   * @param {Object} object Contains email, token, uid.
+   * @promise
+   */
+
+  User.verify = function(email, verifyOptions, options, cb) {
```
bajtos: I see two ways how to identify the user we want to verify: 1) by providing User's id 2) by providing User's email.

I would personally prefer to use `id`, I think it makes it more difficult to spam a specific user, because the spammer would have to know user's id. I may be wrong though.

What are the advantages of using email instead of id?

common/models/user.js
```diff
@@ -863,6 +975,18 @@ module.exports = function(User) {
     );
 
     UserModel.remoteMethod(
+      'verify',
+      {
+        description: 'Trigger user\'s email verification.',
+        accepts: [
+          {arg: 'email', type: 'string', required: true, http: {source: 'form'}},
+          {arg: 'options', type: 'object', http: 'optionsFromRequest'},
```
bajtos: +1 for not exposing `verifyOptions` to the clients.

common/models/user.js
```diff
@@ -428,6 +428,117 @@ module.exports = function(User) {
...
+   *  object, instead simply execute the callback with the token! User saving
+   *  and email sending will be handled in the `verify()` method.
+   * @callback {Function} fn Callback function.
+   * @param {Error} err Error object.
+   * @param {Object} object Contains email, token, uid.
+   * @promise
+   */
+
+  User.verify = function(email, verifyOptions, options, cb) {
+    verifyOptions = verifyOptions || User.verifyOptions;
```
bajtos: Can we push this down to `User.prototype.verify`, i.e. modify the prototype function to honour the default verify options too?

common/models/user.js
```diff
@@ -428,6 +428,117 @@ module.exports = function(User) {
...
+   *      restApiRoot= '/api',
+   *      tokenGenerator: function (user, cb) { cb("random-token"); }
+   *    };
+   *```
+   *
+   * Usually you should only require to modify a subset of these params
+   * See `User.verify()` and `User.prototype.verify()` doc for params reference
+   */
+
+  User.verifyOptions = {
```
bajtos: Let's discuss what's the best approach for setting up this object. Here are few alternatives I see:

 - `User.verifyOptions` as a property
 - `User.getVerifyOptions()` as a getter function
 - `User.settings.verifyOptions` to allow configuration via model.json

Use cases to consider:
 - How to allow custom models to overwrite inherited settings while preserving selected options from the base class
 - How to make it easy for app developers to provide the options

I think the following may be most flexible:

```js
User.getVerifyOptions() = function() { 
  return this.settings.verifyOptions || {type: 'email'};
}
```

Custom models then can partially overwrite these settings:

```js
MyUser.getVerifyOptions = function() {
  const base = MyUser.base.getVerifyOptions();
  return Object.assign({}, base, {
    // custom values
  });
}
```

bajtos:
@raymondfeng @ritch I'd like to hear your opinion on the changes proposed in this pull request too.

@ebarault I think the most important thing to do is to write (integration) tests showing all HTTP requests needed by the full login/verify workflow, including non-happy-path scenarios described in #3303.

common/models/user.js
ebarault: unless i misunderstood, User.verify() is meant to be a remote method.
I believe from a user Xp point of view, we expect him to input his email

common/models/user.js
ebarault: thanks for noting that. i was not sure about using a model property
i will move it as getVerifyOptions()

ebarault:
> have you considered exposing the existing prototype method via the REST API?

well, it's the same debate than using email or id in User.verify() : from a ux pov: many cases involve that the user cannot log in so the code in the client does not have a user instance or an id.

bajtos:
> well, it's the same debate than using email or id in User.verify() : from a ux pov: many cases involve that the user cannot log in so the code in the client does not have a user instance or an id.

Oh, I see. When `login` fails because the email is not verified, the user id is not returned in the error response. Would it make sense to modify `login` to include the `user id` in the case the login failed only because the email is not verified?

The built-in User supports both username and email addresses for logging in. So when I login with my username `bajtos`, we don't have any information about the email address I used to register with the site anyways. Which may be a problem - for example, consider a user that saves usernames/passwords in a password manager and users unique per-site email addresses, e.g. `joe.smith+github@gmail.com`.

I think both userId and email approaches have their pros and cons. @raymondfeng @ritch do you have any opinion on this matter to help us make a better decision?

@ebarault Could you perhaps implement this new method in https://github.com/strongloop/loopback-example-user-management together with all required routes and views, so that we can all try it out in a real app and see what approach works best?

ebarault:
> about using username, or email

overall i think there are implementation use cases for each one of them
maybe we should just offer the 2 options to resolve the user
**todo**: check impact of realms

> about using user id
i'm not a big fan that a public function returns any user id provided that someone has the email or the username. I'd rather object from a preventive security pov. who knows what vulnerabilities could be exploited based on user ids.

> Could you perhaps implement this new method in [..]

yes, before diving in the tests, **if** my previous point does not convince you, it might be a good idea

common/models/user.js
pierreclr: @ebarault Why not User.prototype.verify instead ?? 
However, IMO, this method should be internal only and be triggered on `User.observe('after save')` => case newInstance + user.verifyEmail is true

ebarault: @pierreclr 
User.prototype.verify already exists as an internal method
the debate here is to provide a remote method to ease implementing scenarios where the initial email verification link has been lost or has expired and needs to be generated again

bajtos:
> > about using user id

> i'm not a big fan that a public function returns any user id provided that someone has the email or the username. I'd rather object from a preventive security pov. who knows what vulnerabilities could be exploited based on user ids.

My idea is to return user id only after both username/login AND the password were verified. I.e. we have authenticated the user, but they are not allowed to access the site yet because their email was not verified.

On the second thought, I don't mind that much which approach we choose. I think it's more important to make it easy for application developers to do the right thing and to implement all required flows (scenarios) with minimal effort. That's why I suggested:

>  Could you perhaps implement this new method in https://github.com/strongloop/loopback-example-user-management together with all required routes and views, so that we can all try it out in a real app and see what approach works best?

ebarault:
> My idea is to return user id only after both username/login AND the password were verified. I.e. we have authenticated the user

ahh ok. we got somehow out of sync here :) yeah your proposal makes sense and is wrap all my security concerns. One in all, here's to recap what i'll implement:

- modify the login method to return the user Id in the error object when the login fails because of unverified email
- open a remote method for `User.verify(id)`

This can later be improved with a prototype method when support for scoped access tokens is landed

ebarault:
i went with the modifications we discussed
this is good for review @bajtos

common/models/user.js
```diff
@@ -434,6 +441,126 @@ module.exports = function(User) {
...
+    // Make sure to use the constructor of the (sub)class
+    // where the method is invoked from (`this` instead of `User`)
+    this.findById(uid, options, (err, user) => {
+      if (err) return cb(err);
+
+      if (!user) {
+        const err = new Error(`User ${uid} not found`);
+        Object.assign(err, {
+          code: 'USER_NOT_FOUND',
+          statusCode: 422, // or 404?
```
ebarault: should it be a 404 ?

common/models/user.js
```diff
@@ -434,6 +441,126 @@ module.exports = function(User) {
...
+   * }
+   * ```
+   *
+   * Usually you should only require to modify a subset of these params
+   * See `User.verify()` and `User.prototype.verify()` doc for params reference
+   */
+
+  User.getVerifyOptions = function() {
+    return this.settings.verifyOptions || {type: 'email'};
+  };
```
ebarault: the only missing param required to make the feature run OOTB is email `from`
maybe we should default it to something like 'noreply@noreply.com' in order to have OOTB support

also, based on our previous discussions, i implemented this with a possibility to leverage on this.settings.verifyOptions, but also described extensibility at method level.

i think it's fine to keep both in the doc, although extending via settings will not support the tokenGenerator param which expects a function. thoughts?

common/models/user.js
```diff
@@ -475,13 +602,13 @@ module.exports = function(User) {
    */
 
   User.prototype.verify = function(options, fn) {
+    options = options || this.getVerifyOptions();
```
ebarault: implementing options defaulting in prototype method as well.

if fine with you i will rename options as verifyOptions to make sure there's no confusion with REST options used in User.verify()

common/models/user.js
```diff
@@ -475,13 +606,16 @@ module.exports = function(User) {
...
 
     var user = this;
     var userModel = this.constructor;
     var registry = userModel.registry;
     var pkName = userModel.definition.idName() || 'id';
-    assert(typeof options === 'object', 'options required when calling user.verify()');
+    assert(typeof options === 'object', 'verifyOptions are required when calling ' +
+      'user.verify(), please verify the implementation of your user model\'s ' +
+      'getVerifyOptions() method');
     assert(options.type, 'You must supply a verification type (options.type)');
```
ebarault: now that we default options with `getVerifyOptions()` the only case where options can be undefined is when user does not extend correctly this method

common/models/user.js
```diff
@@ -255,6 +261,7 @@ module.exports = function(User) {
               err = new Error(g.f('login failed as the email has not been verified'));
               err.statusCode = 401;
               err.code = 'LOGIN_FAILED_EMAIL_NOT_VERIFIED';
+              err.userId = user.id;
```
bajtos: In production mode, strong-error-handler discards most error properties and keeps only `code` and `details`. As a result, the client won't be access `err.userId`.

My proposal:

```js
err.details = {
  userId: user.id,
};
```

common/models/user.js
```diff
@@ -434,6 +441,130 @@ module.exports = function(User) {
...
+   * without the `verifyOptions` param. This is specifically useful when calling
+   * this method remotely over rest.
+   *
+   * This is the full list of possible params, with example values
+   *
+   * ```js
+   * {
+   *   type: 'email',
+   *   mailer: anEmailModel,
+   *   to: 'test@email.com',
```
bajtos: This does not seem right to me - the recipient of the verification email should be determined based on the user account we are trying to verify.

common/models/user.js
bajtos: > i think we should default it to something like 'noreply@noreply.com' in order to provide OOTB support

I'd prefer to use `example.com` - that's a domain that should be safe to use.

http://example.com/
> This domain is established to be used for illustrative examples in documents. You may use this domain in examples without prior coordination or asking for permission.

We don't want users of LoopBack apps to start spamming a random domain like `noreply.com`.

> also, based on our previous discussions, i implemented this with a possibility to leverage on this.settings.verifyOptions, but i also described extensibility at model method level.
i think it's fine to keep both in the doc, although extending via settings will not support the tokenGenerator param which expects a function.

I like the flexibility provided by this approach 👍

common/models/user.js
bajtos: It depends on the URL. I would personally expose this new method as `http: {verb: 'post', path: '/:uid/verify'}`, in which case 404 makes perfect sense.

common/models/user.js
bajtos: +1 for renaming `options` to `verifyOptions` eventually.

We should eventually fix this method to support context-based `options` argument like other methods already do, and then `User.verify` should pass `options` down to this method too.

Ideally, the change fixing `User.prototype.verify` to support `options` argument should be implemented in a standalone pull request, accompanied by new tests, and landed before we proceed with this patch #3314.

common/models/user.js
```diff
@@ -869,9 +1003,21 @@ module.exports = function(User) {
     );
 
     UserModel.remoteMethod(
+      'verify',
+      {
+        description: 'Trigger user\'s identity verification.',
+        accepts: [
+          {arg: 'uid', type: 'string', required: true, http: {source: 'form'}},
```
bajtos: I am proposing to map this argument to a path-parameter, see my earlier https://github.com/strongloop/loopback/pull/3314/files#r110362900

What are the arguments for sending `uid` in the request body?

common/models/user.js
```diff
@@ -869,9 +1003,21 @@ module.exports = function(User) {
...
+          {arg: 'options', type: 'object', http: 'optionsFromRequest'},
+        ],
+        http: {verb: 'post', path: '/verify'},
```
bajtos: My preference:

```js
http: {verb: 'post', path: '/:uid/verify'},
```

common/models/user.js
```diff
@@ -869,9 +1003,21 @@ module.exports = function(User) {
...
+      }
+    );
+
+    UserModel.remoteMethod(
       'confirm',
       {
-        description: 'Confirm a user registration with email verification token.',
+        description: 'Confirm a user registration with identity verification token.',
```
bajtos: This change does not make sense to me. Isn't this method invoked when the user clicks on the link sent in the verification email? In which case it is an _email verification token_ IMO.

test/user.test.js
```diff
@@ -833,20 +833,23 @@ describe('User', function() {
       User.settings.emailVerificationRequired = false;
     });
 
-    it('Require valid and complete credentials for email verification error', function(done) {
+    it('require valid and complete credentials for email verification', function(done) {
```
bajtos: while we are changing the test title: `it('requires...`

test/user.test.js
```diff
@@ -833,20 +833,23 @@ describe('User', function() {
...
+        // as login is failing because of invalid credentials it should to return
+        // the user id in the error message
+        assert.equal(err.userId, undefined);
 
         done();
       });
     });
 
-    it('Require valid and complete credentials for email verification error - promise variant',
+    it('require valid and complete credentials for email verification - promise variant',
```
bajtos: ditto - `it requires...`

test/user.test.js
```diff
@@ -1405,6 +1408,31 @@ describe('User', function() {
...
+          'type': 'email',
+        });
+        done();
+      });
+
+      it('can be extended by user', function(done) {
+        User.getVerifyOptions = function() {
+          const base = User.base.getVerifyOptions();
+          return Object.assign({}, base, {
+            from: 'noreply@email.com',
```
bajtos: `noreply@example.com` - see http://loopback.io/doc/en/contrib/style-guide.html#email-examples

test/user.test.js
```diff
@@ -1921,6 +1949,75 @@ describe('User', function() {
       });
     });
 
+    describe('User.verify(id)', function() {
+      beforeEach(()=> {
+        User.getVerifyOptions = function() {
+          const base = User.base.getVerifyOptions();
+          return Object.assign({}, base, {
+            from: 'noreply@email.com',
```
bajtos: http://loopback.io/doc/en/contrib/style-guide.html#email-examples

test/user.test.js
```diff
@@ -1921,6 +1949,75 @@ describe('User', function() {
...
+          }
+        );
+      });
+
+      it('is called over REST method /User/verify', function() {
+        return request(app)
+          .post('/test-users')
+          .send({email: 'bar@bat.com', password: 'bar'})
+          .expect('Content-Type', /json/)
+          .expect(200)
```
bajtos: Since this test is checking the REST API of the verify function only, I'd prefer to create the user using regular JS API - it makes the test easier to read to me.

```js
return User.create(...).then(u => {
  return request(app)
    // ...
});
```

Thoughts?

bajtos:
Hi @ebarault, what's the status of this pull request? I see that you have pushed new commits and addressed some of my feedback. Is this ready for another round of review? Could you please rebase the patch on top of the latest `master` and resolve merge conflicts?

ebarault:
Hi @bajtos, i was waiting for my PR on options propagation in verify to be landed to resume. Now i probably won't have bandwidth to finalize it before early next week.

bajtos:
> i was waiting for my PR on options propagation in verify to be landed to resume.

Makes sense. Are you referring to #3344, which has been landed by now?

> Now i probably won't have bandwidth to finalize it before early next week.

No worries. Would you mind if I take this over, make the final changes myself and then ask you to review them? I am not sure if I'll have bandwidth before you, but I think it's worth to ask just in case :)

ebarault:
Let me find an hour tomorrow morning to do the rebase and merge then i can hand it over to you

common/models/user.js
ebarault: i know. But from the existing code, `to` can be configured. Maybe this was introduced for testing purposes. For sake of clarity we should make it explicit in the jsdoc that `to` defaults to user's email if available but can be overridden for testing purposes.

This makes me think: with the existing implementation a user with no email for whom the email verification method would be triggered would result in the app crashing when asserting `verifyOptions.to`, it's extreme: the app should not crash on such runtime scenarios.
This should be fixed outside the scope of this PR

common/models/user.js
ebarault: although this is the only user's identity verification option implemented as of today, the verify options is not meant to be used only to verify user's email. It could verify user's phone number for example (link sent via SMS). IMO the appropriate wording is identity verification token.

test/user.test.js
```diff
@@ -2019,6 +2070,88 @@ describe('User', function() {
...
+      it('forwards the "options" argument', function() {
+        const options = {testFlag: true};
+        const verifyOptions = User.getVerifyOptions();
+
+        // backup original user.verify() method
+        const originalMethod = User.prototype.verify;
+
+        // injecting mockup user.verify() method to ease option forwarding test
+        User.prototype.verify = function(verifyOptions, options, cb) {
+          expect(options).to.eql({testFlag: true});
```
bajtos: If `User.prototype.verify` is not called at all, then the test will incorrectly pass.

A better solution is to store `options` to `observedOptions` and then check the value in a `.then` block in the main test flow below.

common/models/user.js
```diff
@@ -568,6 +642,83 @@ module.exports = function(User) {
    * @promise
    */
 
+  User.verify = function(uid, verifyOptions, options, cb) {
```
bajtos: I am wondering whether we need to add this method, considering that the remoting layer is exposing `User.prototype.verify` now? It would make sense to me to add `User.verify` if that was the method exposed via REST, but this way I think it's not worth the future maintenance effort.

Thoughts?

test/user.test.js
```diff
@@ -879,42 +882,53 @@ describe('User', function() {
...
-          done();
-        })
-        .catch(function(err) {
-          assert(err);
-          assert.equal(err.code, 'LOGIN_FAILED_EMAIL_NOT_VERIFIED');
-
-          done();
-        });
+    it('does not login a user with unverified email but provides userId ' +
+      '- promise variant', function() {
```
bajtos: IMO, there is no need to test both callback and promise variants - there is no promise/callback specific code in path triggered by this test case, therefore it's enough to test a single variant (preferably the promise variant).

ebarault:
@bajtos thanks for your review.

> AFAICT, there are only two REST-based tests added, both testing/User/:id/verify. I think we should have also a test verifying that when login fails with LOGIN_FAILED_EMAIL_NOT_VERIFIED, the HTTP response data contains the userId. This test should fail with the original code setting err.userId and pass with err.details.userId. Thoughts?

👍  sure
 
> I am wondering whether we need to add this method, considering that the remoting layer is exposing User.prototype.verify now?

well, my initial intention was to provide both alternatives to developers to ease their further integration, but you had me on "future maintenance effort", so OK 

🖐  could you also check my comments [here](https://github.com/strongloop/loopback/pull/3314#discussion_r112663631)

common/models/user.js
bajtos: > i know. But from the existing code, to can be configured. Maybe this was introduced for testing purposes. For sake of clarity we should make it explicit in the jsdoc that to defaults to user's email if available but can be overridden for testing purposes.

I see, makes sense. Let's keep the `to` option in for backwards compatibility.

> with the existing implementation a user with no email for whom the email verification method would be triggered would result in the app crashing when asserting verifyOptions.to, it's extreme: the app should not crash on such runtime scenarios.
This should be fixed outside the scope of this PR

I agree the method should not be crashing, and I am fine with fixing that problem outside of this pull request.

ebarault:
@bajtos : PTAL. 
regarding
> I think we should have also a test verifying that when login fails with LOGIN_FAILED_EMAIL_NOT_VERIFIED, the HTTP response data contains the userId

I modified an existing test verifying that logins fails with LOGIN_FAILED_EMAIL_NOT_VERIFIED when email is not verified

common/models/user.js
```diff
@@ -545,25 +545,30 @@ module.exports = function(User) {
...
    * ```
    *
-   * The default options can be modified in your custom user model json definition
-   * using `settings.verifyOptions` or programmatically like follows:
+   * NOTE: param `to` internally defaults to user's email but can be overriden for
+   * test purposes or advanced customization.
+   *
+   * Static default params can be modified in your custom user model json definition
+   * using `settings.verifyOptions`. Any default param can be programmatically
+   * like follows:
```
bajtos: > Any default param can be programmatically like follows

I think you are missing a verb between "programmatically" and "like". I mean what can be programmatically done?

common/models/user.js
```diff
@@ -693,25 +617,51 @@ module.exports = function(User) {
    * ```
    *
    * @options {Object} verifyOptions
-   * @property {String} type Must be 'email'.
+   * @property {String} type Must be `'email'` in the current implementation.
+   * @property {Function} mailer A mailer function with a static `.send() method.
+   *  It must accept the verifyOptions object, the method'sremoting context options
+   *  object and a callback function with `(err, {email, verificationToken, uid})`
```
bajtos: I find this suspicious, are you sure that `Email.send` passes ` {email, verificationToken, uid}` to the callback? I thought it was passing the email MIME message or something similar.

test/user.test.js
```diff
@@ -991,7 +976,19 @@ describe('User', function() {
           if (err) return done(err);
 
           var errorResponse = res.body.error;
-          assert.equal(errorResponse.code, 'LOGIN_FAILED_EMAIL_NOT_VERIFIED');
+
+          // extracting code and details error response
+          let errorExcerpts = {
```
bajtos: Nitpick: I think one of `expect(Object.assign({}, err)).to.contain({...})` or `expect(Object.assign({}, err)).to.include({...})` should work equally well, but with less code.

See e.g. this exiting test:

```js
    it('fails with 401 for unknown users', () => {
      return User.setPassword('unknown-id', 'pass').then(
        success => { throw new Error('setPassword should have failed'); },
        err => {
          // workaround for chai problem
          //   object tested must be an array, an object, or a string,
          //   but error given
          const props = Object.assign({}, err);
          expect(props).to.contain({
            code: 'USER_NOT_FOUND',
            statusCode: 401,
          });
        });
    });
```

bajtos:
> > AFAICT, the test you modified is executing the method directly. What I am saying is that we need a test that will make an HTTP request
> in fact i was referring to this test: https://github.com/strongloop/loopback/pull/3314/files#diff-a202d1e6aac601eca4c15ad2418135fdR969

You are right, I did missed that you modified a test executing REST request. Consider my comment as resolved then.

common/models/user.js
```diff
@@ -545,25 +545,30 @@ module.exports = function(User) {
...
+   *   templateFn: function(verifyOptions, options, cb) { cb(null, "some body template"); }
    *   redirect: '/',
-   *   verifyHref: '',
+   *   verifyHref: 'http://localhost:3000/api/user/confirm',
    *   host: 'localhost'
    *   protocol: 'http'
    *   port: 3000,
    *   restApiRoot= '/api',
-   *   tokenGenerator: function (user, cb) { cb("random-token"); }
+   *   generateVerificationToken: function (user, cb) { cb(null, "random-token"); }
```
ebarault: thanks for noting this. I omitted this one.

common/models/user.js
ebarault: 👍  you are totally right, i got wrong when reading the cb signature, swapping with the overall verify cb

test/user.test.js
ebarault: no... already tried before fall-backing to this trick.
your proposal does not work with nested objects and leads to 
` Uncaught AssertionError: expected { Object (statusCode, name, ...) } to have a property 'details' of { Object (userId) }, but got { Object (userId) }`

ebarault: plus: errorResponse is not a real error object, Object.assign() should not be required either

bajtos: Oh, I remember, and `.deep.include` and `.deep.contain` does not work either. Bummer :(

Let's keep your current code then.

common/models/user.js
```diff
@@ -527,42 +536,143 @@ module.exports = function(User) {
...
    *
    * ```js
    *    var verifyOptions = {
    *      type: 'email',
-   *      from: noreply@example.com,
+   *      from: 'noreply@example.com'
    *      template: 'verify.ejs',
    *      redirect: '/',
-   *      tokenGenerator: function (user, cb) { cb("random-token"); }
+   *      tokenGenerator: function (user, options, cb) { cb("random-token"); }
```
bajtos: Nitpick: we use single quotes everywhere else. `cb('random-token')`.

bajtos:
> [cis-jenkins] downstream: loopback-getting-started-intermediate@master — Failed! (f779081)

That's a problem in our CI system, please ignore it.

```
[EnvInject] - Loading node environment variables.
FATAL: java.io.IOException: Unexpected termination of the channel
java.io.EOFException
	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2353)
```

ebarault:
landed! 🎉 , thanks @bajtos for your detailed reviews.

PR #3344, opened 2017-04-07
forward context options in user.verify

ebarault:
### Description

change original "options" argument name to "verifyOptions"
adds ctx options argument as "options"
forward context "options" down to relevant downstream functions
review "verifyOptions" assertions
overall code cleaning

#### Related issues

required by https://github.com/strongloop/loopback/pull/3314 (see [comment](https://github.com/strongloop/loopback/pull/3314#discussion_r110363193))

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -437,18 +437,18 @@ module.exports = function(User) {
    * Verify a user's identity by sending them a confirmation email.
    *
    * ```js
-   *    var options = {
+   *    var verifyOptions = {
    *      type: 'email',
-   *      to: user.email,
+   *      from: noreply@example.com,
    *      template: 'verify.ejs',
    *      redirect: '/',
    *      tokenGenerator: function (user, cb) { cb("random-token"); }
    *    };
    *
-   *    user.verify(options, next);
+   *    user.verify(verifyOptions, options, next);
    * ```
    *
-   * @options {Object} options
+   * @options {Object} verifyOptions
    * @property {String} type Must be 'email'.
    * @property {String} to Email address to which verification email is sent.
    * @property {String} from Sender email addresss, for example
@@ -468,130 +468,157 @@ module.exports = function(User) {
    *  callback function. This function should NOT add the token to the user
    *  object, instead simply execute the callback with the token! User saving
    *  and email sending will be handled in the `verify()` method.
-   * @callback {Function} fn Callback function.
+   * @param {Object} options remote context options.
+   * @callback {Function} cb Callback function.
    * @param {Error} err Error object.
    * @param {Object} object Contains email, token, uid.
    * @promise
    */
 
-  User.prototype.verify = function(options, fn) {
-    fn = fn || utils.createPromiseCallback();
+  User.prototype.verify = function(verifyOptions, options, cb) {
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
 
     var user = this;
     var userModel = this.constructor;
     var registry = userModel.registry;
-    var pkName = userModel.definition.idName() || 'id';
-    assert(typeof options === 'object', 'options required when calling user.verify()');
-    assert(options.type, 'You must supply a verification type (options.type)');
-    assert(options.type === 'email', 'Unsupported verification type');
-    assert(options.to || this.email,
-      'Must include options.to when calling user.verify() ' +
-      'or the user must have an email property');
-    assert(options.from, 'Must include options.from when calling user.verify()');
 
-    options.redirect = options.redirect || '/';
+    // final assertion is performed once all options are assigned
+    assert(typeof verifyOptions === 'object',
+      'verifyOptions object param required when calling user.verify()');
+
+    // Set a default template generation function if none provided
+    verifyOptions.templateFn = verifyOptions.templateFn || createVerificationEmailBody;
+    // Set a default token generation function if none provided
+    verifyOptions.generateVerificationToken = verifyOptions.generateVerificationToken ||
+      User.generateVerificationToken;
+    // Set a default mailer function if none provided
+    verifyOptions.mailer = verifyOptions.mailer || userModel.email ||
+      registry.getModelByType(loopback.Email);
+
+    var pkName = userModel.definition.idName() || 'id';
+    verifyOptions.redirect = verifyOptions.redirect || '/';
     var defaultTemplate = path.join(__dirname, '..', '..', 'templates', 'verify.ejs');
-    options.template = path.resolve(options.template || defaultTemplate);
-    options.user = this;
-    options.protocol = options.protocol || 'http';
+    verifyOptions.template = path.resolve(verifyOptions.template || defaultTemplate);
+    verifyOptions.user = user;
+    verifyOptions.protocol = verifyOptions.protocol || 'http';
 
     var app = userModel.app;
-    options.host = options.host || (app && app.get('host')) || 'localhost';
-    options.port = options.port || (app && app.get('port')) || 3000;
-    options.restApiRoot = options.restApiRoot || (app && app.get('restApiRoot')) || '/api';
+    verifyOptions.host = verifyOptions.host || (app && app.get('host')) || 'localhost';
+    verifyOptions.port = verifyOptions.port || (app && app.get('port')) || 3000;
+    verifyOptions.restApiRoot = verifyOptions.restApiRoot || (app && app.get('restApiRoot')) || '/api';
 
     var displayPort = (
-      (options.protocol === 'http' && options.port == '80') ||
-      (options.protocol === 'https' && options.port == '443')
-    ) ? '' : ':' + options.port;
+      (verifyOptions.protocol === 'http' && verifyOptions.port == '80') ||
+      (verifyOptions.protocol === 'https' && verifyOptions.port == '443')
+    ) ? '' : ':' + verifyOptions.port;
 
     var urlPath = joinUrlPath(
-      options.restApiRoot,
+      verifyOptions.restApiRoot,
       userModel.http.path,
       userModel.sharedClass.findMethodByName('confirm').http.path
     );
 
-    options.verifyHref = options.verifyHref ||
-      options.protocol +
+    verifyOptions.verifyHref = verifyOptions.verifyHref ||
+      verifyOptions.protocol +
       '://' +
-      options.host +
+      verifyOptions.host +
       displayPort +
       urlPath +
       '?' + qs.stringify({
-        uid: '' + options.user[pkName],
-        redirect: options.redirect,
+        uid: '' + verifyOptions.user[pkName],
+        redirect: verifyOptions.redirect,
       });
 
-    options.templateFn = options.templateFn || createVerificationEmailBody;
+    verifyOptions.to = verifyOptions.to || user.email;
+    verifyOptions.subject = verifyOptions.subject || g.f('Thanks for Registering');
+    verifyOptions.headers = verifyOptions.headers || {};
 
-    // Email model
-    var Email =
-      options.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
+    // assert the verifyOptions params that might have been badly defined
+    assertVerifyOptions(verifyOptions);
 
-    // Set a default token generation function if one is not provided
-    var tokenGenerator = options.generateVerificationToken || User.generateVerificationToken;
-    assert(typeof tokenGenerator === 'function', 'generateVerificationToken must be a function');
+    // argument "options" is passed depending on verifyOptions.generateVerificationToken function requirements
+    var tokenGenerator = verifyOptions.generateVerificationToken;
+    if (tokenGenerator.length == 3) {
+      tokenGenerator(user, options, addTokenToUserAndSave);
+    } else {
+      tokenGenerator(user, addTokenToUserAndSave);
+    }
 
-    tokenGenerator(user, function(err, token) {
-      if (err) { return fn(err); }
-
+    function addTokenToUserAndSave(err, token) {
+      if (err) return cb(err);
       user.verificationToken = token;
-      user.save(function(err) {
-        if (err) {
-          fn(err);
-        } else {
-          sendEmail(user);
-        }
+      user.save(options, function(err) {
+        if (err) return cb(err);
+        sendEmail(user);
       });
-    });
+    }
 
     // TODO - support more verification types
     function sendEmail(user) {
-      options.verifyHref += '&token=' + user.verificationToken;
+      verifyOptions.verifyHref += '&token=' + user.verificationToken;
+      verifyOptions.verificationToken = user.verificationToken;
 
-      options.verificationToken = user.verificationToken;
+      verifyOptions.text = verifyOptions.text || g.f('Please verify your email by opening ' +
+        'this link in a web browser:\n\t%s', verifyOptions.verifyHref);
 
-      options.text = options.text || g.f('Please verify your email by opening ' +
-        'this link in a web browser:\n\t%s', options.verifyHref);
+      verifyOptions.text = verifyOptions.text.replace(/\{href\}/g, verifyOptions.verifyHref);
 
-      options.text = options.text.replace(/\{href\}/g, options.verifyHref);
+      // argument "options" is passed depending on templateFn function requirements
+      var templateFn = verifyOptions.templateFn;
+      if (templateFn.length == 3) {
+        templateFn(verifyOptions, options, setHtmlContentAndSend);
+      } else {
+        templateFn(verifyOptions, setHtmlContentAndSend);
+      }
 
-      options.to = options.to || user.email;
+      function setHtmlContentAndSend(err, html) {
+        if (err) return cb(err);
 
-      options.subject = options.subject || g.f('Thanks for Registering');
+        verifyOptions.html = html;
 
-      options.headers = options.headers || {};
-
-      options.templateFn(options, function(err, html) {
-        if (err) {
-          fn(err);
-        } else {
-          setHtmlContentAndSend(html);
-        }
-      });
-
-      function setHtmlContentAndSend(html) {
-        options.html = html;
-
-        // Remove options.template to prevent rejection by certain
+        // Remove verifyOptions.template to prevent rejection by certain
         // nodemailer transport plugins.
-        delete options.template;
+        delete verifyOptions.template;
 
-        Email.send(options, function(err, email) {
-          if (err) {
-            fn(err);
-          } else {
-            fn(null, {email: email, token: user.verificationToken, uid: user[pkName]});
-          }
-        });
+        // argument "options" is passed depending on Email.send function requirements
+        var Email = verifyOptions.mailer;
+        if (Email.send.length == 3) {
+          Email.send(verifyOptions, options, handleAfterSend);
+        } else {
+          Email.send(verifyOptions, handleAfterSend);
+        }
+
+        function handleAfterSend(err, email) {
+          if (err) return cb(err);
+          cb(null, {email: email, token: user.verificationToken, uid: user[pkName]});
+        }
       }
     }
-    return fn.promise;
+
+    return cb.promise;
   };
 
-  function createVerificationEmailBody(options, cb) {
-    var template = loopback.template(options.template);
-    var body = template(options);
+  function assertVerifyOptions(verifyOptions) {
+    assert(verifyOptions.type, 'You must supply a verification type (verifyOptions.type)');
+    assert(verifyOptions.type === 'email', 'Unsupported verification type');
+    assert(verifyOptions.to, 'Must include verifyOptions.to when calling user.verify() ' +
+      'or the user must have an email property');
+    assert(verifyOptions.from, 'Must include verifyOptions.from when calling user.verify()');
+    assert(typeof verifyOptions.templateFn === 'function',
+      'templateFn must be a function');
+    assert(typeof verifyOptions.generateVerificationToken === 'function',
+      'generateVerificationToken must be a function');
+    assert(verifyOptions.mailer, 'A mailer function must be provided');
+    assert(typeof verifyOptions.mailer.send === 'function', 'mailer.send must be a function ');
+  }
+
+  function createVerificationEmailBody(verifyOptions, options, cb) {
+    var template = loopback.template(verifyOptions.template);
+    var body = template(verifyOptions);
     cb(null, body);
   }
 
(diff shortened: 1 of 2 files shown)


ebarault:
@bajtos 
commit https://github.com/strongloop/loopback/pull/3344/commits/9a8c9d17333bce26a9b36c72a3632117e0b5f401 addresses the PR's purpose

while commit https://github.com/strongloop/loopback/pull/3344/commits/5d7e23ce0bf2335b7aed5ecdd1157f5f5c359a2e addresses more general code refactor of user.verify() which is sooooo ugly right now....
What I did mostly is grouping all the asserting of the verifyOptions object once all the options are assigned.

The second inherently implies some code churn.
If you don't mind reviewing one then the other in the same PR, you'll make my day :1st_place_medal: . Otherwise i'll cherrypick the second one in another branch.

common/models/user.js
```diff
@@ -556,25 +560,19 @@ module.exports = function(User) {
...
         'this link in a web browser:\n\t%s', verifyOptions.verifyHref);
 
       verifyOptions.text = verifyOptions.text.replace(/\{href\}/g, verifyOptions.verifyHref);
 
-      verifyOptions.to = verifyOptions.to || user.email;
-
-      verifyOptions.subject = verifyOptions.subject || g.f('Thanks for Registering');
-
-      verifyOptions.headers = verifyOptions.headers || {};
-
```
ebarault: these 3 assignments should be done while still sync so we can call assertVerifyOptions() before going async
(here we are already async as we saved the user with new verification token)

common/models/user.js
```diff
@@ -599,9 +598,24 @@ module.exports = function(User) {
...
+      'or the user must have an email property');
+    assert(verifyOptions.from, 'Must include verifyOptions.from when calling user.verify()');
+    assert(typeof verifyOptions.templateFn === 'function',
+      'templateFn must be a function');
+    assert(typeof verifyOptions.generateVerificationToken === 'function',
+      'generateVerificationToken must be a function');
+    assert(verifyOptions.mailer, 'A mailer function must be provided');
+    assert(typeof verifyOptions.mailer.send === 'function', 'mailer.send must be a function ');
+  }
+
```
ebarault: verifyOptions assertions grouped outside the main block

common/models/user.js
```diff
@@ -527,17 +533,15 @@ module.exports = function(User) {
...
-      verifyOptions.mailer || this.constructor.email || registry.getModelByType(loopback.Email);
+    verifyOptions.to = verifyOptions.to || user.email;
+    verifyOptions.subject = verifyOptions.subject || g.f('Thanks for Registering');
+    verifyOptions.headers = verifyOptions.headers || {};
 
-    // Set a default token generation function if one is not provided
-    var tokenGenerator = verifyOptions.generateVerificationToken || User.generateVerificationToken;
-    assert(typeof tokenGenerator === 'function', 'generateVerificationToken must be a function');
+    // assert the verifyOptions params that might have been badly defined
+    assertVerifyOptions(verifyOptions);
```
ebarault: verifyOptions assertion is now done right before generating token and going async

common/models/user.js
```diff
@@ -556,25 +560,19 @@ module.exports = function(User) {
     // TODO - support more verification types
     function sendEmail(user) {
       verifyOptions.verifyHref += '&token=' + user.verificationToken;
-
       verifyOptions.verificationToken = user.verificationToken;
 
       verifyOptions.text = verifyOptions.text || g.f('Please verify your email by opening ' +
         'this link in a web browser:\n\t%s', verifyOptions.verifyHref);
 
       verifyOptions.text = verifyOptions.text.replace(/\{href\}/g, verifyOptions.verifyHref);
```
ebarault: these 4 assignments can't be done before since they depend on user toke generation
we could bring back verifyOptions assertion right after token generation but we'd sacrifice on performance in case of badly defined options

common/models/user.js
```diff
@@ -485,19 +485,25 @@ module.exports = function(User) {
...
+
+    // Set a default template generation function if none provided
+    verifyOptions.templateFn = verifyOptions.templateFn || createVerificationEmailBody;
+    // Set a default token generation function if none provided
+    verifyOptions.generateVerificationToken = verifyOptions.generateVerificationToken ||
+      User.generateVerificationToken;
+    // Set a default mailer function if none provided
+    verifyOptions.mailer = verifyOptions.mailer || userModel.email ||
+      registry.getModelByType(loopback.Email);
+
```
ebarault: grouping `verifyOptions` function defaulting logic at the top

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
-          fn(err);
-        } else {
-          sendEmail(user);
-        }
+      user.save(options, function(err) {
+        if (err) return cb(err);
+        sendEmail(user);
       });
-    });
+    }
```
ebarault: encapsulating actions done one user save in a callback function to ease tokenGenerator call with 2 or 3 args

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
-      });
+      // argument "options" is passed depending on templateFn function requirements
+      if (verifyOptions.templateFn.length == 3) {
+        verifyOptions.templateFn(verifyOptions, options, setHtmlContentAndSend);
+      } else {
+        verifyOptions.templateFn(verifyOptions, setHtmlContentAndSend);
+      }
+
+      function setHtmlContentAndSend(err, html) {
+        if (err) return cb(err);
```
ebarault: adding the handling of templateFn callback error inside this function to ease the call to templateFn with 2 or 3 args

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
+        if (Email.send.length == 3) {
+          Email.send(verifyOptions, options, handleAfterSend);
+        } else {
+          Email.send(verifyOptions, handleAfterSend);
+        }
+
+        function handleAfterSend(err, email) {
+          if (err) return cb(err);
+          cb(null, {email: email, token: user.verificationToken, uid: user[pkName]});
+        }
```
ebarault: encapsulating the handling of actions on email.send callback in this function to ease call to email.send with 2 or 3 args

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
+   * @param {Object} options remote context options
+   * @callback {Function} cb Callback function.
    * @param {Error} err Error object.
    * @param {Object} object Contains email, token, uid.
    * @promise
    */
 
-  User.prototype.verify = function(options, fn) {
-    fn = fn || utils.createPromiseCallback();
+  User.prototype.verify = function(verifyOptions, options, cb) {
```
ebarault: changing "options" to "verifyOption"
adding context options as "options"
changing "fn" to "cb"

common/models/user.js
```diff
@@ -468,130 +468,157 @@ module.exports = function(User) {
...
-          if (err) {
-            fn(err);
-          } else {
-            fn(null, {email: email, token: user.verificationToken, uid: user[pkName]});
-          }
-        });
+        function handleAfterSend(err, email) {
+          if (err) return cb(err);
+          cb(null, {email: email, token: user.verificationToken, uid: user[pkName]});
+        }
```
ebarault: encapsulating the handling of actions on email.send callback in this function to ease call to email.send with 2 or 3 args

common/models/user.js
```diff
@@ -468,130 +468,157 @@ module.exports = function(User) {
...
+      // argument "options" is passed depending on templateFn function requirements
+      var templateFn = verifyOptions.templateFn;
+      if (templateFn.length == 3) {
+        templateFn(verifyOptions, options, setHtmlContentAndSend);
+      } else {
+        templateFn(verifyOptions, setHtmlContentAndSend);
+      }
 
-      options.to = options.to || user.email;
+      function setHtmlContentAndSend(err, html) {
```
ebarault: adding the handling of templateFn callback error inside this function to ease the call to templateFn with 2 or 3 args

common/models/user.js
ebarault: adding the handling of templateFn callback error inside this function to ease the call to templateFn with 2 or 3 args

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
-          }
-        });
+        // argument "options" is passed depending on Email.send function requirements
+        if (Email.send.length == 3) {
+          Email.send(verifyOptions, options, handleAfterSend);
+        } else {
+          Email.send(verifyOptions, handleAfterSend);
+        }
+
+        function handleAfterSend(err, email) {
```
ebarault: encapsulating the handling of actions on email.send callback in this function to ease call to email.send with 2 or 3 args

bajtos:
Nice 👏 I am afraid I don't have bandwidth to review this pull request today, I'll do my best to review it in the next day or two.

common/models/user.js
```diff
@@ -468,130 +468,143 @@ module.exports = function(User) {
...
-            fn(err);
-          } else {
-            fn(null, {email: email, token: user.verificationToken, uid: user[pkName]});
-          }
-        });
+        // argument "options" is passed depending on Email.send function requirements
+        if (Email.send.length == 3) {
+          Email.send(verifyOptions, options, handleAfterSend);
+        } else {
+          Email.send(verifyOptions, handleAfterSend);
```
bajtos: `Email.send` does not support options yet - see https://github.com/strongloop/loopback/blob/95240598b3d7d3b365eaa184f53a0ae26a6aad14/lib/connectors/mail.js#L142

It would be nice to eventually fix Email's API and add support for options too. Let's leave that out of scope of this pull request though.

common/models/user.js
```diff
@@ -605,7 +618,7 @@ module.exports = function(User) {
    * @param {object} user The User this token is being generated for.
    * @param {Function} cb The generator must pass back the new token with this function call
    */
-  User.generateVerificationToken = function(user, cb) {
+  User.generateVerificationToken = function(user, options, cb) {
```
bajtos: Please describe the new `options` argument in the jsdoc comment.

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
+          type: 'email',
+          from: 'noreply@myapp.org',
+        };
+
+        // before save operation hook to capture ctx.options when saving
+        // verification token in user instance
+        User.observe('before save', function(ctx, next) {
+          if (!ctx.isNewInstance) {
+            // not checking equality since other properties are added by user.save()
+            expect(ctx.options).to.contain({testFlag: true});
```
bajtos: This is fragile. If the hook is not called then the assertion is never executed.

Here is a better way for testing the context passed to hooks: save the context to an array of observed contexts in the hook callback/handler and then assert the content of observed contexts after `user.verify` finished.

BTW you can install the observer after `User.create` finished, that way you don't have to check for `isNewInstance`.

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
+      });
+
+      it('forwards the "options" argument to a custom templateFn function', function() {
+        const ctxOptions = {testFlag: true};
+        const verifyOptions = {
+          type: 'email',
+          from: 'noreply@myapp.org',
+          templateFn: (verifyOptions, options, cb) => {
+            // not checking equality since other properties are added by user.save()
+            expect(options).to.contain({testFlag: true});
```
bajtos: ditto - assert after `user.verify()` finished

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
+      });
+
+      it('forwards the "options" argment to a custom token generator function', function() {
+        const ctxOptions = {testFlag: true};
+        const verifyOptions = {
+          type: 'email',
+          from: 'noreply@example.com',
+          generateVerificationToken: function(user, options, cb) {
+            // not checking equality since other properties are added by user.save()
+            expect(options).to.contain({testFlag: true});
```
bajtos: ditto - assert after `user.verify()` finished

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
+      });
+
+      it('forwards the "options" argument to a custom mailer function', function() {
+        const ctxOptions = {testFlag: true};
+
+        // custom mailer function accepting the options argument
+        const mailer = function() {};
+        mailer.send = function(verifyOptions, options, cb) {
+          // not checking equality since other properties are added by user.save()
+          expect(options).to.contain({testFlag: true});
```
bajtos: ditto - assert after `user.verify()` finished

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
           });
       });
+
+      it('forwards the "options" argument to user.save() ' +
+        'when adding verification token', function() {
+        const ctxOptions = {testFlag: true};
+        const verifyOptions = {
+          type: 'email',
+          from: 'noreply@myapp.org',
+        };
```
bajtos: This block is repeated in every test, extract it in a shared variable please.

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
...
+        // verification token in user instance
+        User.observe('before save', function(ctx, next) {
+          if (!ctx.isNewInstance) {
+            // not checking equality since other properties are added by user.save()
+            expect(ctx.options).to.contain({testFlag: true});
+          }
+          next();
+        });
+
+        return User.create({email: 'test@example.com', password: 'pass'})
```
bajtos: This block is repeated in every test, let's extract it into a shared piece please. At minimum, extract the credentials. I think it would be even better to move `User.create` into a `beforeEach` hook.

```js
let user;
beforeEach(givenUser);

// tests ...

function givenUser() {
  return User.create(...).then(u => user = u);
}
```

test/user.test.js
```diff
@@ -1919,6 +1920,89 @@ describe('User', function() {
             expect(token).to.exist();
           });
       });
+
+      it('forwards the "options" argument to user.save() ' +
+        'when adding verification token', function() {
+        const ctxOptions = {testFlag: true};
```
bajtos: `ctxOptions` are shared by all test cases, please extract into a shared constant.

common/models/user.js
```diff
@@ -468,130 +468,157 @@ module.exports = function(User) {
...
+        // nodemailer transport plugins.
+        delete verifyOptions.template;
 
-      options.templateFn(options, function(err, html) {
-        if (err) {
-          fn(err);
+        // argument "options" is passed depending on Email.send function requirements
+        var Email = verifyOptions.mailer;
+        if (Email.send.length == 3) {
+          Email.send(verifyOptions, options, handleAfterSend);
```
bajtos: Email.send does not support options yet - see https://github.com/strongloop/loopback/blob/95240598b3d7d3b365eaa184f53a0ae26a6aad14/lib/connectors/mail.js#L142

It would be nice to eventually fix Email's API and add support for options too. Let's leave that out of scope of this pull request though.

ebarault:
@bajtos : thanks for reviewing 👍 . Your change requests look good to me. I'm in the middle of something right now, but will try applying tomorrow to keep it fresh.

test/user.test.js
```diff
@@ -90,14 +92,13 @@ describe('User', function() {
...
-        validCredentialsEmailVerifiedUser = user;
-        done();
-      });
+    // create 2 users: with and without verified email
+    return Promise.map(
+      [validCredentials, validCredentialsEmailVerified],
+      credentials => User.create(credentials)
+    ).then(users => {
+      validCredentialsUser = user = users[0];
+      validCredentialsEmailVerifiedUser = users[1];
```
ebarault: refactoring cb -> promises when creating users
assigning a shortcut `user` to use in tests rather than the heavy `validCredentialsUser`

test/user.test.js
```diff
@@ -1406,19 +1407,21 @@ describe('User', function() {
...
+      let verifyOptions;
+
+      beforeEach(function() {
+        // reset verifyOptions before each test
+        verifyOptions = {
+          type: 'email',
+          from: 'noreply@example.org',
+        };
+      });
+
```
ebarault: base verifyOptions used in all tests
overloaded specifically in each test

test/user.test.js
```diff
@@ -1406,19 +1407,21 @@ describe('User', function() {
...
 
-          var verifyOptions = {
-            type: 'email',
-            to: user.email,
-            from: 'noreply@myapp.org',
-            redirect: '/',
-            protocol: ctx.req.protocol,
-            host: ctx.req.get('host'),
-          };
-
```
ebarault: refactoring existing tests to use the shared `verifyOptions`
we can simplify as most of these options are defaulted in `user.verify` implementation : `redirect`, `to`, `protocol`, `host`

test/user.test.js
```diff
@@ -1484,15 +1478,7 @@ describe('User', function() {
...
-          var verifyOptions = {
-            type: 'email',
-            to: user.email,
-            from: 'noreply@myapp.org',
-            redirect: '/',
-            protocol: ctx.req.protocol,
-            host: ctx.req.get('host'),
-            headers: {'message-id': 'custom-header-value'},
-          };
+          verifyOptions.headers = {'message-id': 'custom-header-value'};
```
ebarault: actually only the `headers` are not standard in this test

test/user.test.js
```diff
@@ -1516,16 +1502,8 @@ describe('User', function() {
...
-            to: user.email,
-            from: 'noreply@myapp.org',
-            redirect: '/',
-            protocol: ctx.req.protocol,
-            host: ctx.req.get('host'),
-            templateFn: function(verifyOptions, cb) {
-              cb(null, 'custom template  - verify url: ' + verifyOptions.verifyHref);
-            },
+          verifyOptions.templateFn = function(verifyOptions, cb) {
+            cb(null, 'custom template  - verify url: ' + verifyOptions.verifyHref);
```
ebarault: only the `templateFn` function is not standard

test/user.test.js
```diff
@@ -1901,118 +1819,105 @@ describe('User', function() {
...
+          cb(null, 'dummy body');
         };
 
-        return User.create({email: 'test@example.com', password: 'pass'})
-          .then(user => {
-            return user.verify(verifyOptions, ctxOptions);
+        return user.verify(verifyOptions, ctxOptions)
+          .then(() => {
+            // not checking equality since other properties are added by user.save()
+            expect(templateFnOptions).to.contain({testFlag: true});
```
ebarault: now checking options once verify is done

test/user.test.js
```diff
@@ -1901,118 +1819,105 @@ describe('User', function() {
...
-      var options;
+    describe('User.confirm(verifyOptions, fn)', function() {
+      var verifyOptions;
 
       function testConfirm(testFunc, done) {
         User.afterRemote('create', function(ctx, user, next) {
           assert(user, 'afterRemote should include result');
 
-          options = {
+          verifyOptions = {
```
ebarault: changing options name to verifyOptions here too so be consistent

test/user.test.js
```diff
@@ -1862,36 +1789,27 @@ describe('User', function() {
...
+
+        Object.assign(verifyOptions, {
+          redirect: '#/some-path?a=1&b=2',
+          templateFn: (verifyOptions, cb) => {
+            actualVerifyHref = verifyOptions.verifyHref;
+            cb(null, 'dummy body');
+          },
+        });
+
+        return user.verify(verifyOptions)
```
ebarault: reusing one user created in the main beforeEach script

test/user.test.js
```diff
@@ -1901,118 +1819,105 @@ describe('User', function() {
...
+        Object.assign(verifyOptions, {
+          redirect: '#/some-path?a=1&b=2',
+          templateFn: (verifyOptions, cb) => {
+            actualVerificationToken = verifyOptions.verificationToken;
+            cb(null, 'dummy body');
+          },
+        });
+
+        return user.verify(verifyOptions)
+          .then(() => actualVerificationToken)
```
ebarault: ditto

test/user.test.js
```diff
@@ -1901,118 +1819,105 @@ describe('User', function() {
...
+        // custom templateFn function accepting the options argument
+        verifyOptions.templateFn = (verifyOptions, options, cb) => {
+          templateFnOptions = options;
+          cb(null, 'dummy body');
         };
 
-        return User.create({email: 'test@example.com', password: 'pass'})
-          .then(user => {
-            return user.verify(verifyOptions, ctxOptions);
+        return user.verify(verifyOptions, ctxOptions)
```
ebarault: ...

test/user.test.js
```diff
@@ -1828,17 +1760,12 @@ describe('User', function() {
...
           assert(user, 'afterRemote should include result');
 
-          var options = {
-            type: 'email',
-            to: user.email,
-            from: 'noreply@myapp.org',
-            redirect: '/',
+          Object.assign(verifyOptions, {
             host: 'myapp.org',
-            port: 3000,
```
bajtos: I think it's better to keep explicit `port: 3000`, because we are asserting this port number few lines below (`http://myapp.org:3000/test-users/confirm`).

When the default port is used, it's not clear to readers of this test where is the port 3000 coming from.

test/user.test.js
```diff
@@ -1858,77 +1785,139 @@ describe('User', function() {
...
-              redirect: '#/some-path?a=1&b=2',
-              templateFn: (options, cb) => {
-                actualVerificationToken = options.verificationToken;
-                cb(null, 'dummy body');
-              },
-            })
-            .then(() => actualVerificationToken);
-          })
+      it('verify that verifyOptions.templateFn receives ' +
+        'verifyOptions.verificationToken', function() {
```
bajtos: I think the following would be more consistent with our style:

```js
it('verifies that verifyOptions.templateFn receives verifyOptions.verificationToken',
function() {
```

test/user.test.js
```diff
@@ -1858,77 +1785,139 @@ describe('User', function() {
...
+        // verification token in user instance
+        User.observe('before save', function(ctx, next) {
+          onBeforeSaveCtx = ctx;
+          next();
+        });
+
+        return user.verify(verifyOptions, ctxOptions)
+          .then(() => {
+            // not checking equality since other properties are added by user.save()
+            expect(onBeforeSaveCtx.options).to.contain({testFlag: true});
```
bajtos: If the hook was not called, then this would throw a TypeError because `onBeforeSaveCtx` is not defined. A possible solution: initalise `onBeforeSaveCtx` with an empty object.

ebarault:
@bajtos : 2 downstream tests are failing, i can't see why, PTAL

ssh24:
@ebarault Looks like the `make` command failed for `node-gyp`.

@slnode test please

bajtos:
> `[cis-jenkins] PR Builder — Build finished.`

This is the "master" job that triggers build of dependent jobs like `x64 && linux && nvm.4` and all `downstream:`-prefixed jobs. It passes when all dependent pass, fails when at least dependent failed.

> `[cis-jenkins] downstream: gateway-director-bluemix@develop`

This build is failing because the binary addon in `node-rdkafka@0.10.0` cannot be compiled. Since this problem is not related to LoopBack changes at all, we can safely ignore this particular build failure.

bajtos:
Landed, thank you @ebarault for your contribution!

Issue #3349, opened 2017-04-10
New password-reset flow: docs, examples, project template


bajtos:
- [ ] Update https://github.com/strongloop/loopback-example-user-management 
   - https://github.com/strongloop/loopback-example-user-management/pull/58
- [ ] Update our documentation to reflect the changes made in https://github.com/strongloop/loopback/issues/382. Include a guide showing users how to correctly set up password-related endpoints in LoopBack applications.
  (internal task: https://github.com/strongloop-internal/scrum-asteroid/issues/131)
- [ ] Modify the template in loopback-workspace to enable the new, more secure mode
  (internal task: https://github.com/strongloop-internal/scrum-asteroid/issues/132)
- [ ] Print a warning when the feature flag is unset, to force users to make a decision whether to upgrade or stay behind.
  (internal task: https://github.com/strongloop-internal/scrum-asteroid/issues/133)
- [ ] Announce these changes to our ANN mailing list

superkhau:
Create new timeboxed story.

bajtos:
I don't think we will have bandwidth to do this, considering that LB 4.0 has been released.

PR #3350, opened 2017-04-10
Set password with token, disable password changes via patch/replace

bajtos:
### Description

**Add `User.setPassword(id, new, cb)`**

Implement a new method for changing user password with password-reset token but without the old password.

REST API

    POST /api/users/reset-password
    Authorization: your-password-reset-token-id
    Content-Type: application/json

    {"newPassword": "new-pass"}

JavaScript API

    User.setPassword(userId, newPassword[, cb])
    userInstance.setPassword(newPassword[, cb])

**Implement more secure password flow**

Improve the flow for setting/changing/resetting User password to make
it more secure.

 1. Modify `User.resetPassword` to create a token scoped to allow
    invocation of a single remote method: `User.setPassword`.

 2. Scope the method `User.setPassword` so that regular tokens created
    by `User.login` are not allowed to execute it.

For backwards compatibility, this new mode (flow) is enabled only
when User model setting `restrictResetPasswordTokenScope` is set to `true`.

 3. Changing the password via `User.prototype.patchAttributes`
    (and similar DAO methods) is no longer allowed. Applications
    must call `User.changePassword` and ask the user to provide
    the current (old) password.

For backwards compatibility, this new mode (flow) is enabled only
when User model setting `rejectPasswordChangesViaPatchOrReplace` is set to `true`.

#### Related issues

- Connect to #382
- Follow-up issue to write documentation, update the example repo and workspace templates: #3349 

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -421,15 +421,82 @@ module.exports = function(User) {
         return cb(err);
       }
 
-      try {
-        User.validatePassword(newPassword);
-      } catch (err) {
+      this.setPassword(newPassword, options, cb);
+    });
+    return cb.promise;
+  };
+
+  /**
+   * Set this user's password after a password-reset request was made.
+   *
+   * @param {*} userId Id of the user changing the password
+   * @param {string} newPassword The new password to use.
+   * @param {Object} [options] Additional options including remoting context
+   * @callback {Function} callback
+   * @param {Error} err Error object
+   * @promise
+   */
+  User.setPassword = function(userId, newPassword, options, cb) {
+    assert(userId != null && userId !== '', 'userId is a required argument');
+    assert(!!newPassword, 'newPassword is a required argument');
+
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
+
+    // Make sure to use the constructor of the (sub)class
+    // where the method is invoked from (`this` instead of `User`)
+    this.findById(userId, options, (err, inst) => {
+      if (err) return cb(err);
+
+      if (!inst) {
+        const err = new Error(`User ${userId} not found`);
+        Object.assign(err, {
+          code: 'USER_NOT_FOUND',
+          statusCode: 401,
+        });
         return cb(err);
       }
 
-      const delta = {password: newPassword};
-      this.patchAttributes(delta, options, (err, updated) => cb(err));
+      inst.setPassword(newPassword, options, cb);
     });
+
+    return cb.promise;
+  };
+
+  /**
+   * Set this user's password. The callers of this method
+   * must ensure the client making the request is authorized
+   * to change the password, typically by providing the correct
+   * current password or a password-reset token.
+   *
+   * @param {string} newPassword The new password to use.
+   * @param {Object} [options] Additional options including remoting context
+   * @callback {Function} callback
+   * @param {Error} err Error object
+   * @promise
+   */
+  User.prototype.setPassword = function(newPassword, options, cb) {
+    assert(!!newPassword, 'newPassword is a required argument');
+
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
+
+    try {
+      this.constructor.validatePassword(newPassword);
+    } catch (err) {
+      cb(err);
+      return cb.promise;
+    }
+
+    const delta = {password: newPassword};
+    this.patchAttributes(delta, options, (err, updated) => cb(err));
+
     return cb.promise;
   };
 
@@ -933,6 +1000,21 @@ module.exports = function(User) {
           {arg: 'options', type: 'object', http: 'optionsFromRequest'},
         ],
         http: {verb: 'POST', path: '/change-password'},
+      }
+    );
+
+    UserModel.remoteMethod(
+      'setPassword',
+      {
+        description: 'Reset user\'s password via a password-reset token.',
+        accepts: [
+          {arg: 'id', type: 'any',
+            http: ctx => ctx.req.accessToken && ctx.req.accessToken.userId,
+          },
+          {arg: 'newPassword', type: 'string', required: true, http: {source: 'form'}},
+          {arg: 'options', type: 'object', http: 'optionsFromRequest'},
+        ],
+        http: {verb: 'POST', path: '/reset-password'},
       }
     );
 
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -79,12 +79,26 @@ module.exports = function(User) {
    * Create access token for the logged in user. This method can be overridden to
    * customize how access tokens are generated
    *
-   * @param {Number} ttl The requested ttl
-   * @param {Object} [options] The options for access token, such as scope, appId
+   * Supported flavours:
+   *
+   * ```js
+   * createAccessToken(ttl, cb)
+   * createAccessToken(ttl, options, cb);
+   * createAccessToken(options, cb);
+   * // recent addition:
+   * createAccessToken(data, options, cb);
+   * ```
+   *
+   * @options {Number|Object} [ttl|data] Either the requested ttl,
+   * or an object with token properties to set (see below).
+   * @property {Number} [ttl] The requested ttl
+   * @property {String[]} [scopes] The access scopes granted to the token.
+   * @param {Object} [options] Additional options including remoting context
    * @callback {Function} cb The callback function
    * @param {String|Error} err The error string or object
    * @param {AccessToken} token The generated access token object
    * @promise
+   *
    */
   User.prototype.createAccessToken = function(ttl, options, cb) {
     if (cb === undefined && typeof options === 'function') {
@@ -95,17 +109,21 @@ module.exports = function(User) {
 
     cb = cb || utils.createPromiseCallback();
 
-    if (typeof ttl === 'object' && !options) {
-      // createAccessToken(options, cb)
-      options = ttl;
-      ttl = options.ttl;
+    let tokenData;
+    if (typeof ttl !== 'object') {
+      // createAccessToken(ttl[, options], cb)
+      tokenData = {ttl};
+    } else if (options) {
+      // createAccessToken(data, options, cb)
+      tokenData = ttl;
+    } else {
+      // createAccessToken(options, cb);
+      tokenData = {};
     }
-    options = options || {};
-    var userModel = this.constructor;
-    ttl = Math.min(ttl || userModel.settings.ttl, userModel.settings.maxTTL);
-    this.accessTokens.create({
-      ttl: ttl,
-    }, cb);
+
+    var userSettings = this.constructor.settings;
+    tokenData.ttl = Math.min(tokenData.ttl || userSettings.ttl, userSettings.maxTTL);
+    this.accessTokens.create(tokenData, options, cb);
     return cb.promise;
   };
 
@@ -494,6 +512,14 @@ module.exports = function(User) {
       return cb.promise;
     }
 
+    // We need to modify options passed to patchAttributes, but we don't want
+    // to modify the original options object passed to us by setPassword caller
+    options = Object.assign({}, options);
+
+    // patchAttributes() does not allow callers to modify the password property
+    // unless "options.setPassword" is set.
+    options.setPassword = true;
+
     const delta = {password: newPassword};
     this.patchAttributes(delta, options, (err, updated) => cb(err));
 
@@ -806,7 +832,21 @@ module.exports = function(User) {
         return cb(err);
       }
 
-      user.createAccessToken(ttl, function(err, accessToken) {
+      if (UserModel.settings.restrictResetPasswordTokenScope) {
+        const tokenData = {
+          ttl: ttl,
+          scopes: ['reset-password'],
+        };
+        user.createAccessToken(tokenData, options, onTokenCreated);
+      } else {
+        // We need to preserve backwards-compatibility with
+        // user-supplied implementations of "createAccessToken"
+        // that may not support "options" argument (we have such
+        // examples in our test suite).
+        user.createAccessToken(ttl, onTokenCreated);
+      }
+
+      function onTokenCreated(err, accessToken) {
         if (err) {
           return cb(err);
         }
@@ -817,7 +857,7 @@ module.exports = function(User) {
           user: user,
           options: options,
         });
-      });
+      }
     });
 
     return cb.promise;
@@ -1003,6 +1043,9 @@ module.exports = function(User) {
       }
     );
 
+    const setPasswordScopes = UserModel.settings.restrictResetPasswordTokenScope ?
+      ['reset-password'] : undefined;
+
     UserModel.remoteMethod(
       'setPassword',
       {
@@ -1014,6 +1057,7 @@ module.exports = function(User) {
           {arg: 'newPassword', type: 'string', required: true, http: {source: 'form'}},
           {arg: 'options', type: 'object', http: 'optionsFromRequest'},
         ],
+        accessScopes: setPasswordScopes,
         http: {verb: 'POST', path: '/reset-password'},
       }
     );
@@ -1077,6 +1121,46 @@ module.exports = function(User) {
       ctx.query.where.email = ctx.query.where.email.toLowerCase();
     }
     next();
+  });
+
+  User.observe('before save', function rejectInsecurePasswordChange(ctx, next) {
+    const UserModel = ctx.Model;
+    if (!UserModel.settings.rejectPasswordChangesViaPatchOrReplace) {
+      // In legacy password flow, any DAO method can change the password
+      return next();
+    }
+
+    if (ctx.isNewInstance) {
+      // The password can be always set when creating a new User instance
+      return next();
+    }
+    const data = ctx.data || ctx.instance;
+    const isPasswordChange = 'password' in data;
+
+    // This is the option set by `setPassword()` API
+    // when calling `this.patchAttritubes()` to change user's password
+    if (ctx.options.setPassword) {
+      // Verify that only the password is changed and nothing more or less.
+      if (Object.keys(data).length > 1 || !isPasswordChange) {
+        // This is a programmer's error, use the default status code 500
+        return next(new Error(
+          'Invalid use of "options.setPassword". Only "password" can be ' +
+          'changed when using this option.'));
+      }
+
+      return next();
+    }
+
+    if (!isPasswordChange) {
+      return next();
+    }
+
+    const err = new Error(
+      'Changing user password via patch/replace API is not allowed. ' +
+      'Use changePassword() or setPassword() instead.');
+    err.statusCode = 401;
+    err.code = 'PASSWORD_CHANGE_NOT_ALLOWED';
+    next(err);
   });
 
   User.observe('before save', function prepareForTokenInvalidation(ctx, next) {
--- a/test/authorization-scopes.test.js
+++ b/test/authorization-scopes.test.js
@@ -82,7 +82,7 @@ describe('Authorization scopes', () => {
   }
 
   function givenRemoteMethodWithCustomScope() {
-    // Delete any previosly registered instance of the method "scoped"
+    // Delete any previously registered instance of the method "scoped"
     User.sharedClass._methods = User.sharedClass._methods
       .filter(m => m.name !== 'scoped');
 
--- a/test/authorization-scopes.test.js
+++ b/test/authorization-scopes.test.js
@@ -82,6 +82,10 @@ describe('Authorization scopes', () => {
   }
 
   function givenRemoteMethodWithCustomScope() {
+    // Delete any previosly registered instance of the method "scoped"
+    User.sharedClass._methods = User.sharedClass._methods
+      .filter(m => m.name !== 'scoped');
+
     const accessScopes = arguments[0] || [CUSTOM_SCOPE];
     User.scoped = function(cb) { cb(); };
     User.remoteMethod('scoped', {
--- a/test/user-password.test.js
+++ b/test/user-password.test.js
@@ -1,0 +1,242 @@
+// Copyright IBM Corp. 2013,2017. All Rights Reserved.
+// Node module: loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+const expect = require('./helpers/expect');
+const errorHandler = require('strong-error-handler');
+const loopback = require('../');
+const Promise = require('bluebird');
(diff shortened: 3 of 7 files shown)


common/models/user.js
```diff
@@ -421,15 +422,83 @@ module.exports = function(User) {
...
+   *
+   * @param {AccessToken} resetToken The token created by resetPassword().
+   * @param {string} newPassword The new password to use.
+   * @param {object} [options]
+   * @callback {Function} callback
+   * @param {Error} err Error object
+   * @promise
+   */
+  User.setPassword = function(userId, newPassword, options, cb) {
+    assert(!!userId, 'resetToken is a required argument');
```
raymondfeng: It should say `userId` is a required argument.

We should probably test `!!userId || userId ===0` so that `0` can be valid `userId`

common/models/user.js
```diff
@@ -421,15 +422,83 @@ module.exports = function(User) {
...
+
+    if (cb === undefined && typeof options === 'function') {
+      cb = options;
+      options = undefined;
+    }
+    cb = cb || utils.createPromiseCallback();
+
+    // Make sure to use the constructor of the (sub)class
+    // where the method is invoked from (`this` instead of `User`)
+    this.findById(userId, options, (err, inst) => {
```
raymondfeng: IIUC, `userId` should be `resetToken.userId`, right?

common/models/user.js
```diff
@@ -969,6 +1067,47 @@ module.exports = function(User) {
...
+
+    if (ctx.isNewInstance) {
+      // The password can be always set when creating a new User instance
+      return next();
+    }
+    const isPasswordChange = 'password' in (ctx.data || ctx.instance);
+
+    if (ctx.options.setPassword) {
+      // This is the option used by `setPassword()` API
+      // to turn off this password-change check
```
ebarault: i find the comment ambiguous
why not something like
"This is the option set by `setPassword()` API when calling this.patchAttritubes() to change user's password" ?

common/models/user.js
```diff
@@ -421,15 +422,83 @@ module.exports = function(User) {
...
+
+    try {
+      this.constructor.validatePassword(newPassword);
+    } catch (err) {
+      cb(err);
+      return cb.promise;
+    }
+
+    const delta = {password: newPassword};
+    options = Object.assign({setPassword: true}, options);
```
ebarault: please add a comment to reflect why the setPassword option is set
maybe something like "setPassword option is set to true to indicate to downstream operation hooks the origin of the call"
(BTW: isn't there a PR to reflect in operation hooks the upstream call stack?)

curiosity: any particular reason why not the reversed construction?
`Object.assign(options, {setPassword: true});`

test/user-password.test.js
```diff
@@ -0,0 +1,218 @@
...
+    });
+
+    it('allows setPassword-like call with non-password changes', () => {
+      return patchNameAndPasswordDirectly().then(() => {
+        // test passed
+      });
+    });
+  });
+
+  function givenAppWithModernFlow() {
```
ebarault: Can we find something more explicit than Modern vs. Legacy? (what if we change again)

test/user.test.js
```diff
@@ -2171,6 +2278,44 @@ describe('User', function() {
...
+          property: 'testMethod',
+        });
+
+        return triggerPasswordReset(options.email)
+          .then(info => request(app)
+            .get(`/test-users/${info.user.id}/test`)
+            .set('Authorization', info.accessToken.id)
+            .expect(200));
+      });
+
```
ebarault: IIUC these tests are for the legacy call flow. 

I don't see the two opposite tests for the new call flow that
- creates token that prevent patching User with new password
- creates token that allow calling setPassword 
- creates token that prevent calling other methods than setPassword

test/user-password.test.js
```diff
@@ -0,0 +1,218 @@
...
+        return resetPassword(resetToken).expect(204);
+      });
+    });
+
+    it('allows setPassword-like call with non-password changes', () => {
+      return patchNameAndPasswordDirectly().then(() => {
+        // test passed
+      });
+    });
+  });
```
ebarault: very compact style 👍

test/user.integration.js
```diff
@@ -178,6 +180,64 @@ describe('users - integration', function() {
...
+    it('rejects unauthenticated reset password request', function() {
+      return this.post('/api/users/reset-password')
+        .send({
+          newPassword: 'new password',
+        })
+        .expect(401);
+    });
+
+    it('injects reset password options from remoting context', function() {
+      const User = app.models.User;
```
ebarault: forwards vs. injects? (we shall decide which term to use)

test/user.test.js
```diff
@@ -1404,6 +1425,92 @@ describe('User', function() {
...
+          //   but error given
+          const props = Object.assign({}, err);
+          expect(props).to.contain({
+            code: 'USER_NOT_FOUND',
+            statusCode: 401,
+          });
+        });
+    });
+
+    it('forwards the "options" argument', () => {
```
ebarault: see my previous comment, here it's: forwards

common/models/user.js
bajtos: > please add a comment to reflect why the setPassword option is set

Makes sense, will do.

> BTW: isn't there a PR to reflect in operation hooks the upstream call stack?

Not AFAIK.


> curiosity: any particular reason why not the reversed construction?
`Object.assign(options, {setPassword: true});`

I don't want to modify the `options` object that was provided us by the caller.

bajtos:
https://github.com/strongloop/loopback/pull/3350#discussion_r110965803
> i find the createAccessToken() method signature confusing  (...)

Good point 💯 

I am proposing the change the signature of `createAccessToken` in a different way then. Let's keep the currently supported forms:

```js
createAccessToken(ttl, cb)
createAccessToken(ttl, options, cb);
createAccessToken(options, cb);
```

and add a new form that accepts a data object as the first arg:

```
createAccessToken(data, options, cb);
```

thoughts?

@raymondfeng

common/models/user.js
bajtos: See f585d8

common/models/user.js
```diff
@@ -711,7 +780,17 @@ module.exports = function(User) {
         return cb(err);
       }
 
-      user.createAccessToken(ttl, function(err, accessToken) {
+      if (UserModel.settings.legacyPasswordFlow === false) {
+        const params = {
+          ttl: ttl,
+          scopes: ['reset-password'],
+        };
+        user.createAccessToken(params, onTokenCreated);
```
bajtos: See https://github.com/strongloop/loopback/pull/3350#issuecomment-293905818 and 745d20a

common/models/user.js
```diff
@@ -969,6 +1067,47 @@ module.exports = function(User) {
...
+    if (UserModel.settings.legacyPasswordFlow !== false) {
+      // In legacyPasswordFlow, any DAO method can change the password
+      return next();
+    }
+
+    if (ctx.isNewInstance) {
+      // The password can be always set when creating a new User instance
+      return next();
+    }
+    const isPasswordChange = 'password' in (ctx.data || ctx.instance);
```
bajtos: What do you mean by string search? Could you paste a code snippet showing what do you have in mind?

bajtos:
@ebarault thanks for your thoughtful review! I have addressed the first two comments, PTAL. I'll work on the remaining comments later. Most likely I won't get to this patch until next week, because Good Friday is a Public Holiday here in CZ.

test/authorization-scopes.test.js
```diff
@@ -82,6 +82,10 @@ describe('Authorization scopes', () => {
   }
 
   function givenRemoteMethodWithCustomScope() {
+    // Delete any previosly registered instance of the method "scoped"
```
bajtos: I am afraid I don't see the typo, could you please be more specific?

test/user-password.test.js
bajtos: I see your point. Do you have any proposal in mind?

Note that the `User` setting is called `legacyPasswordFlow` right now. This setting controls whether:
 - `POST /api/users/reset-password` is scoped to `['reset-password']`
 - The token created by `POST /api/users/reset` is granted `['reset-password']` scope
 - User's "before save" hooks rejects attempts to change the password via patch/replace

I have difficulties coming with a descriptive name that would cover all of that.

test/user.test.js
bajtos: I am afraid I don't understand. Which previous comment do you mean?

test/user.test.js
bajtos: > these tests are for the legacy call flow.

Yes, I am keeping them here to ensure backward-compatibility is preserved. The test cases you have pointed out are covered by `test/user-password.test.js`:

> creates token that prevent patching User with new password

See https://github.com/strongloop/loopback/blob/745d20a4c9d1161206ff6023a8f97a6abfae6b5a/test/user-password.test.js#L30-L32

> creates token that allow calling setPassword

https://github.com/strongloop/loopback/blob/745d20a4c9d1161206ff6023a8f97a6abfae6b5a/test/user-password.test.js#L58-L60

> creates token that prevent calling other methods than setPassword

https://github.com/strongloop/loopback/blob/745d20a4c9d1161206ff6023a8f97a6abfae6b5a/test/user-password.test.js#L46-L56

test/user.test.js
bajtos: I think I found the other comment: https://github.com/strongloop/loopback/pull/3350#discussion_r110979869

Let me explain why I think we should keep using both "forwards" and "injects":

The test `it forwards the "options" argument` verifies that the implementation of `User.setPassword` is forwarding the`options` argument whenever calling other User and DAO methods, to ensure that context can be correctly propagated.

The second test, "it injects reset password options from remoting context" verifies that remoting metadata of `setPassword` provides correct configuration for the `options` argument, a configuration instructing loopback/strong-remoting to inject options built from the request context.

In my mind, these two test cases are different, and therefore it makes sense to me to use different verbs to describe them.

test/user.integration.js
bajtos: In the light of the discussion under https://github.com/strongloop/loopback/pull/3350#discussion_r110980156, maybe a better test name would be:

```js
it('configures "options" argument to be injected from remoting context'
```

Thoughts?

bajtos:
@ebarault I addressed all of your review comments, either by changing the relevant code, or by starting a discussion. Could you PTAL again?

test/authorization-scopes.test.js
loay: previosly-->previously

bajtos: Ah, I see it now. Thank you @loay 🙇

test/user-password.test.js
ebarault: i think we had a discussion a few month later in another PR
Do you have arguments against naming such options and tests according to versioning and add descriptions in the changes log? (a dedicated page on options is required at this point IMO)
e.g. `passwordFlow@3.5.0` (the version being the first LB version introducing the feature)

test/user.integration.js
ebarault: i reviewed your other comment, let's keep it simple: i agree with **inject** for the root call, and **forwards** for downstream calls

test/user.test.js
ebarault: 👍

test/user.test.js
ebarault: ok, thanks for pointing out these tests
to ease further reading, could you add `context` blocks to segregate tests with "legacy" password flow and tests with "modern" password flow?, as in `user-password.test.js`

common/models/user.js
```diff
@@ -421,15 +439,90 @@ module.exports = function(User) {
...
+    try {
+      this.constructor.validatePassword(newPassword);
+    } catch (err) {
+      cb(err);
+      return cb.promise;
+    }
+
+    // We need to modify options passed to patchAttributes, but we don't want
+    // to modify the original options object passed to us by setPassword caller
+    options = Object.assign({}, options);
```
bajtos: Consider the following code:

```js
const options = { foo: 'bar' };
user.setPassword('pass', options, cb);
console.log(options);
```

In my proposed implementation, the snippet prints `{ foo: 'bar' }`.

Without the `Object.assign` call, the snippet would print `{ foo: 'bar', setPassword: true }`.

In other words, the code inside `setPassword` is changing the value of the `options` argument (the reference to the options object), but it does not mutate the original object referenced by the initial `options` value.

I hope my comment have clarified the problem, please do keep asking if there is anything still not clear yet.

test/user-password.test.js
bajtos: > Do you have arguments against naming such options and tests according to versioning and add descriptions in the changes log? (a dedicated page on options is required at this point IMO)
e.g. `passwordFlow@2.y.z`, `passwordFlow@3.6.0` (the version being the first LB version introducing the feature)

I am not sure if I understand you correctly. What is the model setting name and what are the values that you are proposing?

In general, I prefer feature flags that describe the behaviour of the application. If we cannot come up with a descriptive flag name for this situation, then I guess the second best option is to introduce two feature flags:

 - `scopePasswordResetToken` controlling the behaviour of `POST /api/users/reset-password` and `POST /api/users/reset`
 - `rejectPasswordChangesViaPatchOrReplace` or perhaps `allowInsecurePasswordUpdates` to enable the "before save" hook

Thoughts?

test/user.test.js
bajtos: Well, I consider the tests in `user.test.js` as verifying the "default" password flow. In that light, do you still think I should move them to a `context('default password flow')` block?

bajtos:
@ebarault thank you for another review, I think we are getting close :)

AFAICT, there are three outstanding discussion points:

 - Feature flag naming - https://github.com/strongloop/loopback/pull/3350#discussion_r112149711 
 - Context blocks in user.test.js - https://github.com/strongloop/loopback/pull/3350#discussion_r112154506
 - Object.assign confusion - https://github.com/strongloop/loopback/pull/3350#discussion_r112148476

ebarault:
> Feature flag naming - #3350 (comment)

my proposal was to version the features such as password flow such as: `passwordFlow@3.6.0 : true`, and describe the content elsewhere. But one in all i prefer your proposal to split favor feature flags that describe the behaviour of the application. a similar situations we should use several feature flags:

> scopePasswordResetToken controlling the behaviour of POST /api/users/reset-password and POST /api/users/reset
 
what about restrictResetPasswordTokenScope ? at least something with an explicit **verb**

> rejectPasswordChangesViaPatchOrReplace or perhaps allowInsecurePasswordUpdates to enable the "before save" hook

rejectPasswordChangesViaPatchOrReplace 👍

ebarault:
> Context blocks in user.test.js - #3350 (comment)

hmm, so when the default behavior comes to be the feature flagged one ("modern password flow") we should review the code in user.test.js to match the correct behavior? I believe i'm not aware enough of the logic that leads to gather some tests in user.test.js, compared to dedicating other files to specific features. My only concern is about code maintenance when going to next major release. I leave this to you to choose the approach that minimize the code maintenance IYO.

ebarault:
> Object.assign confusion - #3350 (comment)

I see your point, but i'd need to do some tests on my side to be 100% convinced that assigning the options var inside the function block does not modify the original options arg.
So as your code is nevertheless 100% safe, let's close this discussion and keep your proposal.

bajtos:
**Feature flag naming**
> what about restrictResetPasswordTokenScope 
> rejectPasswordChangesViaPatchOrReplace

Done in  3dc012d

**Context blocks in user.test.js**
> hmm, so when the default behavior comes to be the feature flagged one ("modern password flow") we should review the code in user.test.js to match the correct behavior?

Yes.

> I believe i'm not aware enough of the logic that leads to gather some tests in user.test.js, compared to dedicating other files to specific features. 

Well, I don't think there is much logic here. I am keeping existing tests to ensure full backwards compatibility is preserved.

> My only concern is about code maintenance when going to next major release. 

I think we have two options:
 - Clean the tests right now. Unless I temporarily change the default value of these flags, I may miss some tests that would break when defaults are changed in the future. When moving the tests around, one may accidentally change their behaviour, which would go against the goal of verifying backwards compatibility.
 - Defer test cleanup until we are changing the defaults, even though it will make such change more costly. OTOH if we never change the default, then we saved time. 

> I leave this to you to choose the approach that minimize the code maintenance IYO.

Let's leave test cleanup for later then.

**Object.assign confusion**

> So as your code is nevertheless 100% safe, let's close this discussion and keep your proposal.

Sounds good 👍 

---

@ebarault Could you please review the latest changes in 3dc012d? I think that's the last step before this patch can be landed.

loay:
@slnode test please

ebarault:
> OTOH if we never change the default, then we saved time

oh gosh, let's hope we'll not end there !! ;-)

your changes looks good to me, let's land it !

loay:
@slnode test please

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+var expect = require('./helpers/expect');
+const errorHandler = require('strong-error-handler');
+var loopback = require('../');
+const Promise = require('bluebird');
+var request = require('supertest');
+const waitForEvent = require('./helpers/wait-for-event');
```
superkhau: Why `var`s? Seems like we can use all `const`s.

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+        return resetPassword(resetToken).expect(204);
+      });
+    });
+
+    function givenAppWithRestrictionEnabled() {
+      return givenAppWithUser({restrictResetPasswordTokenScope: true});
+    }
+  });
+
+  context('rejectPasswordChangesViaPatchOrReplace', () => {
```
superkhau: normal sentence for consistency with context names below? 'reject password changes via patchOrReplace'?

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+const Promise = require('bluebird');
+var request = require('supertest');
+const waitForEvent = require('./helpers/wait-for-event');
+
+describe('User.password', () => {
+  const credentials = {email: 'test@example.com', password: 'pass'};
+
+  let app, User, testUser, regularToken, resetToken;
+
+  context('restrictResetPasswordTokenScope', () => {
```
superkhau: 'restrict reset password token scope'?

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+          expect(err.message).to.match(/Invalid use.*options.setPassword/);
+        });
+    });
+
+    function givenAppWithRejectionEnabled() {
+      return givenAppWithUser({rejectPasswordChangesViaPatchOrReplace: true});
+    }
+  });
+
+  context('no limitations', () => {
```
superkhau: with no restrictions/limitations? (consistent wording with line 100)

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+    beforeEach(givenAppWithNoRestrictions);
+
+    context('with regular access token', () => {
```
superkhau: using regular access tokens

test/user-password.test.js
```diff
@@ -0,0 +1,242 @@
...
+      it('allows changing user password', () => {
+        return changePassword(regularToken).expect(204);
+      });
+
+      it('allows resetting user password', () => {
+        return resetPassword(regularToken).expect(204);
+      });
+    });
+
+    context('with password-reset token', () => {
```
superkhau: using password-reset token

test/user.integration.js
```diff
@@ -178,6 +180,64 @@ describe('users - integration', function() {
           expect(injectedOptions).to.have.property('accessToken');
         });
     });
+
+    it('resets user\'s password', function() {
```
superkhau: resets the

test/user.integration.js
```diff
@@ -178,6 +180,64 @@ describe('users - integration', function() {
...
+        .then(() => {
+          return User.findById(this.user.id);
+        })
+        .then(user => {
+          return user.hasPassword('new password');
+        })
+        .then(isMatch => expect(isMatch, 'user has new password').to.be.true());
+    });
+
+    it('rejects unauthenticated reset password request', function() {
```
superkhau: rejects unauthenticated password reset requests

test/user.integration.js
```diff
@@ -178,6 +180,64 @@ describe('users - integration', function() {
...
+      return this.post('/api/users/reset-password')
+        .send({
+          newPassword: 'new password',
+        })
+        .expect(401);
+    });
+
+    it('injects reset password options from remoting context', function() {
```
superkhau: uses password reset options provided by the remoting context

test/user.integration.js
```diff
@@ -178,6 +180,64 @@ describe('users - integration', function() {
...
+      const User = app.models.User;
+      const credentials = {email: 'inject-reset@example.com', password: 'pass'};
+
+      let injectedOptions;
```
superkhau: replace injected with remotingContextOptions? and below

test/user.test.js
```diff
@@ -1405,6 +1425,92 @@ describe('User', function() {
...
+        })
+        .then(user => {
+          return user.hasPassword('new password');
+        })
+        .then(isMatch => {
+          expect(isMatch, 'user has new password').to.be.true();
+        });
+    });
+
+    it('fails with 401 for unknown user', () => {
```
superkhau: unknown users

loay:
linting issues should be fixed now on `gateway-director-bluemix@develop ` .. the next build should be a success.

bajtos:
@ebarault 
> could you please open a new issue/PR so that we don't forget to add these 2 new options in the new project templates ?

I have already done that, see https://github.com/strongloop/loopback/issues/3349

bajtos:
@loay 
> linting issues should be fixed now on `gateway-director-bluemix@develop ` .. the next build should be a success. 

Thanks! 🙇

test/user-password.test.js
bajtos: The idea is that this context disables all new feature flags. I am going to use `context('all feature flags disabled')`

test/user.integration.js
bajtos: Let's use `observedOptions`.

bajtos:
@superkhau thank you for the detailed review, I addressed your comments in b1aeb8c.

bajtos:
> [cis-jenkins] downstream: loopback-getting-started@master — Failed! (c5ca2e1)

That was a (temporary?) CI issue:

```
FATAL: java.io.IOException: Unexpected termination of the channel
java.io.EOFException
	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2353)
```

bajtos:
Landed. Thank you everybody for keeping the fast pace, allowing me to iterate on your feedback and get this complex change landed in less than two weeks 🎉

PR #3381, opened 2017-05-01
Refactor access token to make it extensible

raymondfeng:
1. Make it possible to reuse getAccessTokenId
2. Introduce a flag to control if oAuth2 bearer token should be base64
encoded

#### Related issues

- None

### Checklist

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -79,6 +79,121 @@ module.exports = function(AccessToken) {
   });
 
   /**
+   * Extract the access token id from the HTTP request
+   * @param {Request} req HTTP request object
+   * @options {Object} [options] Each option array is used to add additional keys to find an `accessToken` for a `request`.
+   * @property {Array} [cookies] Array of cookie names.
+   * @property {Array} [headers] Array of header names.
+   * @property {Array} [params] Array of param names.
+   * @property {Boolean} [searchDefaultTokenKeys] Use the default search locations for Token in request
+   * @property {Boolean} [bearerTokenBase64Encoded] Defaults to `true`. For `Bearer` token based `Authorization` headers,
+   * decode the value from `Base64`. If set to `false`, the decoding will be skipped and the token id will be the raw value
+   * parsed from the header.
+   * @return {String} The access token
+   */
+  AccessToken.getIdForRequest = function(req, options) {
+    options = options || {};
+    var params = options.params || [];
+    var headers = options.headers || [];
+    var cookies = options.cookies || [];
+    var i = 0;
+    var length, id;
+
+    // https://github.com/strongloop/loopback/issues/1326
+    if (options.searchDefaultTokenKeys !== false) {
+      params = params.concat(['access_token']);
+      headers = headers.concat(['X-Access-Token', 'authorization']);
+      cookies = cookies.concat(['access_token', 'authorization']);
+    }
+
+    for (length = params.length; i < length; i++) {
+      var param = params[i];
+      // replacement for deprecated req.param()
+      id = req.params && req.params[param] !== undefined ? req.params[param] :
+        req.body && req.body[param] !== undefined ? req.body[param] :
+        req.query && req.query[param] !== undefined ? req.query[param] :
+        undefined;
+
+      if (typeof id === 'string') {
+        return id;
+      }
+    }
+
+    for (i = 0, length = headers.length; i < length; i++) {
+      id = req.header(headers[i]);
+
+      if (typeof id === 'string') {
+        // Add support for oAuth 2.0 bearer token
+        // http://tools.ietf.org/html/rfc6750
+        if (id.indexOf('Bearer ') === 0) {
+          id = id.substring(7);
+          if (options.bearerTokenBase64Encoded) {
+            // Decode from base64
+            var buf = new Buffer(id, 'base64');
+            id = buf.toString('utf8');
+          }
+        } else if (/^Basic /i.test(id)) {
+          id = id.substring(6);
+          id = (new Buffer(id, 'base64')).toString('utf8');
+          // The spec says the string is user:pass, so if we see both parts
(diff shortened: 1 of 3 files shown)


common/models/access-token.js
```diff
@@ -78,6 +78,77 @@ module.exports = function(AccessToken) {
     });
   });
 
+  AccessToken.getAccessTokenId = function(req, options) {
```
ritch: Public api now? Needs doc before we merge.

ritch: Since the method moved its impossible to do a diff...? Is this unchanged otherwise?

server/middleware/token.js
```diff
@@ -88,6 +88,8 @@ function escapeRegExp(str) {
  * @property {Boolean} [overwriteExistingToken] only has effect in combination with `enableDoublecheck`. If truthy, will allow to overwrite an existing accessToken.
  * @property {Function|String} [model] AccessToken model name or class to use.
  * @property {String} [currentUserLiteral] String literal for the current user.
+ * @property {Boolean} [bearerTokenBase64Encoded] oAuth2 is not clear if bearer token in Authorization header should be base64 encoded.
```
ritch:
```
oAuth2 is not clear if bearer token in Authorization header should be base64 encoded.
If set, we'll decode the token first. Default to `true` for backward compatibility.
```

I would reword this a little bit:

```
Defaults to `true`. For `Bearer` token based `Authorization` headers, decode the value from `Base64`. If set to `false` the decoding will be skipped and the token id will be the raw value parsed from the header.
```

common/models/access-token.js
raymondfeng: Yes. It's moved out as is. The jsdoc is added now.

raymondfeng:
@ritch PTAL.

test/access-token.test.js
```diff
@@ -462,6 +473,11 @@ describe('loopback.token(options)', function() {
 describe('AccessToken', function() {
   beforeEach(createTestingToken);
 
+  it('should have getAccessTokenId method', function() {
+    assert(typeof Token.getAccessTokenId === 'function');
+    assert(typeof Token.findForRequest === 'function');
+  });
```
ebarault: - there should be one assertion per test (test on `.findForRequest()` is unrelated)
- new tests should use expect, not assert
- new test descriptions do not mention 'should': `it has getAccessTokenId method`
- test on `.getAccessTokenId()` should actually verify that the method can be customized by the user (main purpose on the PR)
- while at it, test on `. findForRequest()` should do likewise (as it's public api)

raymondfeng:
@ebarault Thanks for the feedback. PTAL.

common/models/access-token.js
```diff
@@ -79,6 +79,119 @@ module.exports = function(AccessToken) {
...
+    }
+    return null;
+  };
+
+  /**
+   * Validate the access token
+   * @param {String} id Access token
+   * @callback {Fuction} cb Callback function
+   */
+  AccessToken.validate = function(id, cb) {
```
ebarault: - the signature of this new api is inconsistent with `.prototype.validate()` which returns a **Boolean**, i suggest returning a boolean, as the token is already provided as param.
- please add also the @callback signature in the jsdoc.
```
// as in
  @callback (Function) callback
  @param {Error} err
  @param {isValid} is the token valid
```
- you also have a typo (see: _Fuction_)

common/models/access-token.js
```diff
@@ -79,6 +79,119 @@ module.exports = function(AccessToken) {
...
+    this.findById(id, function(err, token) {
+      if (err) {
+        cb(err);
+      } else if (token) {
+        token.validate(function(err, isValid) {
+          if (err) {
+            cb(err);
+          } else if (isValid) {
+            cb(null, token);
```
ebarault: `cb(null, true);`

common/models/access-token.js
```diff
@@ -87,40 +200,18 @@ module.exports = function(AccessToken) {
...
-              e.status = e.statusCode = 401;
-              e.code = 'INVALID_TOKEN';
-              cb(e);
-            }
-          });
-        } else {
-          cb();
-        }
-      });
+      this.validate(id, cb);
```
ebarault:
```
this.validate(id, function(err, isValid) {
  if (err) cb(err);
  cb(null, isValid);
});
```

test/access-token.test.js
```diff
@@ -172,6 +173,16 @@ describe('loopback.token(options)', function() {
       .end(done);
   });
 
+  it('should populate req.token from an authorization header with bearer token', function(done) {
```
ebarault: ditto

test/access-token.test.js
```diff
@@ -162,7 +162,8 @@ describe('loopback.token(options)', function() {
     });
   });
 
-  it('should populate req.token from an authorization header with bearer token', function(done) {
+  it('should populate req.token from an authorization header with bearer token with base64',
```
ebarault: while at it could you remove the `should`?

test/access-token.test.js
```diff
@@ -525,6 +544,54 @@ describe('AccessToken', function() {
...
+        if (err) return done(err);
+
+        expect(token.id).to.eql(expectedTokenId);
+        expect(called).to.be.true();
+
+        done();
+      });
+    });
+
     function mockRequest(opts) {
```
ebarault: this should do

raymondfeng:
@ebarault I made the further changes per your request:

1. Rename `AccessToken.validate` to `AccessToken.resolve`. Please note that `resolve` takes `id` and returns the full access token object.

2. Fix the jsdoc

3. Remove `should` for all tests

raymondfeng:
We cannot change `findForRequest` as it's already there.

I can rename as follows:

1. getAccessTokenId --> getIdForRequest
2. resolve (as-is, resolve already hints by id)

PR #3433, opened 2017-06-06
Make bulkUpdate non atomic function

ssh24:
### Description

Make bulkUpdate a non atomic function. Allow connectors to perform bulkUpdates now.

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/35


bajtos:
@ssh24 Could you please explain what problem is this pull request trying to solve?

`bulkUpdate` API is an integral part of change replication/offline sync implementation. I am concerned that by moving this implementation down to connectors, we make it easy for 3rd-party/community connectors to break change replication by implementing an incomplete version of `bulkUpdate`. To address that, I would like to see a robust test suite covering `bulkUpdate` in juggler's shared test suite run by all connectors.

I also think this is a breaking (semver-major) change - existing applications will stop working after they update to a newer loopback version (with bulkUpdate removed) in case their connector does not provide its own `bulkUpdate` implementation. A possible solution for preserving backwards-compatibility is to keep `bulkUpdate` implementation inside `loopback`, mark it as deprecated, and let connectors to override this loopback-provided version with their own implementation, similarly how they are already overriding other DAO methods.

ssh24:
@bajtos This PR and the subset PR's are not complete. Thank you for joining the conversation. I would like @raymondfeng and @kjdelisle to join as well.

PR in juggler: https://github.com/strongloop/loopback-datasource-juggler/pull/1399
PR in cloudant: https://github.com/strongloop/loopback-connector-cloudant/pull/150

The goal here was to allow cloudant connector perform `bulkUpdate`. This is definitely a breaking change in `loopback`. The tests are failing as well on `bulkUpdate`. So I want opinion on this and what would be the best way to do it.

ssh24:
I realized we can override this function in DAO. Closing this PR.

PR #3455, opened 2017-06-20
Remove unnecessary else, use early return

andreipfeiffer:
### Description
No need to use else branch when returning from if branch.
Code is less indented this way and easier to extend (no need for if ... else if ... elseif ... etc).

#### Related issues
- None

### Checklist

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -617,9 +617,9 @@ module.exports = function(registry) {
 
     if (ds.idName) {
       return ds.idName(Model.modelName);
-    } else {
-      return 'id';
     }
+
+    return 'id';
   };
 
   PersistedModel.setupRemoting = function() {


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
Hello, thank you @andreipfeiffer for the pull request. While we value clean code, we also value a clean git history that makes it easy to find why a certain block of code was written the way it is. Refactoring changes like the one you are proposing make that task more difficult, because the person performing code archaeology has to skip/exclude refactoring commits. For that reasons, we are not accepting such changes.

andreipfeiffer:
Got it, no problem.

PR #3457, opened 2017-06-21
Support for IBM Object Storage

hacksparrow:
### Description

Adds support for `object-storage` connector.

#### Related issues

- https://github.com/strongloop-internal/scrum-loopback/issues/1537

### Checklist

### DIFF
--- a/test/object-storage.test.js
+++ b/test/object-storage.test.js
@@ -1,0 +1,23 @@
+'use strict';
+var assert = require('assert');
+
+describe('object-storage connector', function() {
+  it('should be set up as loopback-component-storage', function(done) {
+    this.timeout(10000);
+    var app = require('./fixtures/object-storage-app');
+    var booted = false;
+    app.on('booted', function() {
+      if (booted) return;
+      assert('MyObjectStorage' in app.datasources);
+      var ds = app.datasources.MyObjectStorage;
+      assert.equal(ds.name, 'loopback-component-storage');
+      assert.equal(ds.settings.connector, 'loopback-component-storage');
+      assert.equal(ds.settings.provider, 'openstack');
+      assert.equal(ds.settings.useServiceCatalog, true);
+      assert.equal(ds.settings.useInternal, false);
+      assert.equal(ds.settings.keystoneAuthVersion, 'v3');
+      booted = true;
+      done();
+    });
+  });
+});
--- a/lib/application.js
+++ b/lib/application.js
@@ -225,6 +225,15 @@ app.models = function() {
  * @param {Object} config The data source config
  */
 app.dataSource = function(name, config) {
+  // some connectors are aliases to loopback-component-storage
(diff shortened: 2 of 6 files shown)


raymondfeng:
I don't think it's wise to add specific handling for object-storage here in this module. Can you elaborate why?

test/object-storage.test.js
```diff
@@ -0,0 +1,23 @@
...
+var assert = require('assert');
+
+describe('object-storage connector', function() {
+  it('should be set up as loopback-component-storage', function(done) {
+    this.timeout(10000);
+    var app = require('./fixtures/object-storage-app');
+    var booted = false;
+    app.on('booted', function() {
+      if (booted) return;
+      assert('MyObjectStorage' in app.datasources);
```
bajtos: When this check fails, it will produce an error message that's not helpful. Here is a better solution:

```js
expect(Object.keys(app.datasources)).to.contain('MyObjectStorage');
```

(Or maybe `.to.have.member`? I am never certain about which Chai assertion to use for this.)

test/object-storage.test.js
```diff
@@ -0,0 +1,23 @@
...
+      var ds = app.datasources.MyObjectStorage;
+      assert.equal(ds.name, 'loopback-component-storage');
```
bajtos: Similarly here, the assertion errors don't specify the property name that was being tested.

```js
expect(ds).to.have.property('name', 'loopback-component-storage');
```

The same comment applies to the lines below too.

test/object-storage.test.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+var assert = require('assert');
+
+describe('object-storage connector', function() {
+  it('should be set up as loopback-component-storage', function(done) {
+    this.timeout(10000);
+    var app = require('./fixtures/object-storage-app');
+    var booted = false;
+    app.on('booted', function() {
+      if (booted) return;
```
bajtos: First of all, I find it rather weird that `booted` is emitted more than once.

Let's not worry about that though. If you want your event listener to be called only once, then you should use `once` instead of `on`. That way you don't need to maintain the flag `booted`.

```js
app.once('booted', () => {
  // ...
});
```

hacksparrow:
As discussed on Slack, specific-datasource code should not be added here. Closing this.

test/object-storage.test.js
bajtos: On the second thought, I think using `loopback-boot` is an overkill in this scenario. Also in general, the tests in LoopBack should not be relying on loopback-boot, but call LoopBack APIs directly.

Here is how you can simplify your test:

```js
it('should be set up as loopback-component-storage', () => {
  const app = loopback({localRegistry: true});
  app.dataSource('My-Object-Storage', {
    username: '0',
    // ...
    connector: 'object-storage',
  });

  // asserts
});
```

Issue #1364, opened 2015-05-05
User.validatePassword throws uncaught exception in User.create


cleas:
related to #941 

Creating user with empty password : loopback@2.8 user created , loopback@2.9 application crash due to uncaught exception

```
User.create({username: 'John', email: 'john@doe.com', password:''},function(err,user){
//invalid password error should be returned in callback
 });
```

overriding validatePassword to check for minimum length will also give uncaught exception

see also example https://github.com/cleas/loopback-sandbox
https://github.com/cleas/loopback-sandbox/blob/master/server/boot/create-model-instances.js

superkhau:
I tried running your example. I get:

```
Error: Invalid password:
  ...
```

Can I close this?

cleas:
I tried with loopback 2.18.0 => error is still there.
looking at the source nothing changed to User.hashPassword and User.validatePassword.
function User.ValidatePassword can throw exception , this exception is not handled in UserModel.setter.password

rawrmaan:
I'm having this issue as well. I overrode validatePassword in my own User model subclass and used the default implementation as a reference. Now my instance crashes whenever there's an invalid password. Should I just return false instead of throwing an error?

jlvdh:
I also encountered this, and used the following as a quickfix in user.js

    User.observe('before save', function checkEmptyPasswordString(ctx, next) {
      if (ctx.data && ctx.data.password === '') {
        next('Password is empty string');
      }
      next();
    });

superkhau:
@cleas Are you still running into issues? Seems to be fix by #941 now as I can't reproduce the error in your `loopback-sandbox` for anymore.

jlvdh:
I'm still experiencing the error on Loopback 2.21.0 with the following request:

    PUT /user

    {"id":140, "password":""}

cleas:
The error in the ````loopback-sandbox example```` is now catched due to the changes in commit: 

https://github.com/strongloop/loopback-datasource-juggler/commit/21c0067462f48210f156bb043cb4076e979ff94d
"Report deferred exceptions via callback"

But still  validatePassword exceptions are not caught in User.create function it self and returned in cb.
For example if the create functions are wrapped in a dataSource.autoupdate function, the  application crashes.
see updated ````loopback-sandbox```` example.

gunjpan:
@cleas: I believe this should have been resolved in newer versions of LoopBack. Could you please confirm?

vijay22sai:
[...]

PR #3481, opened 2017-07-11
Add unit test for empty password

loay:
### Description

Passwords with empty strings should be rejected
#### Related issues
connect to https://github.com/strongloop/loopback/issues/1364

### Checklist

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -401,16 +401,31 @@ describe('User', function() {
     var pass73Char = pass72Char + '3';
     var passTooLong = pass72Char + 'WXYZ1234';
 
-    it('rejects passwords longer than 72 characters', function(done) {
-      try {
-        User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
-          if (err) return done(err);
-          done(new Error('User.create() should have thrown an error.'));
+    it('rejects empty passwords creation', function(done) {
+      User.create({email: 'b@c.com', password: ''}, function(err) {
+          expect(err.code).to.equal('INVALID_PASSWORD');
+          expect(err.statusCode).to.equal(422);
+          done();
         });
-      } catch (e) {
-        expect(e).to.match(/Password too long/);
+    });
+
+    it('rejects updating with empty password', function(done) {
+        User.create({email: 'blank@c.com', password: pass72Char}, function(err, userCreated) {
+          if (err) return done(err);
+          userCreated.updateAttribute('password', '', function(err, userUpdated) {
+            expect(err.code).to.equal('INVALID_PASSWORD');
+            expect(err.statusCode).to.equal(422);
+            done();
+          });
+        });
+      });
+
+    it('rejects passwords longer than 72 characters', function(done) {
+      User.create({ email: 'b@c.com', password: pass73Char }, function(err) {
+        expect(err.code).to.equal('PASSWORD_TOO_LONG');
+        expect(err.statusCode).to.equal(422);
         done();
-      }
+      });
     });
 
     it('rejects a new user with password longer than 72 characters', function(done) {


common/models/user.js
```diff
@@ -870,6 +870,16 @@ module.exports = function(User) {
     next();
   });
 
+  User.observe('before save', function emptyPasswordUpdate(ctx, next) {
+    if ((ctx.instance && ctx.instance.password === '') || (ctx.data && ctx.data.password === '')) {
```
bajtos: So far, the password validation is happening in `User.validatePassword`, see https://github.com/strongloop/loopback/blob/c7cd43c90ad4f8e52d70289468264a35a3df0efe/common/models/user.js#L646

Is there any particular reason why this new check cannot be performed by that method too?

test/user.test.js
```diff
@@ -401,6 +401,29 @@ describe('User', function() {
...
     var passTooLong = pass72Char + 'WXYZ1234';
 
+    it('rejects empty passwords creation', function(done) {
+      try {
+        User.create({email: 'b@c.com', password: ''}, function(err) {
+          if (err) return done(err);
+          done(new Error('User.create() should have thrown an error.'));
+        });
+      } catch (e) {
+        expect(e).to.match(/Invalid password/);
```
bajtos: Please verify `err.code`, not the error message. The message can change in the future (or when localized).

test/user.test.js
```diff
@@ -401,6 +401,29 @@ describe('User', function() {
     var pass73Char = pass72Char + '3';
     var passTooLong = pass72Char + 'WXYZ1234';
 
+    it('rejects empty passwords creation', function(done) {
+      try {
+        User.create({email: 'b@c.com', password: ''}, function(err) {
+          if (err) return done(err);
+          done(new Error('User.create() should have thrown an error.'));
+        });
+      } catch (e) {
```
bajtos: This is pretty weird. A callback/promise-based function should report errors via callback/promise rejection, not by throwing synchronously to the caller.

```js
User.create({email: 'b@c.com', password: ''}, function(err) {
  expect(err).to.have.property('code', 'INVALID_PASSWORD');
  done();
});
```

test/user.test.js
```diff
@@ -401,6 +401,29 @@ describe('User', function() {
...
+        done();
+      }
+    });
+
+    it('rejects updating with empty password', function(done) {
+        User.create({email: 'blank@c.com', password: pass72Char}, function(err, userCreated) {
+          if (err) return done(err);
+          userCreated.updateAttribute('password', '', function(err, userUpdated) {
+            assert(err);
+            expect(err).to.match(/Invalid Password/);
```
bajtos: Please verify `err.code`, not the message.

I think both lines 420 and 421 can be compressed to the following:

```js
expect(err).to.have.property('code', 'INVALID_PASSWORD');
```

test/user.test.js
loay: It is for consistency since the `try/catch` was used for the rest of the test cases for the same topic.

common/models/user.js
loay: Yes, it does not wrap the cb. I have couple of examples to show you.

bajtos:
@loay I am afraid I still think this pull request is changing wrong parts of the code.

In https://github.com/strongloop/loopback/issues/1364, users complain that `User.create` is throwing an exception that should have been passed via callback.

From the issue description:
> Creating user with empty password : loopback@2.8 user created , loopback@2.9 application crash due to uncaught exception

https://github.com/strongloop/loopback/issues/1364#issuecomment-120112845
> I overrode validatePassword in my own User model subclass and used the default implementation as a reference. Now my instance crashes whenever there's an invalid password. 

https://github.com/strongloop/loopback/issues/1364#issuecomment-137026078
> But still validatePassword exceptions are not caught in User.create function it self and returned in cb.
> For example if the create functions are wrapped in a dataSource.autoupdate function, the application crashes.

Similar effects are observed for `udpateAttributes`:

https://github.com/strongloop/loopback/issues/1364#issuecomment-304522286
> When calling `User.updateAttribute('password','someinvalidpassword',cb)`, `User.validatePassword()` exception is not passed to the callback, thus the application crashes.

Notice that the unhandled error is thrown not only for empty passwords, but for any invalid password too. This was reported in https://github.com/strongloop/loopback/issues/1364#issuecomment-120112845 too.

My conclusion is that we need to fix DAO methods in loopback-datasource-juggler to correctly handle the situation when a property setter throws an error.

For example, take a look at this snippet from the `create()` method:

https://github.com/strongloop/loopback-datasource-juggler/blob/6a1b55568cab3fabd7bd5823f8749684ad75a1fc/lib/dao.js#L315-L323
```js
  // if we come from save
  if (data instanceof Model && !idValue) {
    obj = data;
  } else {
    obj = new Model(data);
  }

  this.applyProperties(enforced, obj);
  obj.setAttributes(enforced);
```

I believe `new Model` and/or `setAttributes` invoke custom property setters and if a custom setter throws an error, the error is left unhandled. The solution is to wrap all of these calls in a `try/catch` block reporting errors via the callback.

```js
try {
  // if we come from save
  if (data instanceof Model && !idValue) {
    obj = data;
  } else {
    obj = new Model(data);
  }

  this.applyProperties(enforced, obj);
  obj.setAttributes(enforced);
} catch(err) {
  return cb(err);
}
```

Once this fix is landed in juggler, you need to fix `User` tests in LoopBack to expect password-validation errors to be reported via callback.

common/models/user.js
bajtos: One reason why I am opposing this new "before save" hook is inconsistency with existing password validations (e.g. max length). Even if this new hook fixed the issue for an empty password, it will most likely not fix it for other validations like max length. Additionally, it will make the code more difficult to maintain, because the validation logic will be spread across multiple places in the source code file.

test/user.test.js
bajtos: > since the `try/catch` was used for the rest of the test cases for the same topic

In my understanding, this `try/catch` approach is actually documenting the bug reported in https://github.com/strongloop/loopback/issues/1364. We need to fix the code so that password-validation errors are reported via callback, so that `try/catch` based tests in this file can be reworked to the usual callback-based style.

loay:
@bajtos 

I think there was a little misunderstanding, so I guess I was not clear. Let me explain a little bit further.

The create part was not a problem. Apparently, it was fixed somewhere and the issue is not occurring there anymore, as per the main issue. The actual problem happens when we try to update. 

I have created a test application to test the effect of empty password. 

- Create a new user with proper password: PASS
- create a new user with empty password: error caught and app does not crash. 
- Update an existing user's password with empty password: error not caught, and app crashes. 
- Update an existing user's password with long  password: error not caught, and app crashes. 

That's where I tried the `before save` part to fix the issue. 

I created a new pr in juggler with your proposed solution and I added another `try/catch` block for the update method. I did not use the `try/catch` for the `create` method since it broke some of the datatypes test cases. 

I linked with loopback and tests here pass, as well as I tested with the previously created app

- Create a new user with proper password: PASS
- create a new user with empty password: error caught and app does not crash. 
- Update an existing user's password with empty password: error caught and app does not crash. 
- Update an existing user's password with long  password: error caught and app does not crash. 

I will trigger the tests here once the juggler pr is merged.

loay:
@slnode test please

loay:
@slnode test please

Issue #2860, opened 2016-10-14
White/Black listing methods generated by relations


gunjpan:
### Bug or feature request

- [x] Bug
- [ ] Feature request
### Description of feature (or steps to reproduce if bug)

Continuing from #651, the feature is a limited one. It does not allow to white/black list methods imposed on a model by relations.
### Expected result

white/black listing of methods should be applied to related model method endpoints along with model's own methods
### Actual result (if bug)

For example: `student.__get__projects`, a related model method for endpoint:`/api/students/:id/projects`, is not affected when following is set in `model-config.json`:

``` javascript
"MyModel": {
    "dataSource": "db",
    "public": true,
    "options":{
      "remoting": {
      "sharedMethods": {
        "*": false, // this is not effective
        "create": true,
        "__get__projects": false //neither this for related model methods
      }
     }
    }
  }
```

gunjpan:
cc/@bajtos

gunjpan:
Another flaw, reported by @ahmed-abdulmoniem : 
@gunjpan Another issue ..

Can we hide PATCH end points?

![image](https://cloud.githubusercontent.com/assets/6064145/19409002/d1c3d5a4-92ca-11e6-95ed-e79a0c9e6e22.png)

I see for example PATCH and PUT has the same method names but with index?! like `upsert_0` and `upsert`?

When I hide `upsert` it hides both PATCH and PUT .. can we keep one of them and hide the others?

superkhau:
@ritch I think we should get @jannyHou to work on this feature as she is more familiar in this area.

jannyHou:
@superkhau 
According to the bug described in the first comment:
> Continuing from #651, the feature is a limited one. It does not allow to white/black list methods imposed on a model by relations.

It seems more like a loopback-model/strong-remoting issue, I am not familiar with how the white/black list work :(

relation relevant part:
[`loopback/model.js`](https://github.com/strongloop/loopback/blob/master/lib/model.js#L232-L252) resolves relation functions in model.

superkhau:
>It seems more like a loopback-model/strong-remoting issue, I am not familiar with how the white/black list work :(

NP, I'm not sure either, I believe @richardpringle worked this feature awhile back. It'll probably touch both areas: remoting + the relation part you mentioned above. Anyways, let's leave it in the backlog and figure out when it actually gets prioritized. TY for chiming in. ;)

richardpringle:
[...]

PR #3504, opened 2017-07-18
Allow glob-style patterns for remote options

zbarbuto:
### Description

This PR achieves two things:
1. By moving `setSharedMethodSharedProperties` to after registry configuration, shared methods on model relations can correctly be disabled (#2860)
2. By using a regex instead of hard-coded '*' we can use glob-style expressions to blanket disable methods (e.g. `__*__someRelation` to disable all remote methods of a model relationship)

Note, the reason I used string replacement for `*` instead of straight regular expressions is to make it backwards compatible ('*' will still enable/disable all methods)

#### Related issues

- 2860

### Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -533,9 +533,9 @@ function configureModel(ModelCtor, config, app) {
   config = extend({}, config);
   config.dataSource = dataSource;
 
-  setSharedMethodSharedProperties(ModelCtor, app, config);
-
   app.registry.configureModel(ModelCtor, config);
+
+  setSharedMethodSharedProperties(ModelCtor, app, config);
 }
 
 function setSharedMethodSharedProperties(model, app, modelConfigs) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
Hello @zbarbuto, thank you for the pull request! The proposed changes make sense at high level. Could you please add few tests to verify them? Also, I think it's not enough to replace `*` with `.*`, you should escape other characters that have special meaning in RegExp (e.g. `.` and `?`).

zbarbuto:
@bajtos Have sanitized incoming strings for reserved regex characters (excluding `*` of course).

Have added tests but this has revealed a problem that the related method globs still don't seem to be working. I can't figure out why. In debugging I can see that it's definitely setting the correct methods to `shared = false` but `TestModel.sharedClass.methods()` in the test is still returning them as `shared: true`... 

Possibly something further down the chain is re-enabling them again (such as `strong-remoting`? Any ideas?

test/loopback.test.js
```diff
@@ -739,4 +739,157 @@ describe('loopback', function() {
...
+      });
+
+      var publicMethods = TestModel.sharedClass
+        .methods()
+        .filter(function(m) {
+          return m.shared === true;
+        })
+        .map(function(m) {
+          return m.stringName.replace(/^[^.]+\./, ''); // drop the class name
+        });
```
bajtos: Please extract these 8 lines of code into a shared helper function. E.g. `listPublicMethods(TestModel)`.

zbarbuto: Done

bajtos:
> Have added tests but this has revealed a problem that the globs still don't seem to be working for related methods. I can't figure out why. In debugging I can see that it's definitely setting the correct methods to shared = false but TestModel.sharedClass.methods() in the test is still returning them as shared: true

Remoting metadata for relation methods is defined via strong-remoting "resolver", which is a function called by `sharedClass.methods()` to describe additional remote methods that were not defined via `SharedMethod` API nor via properties on the function objects (e.g. `MyModel.myMethod.shared = true`).

The part in strong-remoting: https://github.com/strongloop/strong-remoting/blob/5b89fb2ab79a77c5b4d25b4009e8033951eb4852/lib/shared-class.js#L133-L136

The part in LoopBack: https://github.com/strongloop/loopback/blob/839c58639d83f7c15477ade2285aa111f74378eb/lib/model.js#L261-L296

As I understand this code, the `SharedMethod` entries created by these resolvers are temporary, they are always created from fresh whenever `sharedClass.methods()` is called, and discarded by GarbageCollector afterwards.

BTW similar issue exists for methods with remoting metadata defined via function properties, e.g. `MyModel.myMethod.shared = true` - new temporary `SharedMethod` entries are created for them, any changes made to these temporary entries are discarded.

You may be wondering why we don't define relation remote methods immediately at model setup time, similarly how regular PersistedModel methods are defined. So am I :) I think the original reason was that relations are defined by connectors when the model is attached to a datasource, which typically happens later, possibly even asynchronously (e.g. for connectors like SOAP or Swagger that need to fetch WSDL/Swagger description of the API first). I am not sure if that's still relevant though.

As a short-term workaround, perhaps you could use a different approach for disabling remote methods and call `sharedMethod.sharedClass.disableRemoteMethodByName(...)` instead of `sharedMethod.shared = false`.

Thoughts?

cc @ritch as the original author of strong-remoting resolvers (IIRC)
cc @superkhau as the author of `setSharedMethodSharedProperties` (IIRC)

zbarbuto:
@bajtos Sorry for the delay in updating this. Finally managed to get it working with `disableMethodByName` as you suggested. Have also rebased on master.

lib/application.js
```diff
@@ -568,15 +568,38 @@ function setSharedMethodSharedProperties(model, app, modelConfigs) {
...
     if (hasSpecificSetting) {
-      sharedMethod.shared = settings[sharedMethod.name];
-    } else { // otherwise, use the default setting if it exists
-      var hasDefaultSetting = settings.hasOwnProperty('*');
-      if (hasDefaultSetting)
-        sharedMethod.shared = settings['*'];
+      if (settings[methodName] === false) {
+        sharedMethod.sharedClass.disableMethodByName(methodName);
+      } else {
+        sharedMethod.shared = true;
```
zbarbuto: FTR - this line needed to be added otherwise tests relating to `default definitions in model-config.json` were failing.

lib/application.js
```diff
@@ -568,15 +568,38 @@ function setSharedMethodSharedProperties(model, app, modelConfigs) {
...
+      }
+    } else {
+      tests.forEach(function(glob) {
+        if (glob.regex.test(methodName)) {
+          if (glob.setting === false) {
+            sharedMethod.sharedClass.disableMethodByName(methodName);
+          } else {
+            sharedMethod.shared = true;
```
zbarbuto: Same again here.

zbarbuto:
@bajtos No problems. Done.

zbarbuto:
Rebased off latest master.

bajtos:

Issue #3529, opened 2017-07-29
Inheriting remote methods triggers deprecation warning about use of isStatic


lehni:
# Description/Steps to reproduce

- Define two models of which one inherits from the other
- Define a remote method in the base model (it doesn't matter wether it is static or not):

#### base-class.json
```
{
  "name": "BaseClass",
  "base": "Model",
  "methods": {
    "test": {}
  }
}
```

#### extending-class.json
```
{
  "name": "ExtendingClass",
  "base": "BaseClass"
}
```

When running the app, this deprecation warning appears:

```
strong-remoting deprecated Remoting metadata "isStatic" is deprecated. Please specify "prototype.name" in method name instead for isStatic=false
```

# Expected result

No deprecation warning should be triggered

# Additional information

Looking into what's causing this, it looks like `Registry.prototype.createModel()` / `BaseClass.extend()` is passing on `BaseClass.settings.methods` to `ExtendingClass` (merged with it with its own method definitions, which in this case here is none). 

The problem is that `Model.settings.methods` is internally using the now deprecated notation with `isStatic` boolean values. So when they get passed on, the deprecation warning is triggered, although none of the app's own code is specifying `isStatic` anywhere.

lehni:
This is tricky to fix, due to the use of a different format internally than in the JSON file. I'd suggest to either change the internal format as well, or to somehow mark inherited methods as inherited in their settings (or perhaps better: already parsed from JSON?) and not trigger a deprecation warning for such method definitions.

bajtos:
Hello @lehni, thank you for reporting this issue. 

I think the problem may be actually easier to fix. The `isStatic` flag is added by `Registry.prototype._defineRemoteMethods`, see [here](https://github.com/strongloop/loopback/blob/8589629e7af41dbd02a1ff2fa57d549c37656d61/lib/registry.js#L276). It looks like this method is modifying the original metadata specified in the JSON file. The inheriting model is then receiving the modified settings instead of what has been specified in the JSON file.

If my analysis is correct, then I am proposing to simply fix `_defineRemoteMethods` to create a new object before adding `isStatic` flag. E.g.


```diff
- meta.isStatic = isStatic;
+ meta = Object.assign({}, meta, {isStatic});
```

Thoughts?

lehni:
Interesting! I will test this right now.

lehni:
Perfect! That simple change does indeed address the issue, the deprecation warning is gone. Shall I create a PR for it, or will you?

[...]

PR #3540, opened 2017-08-02
Do not add isStatic properties to method settings

lehni:
### Description

As discussed in #3529, here the PR to fix the issue where creating models from JSON descriptions modifies their remote method settings by adding `isStatic` properties to them. Once a model inherits from another, these `isStatic` properties would trigger deprecation warnings.

#### Related issues

- #3529

### Checklist

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -1032,4 +1032,18 @@ describe.onServer('Remote Methods', function() {
         });
     }
   });
+
+  describe('Create Model with remote methods from JSON description', function() {
+    it('does not add isStatic properties to the method settings', function() {
+      const app = loopback();
+      const Foo = app.registry.createModel({
+        name: 'Foo',
+        methods: {
+          staticMethod: {},
+        },
+      });
+      app.model(Foo);
+      expect(app.models.Foo.settings.methods.staticMethod).to.eql({});
+    });
+  });
 });
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -273,7 +273,7 @@ Registry.prototype._defineRemoteMethods = function(ModelCtor, methods) {
 
     if (typeof meta.isStatic !== 'boolean') {
       key = isStatic ? key : m[1];
-      meta.isStatic = isStatic;
+      meta = Object.assign({}, meta, {isStatic});
     } else if (meta.isStatic && m) {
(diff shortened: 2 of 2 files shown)


test/fixtures/remote-methods/server/server.js
```diff
@@ -0,0 +1,12 @@
...
+// Node module: loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+var boot = require('loopback-boot');
+var loopback = require('../../../../index');
+
+var app = module.exports = loopback();
+boot(app, __dirname);
```
bajtos: Adding `test/fixtures` apps was a mistake that we should have fixed long time ago, see 
https://github.com/strongloop/loopback/commit/716ed4569fe79b4b153a46ce8243785a00735705#commitcomment-15931869. Unit tests should be calling `app.registry.createModel`, `app.dataSource` and `app.model` to setup the application. The test setup should be as small as possible, containing only models/datasources necessary for the test to fail with the expected error (before the changes) and pass (after the changes).

I think in your test, it should be enough to create an empty app with a single model `Foo` attached to `db` datasource.

```js
const app = loopback();
const Foo = app.registry.createModel({
  name: 'Foo',
  base: 'PersistedModel',
  methods: {
    'staticMethod': {}
  }
});
app.model(Foo, {dataSource: 'db'});

// do the checks now
```

test/model.test.js
```diff
@@ -1032,4 +1033,12 @@ describe.onServer('Remote Methods', function() {
         });
     }
   });
+
+  describe('Create Model with remote methods from JSON description', function() {
+    it('does not add isStatic properties to the method settings', function() {
+      var app = require(path.join(__dirname,
+        'fixtures/remote-methods/server/server.js'));
+      assert.deepEqual(app.models.Foo.settings.methods.staticMethod, {});
```
bajtos:
```js
expect(app.models.Foo.settings.methods.staticMethod).to.eql({});
```

bajtos:
@slnode test please

bajtos:
@lehni ping, what's the status of this pull request?

lehni:
@bajtos I've updated the PR to address all your input. Thanks!

bajtos:
Landed, thank you!

PR #3541, opened 2017-08-02
Validate on updateAll

ssh24:
### Description

Juggler has added validation on `update`/`updateAll`. This causes failure in the tests of loopback core because the `updateAll` function does not use all the model **properties**.

This PR fixes it by removing the default validation and keeping the test as is.

connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1445

### DIFF
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -2615,6 +2615,7 @@ describe('User', function() {
           });
         },
         function updatePartialUser(next) {
+          delete User.validations;
           User.updateAll(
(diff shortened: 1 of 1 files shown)


bajtos:
@ssh24 could you please revert this pull request and fix juggler to preserve backwards compatibility, per my https://github.com/strongloop/loopback-datasource-juggler/pull/1445#issuecomment-319889733?

Issue #2924, opened 2016-11-09
Incorrect swagger definition for models with forceId:true


bajtos:
### Description of feature (or steps to reproduce if bug)

The swagger specification generated for LoopBack models with `forceId` set to `true` (the default in LoopBack 3.0) is incorrect. The model definition marks the `id` property as required, but the `create` operation is rejecting requests with `id` property set.

This causes problems in API Connect, where the sample request data generated by Editor contains `id` now.

### Expected result

The sample request generated by APIC Editor for `create` operation does not contain `id` property.

### Actual result (if bug)

The sample request generated by APIC Editor for `create` operation contains `id` property.

### Additional information (Node.js version, LoopBack version, etc)

 - loopback@3 
 - loopback-swagger@latest

The default value of `forceId` was changed to `true` in LoopBack 3.0 for security reasons, see https://github.com/strongloop/loopback-datasource-juggler/pull/982 and https://github.com/strongloop/loopback-datasource-juggler/issues/849

bajtos:
The petstore example shows the approach we can use too.

https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/yaml/petstore-expanded.yaml

Check out their POST /addPet - it takes a body of type NewPet. GET /pet returns a type Pet, which is allOf NewPet, plus id.

bajtos:
Instead of hard-coding detection of `forceId`, we should come up with a property flag indicating whether a property may or may not be in `create` data. This flag then can be used e.g. for Cloudant's `_rev` field, see https://github.com/strongloop/loopback-connector-cloudant/issues/55

ritch:
This is not a blocker for 3.x because we can fix this in the swagger generator.

bajtos:
> Instead of hard-coding detection of `forceId`, we should come up with a property flag indicating whether a property may or may not be in `create` data. This flag then can be used e.g. for Cloudant's `_rev` field, see https://github.com/strongloop/loopback-connector-cloudant/issues/55

Similarly, we need to specify that `User.create` body must contain a required `password` property, that's otherwise hidden from all other payloads.

bschrammIBM:
This issue is producing a 422 error in the Testing a LoopBack project tutorial. It is causing confusion for the customers. Any chance it could get fixed?

bajtos:
@bschrammIBM Thank you for bringing this issue into our attention. Could you please give us the link to "Testing a LoopBack project tutorial" you have mentioned?

[...]

PR #3548, opened 2017-08-05
Support createOnlyInstance in model

rashmihunt:
This is part of solution for https://github.com/strongloop/loopback/issues/2924. 

Other PRs involved in the solution of issue https://github.com/strongloop/loopback/issues/2924
 - https://github.com/strongloop/loopback-datasource-juggler/pull/1453
 - https://github.com/strongloop/loopback-swagger/pull/92

@bajtos Work In Progress PR. Please take a initial look.

Changes include
- setting up createOnlyInstance in PersistentModel.js and model.js

TODO: 
- Add a test case

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -758,13 +758,11 @@ module.exports = function(registry) {
     var pathName =
       (scope.options && scope.options.http && scope.options.http.path) || scopeName;
 
-    var updateOnlyProps = this.getUpdateOnlyProperties();
-    var createOnly = false;
-    // if there is atleast one updateOnly property, then we set createOnlyInstance flag in __create__
-    // to indicate loopback-swagger code to create a separate model instance for create operation only
-    if (updateOnlyProps && updateOnlyProps.length > 0) {
-      createOnly = true;
-    }
+    // if there is atleast one updateOnly property, then we set
+    // createOnlyInstance flag in __create__ to indicate loopback-swagger
+    // code to create a separate model instance for create operation only
+    const updateOnlyProps = this.getUpdateOnlyProperties();
+    const hasUpdateOnlyProps = updateOnlyProps && updateOnlyProps.length > 0;
 
     var isStatic = scope.isStatic;
     var toModelName = scope.modelTo.modelName;
@@ -799,7 +797,7 @@ module.exports = function(registry) {
           type: 'object',
           allowArray: true,
           model: toModelName,
-          createOnlyInstance: createOnly,
+          createOnlyInstance: hasUpdateOnlyProps,
           http: {source: 'body'},
         },
         {arg: 'options', type: 'object', http: 'optionsFromRequest'},
--- a/lib/model.js
+++ b/lib/model.js
@@ -758,7 +758,7 @@ module.exports = function(registry) {
     var pathName =
       (scope.options && scope.options.http && scope.options.http.path) || scopeName;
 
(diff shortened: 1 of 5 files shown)


lib/model.js
```diff
@@ -758,6 +758,14 @@ module.exports = function(registry) {
     var pathName =
       (scope.options && scope.options.http && scope.options.http.path) || scopeName;
 
+    var updateOnlyProps = this.getUpdateOnlyProperties();
+    var createOnly = false;
```
bajtos: Let's use a more descriptive name please, e.g. `hasUpdateOnlyProps`.

Also please use `const` instead of `var`, see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations

```js
const updateOnlyProps = this.getUpdateOnlyProperties();
const hasUpdateOnlyProps = updateOnlyProps && updateOnlyProps.length > 0;
```

lib/persisted-model.js
```diff
@@ -627,6 +627,14 @@ module.exports = function(registry) {
     var typeName = PersistedModel.modelName;
     var options = PersistedModel.settings;
 
+    var updateOnlyProps = PersistedModel.getUpdateOnlyProperties();
+    var createOnly = false;
```
bajtos: Ditto.

rashmihunt:
@bajtos PTAL
- Incorporated code review comments
- Added new tests

lib/model.js
bajtos: This comment does not seem to be addressed yet?

test/remoting.integration.js
```diff
@@ -316,6 +331,12 @@ function getFormattedMethodsExcludingRelations(methods) {
   });
 }
 
+function getCreateMethod(methods) {
+  return methods.filter(function(m) {
```
bajtos: I find it confusing that a function called `getCreateMethod` returns an array of instances. Please use `methods.find` instead of `methods.filter` - see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find

lib/model.js
rashmihunt: @bajtos I have two environment 1) loopback application where I can test with the API designer end to end 2) my git hub branch where I can do run tests and checkins etc. I addressed above comments in the first env to make sure I test end-to-end through API Explorer with these code review comment changes and forgot to do same change in 2nd env/local git hub branch. Tests were passing since my previous code was logically correct. Anyway, done the above change now...

test/remoting.integration.js
rashmihunt: @bajtos Done. PTAL

rashmihunt:
@slnode test please

test/remoting.integration.js
```diff
@@ -316,6 +331,12 @@ function getFormattedMethodsExcludingRelations(methods) {
   });
 }
 
+function getCreateMethod(methods) {
+  return methods.find(function(m) {
+    return (m.name === 'create');
```
bajtos: FWIW, this can be simplified to a one-liner:

```js
return methods.find(m => m.name === 'create');
```

rashmihunt:
@slnode test please

rashmihunt:
loopback-connector-swagger failures are not related to this PR.  Looks like 6 tests are failing on master branch for this connector for sometime. Opened an issue for this - https://github.com/strongloop/loopback-connector-swagger/issues/45

PR #3556, opened 2017-08-10
fix(validatePassword): Reword password too long error.

STRML:
This password is hashed before entering the database,
there is no need for a hardcoded size limit.

Additionally, it looks very bad when a server error
echoes back a plaintext password.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -16,6 +16,8 @@ var path = require('path');
 var qs = require('querystring');
 var SALT_WORK_FACTOR = 10;
 var crypto = require('crypto');
+// bcrypt's max length is 72 bytes;
+// See https://github.com/kelektiv/node.bcrypt.js/blob/45f498ef6dc6e8234e58e07834ce06a50ff16352/src/node_blf.h#L59
 var MAX_PASSWORD_LENGTH = 72;
 var bcrypt;
 try {
@@ -993,18 +995,22 @@ module.exports = function(User) {
 
   User.validatePassword = function(plain) {
     var err;
-    if (plain && typeof plain === 'string' && plain.length <= MAX_PASSWORD_LENGTH) {
-      return true;
-    }
-    if (plain.length > MAX_PASSWORD_LENGTH) {
-      err = new Error(g.f('Password too long: %s', plain));
-      err.code = 'PASSWORD_TOO_LONG';
-    } else {
-      err =  new Error(g.f('Invalid password: %s', plain));
+    if (!plain || typeof plain !== 'string') {
+      err = new Error(g.f('Invalid password.'));
       err.code = 'INVALID_PASSWORD';
+      err.statusCode = 422;
+      throw err;
     }
-    err.statusCode = 422;
-    throw err;
+
+    // Bcrypt only supports up to 72 bytes; the rest is silently dropped.
+    var len = Buffer.byteLength(plain, 'utf8');
+    if (len > MAX_PASSWORD_LENGTH) {
+      err = new Error(g.f('The password entered was too long. Max length is %d (entered %d)',
+        MAX_PASSWORD_LENGTH, len));
+      err.code = 'PASSWORD_TOO_LONG';
+      err.statusCode = 422;
+      throw err;
+    }
   };
 
   User._invalidateAccessTokensOfUsers = function(userIds, options, cb) {
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -430,7 +430,7 @@ describe('User', function() {
         var u = new User({username: 'foo', password: pass73Char});
         assert(false, 'Error should have been thrown');
       } catch (e) {
-        expect(e).to.match(/Password too long/);
+        expect(e).to.match(/password entered was too long/);
         done();
       }
     });
@@ -462,7 +462,7 @@ describe('User', function() {
         if (err) return done(err);
         User.resetPassword({email: 'b@c.com', password: pass73Char}, function(err) {
           assert(err);
-          expect(err).to.match(/Password too long/);
+          expect(err).to.match(/password entered was too long/);
           done();
         });
       });
@@ -474,7 +474,7 @@ describe('User', function() {
         .then(
           success => { throw new Error('changePassword should have failed'); },
           err => {
-            expect(err.message).to.match(/Password too long/);
+            expect(err.message).to.match(/password entered was too long/);
 
             // workaround for chai problem
             //   object tested must be an array, an object, or a string,
@@ -493,7 +493,7 @@ describe('User', function() {
         .then(
           success => { throw new Error('setPassword should have failed'); },
           err => {
-            expect(err.message).to.match(/Password too long/);
+            expect(err.message).to.match(/password entered was too long/);
 
             // workaround for chai problem
             //   object tested must be an array, an object, or a string,


dhmlau:
@loay @bajtos , I recalled there was a specific reason we set the password length limit.  Is it still applicable? 
Either way, I agree that we shouldn't print out the actual password.

bajtos:
Hello @STRML, thank you for the pull request.

>This password is hashed before entering the database, there is no need for a hardcoded size limit.

This is unfortunately not true. We are using `bcrypt` to hash the password, and apparently bcrypt (or at least the implementation we use) truncates the input to the first 72 characters. In other words, if you use a password that's 73 characters long, then at login, you can replace the last character with any other one and as long as the first 72 chars match your password, the system will let you through.

The change was implemented by #2580, see https://loopback.io/doc/en/lb2/Security-advisory-08-16a-2016.html for more details.

I am afraid we cannot accept this part of your pull request.

> Additionally, it looks very bad when a server error echoes back a plaintext password.

Good point, let's fix the error messages then!

Last but not least - please make sure `npm test` is passing.

STRML:
Okay - a few things to do:

* Bcrypt uses the first 72 bytes, not first 72 characters, so we need to use `Buffer.byteLength` if we're really going to validate this.
* The error message really shouldn't echo back the password & perhaps should explain why this limit exists
* Test fixes

bajtos:
> Okay - a few things to do:

Sounds like a good plan to me!

STRML:
Please re-review - I have another suggestion;

In `User.hashPassword`, we could safely add the following:

```js
  this.validatePassword(plain);
  if (plain > MAX_BCRYPT_PASSWORD_LENGTH) {
    plain = crypto.createHash('sha256').update(plain).digest('latin1');
  }
  var salt = bcrypt.genSaltSync(this.settings.saltWorkFactor || SALT_WORK_FACTOR);
  return bcrypt.hashSync(plain, salt);
};
```

This wouldn't break backcompat.

STRML:
Additionally, could your team please comment on the significance of the `$` in the `email` and `password` setters?

```
      if (typeof plain !== 'string') {
        return;
      }
      if (plain.indexOf('$2a$') === 0 && plain.length === 60) {
        // The password is already hashed. It can be the case
        // when the instance is loaded from DB
        this.$password = plain;
      } else {
      }
    };
```

I can't find any documentation on getters/setters anyway (just some SO answers & GH issues), nor anything on why the set is to `this.$password` rather than `this.password`.

bajtos:
> In User.hashPassword, we could safely add the following:

```js
  if (plain > MAX_BCRYPT_PASSWORD_LENGTH) {
    plain = crypto.createHash('sha256').update(plain).digest('latin1');
  }
```

Well, while I am not an expert on security and crypto, I have read enough on storing password securely to be not so sure about the safety of this approach.

AFAIK, the hashing algorithm `bcrypt` is designed specifically to make password hashing _reasonably slow_. Not too slow to hurt server performance, but slow enough to significantly slow down any attacker trying a brute-force attack to obtain plaintext passwords from a compromised hashed values. It also uses computations that are difficult to run on GPUs.

Compare that with sha256, which was designed to run fast and which is easy to run on GPUs. A quick search on the internet found [this article](https://pthree.org/2014/12/26/sha512crypt-versus-bcrypt/), where they claim a 25-GPU cluster can work through 50 billion SHA-512 cryptographic hashes per second.

Having said that, trying all combinations of a 73-characters long password where each character can have one of 100 possible values, with the rate of 50 billion checks per seconds, will still take millions of years, so I guess the speed of the hashing algorithm is no longer relevant for such long password.

In that light, I suppose we can work around bcrypt password-length limitation by using your approach. Are you sure it's worth the extra complexity (and future maintenance costs) though? I mean do you have real users demanding passwords longer than 72 characters?

bajtos:
> I can't find any documentation on getters/setters anyway (just some SO answers & GH issues), nor anything on why the set is to this.$password rather than this.password.

My knowledge of this part is pretty minimal, it's possible that `$`-prefixed properties is something we inherited from the original `juggler` module.

> I want to propose making hashPassword async, which would require using a before save hook, but would have the benefit of not blocking the event loop on registration.

I think `persist` hook may be a better one. Other than that, if we can find a reasonable way how to refactor the current password handling to make it async, ideally to support proper validation too (see #251 reported by yourself @STRML), then I am all for it!

raymondfeng:
The `$password` is probably related to https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L589-L604.

common/models/user.js
```diff
@@ -993,18 +995,22 @@ module.exports = function(User) {
 
   User.validatePassword = function(plain) {
     var err;
-    if (plain && typeof plain === 'string' && plain.length <= MAX_PASSWORD_LENGTH) {
-      return true;
+    if (!plain || typeof plain !== 'string') {
+      err = new Error(g.f('Invalid password.'));
+      err.code = 'INVALID_PASSWORD';
```
loay: @STRML @bajtos 
I agree about removing the print out of the password and unifying the error code as well, however, here is a thought, I think we should not even point to the direction of the error .. In other words, we should not let the error code say if it is wrong username and/or password .. It should be more vague for security purposes .. Like 
`INVALID_LOGIN_INFO` or something similar.. That's my 2 cents .. the rest of the PR is LGTM

kjdelisle: I disagree; this isn't about telling the user (or potential attacker) that the password was _correctly formatted, but invalid_, only that the format itself is what's problematic. The only information it reveals is information they should've already had (the password should be a string!)

loay: That validation is not only for newly registered users, but for resetting passwords as well, so we can be pointing a potential hackers to to where they should dig  deeper :)

STRML:
@bajtos 

> Well, while I am not an expert on security and crypto, I have read enough on storing password securely to be not so sure about the safety of this approach.

What I'm proposing is not in *replacing* bcrypt with SHA256, but rather in running a round of SHA256 before bcrypt on passwords > 72.

There are a few motivations for this. For one, I don't want to refuse passwords over 72 characters (especially with the current error message). Users who see this think "Are you even hashing my password? Why would this be an issue?" - I would think the same about any service that does this. Inputting large passwords is a basic step pentesters use to feel out a system, and I don't want to give them, or customers, the impression we're doing something improper here.

So if we're going to accept > 72char passwords, there are two options:

1. Silently truncate, which may be dangerous depending on the user's password scheme. 
  * Imagine a scheme with a long prefix, perhaps generated via the static password feature on a yubikey, plus a site-specific identifier; pretty popular in some circles). I know people are doing this.
2. Run *only* passwords over 72 characters through a round of SHA256 first. `latin1` encoding produces 48 bytes of output; unfortunately we can't pass raw Buffers into bcrypt. By my understanding of SHA, it is perfectly fine to also use SHA512 and let bcrypt truncate.
  * This has the distinct advantage of giving us as much as 472 bits of entropy without a possible prefix attack, and it's *very* fast.
  * It also will not break old clients as long as it's only applied to passwords over 72 characters.
  * There is an edge case around the 50-60 character mark, as Loopback was doing the 72-char check incorrectly; it should have been using `Buffer.byteLength` instead of `String#length`. Some of those passwords, if they are using non-ASCII characters, could occupy an uncanny valley where they are still being truncated. This is an edge case small enough that I'm willing to ignore it.

STRML:
@bajtos:

> I think persist hook may be a better one.

I checked that out - unfortunately it appears the model is read-only at that point (`ctx.currentInstance`). I'm using the following in our own app:

```js
User.coercePassword = function(plain: string): string {
  if (Buffer.byteLength(plain) > 72) {
    // If we've gone beyond the bcrypt max, run through a round of sha512.
    return crypto.createHash('sha512').update(plain).digest('latin1');
  }
  return plain;
};

User.hashPassword = function(_plain: string) {
  throw new Error('Synchronous bcrypt is no longer supported.');
};

User.hashPasswordAsync = function(plain: string, cb?: ?NodeCallback<string>): Promise<string> {
  return Promise.try(() => {
    this.validatePassword(plain);
    return bcrypt.genSalt(this.settings.saltWorkFactor || 10);
  })
  .then((salt) => bcrypt.hash(User.coercePassword(plain), salt))
  .nodeify(cb);
};

User.prototype.hasPassword = function(plain: string): Promise<boolean> {
  if (this.password && plain) {
    return bcrypt.compare(plain, this.password);
  }
  return Promise.resolve(false);
};

// This does a synchronous hash, we don't want that.
delete User.setter.password;

User.observe('before save', function(ctx, next) {
  const model = ctx.instance;
  if (!model) return next();
  model.lastUpdated = new Date();
  const isHashed = (model.password.indexOf('$2a$') === 0 && model.password.length === 60);
  if (isHashed) return next();
  model.constructor.hashPasswordAsync(model.password, function(err, pass) {
    if (err) return next(err);
    model.password = pass;
    return next();
  });
});
```

If that seems sane to you, I'll port it into Loopback (including callback compatibility, of course, and without deprecating `hashPassword`).

common/models/user.js
bajtos: @loay thank you for the comment. I am afraid I'm with @kjdelisle here, but let me explain in more detail:

In my understanding, `validatePassword` is called only when setting/updating the password value, which something performed only by the user themselves, and therefore it's ok to tell them they provided an invalid password value.

> but for resetting passwords as well, so we could be pointing a potential hackers to where they should dig deeper

Assuming you mean "reset password" functionality - the new password is entered only after the person opened the link sent by email from the initial "password reset" page. In order to allow the attacker to access the page for entering a new password, they must have been able to read that password-reset email, in which case we have much bigger trouble than letting the attacker know the maximum password length!

bajtos:
> Run only passwords over 72 characters through a round of SHA256 first.

Sounds reasonable, could you please open a new pull request for that?

Regarding async hashing - let's discuss this in a new GH issue please. One thing to note: recently we have introduced a change (behind a feature flag `UserModel.settings.rejectPasswordChangesViaPatchOrReplace`) that requires clients to call `changePassword` or `setPassword` when changing the password value. It makes me wonder if this could be leveraged instead of the "before save" or "persist" hook, would it make the code simpler?

> I checked that out - unfortunately it appears the model is read-only at that point (`ctx.currentInstance`)

You should use `ctx.data`. `ctx.currentInstance` is not set when creating a new model (for example).

bajtos:
@STRML BTW I have improved the commit message and `git push -f` into your feature branch. I'll land this pull request after the CI build finishes in green.

bajtos:
Landed, thank you @STRML for the contribution!

STRML:
Thanks @bajtos, will make new PRs for the rest.

PR #3562, opened 2017-08-14
Rectify to model

Traksewt:
### Description
Allow after save and after delete to be overwritten on the model. This is needed for updateAll when using change detection, as the rectify step cannot determine which items have changed so it rectifies the entire model table. For example, if there are 100,000 items in a model table, and you do an updateAll with a where clause to update 20 items, the rectify will run against all 100,000 items as it does not know which items have changed.

If this function can be overwritten, then the application can handle rectify intelligently (e.g. by looking at a custom tenantID column).

#### Related issues

- None

### Checklist

- [x] New tests added or existing tests modified to cover all changes.

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -1648,9 +1648,12 @@ module.exports = function(registry) {
         'which requires a string id with GUID/UUID default value.');
     }
 
-    Model.observe('after save', rectifyOnSave);
-
-    Model.observe('after delete', rectifyOnDelete);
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

Traksewt:
Will redo it neatly with delete unit tests too.

PR #3563, opened 2017-08-14
Update all rectify

Traksewt:
### Description
Allow getIdsFromWhere to be overwritten on the model. This is needed for updateAll when using change detection, as the rectify step cannot determine which items have changed so it rectifies the entire model table. For example, if there are 100,000 items in a model table, and you do an updateAll with a where clause to update 20 items, the rectify will run against all 100,000 items as it does not know which items have changed.

The getIdsFromWhereByModelId handles finding ids in 'inq' where clauses by default.

Now that this function can be overwritten, the application can handle more complex rectify where clauses more intelligently (e.g. by looking at a custom tenantID column to find all ids that would have changed).

#### Related issues

- None

### Checklist

- [x] New tests added or existing tests modified to cover all changes.

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -1648,9 +1648,12 @@ module.exports = function(registry) {
         'which requires a string id with GUID/UUID default value.');
     }
 
-    Model.observe('after save', rectifyOnSave);
-
-    Model.observe('after delete', rectifyOnDelete);
+    Model.observe('after save', function (ctx, next) {
+      Model.rectifyOnSave(ctx, next);
+    });
+    Model.observe('after delete', function (ctx, next) {
+      Model.rectifyOnDelete(ctx, next);
+    });
 
     // Only run if the run time is server
     // Can switch off cleanup by setting the interval to -1
@@ -1671,7 +1674,7 @@ module.exports = function(registry) {
     }
   };
 
-  function rectifyOnSave(ctx, next) {
+  PersistedModel.rectifyOnSave = function(ctx, next) {
     var instance = ctx.instance || ctx.currentInstance;
     var id = instance ? instance.getId() :
       getIdFromWhereByModelId(ctx.Model, ctx.where);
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
Hello @Traksewt, thank you for the pull request. I'll try to take a closer look later this week.

Traksewt:
I added the context to the getIdsFromWhere, because since this is done in the after save, the data may have been changed already so querying again to get the IDs will not give the same data as querying before the update.

Now with the context being passed in, you can setup your ID query in the before save, put the IDs in the context, and then retrieve them in the after save to pass on the the rectify. In this case, it means that getIdsFromWhere does not need to be async, but I left it async to keep it flexible as needed. Open to your thoughts on this.

bajtos:
@Traksewt I think your use case is a valid one, especially support for `where` with `inq` operator used to restrict the list of `id` values affected. At the same time, I am worried about the additional complexity this change is bringing into LoopBack core, and also about subtle breaking changes this may introduce.

I have a much simpler counter proposal for you: let's expose `rectifyOnSave` and `rectifyOnDelete` as static PersistedModel methods that can be customized by users (e.g. through a mixin that can implement the improved rectification you have proposed in this pull request).

That way we can keep the complexity outside of loopback core, make it easier for people to customize the behavior even further, or quickly fix a bug in the algorithm deciding what instances to rectify.

Thoughts?

bajtos:
@Traksewt ping, what's the status of this pull request? Are you still keen to get it finished? Should we close it as abandoned?

slnode:

slnode:

slnode:

slnode:

bajtos:
Closing as abandoned. Feel free to reopen if you get time to work on this again.

PR #3565, opened 2017-08-15
Fix relation race condition in model glob

zbarbuto:
### Description

#3504 added support for glob-style patterns when disabling remote methods. However, if a related model is attached _after_ the initial model is configured, the globs do not work. I have updated to reconfigure shared method properties after each new model is attached to remoting.

This was not caught in testing of #3504 as the root model was attached after the related model. Have updated the test to fail without the fix.

### Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -534,83 +534,10 @@ function configureModel(ModelCtor, config, app) {
   config.dataSource = dataSource;
 
   app.registry.configureModel(ModelCtor, config);
-
-  setSharedMethodSharedProperties(ModelCtor, app, config);
-}
-
-function setSharedMethodSharedProperties(model, app, modelConfigs) {
-  var settings = {};
-
-  // apply config.json settings
-  var config = app.get('remoting');
-  var configHasSharedMethodsSettings = config &&
-      config.sharedMethods &&
-      typeof config.sharedMethods === 'object';
-  if (configHasSharedMethodsSettings)
-    util._extend(settings, config.sharedMethods);
-
-  // apply model-config.json settings
-  var modelConfig = modelConfigs.options;
-  var modelConfigHasSharedMethodsSettings = modelConfig &&
-      modelConfig.remoting &&
-      modelConfig.remoting.sharedMethods &&
-      typeof modelConfig.remoting.sharedMethods === 'object';
-  if (modelConfigHasSharedMethodsSettings)
-    util._extend(settings, modelConfig.remoting.sharedMethods);
-
-  // validate setting values
-  Object.keys(settings).forEach(function(setting) {
-    var settingValue = settings[setting];
-    var settingValueType = typeof settingValue;
-    if (settingValueType !== 'boolean')
-      throw new TypeError(g.f('Expected boolean, got %s', settingValueType));
-  });
-
-  // set sharedMethod.shared using the merged settings
-  var sharedMethods = model.sharedClass.methods({includeDisabled: true});
-
-  // re-map glob style values to regular expressions
-  var tests = Object
-    .keys(settings)
-    .filter(function(setting) {
-      return settings.hasOwnProperty(setting) && setting.indexOf('*') >= 0;
-    })
-    .map(function(setting) {
-      // Turn * into an testable regexp string
-      var glob = escapeRegExp(setting).replace(/\*/g, '(.)*');
-      return {regex: new RegExp(glob), setting: settings[setting]};
-    }) || [];
-  sharedMethods.forEach(function(sharedMethod) {
-    // use the specific setting if it exists
-    var methodName = sharedMethod.isStatic ? sharedMethod.name : 'prototype.' + sharedMethod.name;
-    var hasSpecificSetting = settings.hasOwnProperty(methodName);
-    if (hasSpecificSetting) {
-      if (settings[methodName] === false) {
-        sharedMethod.sharedClass.disableMethodByName(methodName);
-      } else {
-        sharedMethod.shared = true;
-      }
-    } else {
-      tests.forEach(function(glob) {
-        if (glob.regex.test(methodName)) {
-          if (glob.setting === false) {
-            sharedMethod.sharedClass.disableMethodByName(methodName);
-          } else {
-            sharedMethod.shared = true;
-          }
-        }
-      });
-    }
-  });
 }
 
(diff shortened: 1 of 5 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

lib/application.js
```diff
@@ -124,6 +124,9 @@ app.model = function(Model, config) {
   if (arguments.length > 1) {
     config = config || {};
     configureModel(Model, config, this);
+    this.on('modelRemoted', function() {
+      setSharedMethodSharedProperties(Model, this, config);
```
bajtos: This will add one listener for each application model, which may trigger maxListener warning.

Is there a way how to install a single listener which will walk through application's models to apply "sharedMethods" configuration?

test/loopback.test.js
```diff
@@ -853,6 +852,7 @@ describe('loopback', function() {
           },
         },
       });
+      app.model(RelatedModel, {dataSource: 'test'});
```
bajtos: I think we should have two tests: one checking the execution path where the related model has been already configured, the second checking the case when the related model is configured later.

bajtos:
@slnode ok to test

lib/application.js
zbarbuto: So the problem was that `setSharedMethodSharedProperties` wasn't exported anywhere. This made it difficult to call it from somewhere useful and setup only one watcher. I've moved it to its own file so it can be called from `loopback.js` and `application.js`. Let me know if you can think of a better solution.

test/loopback.test.js
zbarbuto: Done

lib/application.js
zbarbuto: The other issue is that the model config only lives for the duration of `app.model` run. I stuck it on the `Model.config` property so I could get to it from `loopback.js` but this is probably not the best way to do it (and is causing another test to fail). Is there a better way to make the `config` available on the model itself at a later time?

bajtos: Hey, sorry for the delay.

> So the problem was that setSharedMethodSharedProperties wasn't exported anywhere. This made it difficult to call it from somewhere useful and setup only one watcher. I've moved it to its own file so it can be called from loopback.js and application.js. Let me know if you can think of a better solution.

I like that 👍 

> The other issue is that the model config only lives for the duration of app.model run. I stuck it on the Model.config property so I could get to it from loopback.js but this is probably not the best way to do it (and is causing another test to fail). Is there a better way to make the config available on the model itself at a later time?

I am proposing to include the config in the arguments of `modelRemoted` event, see https://github.com/strongloop/loopback/blob/6ba35c297b178aaa24272b7b04fa4eb3d7f26e41/lib/application.js#L149

```js
// my proposal
this.emit('modelRemoted', Model.sharedClass, config);
```

lib/application.js
```diff
@@ -17,6 +17,7 @@ var extend = require('util')._extend;
 var RemoteObjects = require('strong-remoting');
 var classify = require('underscore.string/classify');
 var camelize = require('underscore.string/camelize');
+var setSharedMethodSharedProperties = require('./set-shared-method-properties');
```
bajtos: Nitpick: Perhaps "configure-shared-methods" would be a better name?

test/loopback.test.js
```diff
@@ -832,7 +832,36 @@ describe('loopback', function() {
       expect(publicMethods).to.be.empty();
     });
 
-    it('hides methods for related models using globs', function() {
+    it('hides methods for related models using globs (model configured first)', function() {
+      var TestModel = app.registry.createModel(uniqueModelName);
+      var RelatedModel = app.registry.createModel(uniqueModelName);
```
bajtos: Both `TestModel` and `RelatedModel` are using the same model name, that's looks suspicious to me. Since we are using local per-app registry now, there is no need to use unique model names anymore.  You can simply do the following:

```js
const TestModel = app.registry.createModel('TestModel');
const RelatedModel = app.registry.createModel('RelatedModel');
// ...
```

(Also note that we are using ES6 syntax now, see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations)

slnode:

slnode:

slnode:

slnode:

lib/application.js
zbarbuto: >I am proposing to include the config in the arguments of modelRemoted event, see
>`this.emit('modelRemoted', Model.sharedClass, config);`

The problem I see here is that that only emits the config for the model being remoted. We need to re-configure every model in the app and - therefore - get at the config for every model somehow. This is why I was storing it on the models individually. Maybe it should be a sub-property of `sharedClass` ie `Model.sharedClass.config`?

test/loopback.test.js
zbarbuto: Ah, I see, wasn't aware this was the case. Would like like me to refactor the whole PR or just these is fine?

bajtos: I think fixing it just here (as you have already done) is fine 👍

lib/application.js
bajtos: You are right, I did not realize that all classes need to be reconfigured. Let me take a closer look to better understand why some tests are failing with your current approach.

bajtos:
@zbarbuto I fixed the test failure in dc12f54 and also cleaned the code a bit. PTAL and let me know if you are ok with the result.

zbarbuto:
@bajtos variable name updated and rebased.

lib/application.js
```diff
@@ -17,6 +17,7 @@ var extend = require('util')._extend;
 var RemoteObjects = require('strong-remoting');
 var classify = require('underscore.string/classify');
 var camelize = require('underscore.string/camelize');
+var configureSharedMethods = require('./configure-shared-methods');
```
bajtos: This `require()` is no longer needed, let's remove it. (I should have removed it myself, sorry!)

zbarbuto: done

bajtos:

calebfroese:
@bajtos Is there a minor release coming soon with this fix?

bajtos:
@calebfroese thank you for the poke, here is the new release: `loopback@3.12.0`

PR #3579, opened 2017-08-23
Handle missing getUpdateOnlyProperties fn

kjdelisle:
### Description

If the current scope (`this`) does not define a `.getUpdateOnlyProperties`
function, the `updateOnlyProps` value will now be set to false.

#### Related issues
Fixes regression in https://github.com/strongloop/loopback/pull/3548

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -630,7 +630,8 @@ module.exports = function(registry) {
     // if there is atleast one updateOnly property, then we set
     // createOnlyInstance flag in __create__ to indicate loopback-swagger
     // code to create a separate model instance for create operation only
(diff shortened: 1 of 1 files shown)


BendingBender:
Same problem still persists in lib/model.js:764

supermafete:
Same problem in lib/model.js:764. I think must be reviewed.

lehni:
I agree with @BendingBender and @supermafete, the problem also exists here and can be fixed in the exact same way:

https://github.com/strongloop/loopback/blob/8488da2e26462ec774548dd0a14f5d463246f9c8/lib/model.js#L769

I will create a follow up PR

Issue #3584, opened 2017-08-25
Nodemailer needs update to latest v4


mitsos1os:
# Description/Steps to reproduce

**Nodemailer** dependency warns for deprecation of used v2 version.
Also latest v4 encapsulates all functionality and results to less modules installation

wzup:
> npm WARN deprecated nodemailer@2.7.2: All versions below 4.0.1 of Nodemailer are deprecated. See https://nodemailer.com/status/

When are you going to update it? 
Does everything work with deprecated Nodemailer?

dhmlau:
Closing in favor of https://github.com/strongloop/loopback/pull/3585

PR #3585, opened 2017-08-25
Update nodemailer to the latest v4 version

mitsos1os:
### Description
Upon installation, **nodemailer** complained for deprecated version v2... Updated to latest version.
Also everything is packed in single file, so this is better for less modules installed

#### Related issues

- connect to #3584

### Checklist


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
Hello @mitsos1os, thank you for the pull request. (Thank you @jannyHou for #3586 too!)

What are the breaking changes between nodemailer 2.x and 4.x? I remember 3.x used an incompatible license, while 4.x reverted back to something we can use again, which is good.

In #3586, @jannyHou mentions that node-mailer 4.x does not support Node.js 4.x. I am afraid this is a show stopper for us. LoopBack 3.x does support Node.js 4.x and dropping this support requires a new semver-major release of LoopBack. (Also note that Node.js 4.x is still officially maintained by the Node.js project until April 2018, see their [LTS schedule](https://github.com/nodejs/LTS)).

What are the downsides of keeping nodemailer version 2.x?

If we decide to bite the bullet and release a new semver-major version of LoopBack, then I think it's better to remove the email support from loopback core into a standalone package/component, to make it easier to introduce breaking changes in the future.

@raymondfeng @kjdelisle thoughts?

mitsos1os:
@bajtos To be sincere, no actual need, I was just led from the notice that version 2.x of nodemailer is reported as deprecated during npm install. Also since everything is packed during a single package, there is the improvement of installing fewer dependencies...
Other than that, I can understand the need to still support NodeJS v4.... I leave it up to you to handle the request accordingly

raymondfeng:
There are a few places in nodemailer 4.x that uses [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)  (`...args`). 

See http://node.green/#ES2015-syntax-rest-parameters. Node 4.x only supports rest parameters under a feature flag.

raymondfeng:
My take is that `email` support is not a key feature of `loopback` core. If upgrading nodemailer to 4.x won't require the rest of `loopback` to use Node 6.x, I'm fine to go ahead.

If a LoopBack user uses `email`, we can say Node 6.x is needed.

bajtos:
> My take is that `email` support is not a key feature of `loopback` core. If upgrading nodemailer to 4.x won't require the rest of `loopback` to use Node 6.x, I'm fine to go ahead.

I am afraid I don't agree with you, @raymondfeng. If there are people running a LoopBack 3.x application on Node.js 4.x and using our email feature, then their application will break after running `npm update`. I don't know how do you feel about that, but I hate it when my dependencies are violating the semver contract and making breaking changes in minor (or even patch) releases. We have been through this pain two years ago, when the support for Node.js 0.10 and later 0.12 was coming to end and packages (loopback dependencies) started to drop support for these Node.js versions.

If we want to upgrade nodemailer to v4.x but don't spend a lot of time/effort on the upgrade, then I am ok with dropping support for sending emails on Node.js v4.x as long as this change is released as `loopback@4.0.0`. We will still need to fix at least our test code to make sure that `npm test` is passing on Node.js 4.x.

bajtos:
Considering there is no pressing need to upgrade nodemailer right now, and because it looks like nobody is willing to invest the time needed to make this upgrade happen _the proper way_, I am proposing to close this pull request for now.

@mitsos1os @jannyHou @raymondfeng Any objections?

mitsos1os:
@bajtos If you need support for NodeJS v4 then there is no need to talk about the merge request... nodemailer v4 requried NodeJS 6+ If you choose to disable email sending for nodejs v4, and accept the merge request, the only thing that could be done is create an extra check in the mail model for node -
 nodemailer version and just forbid the sending... If you would like I could implement it. As for later node versions, I have changed the tests appropriately and everything is already working

bajtos:
We had an internal discussion about the possibility of making a new semver-major loopback release from the current codebase. Unfortunately, a new major release has quite a lot of overhead, for example the documentation on http://loopback.io/doc/ provides different content for each major version. Considering that the next major version of LoopBack will be based on an entirely new codebase (see https://github.com/strongloop/loopback-next), we feel the benefits of upgrading nodemailer via a new semver-major version are not worth the effort a new major version requires.

I am closing this pull request as rejected. We appreciate your contribution though!

slnode:

bajtos:
Node.js 4.x is no longer supported by the Node.js project, it's time to revisit this pull request again 👍 
cc @dhmlau

bajtos:
We need to land https://github.com/strongloop/loopback/pull/3908 first, then I'll rebase this pull request on top of the latest master.

bajtos:
Landed, thank you @mitsos1os for the contribution 😄

PR #3595, opened 2017-08-30
CODEOWNERS: add lehni as a collaborator

bajtos:
🎉 @lehni welcome on board 🎉 

We use CODEOWNERS files to communicate who is maintaining which repository. As a side effect, people listed on the `*` line receive notifications about all new pull requests opened against the repository, thus having a chance to review and comment.

@lehni Please let me know if you are ok with these notifications (IIRC, we are receiving less than one new pull request a week here in `strongloop/loopback`). Alternatively, we can tweak CODEOWNERS to list you as an owner of a specific subset of source files, or even a file that does not exist (in which case you should not be receiving any notifications at all).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,4 +2,4 @@
 # Each line is a file pattern followed by one or more owners,
 # the last matching pattern has the most precendence.
 
-* @bajtos @fabien @clarkorz @ebarault
+* @bajtos @fabien @clarkorz @ebarault @lehni


lehni:
@bajtos thanks for the welcome! And yes, let's add my name to this list the way you suggest. If it gets too much, I can always suggest the change to have a more specific filter. 

This only concerns the `loopback` module, not the other ones I assume. Will the same be done on the other modules that you mentioned in the invite?

bajtos:
>  If it gets too much, I can always suggest the change to have a more specific filter.

Exactly :)

> This only concerns the loopback module, not the other ones I assume. Will the same be done on the other modules that you mentioned in the invite?

Yes, of course. I just wanted to know your preference about notifications before opening pull requests against the other modules.

PR #3609, opened 2017-09-08
Fix handling of user verification options

### LINKED ISSUE #3599: Verification mail fail when is sent more than once
When a user is already created and the verify endpoint is executed more than ONCE, then the verifyHref arrived with 2 or more tokens, the previous generated tokens and the last one. The last token is added at the end. That means, "...token=dsjadha...&token=plndsan...&token=thisisthelastone".

Thanks,
Sebastián.

sebastianfelipe:
### Description
Hi,

I submit this pull request because the Issue #3599, when the verification functionality is sent more than once, then the data change drastically. That is because the copy by reference that the User.getVerifyOptions() do inside its definition (when the correct verifyOptions settings are returned). After that, the data is modified and that modification implies a global modification (that means), information can reload and have some problems, like the issue #3599.

The change is in one line and its test should consider the result in the email or a debug of the generated html by User.verify() function.

#### Related issues

- fixes #3599 

### Checklist

- [x] New tests added or existing tests modified to cover all changes in user.test.js -> 'returns same custom verify options after verify user model'

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -603,11 +603,11 @@ module.exports = function(User) {
    */
 
   User.getVerifyOptions = function() {
-    const verifyOptions = {
+    const defaultOptions = {
       type: 'email',
       from: 'noreply@example.com',
     };
-    return this.settings.verifyOptions || verifyOptions;
+    return Object.assign({}, this.settings.verifyOptions || defaultOptions);
   };
 
   /**
@@ -699,10 +699,14 @@ module.exports = function(User) {
     var user = this;
     var userModel = this.constructor;
     var registry = userModel.registry;
-
+    verifyOptions = Object.assign({}, verifyOptions);
     // final assertion is performed once all options are assigned
     assert(typeof verifyOptions === 'object',
       'verifyOptions object param required when calling user.verify()');
+
+    // Shallow-clone the options object so that we don't override
+    // the global default options object
+    verifyOptions = Object.assign({}, verifyOptions);
 
     // Set a default template generation function if none provided
     verifyOptions.templateFn = verifyOptions.templateFn || createVerificationEmailBody;
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -1606,6 +1606,34 @@ describe('User', function() {
           done();
         });
 
+        it('returns same verifyOptions after verify user model', () => {
+          const defaultOptions = {
+            type: 'email',
+            from: 'test@example.com',
+          };
+          var verifyOptions = Object.assign({}, defaultOptions);
+          const user = new User({
+            email: 'example@example.com',
+            password: 'pass',
+            verificationToken: 'example-token',
+          });
+          return user
+            .verify(verifyOptions)
+            .then(res => expect(verifyOptions).to.eql(defaultOptions));
+        });
+
+        it('getVerifyOptions() always returns the same', () => {
+          const defaultOptions = {
+            type: 'email',
+            from: 'test@example.com',
+          };
+          User.settings.verifyOptions = Object.assign({}, defaultOptions);
+          var verifyOptions = User.getVerifyOptions();
+          verifyOptions.from = 'newTest@example.com';
+          verifyOptions.test = 'test';
+          expect(User.getVerifyOptions()).to.eql(defaultOptions);
+        });
+
         it('can be extended by user', function(done) {
           User.getVerifyOptions = function() {
             const base = User.base.getVerifyOptions();


slnode:

slnode:

slnode:

slnode:

slnode:

dhmlau:
@sebastianfelipe, thanks for your contribution.  There is failure in `Commit Linter` because the commit message is too long, the first line only accepts 50 characters or less.  You can split the commit message into more than 1 line to fix this problem. 

```
**************************************************
**
**  Linting commit logs
**
**  3 problems found:
**    ae1b351 - Fix issue 3599. Copy by value is needed to be retur: First line should be 50 characters or less (saw 94)
**    a3e6992 - Fix issue 3599. Copy by value is needed to be retur: First line should be 50 characters or less (saw 94)
**    44a9f13 - Fix issue 3599. Copy by value is needed to be retur: First line should be 50 characters or less (saw 94)
**
**************************************************
```

It would be good to squash the commits when this PR is ready to land.  Before that, could you please provide tests related to your fix.  A place to add would be `test\user.test.js`.  Thanks!

bajtos:
@slnode ok to test

bajtos:
Last but not least, please make sure that `npm test` is passing locally on your computer. We are running code linter as part of the test workflow and your code is violating some of our coding style rules.

>    /home/travis/build/strongloop/loopback/common/models/user.js
>     610:1  error  Trailing spaces not allowed  no-trailing-spaces

sebastianfelipe:
Hi, this is my first time doing a PR. Is just one commit, but I made another one to delete a line and after that I added a line, but I was wrong. I'll make some time to finish this PR.

sebastianfelipe:
Ok guys, I submitted a new test and I run "npm run lint" to check code linter problems and "npm test" (that includes lint), to test. I don't know why I have commit linter problems, I submitted the same copy today but with minimal changes to pass that check but it still notifying me commits formatting problems. The last commit is the only one that matters.

sebastianfelipe:
I think commit a3c010d is valid. I hope it works. At least the test I added should demonstrate that the change is useful you can compare the before and after to see that the reference is modified and everything is sent wrong.

bajtos:
@sebastianfelipe thank you for the update and the added test! I'll try to take a closer look by the end of the next week.

In order to fix the commit message, you need to rewrite your git history - `git rebase -i master` and then `git push -f`. But don't worry about that, I can fix the commit message myself when the code is good to be merged.

bajtos:
While we are discussing commit messages, we try to use a descriptive commit title that works on its own, e.g. in a changelog summary. A reference to an issue number is not very helpful. See http://loopback.io/doc/en/contrib/git-commit-messages.html

In your case, a good commit message title can be e.g. "Fix token duplication in User.verify()"

sebastianfelipe:
Thanks @bajtos, for next times I hope to have learnt better about it.

bajtos:
@sebastianfelipe 
> Thanks @bajtos, for next times I hope to have learnt better about it.

No worries, we all were new to GitHub and open source contributing once; I still remember how many unknowns I had to learn myself :) Keep up the good work and let me know if there is anything I can help you with to make your journey easier.

common/models/user.js
```diff
@@ -607,7 +607,7 @@ module.exports = function(User) {
       type: 'email',
       from: 'noreply@example.com',
     };
-    return this.settings.verifyOptions || verifyOptions;
+    return Object.assign({}, this.settings.verifyOptions || verifyOptions);
```
bajtos: This is actually a wrong place where to fix the problem. We need to change `User.verify` so that it does not modify the provided `verifyOptions` argument.

test/user.test.js
```diff
@@ -1606,6 +1606,28 @@ describe('User', function() {
...
+          });
+          user
+            .verify(User.getVerifyOptions())
+            .then(function(res) {
+              expect(verifyOptons).to.eql(User.settings.verifyOptions);
+              done();
+            })
+            .catch(function(err) {
+              done(err);
+            });
```
bajtos: Mocha supports promises, there is no need to mix promises with callbacks.

sebastianfelipe:
Hi @bajtos, I had a look at your change, but I disagree. The problem root is in getVerifyOptions, I think the change must be there, because that function could be used in other places, for example, to develop custom verify endpoints. If somebody has good programming practices, probably they will reuse that code and that bug wouldn't be fixed at all then. A good fix should be at User.getVerifyOptions() directly.

bajtos:
@sebastianfelipe Makes sense. In that case we need to fix both places and there should be two unit tests - one checking that `verify` does not modify the content of `verifyOptions` argument, and the other checking that `getVerifyOptions` always returns a fresh object.

Could you make the change yourself please?

sebastianfelipe:
The idea is when getVerifyOptions is used, and User.verify is emitted, then User.settings.verifyOptions (that requires it existence), should be the same always. What should be the difference with respect to the last test you provide? I see that works for both cases very well, is self-explained.

slnode:

slnode:

slnode:

bajtos:
@sebastianfelipe 

> The idea is when getVerifyOptions is used, and User.verify is emitted, then User.settings.verifyOptions (that requires it existence), should be the same always. What should be the difference with respect to the last test you provide? I see that works for both cases very well, is self-explained.

Let me try to explain this in more details. In my mind, there are three parts in play here:

 1. `User.verify` accepting `verifyOptions`
 2. `User.getVerifyOptions` returning a default value for these options
 3. A glue that calls `getVerifyOptions` to obtain the value to pass to `User.verify` when it's invoked via REST API

In order to fix the issue when `User.verify` is invoked as a remote method via REST API, your original fix would be sufficient.

However, `User.verify` can be called from JavaScript code too, possibly with custom `verifyOptions` that are not created via `User.getVerifyOptions`. Consider the following code:

```js
const VERIFY_EMPLOYEE_OPTS = {
  type: 'email',
  from: 'admin@internal.example.com'
};

MyUser.prototype.verifyEmployee = function(options, cb) {
  return this.verifyUser(VERIFY_EMPLOYEE_OPTS, options, cb);
};
// setup remoting, e.g. as `POST /api/users/:id/verify-employee
```

If we fix `getVerifyOptions` only and don't fix `verify` to not modify the options passed in, then the code above will be still not working as expected.

That's why I think we need to fix both places - `getVerifyOptions` but also in `verify`.

sebastianfelipe:
Ok @bajtos, I understand your point, is when verifyUser is used. Ok, so it is necessary to make a test to that case of use, instead to the verifyOptions or not? Cause the last test implies the getVerifyOptions.

bajtos:
> Ok, so it is necessary to make a test to that case of use, instead to the verifyOptions or not?

There should be two tests:

 - one verifying that `getVerifyOptions` always return a new object
 - the other verifying that `verifyUser` does not modify the content of `verifyOptions` argument

test/user.test.js
```diff
@@ -1606,6 +1606,24 @@ describe('User', function() {
...
+            from: 'test@example.com',
+          };
+          User.settings.verifyOptions = Object.assign({}, verifyOptons);
+          const user = new User({
+            email: 'example@example.com',
+            password: 'pass',
+            verificationToken: 'example-token',
+          });
+          return user
+            .verify(User.getVerifyOptions())
```
bajtos: In one test, we should call `user.verify(verifyOptions)` - don't use `getVerifyOptions`.

In the second test, we should call `getVerifyOptions` and modify the returned value directly from the test, without calling `user.verify`.

sebastianfelipe:
Done @bajtos, changes were added and tested (npm test). I hope it works now. Let me know if everything comes up :)

sebastianfelipe:
Tried to rebase as you said, but the 15 commits before weren't avoid. GG. I hope the changes works.

sebastianfelipe:
I'm done, cannot make a good rebase :c. Please review my test, and also the loopback.test.js because it didn't pass the "npm test". I added 2 lines and solved it (2 values in a array to be as expected). Let me know what you think about it @bajtos.

test/loopback.test.js
```diff
@@ -94,6 +95,7 @@ describe('loopback', function() {
         'template',
         'token',
         'urlNotFound',
+        'urlencoded',
```
bajtos: This failing test has been already fixed by https://github.com/strongloop/loopback/pull/3637 in a way that should be more resilient to future changes. I'll revert this file as part of final cleanup of the patch.

test/user.test.js
```diff
@@ -1606,6 +1606,42 @@ describe('User', function() {
...
+          const user = new User({
+            email: 'example@example.com',
+            password: 'pass',
+            verificationToken: 'example-token',
+          });
+          user
+            .verify(verifyOptions)
+            .then(function(res) {
+              expect(verifyOptions).to.eql(defaultOptions);
+              done();
```
bajtos: Mocha supports promises out of the box, there is no need to use `done` callback. Just return the promise from the test function.

```js
it('returns same verifyOptions after verify user model', () => {
  // ...
  return user
    .verify(verifyOptions)
    .then(res => expect(verifyOptions).to.eql(defaultOptions));
});
```

Also note that we prefer arrow functions in new code, see http://loopback.io/doc/en/contrib/style-guide.html#arrow-functions

I'll fix this myself to speed things up.

bajtos:

PR #3632, opened 2017-09-29
[WIP-Don't Merge] Fix parallel user pass

loay:
#### Related issues

connect to https://github.com/strongloop-internal/scrum-apex/issues/309

### Checklist


test/user.test.js
```diff
@@ -80,14 +99,21 @@ describe('User', function() {
 
     // Update the AccessToken relation to use the subclass of User
     AccessToken.belongsTo(User, {as: 'user', foreignKey: 'userId'});
+    AccessToken.belongsTo(BasicUser, {as: 'user', foreignKey: 'userId'});
+    AccessToken.belongsTo(AdminUser, {as: 'user', foreignKey: 'userId'});
```
bajtos: The setup where a single access token is shared by multiple User models triggers different code paths in many auth-related methods. We need to preserve the tests in this file to keep using a single User model.

Please add your new test(s) to [test/multiple-user-principal-types.test.js](https://github.com/strongloop/loopback/blob/c9913927e53ddf27fb5b6be211e46cfdf572380b/test/multiple-user-principal-types.test.js) instead.

test/user.test.js
```diff
@@ -2428,6 +2454,47 @@ describe('User', function() {
             .expect(200));
       });
 
+      it('should reject password reset role/user relations', function(done) {
+        var accessToken2, AdminUserReset;
+        async.series([
```
bajtos: Please use promise-based API in the tests to make simpler.

```js
BasicUser.create(...)
  .then(() => AdminUser.create(...))
  .then(() => BasicUser.login(...))
  .then(() => {
    // ...
  });
```

test/user.test.js
```diff
@@ -2428,6 +2454,47 @@ describe('User', function() {
...
+              accessToken2 = accessToken1;
+              if (err) return next(err);
+              assert(accessToken1.id);
+              next();
+            });
+          },
+          function(next) {
+            request(app)
+              .post('/test-basic/reset-password')
+              .set('Authorization', accessToken2.id)
```
bajtos: IIRC, the `reset-password` endpoint does not require authentication - how is the user supposed to provide a valid access token when they cannot remember their password needed to create one?

bajtos: Ah sorry, I got confused because there are two similar endpoints: `/reset` is called to issue a password-reset token, `/reset-password` is called to perform password reset.

Note that using a token created by `login` to reset a password is a design flaw that's supported for backwards compatibility only. Your test should be calling `BasicUser.resetPassword` instead of `BasicUser.login` to create the token.

I also think that in order to reproduce the problem you are trying to fix, there is no need to use REST API and you can invoke `BasicUser.setPassword`/`AdminUser.setPassword` instead.

PR #3637, opened 2017-10-04
test: fix too strict test assertion

bajtos:
Rework the test verifying properties of `loopback` to ignore new express properties added after the test was written.

This change fixes the currently broken CI build and prevents similar breakages to happen in the future.


bajtos:
For future reference, here is the related ExpressJS issue that added back body-parsing middleware: https://github.com/expressjs/express/issues/2211

PR #3647, opened 2017-10-09
Handle missing getUpdateOnlyProperties fn on Model

lehni:
### Description

This is a follow-up PR for #3579, which only addressed `PersistedModel`, while still leaving `Model` vulnerable to the issue. See https://github.com/strongloop/loopback/pull/3579#issuecomment-335081112

If the current scope (`this`) does not define a `.getUpdateOnlyProperties`
function, the `updateOnlyProps` value will now be set to `false`.

#### Related issues
- Finishes incomplete fix from #3579
- Fixes regression in https://github.com/strongloop/loopback/pull/3548

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -766,7 +766,8 @@ module.exports = function(registry) {
     // if there is atleast one updateOnly property, then we set
     // createOnlyInstance flag in __create__ to indicate loopback-swagger
     // code to create a separate model instance for create operation only
-    const updateOnlyProps = this.getUpdateOnlyProperties();
+    const updateOnlyProps = this.getUpdateOnlyProperties ?
+      this.getUpdateOnlyProperties() : false;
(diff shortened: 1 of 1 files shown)


lehni:
@bajtos I get a lot of failing checks but can't look at the details and it's all Jenkins. It looks like the site is down?

bajtos:
>  I see a lot of failing checks but can't look at the details and it's all cis-jenkins. It looks like the site is down?

Unfortunately the Jenkins server is accessible only from within IBM network :( I checked few of the downstream failures, they don't seem to be related to this change. Feel free to ignore them.

> how many reviews should PRs like these get before I can merge?

In general, I think it's a good idea to allow 24 hours for people in different timezones to chime in. However, for a small change like this, I think my approval should be sufficient :)

Let's go ahead and land this pull request!

PR #3650, opened 2017-10-12
update strong-globalize to 3.1.0

shimks:
Testing whether the package works with the latest version of strong-globalize

- connect to strongloop-internal/scrum-apex#327


bajtos:
Landed, thanks!

PR #3666, opened 2017-10-24
Fix "POST /reset-password" for multi-user setup

kjdelisle:
### Description

Fix the code extracting current user id from the access token provided
in the HTTP request, to allow only access tokens created by the target
user models to execute the action.

This fixes the following security vulnerability:

* A UserA with id 1 (for example), requires a resetToken1

* A UserB with the same id requires a resetToken2.

* Using resetToken2, use the UserAs/reset-password endpoint and change
  the password of UserA and/or vice-versa

Fix by @bajtos, I'm just here ~so I don't get fined~ to merge it.

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -1192,9 +1192,7 @@ module.exports = function(User) {
       {
         description: 'Reset user\'s password via a password-reset token.',
         accepts: [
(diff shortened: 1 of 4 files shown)


kjdelisle:
released under loopback@3.16.0

Issue #3615, opened 2017-09-13
Swagger: skipping unknown type "$new_AccessToken".


agung-wete:
# Description/Steps to reproduce
Fresh install loopback, then run `node .`

# Expected result
No error

# Result
```
Swagger: skipping unknown type "$new_AccessToken".
Web server listening at: http://localhost:3000
Browse your REST API at http://localhost:3000/explorer
```

# Additional information
Version:
loopback 3.11.1
loopback-swagger 4.3.0
loopback-explorer 2.0.2

win32 x64 6.11.3

george-oakling:
Same problem here

iorlandini:
Me too! I have the same problem with all hasMany relations!

ArturoBurela:
I have the same issue, but does not seem to cause malfunction.

george-oakling:
it causes malfunction when you generate swagger for external use, you get errors in parsing JSON/YAML definition

fabiorecife:
Me too!
Swagger: skipping unknown type "$new_AccessToken".
Web server listening at: http://localhost:3000
Browse your REST API at http://localhost:3000/explorer

loopback@3.11.1
loopback-swagger@5.0.0

macOs Sierra
Darwin Kernel Version 16.7.0
Node v7.10.0

septk:
same here

Balipeace:
same

gpzenpy:
same

vjtc0n:
same here, dont' know what's going on?

ahelliet:
Solve this by update the package.json with "loopback": "3.2.1", i'm not sure this is the right way but it's works for me

ArturoBurela:
@ahelliet, I tried your solution and the message disappears, but running lb export-api-def still produces invalid definitions.

agung-wete:
Updated to
+ loopback@3.14.0
+ loopback-component-explorer@5.1.0

message still there

maximedubois:
same here... ??

biskandar:
Same here .

├── eslint-config-loopback@8.0.0
├─┬ loopback@3.14.0
│ ├── loopback-connector-remote@3.1.1
│ ├─┬ loopback-datasource-juggler@3.12.0
│ │ ├─┬ loopback-connector@4.3.0
│ ├─┬ loopback-phase@3.0.0
│ │ ├── loopback-datatype-geopoint@1.0.0
│ │ ├─┬ loopback-phase@1.4.1
[...]

PR #3674, opened 2017-10-27
Make sure createOnlyInstance flag is calculated on the target model

raymondfeng:
For scoped or related create method, the createOnlyInstance flag should
be calculated on the target model. For example, User.createAccessTokens
should set the flag only if AccessToken has updateonly properties.

#### Related issues

- connect to https://github.com/strongloop/loopback/issues/3615

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+2017-10-28, Version 3.16.1
+==========================
+
+ * Fix createOnlyInstance for related methods (Raymond Feng)
+
+
 2017-10-24, Version 3.16.0
 ==========================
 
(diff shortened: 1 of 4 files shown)


rashmihunt:
@raymondfeng May be we should add a test case in CI for this specific failure since it's not covered with existing CI.

raymondfeng:
@rashmihunt PTAL

bajtos:
@raymondfeng I can see that 6570b94 has been released as v3.16.1, but this pull request was not landed to master 😱 I am going to run `git merge v3.16.1` to fix that.

PR #3675, opened 2017-10-30
Fix "POST /change-password" for multi-user setup

bajtos:
Fix the code extracting current user id from the access token provided
in the HTTP request, to allow only access tokens created by the target
user models to execute the action.

This fixes the following security vulnerability:

* We have two user models, e.g. Admin and Customer

* We have an Admin instance and a Customer instance with the same
  id and the same password.

* The Customer can change Admin's password using their
  regular access token.

This is a follow-up for https://github.com/strongloop/loopback/pull/3666.

#### Related issues

Connect to strongloop-internal/scrum-apex#309

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -1173,9 +1173,7 @@ module.exports = function(User) {
       {
         description: 'Change a user\'s password.',
         accepts: [
-          {arg: 'id', type: 'any',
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

Issue #3644, opened 2017-10-06
Role.isOwner() does not support polymorphic belongsTo relations to multiple User models


lehni:
# Description/Steps to reproduce

See [this comment](https://github.com/strongloop/loopback/pull/3140/commits/e17132d06182e421b3da9a5544f68fc04c61ea17#r104612589
) by @ebarault outlining the problem and how to reproduce it by activating a currently skipped test in [`multiple-user-principal-types.test.js`](https://github.com/strongloop/loopback/blob/8488da2e26462ec774548dd0a14f5d463246f9c8/test/multiple-user-principal-types.test.js#L479-L546):

> This test currently fails, see the inlined comments, support for polymorphic belongsTo should be added to isOwner() to solve this issue.
> I shook my head for almost an hour on this, trying to find a solution as-is or wondering if it could just be solved at all, before realizing this was a design issue. Using standard belongsTo relations between persisted models and user when using multiple user models in an app just won't do it, expect under particular circumstances.
>
> Let's address belongsTo + polymorphic in a follow-up PR.
> I marked the faulty tests as skipped to keep track

# Link to reproduction sandbox

Simply activate this test by changing `it.skip` to `it` here:
https://github.com/strongloop/loopback/blob/8488da2e26462ec774548dd0a14f5d463246f9c8/test/multiple-user-principal-types.test.js#L479

# Expected result

The owner should be resolved using the correct belongsTo relation

# Additional information

Related issues:

- #3140 

I believe as of v3.14.0, this situation is now triggering the following warning on the console:

> The app configuration follows the multiple user models setup as described in http://ibm.biz/setup-loopback-auth The built-in role resolver $owner is not currently compatible with this configuration and should not be used in production

Once this issue is resolved, this warning can be removed.

louie007:
Any updates ?

lehni:
If there were updates, they would be listed in this issue... I personally don't know how to fix this, I just created the issue to make sure it does not get forgotten.

L-David:
sorry, I don't quite understand the warning, does it means we can't have two CustomUserModels having an ownership relationship to the same model instance or we should not use any of the built-in role resolver at all. Or shouldn't be using multiple user config in production and try to supercharge a simple user config.

shaheero:
[...]

PR #3679, opened 2017-10-31
Correct PrinciplaType for multiple extension of User models

akki-ng:
### Get the polymorphic PrincipalType while creating pricipal when multiple extensions of `User` model is used 

```The app configuration follows the multiple user models setup as described in http://ibm.biz/setup-loopback-auth The built-in role resolver $owner is not currently compatible with this configuration and should not be used in production.```

With this pull request $owner can be used.

### DIFF
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -88,7 +88,9 @@ function AccessContext(context) {
   var token = this.accessToken || {};
 
   if (token.userId) {
-    this.addPrincipal(Principal.USER, token.userId);
+    var userPrincipalType =
+      (this.accessToken && this.accessToken.principalType) || Principal.USER;
+    this.addPrincipal(userPrincipalType, token.userId);
   }
   if (token.appId) {
     this.addPrincipal(Principal.APPLICATION, token.appId);

### CODE REVIEW

**Comment by slnode** [comment]

**Comment by slnode** [comment]

**Comment by slnode** [comment]

**Comment by slnode** [comment]

**Comment by slnode** [comment]

**Comment by ebarault** [comment]
hi @akki-ng,
thanks for trying to fix this.
Could you unskip this [test](https://github.com/strongloop/loopback/blob/master/test/multiple-user-principal-types.test.js#L485) and check if your fix passes it ?

**Comment by lehni** [comment]
@ebarault I'm not part of the LB efforts anymore... I've switched to a selection of smaller components, mainly Koa2 and Objection.js.

**Comment by akki-ng** [comment]
Hi @ebarault 
I checked this test and its failing 
```// these 2 tests fail because we cannot resolve ownership with
              // multiple owners on a single model instance with a classic
              // belongsTo relation, we need to use belongsTo with polymorphic
              // discriminator to distinguish between the 2 models
```

I'm not sure if the fix is causing it to fail. But I believe the above fix was not to solve this test case rather to put the correct principalType while adding a principal. Please correct me if I am wrong.

ebarault:
hi @akki-ng 
the PR description states: "With this pull request $owner can be used", so this test needs to pass to comply with this expectation.

knowing the root cause, i was suspecting this PR wouldn't be enough to fix it, but i wanted to give it a try. 

that being said, as it does not fix this test, i'm wondering what the PR fixes, which I can't tell until more tests are added.

so without knowing the added value or the side effects, we cannot move on with this PR as is i'm afraid.

akki-ng:
Hi @ebarault,

After seeing this test I understand that its a not a complete solution to said problem. 

```
With this pull request $owner can be used.
```
By this statement I meant that I could user $owner successfully on the extended User model.
For example, lets us assume that we extend 4 UserModels for an e-commerce platform

1. BusinessUser - InternalUser($businessAdmin, $businessOperation are the roles)
2. Retailer - People who bring their shop online on the platform($retailOwner is role)
3. StoreUser - Users which belong to a particular online store($storeManager, $storeOperations, $storeDeliveryGuy)

A ```Retailer``` user can be added by BusinessUser only and can be modified by ```$owner``` and ```BusinessUser``` so acls can user $owner

A StoreUser can be added by ```BusinessUser and Retailer``` but can be modified by ```$owner $businessOperation $retailOwner``` $retailOwner can be a dynamic role to check if ```effected storeUser instance``` belongs to a retail shop whose owner is in the current accessContext

In this cases getting the correct PrincipalType in role resolver will give information about the current logged in User and such decision over the role of ```$retailOwner``` can be calculated

Perhaps ```With this pull request $owner can be used.``` was not the right phrase to highlight thi PR. Better words could be ``` correct PrincipalType``` in roleResolver

@ebarault Let me know if this PR makes any sense else we can close it down.

ebarault:
I think it is a step forward the solution, but i would need to dive in the code again to make sure the objective is met the proper way.
Would you like to try fixing the built-in $owner role all the way down? then I could assist in reviewing.

akki-ng:
@ebarault I would love to solve the issue completely. Can you help me with some of the use-cases of test case with example. And how exactly belongsTo relation configuration being changed to support polymorphic behaviour and is there any config which says 'isOwner' in belongsTo config? These info will help me get started.

Mr-Wiredancer:
@akki-ng Thanks for the fix. I have the same problem of having multiple extended models of `User`, and the app doesn't properly set `principalType`. But I don't understand why this cannot be merged.

akki-ng:
@Mr-Wiredancer Welcome :-) This fix will solve many use-cases. But I believe the real issue is a bigger one where a single entity may have many ```belongsto``` relationship with multiple ```extended user models``` with the support of multiple owners/co-owners. So probably maintainers decided to wait for the complete bug fix before merging this PR.

I was busy earlier so could not do a complete fix. I'll try to fix the real issue and submit a PR soon.

akki-ng:
@Mr-Wiredancer looks like the bigger problem is already solved. Have a look.
https://github.com/strongloop/loopback/issues/3644

lib/access-context.js
```diff
@@ -88,7 +88,9 @@ function AccessContext(context) {
   var token = this.accessToken || {};
 
   if (token.userId) {
-    this.addPrincipal(Principal.USER, token.userId);
+    var userPrincipalType =
+      (this.accessToken && this.accessToken.principalType) || Principal.USER;
+    this.addPrincipal(userPrincipalType, token.userId);
```
bajtos: @akki-ng 

This change looks very reasonable to me 👍 and I am ok with landing it as an incremental improvement.

However, I need you to add a test that fails on the current master, passes with your fix in place - this will verify your fix and prevent regressions in the future.

lib/access-context.js
```diff
@@ -88,7 +88,9 @@ function AccessContext(context) {
   var token = this.accessToken || {};
 
   if (token.userId) {
-    this.addPrincipal(Principal.USER, token.userId);
+    var userPrincipalType =
+      (this.accessToken && this.accessToken.principalType) || Principal.USER;
```
bajtos: I think this can be shortened by using `token` instead of `accessToken`?

Also please use `const` instead of `var` in new code - see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations

```ts
const userPrincipalType = token.principalType || Principal.USER;
```

akki-ng: I'll complete the patch soon.

akki-ng:
@MartinCerny-awin @bajtos @Mr-Wiredancer Guys, please follow up on new PR #3823 
New PR has complete patch and the test case for which will fail without the patch.

lib/access-context.js
akki-ng: @bajtos Please follow on #3823

bajtos:
See https://github.com/strongloop/loopback/pull/3883

Issue #3680, opened 2017-10-31
nestRemoting fails to pass optionsFromContext value to getters


bmatson:
# Description/Steps to reproduce

When the Model.nestRemoting function is used to "nest" a second model under the first, the calls to navigate the relation fail to follow the optionsFromContext passing pattern.  This causes the request context to be lost when navigating the nested relations.

The logic in lib/model.js looks for a function as the last argument, but needs to back up one more argument to retrieve and pass the option value to the relation getter as well. 

Pull request with fix on the way....

PR #3681, opened 2017-10-31
Update nestRemoting to pass optionsFromContext 

bmatson:
### Description

Passes context values through when navigating relationships in nested models.

#### Related issues

- connect to #3680

### Checklist

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -947,38 +947,45 @@ module.exports = function(registry) {
             }
           });
 
+          const lastArg = opts.accepts[opts.accepts.length - 1] || {};
+          const hasOptionsFromContext =
+            (lastArg.arg || lastArg.name) === 'options' &&
+            lastArg.type === 'object' && lastArg.http;
+
           if (relation.multiple) {
             sharedClass.defineMethod(methodName, opts, function(fkId) {
-              var args = Array.prototype.slice.call(arguments, 1);
-              var last = args[args.length - 1];
-              var cb = typeof last === 'function' ? last : null;
-              this[getterName](fkId, function(err, inst) {
-                if (err && cb) return cb(err);
+              const args = Array.prototype.slice.call(arguments, 1);
+              const cb = args[args.length - 1];
+              const contextOptions =
+                hasOptionsFromContext && args[args.length - 2] || {};
+              this[getterName](fkId, contextOptions, function(err, inst) {
+                if (err) return cb(err);
                 if (inst instanceof relation.modelTo) {
                   try {
                     nestedFn.apply(inst, args);
                   } catch (err) {
-                    if (cb) return cb(err);
+                    return cb(err);
                   }
-                } else if (cb) {
+                } else {
                   cb(err, null);
                 }
               });
             }, method.isStatic);
           } else {
             sharedClass.defineMethod(methodName, opts, function() {
-              var args = Array.prototype.slice.call(arguments);
-              var last = args[args.length - 1];
-              var cb = typeof last === 'function' ? last : null;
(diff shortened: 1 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

bajtos:
Hello @bmatson, thank you for the pull request. Please add tests to verify correctness of your patch and prevent regressions in the future.

bajtos:
@slnode test please

bmatson:
Ok, updated existing nestRemoting tests, added an observer that throws if it doesn't get a context, updated defined remote method to collect the context from the request.

bajtos:
@bmatson thank you for adding a test. I am afraid we will need to improve it, because right now when the options are not passed down then the tests fail with pretty cryptic messages:

```
1) relations - integration nested relations enables nested relationship routes - hasMany find:
     Error: expected 200 "OK", got 500 "Internal Server Error"

2) relations - integration nested relations enables nested relationship routes - hasMany findById:
     Error: expected 200 "OK", got 500 "Internal Server Error"

3) relations - integration nested relations should catch error if nested function throws:

      Uncaught AssertionError: expected 'missing options' to equal 'This should not crash the app'
      + expected - actual

      -missing options
      +This should not crash the app
```

I'll take a look at this myself to speed up the process of landing your patch.

bajtos:
I cleaned up the code in 58a2476:
    - improve the test to produce more helpful error messages
    - simplify the implementation to detect presence of options
      by checking "accepts" definition
    - remove unnecessary checks for callback presence - strong-remoting
      always provides a callback argument when invoking remote methods

@bmatson PTAL if you are ok with my proposed changes, in which case I'll go squash the commits and land the patch.

BTW your commit message says "Updated nestRemoting to pass optionsFromContext" which is not following our [50/72 style](http://loopback.io/doc/en/contrib/git-commit-messages.html) asking for imperative style. Nothing to worry about here, I'll fix the commit message as part of git history cleanup, but it would be nice if you could follow our conventions next time :)

bajtos:
@slnode ok to test

bmatson:
@bajtos The idea of checking for optionsFromContext is good (much better than just assuming that the if an object precedes the callback it must be the context, which is what I'm doing).  I think the check you're doing (checking the name === 'options') is more brittle than checking that the http key === 'optionsFromRequest'.  Your check will break if someone named the variable something other than 'options' (eg. 'opts').  What really gets the options there is the following

- http: 'optionsFromRequest'
- type: 'object' (the comparison for this needs to be case-insensitive per https://loopback.io/doc/en/lb3/LoopBack-types.html)

bajtos:
@bmatson thank you for a thoughtful comment, you are right!

> http: 'optionsFromRequest'

Unfortunately this is just a shortcut that's converted to a function before we get to read this value, see https://github.com/strongloop/loopback/blob/3bf84bacdec75253fdca64426eaaa948ceba616b/lib/model.js#L479-L491

> type: 'object' (the comparison for this needs to be case-insensitive per https://loopback.io/doc/en/lb3/LoopBack-types.html)

Good point :+1:

I am not sure how to proceed here. Should we perhaps simply assume the options object will be always there? This is based on the assumption that we are nesting only remote methods that we have created ourselves inside loopback.

Thoughts?

bmatson:
I think you could test if http === createOptionsViaModelMethod, but your point about this only being code that loopback put on the model itself is also valid.  What you have is probably ok given that caveat.

bmatson:
Works for me, thanks

Bill

bajtos:

Issue #2374, opened 2016-05-24
To replace `if(id)` behaviors across loopback


davidcheung:
convert `if(id)` behaviors to `id != null` 

as @shaharz suggested, when using if(id) if the `id` value is **0** it will cause the operation to fail
related https://github.com/strongloop/loopback/issues/2356
- [ ] [model.js](https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/lib/model.js#L163)
- [ ] [access-context](https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/lib/access-context.js#L70)

davidcheung:
related https://github.com/strongloop/loopback/issues/3130

PR #3693, opened 2017-11-15
fix(AccessContext): Tighten falsy userid/appid check

STRML:
### Description

An application may have a use for a falsy ID. Tightening the check from `(!id)` to `(id != null)` solves this problem.

This PR is intended for the 2.x branch.

See also #2374

### Checklist

### DIFF
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -61,16 +61,16 @@ function AccessContext(context) {
   var principalType = context.principalType || Principal.USER;
   var principalId = context.principalId || undefined;
   var principalName = context.principalName || undefined;
-  if (principalId) {
+  if (principalId != null) {
     this.addPrincipal(principalType, principalId, principalName);
   }
 
   var token = this.accessToken || {};
 
-  if (token.userId) {
+  if (token.userId != null) {
     this.addPrincipal(Principal.USER, token.userId);
   }
-  if (token.appId) {
+  if (token.appId != null) {
     this.addPrincipal(Principal.APPLICATION, token.appId);
   }
   this.remotingContext = context.remotingContext;
@@ -151,7 +151,7 @@ AccessContext.prototype.getAppId = function() {
  * @returns {boolean}
  */
 AccessContext.prototype.isAuthenticated = function() {
-  return !!(this.getUserId() || this.getAppId());
+  return this.getUserId() != null || this.getAppId() != null;
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


lib/access-context.js
```diff
@@ -151,7 +151,7 @@ AccessContext.prototype.getAppId = function() {
  * @returns {boolean}
  */
 AccessContext.prototype.isAuthenticated = function() {
-  return !!(this.getUserId() || this.getAppId());
+  return Boolean(this.getUserId() != null || this.getAppId() != null);
```
bajtos: Why is it necessary to cast this value to `Boolean`? I believe the inner condition will be always a boolean, since we are using `!=` operators,

bajtos:
@slnode test please

bajtos:
@STRML also please fix the commit message to follow our [50/72 style](http://loopback.io/doc/en/contrib/git-commit-messages.html). Right now, your fist line is 52 chars long.

lib/access-context.js
STRML: Yep, that was a bit of knee-jerk replacement of the `!!`.

STRML:
Yeah, the test is more or less copying the existing callback hell in other tests. Compared to them, it's consistent.

STRML:
Amended

bajtos:
cis-jenkins test are choking on "const" keyword:

```
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  SyntaxError: Unexpected token 'const'
  at /tmp/38d2ce2691b51812f68582f2b7a2100a.browserify:15139
PhantomJS 2.1.1 (Linux 0.0.0): Executed 0 of 0 ERROR (0.211 secs / 0 secs)
```

Strangely enough, Travis CI tests are passing. Since your patch does not contain any `const` and does not change dependencies, I'll assume these failures are unrelated to this pull request and ignore them.

bajtos:
Landed, thank you for the contribution! Would you like to open a new PR to forward-port this change to `master`?  Also according to #2374, there is another place ([model.js](https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/lib/model.js#L163)) where we are using `if(id)` instead of `if(id != null)`, would you like to fix it too?

STRML:
Sure, PR incoming for the rest of the changes in 2.x then I'll evaluate for master.

Issue #3130, opened 2017-01-23
Model relation call fails when id is 0


ExTheSea:
### Bug/Feature request

When creating a relation (tested with hasMany and belongsTo relations) between two models it will crash the application if you try to get the related items from a model with id = 0. These are possible due to them coming from a database where ids can be 0.

### Expected result
Imagine i have a modelA with a hasMany relation to modelB called "child". When i try to use the relation using: 
http://localhost:3000/api/modelA/1/child 
it will work but when i want to get the child elements of the modelA item with the id 0 like 
http://localhost:3000/api/modelA/0/child
it will cause a 500 error saying
```
Error: must specify id or data
at Function.ModelCtor.sharedCtor (<Workspace>\TestLoopback\loopback-test\node_modules\loopback\lib\model.js:185:12)
    at SharedMethod.invoke (<Workspace>\TestLoopback\loopback-test\node_modules\strong-remoting\lib\shared-method.js:262:25)
    at HttpContext.invoke (<Workspace>\TestLoopback\loopback-test\node_modules\strong-remoting\lib\http-context.js:295:12)
    at restPrototypeMethodHandler (<Workspace>\TestLoopback\loopback-test\node_modules\strong-remoting\lib\rest-adapter.js:411:9)
    at Layer.handle [as handle_request] (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\layer.js:95:5)
    at next (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\route.js:131:13)
    at Route.dispatch (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\route.js:112:3)
    at Layer.handle [as handle_request] (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\layer.js:95:5)
    at <Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\index.js:277:22
    at param (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\index.js:349:14)
    at param (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\index.js:365:14)
    at Function.process_params (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\index.js:410:3)
    at next (<Workspace>\TestLoopback\loopback-test\node_modules\express\lib\router\index.js:271:10)
[...]

Issue #2356, opened 2016-05-19
Token middleware user literal replacement fails when user ID is 0


shaharz:
Happened to run into this edge case. The culprit is this if statement, obvious in hindsight: https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/server/middleware/token.js#L24

Happy to send a PR if you need a hand.

P.S. I noticed another issue with ID == 0 where an include filter fails (doesn't show relations) but haven't had time to look into it yet.

davidcheung:
@shaharz I believe this is not a one-off scenario but a general loopback behavior. In order to keep behaviors consistent, I feel the support for `id==0` might be quite a large scope, if the decision is to support `Boolean(id) == false` I think it will need to be done across the board for all ID instances

/cc @strongloop/loopback-dev any thoughts?

For example: [model](https://github.com/strongloop/loopback/blob/70cec0755acc2b670987675c3bc63ecbfe9b473f/lib/model.js#L163), [access-token](https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/lib/access-context.js#L70)

jannyHou:
I agree with
> support for id==0 might be quite a large scope

And btw, now when we open `Explorer` and try `POST/ MyModels`, the default value is `id = 0`, this kinds of hints users that we support `id = 0`, which is not expected.

davidcheung:
Upon discussion with @raymondfeng, he agrees that we should fix it, would you mind submitting a patch? that would be great! 
I will create a separate issue attempting to track all the other places in LoopBack that has that behavior.

rmg:
**Heads up in case it hasn't occurred to anyone yet...**

Relational databases generally avoid using 0 as a primary key. MySQL requires changing the server mode, sqlite3 just doesn't allow it, in Postgres and SQL Server you have to override the default when you define the table/column. I'd be surprised if fixing this in loopback is the end of someone's problems if they are using multiple data sources _and_ have IDs that start at 0. Might be worth a mention in some docs somewhere in case people are used to working with NoSQL and don't realize this might bite them later.

shaharz:
PostgreSQL allows id 0 - granted, I happened to encounter the issue due to automated testing (which generated & inserted that id) so it's not something that would normally happen.

Deciding to not support id 0 sounds completely reasonable to me, however in that case I'd expect Loopback to assert() on that, and at the very least it should be mentioned in the docs.

[...]

PR #3719, opened 2017-12-05
fix(id): replace with != null

STRML:
### Description
Tighter null check on ids in 2.x.

#### Related issues

#2356, #2374, #3130, #3693

ping @bajtos

### DIFF
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -103,7 +103,7 @@ module.exports = function(AccessToken) {
 
     var id = tokenIdForRequest(req, options);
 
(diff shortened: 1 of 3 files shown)


bajtos:
This is great, thank you @STRML.

It makes me wonder, is there any easy way how to add unit tests to cover these changes? I am thinking about creating model instances with `id=0` and then check that the functions you are changing are still working as expected.

Thoughts?

bajtos:
Never mind, I am ok to land this pull request without tests, at the risk of regressions creeping it later.

PR #3720, opened 2017-12-05
Fix/falsy id 3.x

STRML:
### Description

Port of #3693 and #3719 into 3.x

Ping @bajtos - in particular, unsure how to get a 0 id set on User.create, appears new validation was added in 3.x.

Additionally it appears something is failing all PRs in your CI and we in the public lack visibility.

#### Related issues

#2356, #2374, #3130, #3693

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #3725, opened 2017-12-08
fix(test): rem exclusive test

STRML:
Ref: #3720


bajtos:
Landed, thank you very much!

Issue #3732, opened 2017-12-16
Data source names are not correctly assigned or retained


nitro404:
# Description/Steps to reproduce

When a data source is created either via the data sources configuration file or programmatically, the name attribute is not correctly retained and defaults to the name of the connector used instead. Additionally, the settings attribute does not exist for programmatically created data sources, so there is no way to obtain the original name (that I know of) of the data source, which is needed in some situations.

# Link to reproduction sandbox

https://github.com/nitro404/retain-datasource-name-sandbox

# Expected result

In the sandbox environment, the test boot script is expected to output the correct data source name for both the data source configuration file and programmatically created data sources, but it always defaults to the name of the connector. Additionally, the string case seems to vary between the two types, for whatever reason.

# Additional information

This appears to be a fairly substantial bug, but my proposed changes will break any code that makes use of this attribute in its current state. However, I don't believe it's a very commonly used attribute and technically seems to have the wrong value, so it should be addressed. I also have a fix and updated unit test here: https://github.com/nitro404/loopback/tree/hotfix/retain-datasource-name

nitro404:
Issue has been resolved. Thanks!

PR #3733, opened 2017-12-16
Fixed data sources not retaining the correct name value

nitro404:
### Description

Data source names are not correctly assigned or retained in the resulting object.

#### Related issues

- connect to #3732

### Checklist

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -505,21 +505,21 @@ function dataSourcesFromConfig(name, config, connectorRegistry, registry) {
     'can not create data source without config object');
 
   if (typeof config.connector === 'string') {
-    name = config.connector;
-    if (connectorRegistry[name]) {
-      config.connector = connectorRegistry[name];
+    const connectorName = config.connector;
+    if (connectorRegistry[connectorName]) {
+      config.connector = connectorRegistry[connectorName];
     } else {
-      connectorPath = path.join(__dirname, 'connectors', name + '.js');
+      connectorPath = path.join(__dirname, 'connectors', connectorName + '.js');
 
       if (fs.existsSync(connectorPath)) {
         config.connector = require(connectorPath);
       }
     }
     if (config.connector && typeof config.connector === 'object' && !config.connector.name)
-      config.connector.name = name;
+      config.connector.name = connectorName;
   }
 
-  return registry.createDataSource(config);
+  return registry.createDataSource(name, config);
 }
 
 function configureModel(ModelCtor, config, app) {
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -787,7 +787,7 @@ describe('app', function() {
     it('looks up the connector in `app.connectors`', function() {
       app.connector('custom', loopback.Memory);
       app.dataSource('custom', {connector: 'custom'});
-      expect(app.dataSources.custom.name).to.equal(loopback.Memory.name);
+      expect(app.dataSources.custom.name).to.equal('custom');
     });
 
     it('adds data source name to error messages', function() {


slnode:

slnode:

slnode:

slnode:

slnode:

nitro404:
Updated commit message.

bajtos:
Hello, thank you for the contribution and sorry for the long delay caused primarily by the Holiday Season.

I agree this a good change to make, I have been personally always annoyed that datasources had incorrect name. However, I am little bit worried about the consequences of landing this change as a non-major version - what is the probability of an application out there that is reading the datasource name and relying on the name to match the connector?

cc @raymondfeng @kjdelisle @jannyHou thoughts?

bajtos:
@slnode ok to test

lib/application.js
```diff
@@ -505,21 +505,21 @@ function dataSourcesFromConfig(name, config, connectorRegistry, registry) {
     'can not create data source without config object');
 
   if (typeof config.connector === 'string') {
-    name = config.connector;
-    if (connectorRegistry[name]) {
-      config.connector = connectorRegistry[name];
+    var connectorName = config.connector;
```
bajtos: Please use `const`, see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations

lib/application.js
```diff
@@ -505,21 +505,21 @@ function dataSourcesFromConfig(name, config, connectorRegistry, registry) {
...
         config.connector = require(connectorPath);
       }
     }
     if (config.connector && typeof config.connector === 'object' && !config.connector.name)
-      config.connector.name = name;
+      config.connector.name = connectorName;
   }
 
-  return registry.createDataSource(config);
+  return registry.createDataSource(name, config);
```
jannyHou: @nitro404 Thank you for the contribution.

```js
"myds": {
  "name": "temp",  // datasource name
  "connecor": "mysql",  // connector name
  "url": "localhost:3306/mydb"
}
```
Given the example above, I agree that `server.datasources.myds.name` implies the datasource name NOT the connector name.

Most of your change looks good to me 👍 , my only concern is Line 522, I think the first param here should be the connector name not datasource name:
```js
// IIUC, here
// `name` is datasource name
// `connectorName` is the connector name
return registry.createDataSource(connectorName, config);
```

It eventually instantiates `loopback-datasource-juggler/lib/datasource.js: DataSource(name, settings, modelConfig)`, my understanding is that function expects `name` to be a connector name.

Or you can omit it like the old code.
Does test fail if you keep it as `return registry.createDataSource( config);`?

nitro404:
Hello @bajtos! Not a problem. Thank you for taking the time to respond! I totally understand and agree your concern regarding this being a breaking change. I would suspect it would be fairly uncommon for anyone to be referring to a data source by its name, especially since it's the name of the connector anyways, but it's hard to say. The use case where this became a problem for myself is when I was programatically linking a model to a data source at runtime, which requires the data source name explicitly. This code was being used dynamically in a mixin, so the current workaround was to obtain the data source name from the settings attribute.. but this is not present if the data source was also programatically generated, so therein lies the issue.

@jannyHou You are welcome! Hmm I see your point, that is a fair concern. It appears the name is present in the config object, but **only** if it is created from a data source configuration file. I'm not sure if there's another way to ensure that the expected data source name is ultimately passed down the line to the registry.createDataSource function. If I remove the name from being passed as the first argument, the code will be reverted back to it's original state and thus the test will indeed fail. Perhaps the name could be appended to the config object and passed along and handled accordinly instead? This would require a change to the datasource juggler as well as far as I can tell, but I'm not sure. The name is simply discarded otherwise if it is not explicitly passed along in some form which is not good. I am open to suggestions if you have any!

bajtos:
Thank you @jannyHou for joining the discussion, your thoughts are valuable!

It seems to me this matter is more tricky than it looked at the first sight.

```js
"myds": {
  "name": "temp",  // datasource name
  "connector": "mysql",  // connector name
  "url": "localhost:3306/mydb"
}
```

I have two comments to the example above:

IMO "name" should not be part of datasource entry in `datasources.json`, I think it's a bug of loopback-workspace. However, that's completely out of scope of this pull request, so let's not worry about it now.

Since it is possible to have "name" in datasource config object, we should be prepared to handle it. Now the question is, what should take the precedence in the example above? Should the datasource be called "myds" or "temp"? 
 - IIRC, the registry of datasource maintained by loopback (see [lib/registry.js](https://github.com/strongloop/loopback/blob/0a4940e31ef99ffbae98f91812f631a7d9a524cc/lib/registry.js)) always uses the key ("myds"). 
 - When calling `app.model(MyModel, {dataSource: 'db'})`, we use the key ("myds") to resolve datasource string into a datasource instance ([code](https://github.com/strongloop/loopback/blob/0a4940e31ef99ffbae98f91812f631a7d9a524cc/lib/application.js#L532-L534))

In this light, I think it will be best to prefer the key ("myds") over the name config property ("temp"). To make it easier for our users to troubleshoot possible problems, it would be great if we could print a warning (`console.warn`) whenever a datasource has both name key and name config property set and these two values are not the same.

Thoughts?

lib/application.js
bajtos: The jsdoc comment for `Registry.prototype.createDataSource()` says that the first argument `name` is an "Optional name" (not very useful) and then expects `options.connector` to hold the connector name:

https://github.com/strongloop/loopback/blob/0a4940e31ef99ffbae98f91812f631a7d9a524cc/lib/registry.js#L349-L359

Down the line, both `name` and `options` are passed to DataSource constructor. AFAICT, the `name` is treated as the datasource name (not a connector name) in juggler, with the exception of this part, which should not be triggered when a DataSource is created from LoopBack (at least I think so):

https://github.com/strongloop/loopback-datasource-juggler/blob/00cf01f901e5ff20f3a576f10f41779e70cf536f/lib/datasource.js#L371-L376

If the connector is not provided, then it seems like the name is treated as a connector name, but again, I believe loopback's Registry always converts string connector name to a function/object before calling DataSource constructor.

https://github.com/strongloop/loopback-datasource-juggler/blob/00cf01f901e5ff20f3a576f10f41779e70cf536f/lib/datasource.js#L378-L393

nitro404:
Yeah, I kind of suspected as much but I'm not super familiar with all of the inner workings of Loopback unfortunately, haha. Agreed, I always thought it was rather strange that data sources had a name attribute and that it wasn't auto-populated using the key. I assumed that they were required to be equal, but it seems this is not the case. They are definitely populated with equal values, and in my opinion I think the key should always take precedence and the name should be ignored. My reasoning is that since you're defining the data sources as objects in the configuration file, you should in turn be able to access them using those same keys when retrieving them from the application data sources object at runtime. There could be a possibility for setting up the name attribute as an alias name of the data source, but that could get complicated and lead to other issues as well that would need proper handling. I agree with your suggestion for printing a warning if the name attribute is present and does not match, would certainly be the solution that makes the most sense until this is perhaps eventually removed entirely! We could also ensure that the name attribute is overridden with the value of the key as well, just to be safe unless we wanted to keep it as is for whatever reason.

bajtos:
@nitro404 cool, looks like we are on the same page then. Could you please add the warning we discussed above and fix my comment https://github.com/strongloop/loopback/pull/3733#discussion_r159203531? Other than that, I think the patch is good to be landed. (Please use `git commit --amend && git push -f`, it will save us time when merging.)

nitro404:
Awesome! Sure thing, not a problem. Commit should be updated accordingly.

bajtos:
@nitro404 thank you for the update. Could you please add the warning I mentioned earlier?

> To make it easier for our users to troubleshoot possible problems, it would be great if we could print a warning (console.warn) whenever a datasource has both name key and name config property set and these two values are not the same.

Also the commit message is not following our conventions, see http://loopback.io/doc/en/contrib/git-commit-messages.html. Could you please reword it?

bajtos:
> It looks like we could either check for this at the Loopback application level or the actual data source object itself, perhaps in the setup function. Would it perhaps make more sense to check it in the datasource object instead? Let me know what you would prefer, and I can set up a pull request on the datasource juggler instead if preferred. Thanks!

I like the idea of implementing the check in `DataSource` object in juggler, it makes a lot of sense. Let's do it!

I am going to fix the commit message here myself so that this PR can be landed quicker.

bajtos:
Landed, thank you for the contribution! Please mention my GH handle in the juggler pull request so that I can review it. I am not following new PRs in juggler repo.

Issue #3717, opened 2017-12-05
PersistedModel's updateAll method crashes the server when invoked via the remote connector


angfal:
# Description

I have a model with a remote datasource. When I tries to execute `updateAll` method my server crashes. 

The problem is loopback registers `updateAll` remote method with an object type, which is not supported by strong-remoting type registry.

```
setRemoting(PersistedModel, 'updateAll', {
   ...
   returns: {
     arg: 'info',
     description: 'Information related to the outcome of the operation',
     type: { // THE PROBLEM PLACE
       count: {
         type: 'number',
         description: 'The number of instances updated',
       },
     },
     root: true,
   },
   http: {verb: 'post', path: '/update'},
 });
```

# Reproduction

To reproduce the problem, just run this test in any loopback application. It fails on `updateAll` shared method.

```
describe('Type Registry Tests', () => {
  let Model;

  before('init a memory datasource', () => {
    if (!app.dataSources.db) {
      app.dataSource('db', {connector: 'memory'});
    }
  });

  beforeEach('create a model and attach it to the app', () => {
    Model = app.registry.createModel({
      name: specs.modelName()
    });
    app.model(Model, {dataSource: 'db'});
  });

  it('should be compatible with get converter method', () => {
    _.forEach(app.models[Model.modelName].sharedClass.methods(), (method) => {
      should(() => {
        _.forEach(method.returns, ({type}) => {
          
          // The type is {count: {type: 'number', description: '...'} 
          // for updateAll method, but it supports only string and array
          app.remotes()._typeRegistry.getConverter(type);
        });
      }).not.throw();
    });
  });
});
```

# Additional information

"loopback": "3.17.0",
"strong-remoting": "3.6.0",

jannyHou:
Hi @angfal I am trying to reproduce your error, are you running any endpoint from the Explorer(usually http://localhost:3000/explorer/)? If so can you post the endpoint name here? Thanks.

> I have a model with a remote datasource. When I tries to execute updateAll method my server crashes.

I am trying to figure out how do you execute the `updateAll` method.

angfal:
@jannyHou I have two loopback servers A and B. The server A contains a model with a remote connector to the server B. When a function from the server A executes 'Model.updateAll' request to the server B, I receive success result, but the server A crashes when it tries to process the response.

I can provide a sandbox example if it doesn't help

angfal:
The links to sandbox:
- https://github.com/angfal/remote-updateall-crash
or
[...]

PR #3718, opened 2017-12-05
Revert "Fix description of updateAll response"

angfal:
### Description
This reverts commit 1439446

#### Related issues
- connect to #3717

### Checklist
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #3741, opened 2017-12-27
Fix updateAll remote

jannyHou:
#### Related issues

connect to strongloop/loopback#3717

I manually tested the code, it fixes the bug in #3717, while I appreciate some opinion from `strong-remoting` expert before adding the test:

In this case IIUC we need the `info` object to have a property `count`, should we fix it by omitting the object details(like the current code change)? Or keep the detail as it is and create a function to resolve the type of an object?


jannyHou:
Closed as a dup of https://github.com/strongloop/loopback/pull/3718.

PR #3746, opened 2018-01-03
Update Copyright Years

JustinTRoss:
### Description
Update copyright years to include 2018

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Copyright IBM Corp. 2014,2018. All Rights Reserved.
 // Node module: loopback
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) IBM Corp. 2013,2017. All Rights Reserved.
+Copyright (c) IBM Corp. 2013,2018. All Rights Reserved.
 Node module: loopback
 This project is licensed under the MIT License, full text below.
 
--- a/common/models/access-token.js
+++ b/common/models/access-token.js
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Copyright IBM Corp. 2014,2018. All Rights Reserved.
 // Node module: loopback
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Copyright IBM Corp. 2014,2018. All Rights Reserved.
 // Node module: loopback
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
--- a/common/models/application.js
+++ b/common/models/application.js
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Copyright IBM Corp. 2014,2018. All Rights Reserved.
 // Node module: loopback
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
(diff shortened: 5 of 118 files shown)


slnode:

bajtos:
Thank you for the pull request, @JustinTRoss.

I think we should eventually update copyright years in the license headers in all source files (but not necessarily in this pull request). @rmg do you happen to have an automated script for that?

JustinTRoss:
@bajtos The following works on my end. I'll add the updated files to the PR.
``` LC_CTYPE=C && LANG=C && find . -not \( -path ./.git -prune \) -type f -exec sed -i '' s/Copyright\ IBM\ Corp.\ 2014,2016./Copyright\ IBM\ Corp.\ 2014,2018./ {} +```
If you would like, I can add the script in there as well. I don't see any obvious place to put it.

**Comment by JustinTRoss** [comment]
Note - On OSX, `sed` adds a newline char at end of files that are missing it. This is why the pngs and json files show changes.

**Comment by b-admike** [comment]
@JustinTRoss Thank you for your contribution. @shimks I also got 76 hits for JS files searching for `2016`, so perhaps there are a few more. Unfortunately, forked repositories are not searchable, so I was not to do the same on @JustinTRoss's branch :(.

**Comment by JustinTRoss** [comment]
@shimks You're right. I missed a few. There are a number of `2013,2016,` and `2015,2016`. If these different start dates are intentional, we can change the match to `,2016. All Rights Reserved.`, and that should do the trick.

**Comment by bajtos** [comment]
> There are a number of `2013,2016,` and `2015,2016`. If these different start dates are intentional, we can change the match to `,2016. All Rights Reserved.`, and that should do the trick.
  
Yes please, let's preserve the start dates.

While you are making these changes, would you mind to exclude those image files from the changes, perhaps by restricting the script to `*.js` files?

If you could post the final command here then I'll update the wiki page (I think it's editable only by people with the commit bit set).

**Comment by JustinTRoss** [comment]
Sure. No problem. I realized there were still some edge cases I was missing. There are a few copyrights terminating in 2015 and 2017, and there are some copyrights that started in certain years that need extended rather than overwritten. I'm updating to include those now.

**Comment by virkt25** [comment]
Not a blocker for this PR but just realizing we have some files with inconsistent header which are missed as a result. 

Example: https://github.com/strongloop/loopback/blob/master/server/middleware/url-not-found.js

```
// Copyright IBM Corp. 2014. All Rights Reserved.
```

**Comment by JustinTRoss** [comment]
Because of the two patterns involving copyrights starting in past year and copyrights extending through past year, there are two scripts. You could run them together if desired.
- Update copyrights starting in past year
```LC_CTYPE=C && LANG=C && find . -not \( -path ./.git -prune \) -not \( -path *.png -prune \) -not \( -path *.ico -prune \) -type f -exec sed -i '' s/\ 2017.\ All\ Rights\ Reserved./\ 2017,2018.\ All\ Rights\ Reserved./ {} +```
- Update copyrights extending through past year
```LC_CTYPE=C && LANG=C && find . -not \( -path ./.git -prune \) -not \( -path *.png -prune \) -not \( -path *.ico -prune \) -type f -exec sed -i '' s/,2017.\ All\ Rights\ Reserved./,2018.\ All\ Rights\ Reserved./ {} +```

**Comment by JustinTRoss** [comment]
@bajtos We can target file types instead of excluding if desired. We would just need to make sure to add something like \( -name \*.js -o -name \*.txt -o -name \*.md \).

**Comment by JustinTRoss** [comment]
There are still a couple errors. e.g. in lib/globalize.js. 1 sec.

**Comment by JustinTRoss** [comment]
I apologize for the back and forth here. This should be the last one. :)

**Comment by JustinTRoss** [comment]
Reverting those commits is breaking your commit header lint convention. Would you prefer I truncate the commit name we're reverting?

**Comment by raymondfeng** [comment]
@rmg IIRC, we have scripts to maintain the headers. Can you confirm?

**Comment by rmg** [comment]
Yes, there's a tool for updating our copyrights: `strong-tools` :-)

```
$ npm install -t strong-tools
$ SLT_OWNER="IBM Corp." slt copyright
$ SLT_OWNER="IBM Corp." slt license
$ git commit -a -m "chore: update copyright notice years"
```

Since the years are determined by git commit, and this is being run at the beginning of 2018 instead of the end of 2017, the above will actually have to be run a second time (probably followed by `git commit -a --amend` to avoid a second commit) so that the copyright headers take in to account the  start-of-year update and put 2018 in all the correct places.

The need for the `SLT_OWNER` env var can be removed by adding a `copyright.owner` property to `package.json`.

bajtos:
I run the tool pointed out by Ryan and added the changes to this pull request in 725b2eb. Now I am going to squash all commits into a single one before landing.

bajtos:
@slnode ok to test

bajtos:

Issue #3772, opened 2018-01-22
Critical crash when setting user email to undefined where case sensitive email is disabled


nitro404:
# Description/Steps to reproduce

If you disable case sensitive emails for an extended user model and attempt to set the value of the extended model's email to undefined, the user email setter attempts to convert an undefined value to lower case and invokes the extended user created callback twice.

# Link to reproduction sandbox

https://github.com/nitro404/loopback-case-insensitive-user-email-sandbox

# Expected result

API should emit an error for the member object not having an email address.

It should also not attempt to convert an email address that is not a string to lower case or invoke the member created callback twice.

# Additional information

Upon investigation it appears as though a simple type check is missing from the user model email setter when case sensitive email addresses are disabled.

Additionally, it seems as though the case insensitive email check in the access operator of the user model does not account for patterns used in the like operator when applying a where filter to a query, ie. ctx.query.where.email.like since the email address value is currently forced to lower case.

I suspect this should be a fairly simple fix, but I'm not sure what the preferred approach might be for both of these situations. If desired, I would be happy to submit a pull request if it is quicker.

bajtos:
@nitro404 thank you for reporting this issue. I am afraid we don't have much bandwidth right now to look into the problem. It would be great if you could go ahead and submit a pull request fixing the problem, we can discuss more details there. Thanks!

bajtos:
Fixed by #3779.

PR #3779, opened 2018-01-29
Fix bugs with case insensitive user emails

nitro404:
### Description

Fixes user email setter not accounting for the presence of undefined email addresses when attempting to convert them to lowercase when email case sensitivity is disabled.

Fixes like where query filter not accounting for case insensitive email addresses.

#### Related issues

- connect to #3772

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -1061,7 +1061,7 @@ module.exports = function(User) {
     this.settings.ttl = this.settings.ttl || DEFAULT_TTL;
 
     UserModel.setter.email = function(value) {
-      if (!UserModel.settings.caseSensitiveEmail) {
+      if (!UserModel.settings.caseSensitiveEmail && typeof value === 'string') {
         this.$email = value.toLowerCase();
       } else {
         this.$email = value;
--- a/test/user.test.js
+++ b/test/user.test.js
@@ -138,16 +138,29 @@ describe('User', function() {
       });
     });
 
-    it('Email is required', function(done) {
-      User.create({password: '123'}, function(err) {
-        assert(err);
-        assert.equal(err.name, 'ValidationError');
-        assert.equal(err.statusCode, 422);
-        assert.equal(err.details.context, User.modelName);
-        assert.deepEqual(err.details.codes.email, ['presence']);
+    it('fails when the required email is missing (case-sensitivity on)', () => {
+      User.create({password: '123'})
+        .then(
+          success => { throw new Error('create should have failed'); },
+          err => {
+            expect(err.name).to.equal('ValidationError');
+            expect(err.statusCode).to.equal(422);
+            expect(err.details.context).to.equal(User.modelName);
+            expect(err.details.codes.email).to.deep.equal(['presence']);
+          });
+    });
 
-        done();
-      });
+    it('fails when the required email is missing (case-sensitivity off)', () => {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: undefined, password: '123'})
+        .then(
+          success => { throw new Error('create should have failed'); },
+          err => {
+            expect(err.name).to.equal('ValidationError');
+            expect(err.statusCode).to.equal(422);
+            expect(err.details.context).to.equal(User.modelName);
+            expect(err.details.codes.email).to.deep.equal(['presence']);
+          });
     });
 
     // will change in future versions where password will be optional by default


slnode:

bajtos:
Thank you for the pull request, @nitro404. I'll try to review your changes by the end of this week.

nitro404:
Sounds good, thank you! I'm not sure 100% about my modifications to the like filter, as it might possibly break some patterns so it might be worth having an extra set of eyes look it over. There is also the options attribute which could possibly be used in place of modifying the pattern directly, but I'm not sure if the best approach or practices for handling this case.

common/models/user.js
```diff
@@ -1270,9 +1270,13 @@ module.exports = function(User) {
 
   // Access token to normalize email credentials
   User.observe('access', function normalizeEmailCase(ctx, next) {
-    if (!ctx.Model.settings.caseSensitiveEmail && ctx.query.where &&
-        ctx.query.where.email && typeof(ctx.query.where.email) === 'string') {
-      ctx.query.where.email = ctx.query.where.email.toLowerCase();
+    if (!ctx.Model.settings.caseSensitiveEmail && ctx.query.where) {
+      if (typeof ctx.query.where.email === 'string') {
+        ctx.query.where.email = ctx.query.where.email.toLowerCase();
+      } else if (ctx.query.where.email !== null && ctx.query.where.email instanceof Object &&
```
raymondfeng: There might be other operators in addition to `like`. Do you want to handle them all? For example, `{inq: ['x@Y.com', 'a@B.com']}`

nitro404: Ahh that's very true, I had not considered that initially. This definitely opens up a large can of worms though. I think it would be correct to handle all possible cases to prevent unintended behaviour as is currently exhibited. We would have to account for all possible combinations of operators through some sort of recursive function, and there's also the possibility of different types of patterns which as I previously mentioned should possibly not be modified due to case sensitivity of special regular expression operators or character sets. It seems like there's also different ways to specify patterns and regular expressions which would also need to be detected and accounted for. I have some ideas for going about doing this, but this is not a simple problem. Do you have any thoughts or suggestions for traversing and modifying a filter? I'm not sure if anything similar to this might already exist or not in the Loopback framework. This is something that would also need to be maintained as new operators are added.

bajtos:
@slnode ok to test

test/user.test.js
```diff
@@ -138,7 +138,7 @@ describe('User', function() {
       });
     });
 
-    it('Email is required', function(done) {
+    it('Email is required (email case-sensitivity on)', function(done) {
```
bajtos: While you are changing this line, please edit the test name to read as an English sentence.

```js
it('fails when the required email is missing (email case-sensitivity on)', function(done) {
  //...
});
```

test/user.test.js
```diff
@@ -150,6 +150,19 @@ describe('User', function() {
       });
     });
 
+    it('Email is required (email case-sensitivity off)', function(done) {
```
```js
it('fails when the required email is missing (email case-sensitivity off)', function(done) {
```

test/user.test.js
```diff
@@ -138,7 +138,7 @@ describe('User', function() {
       });
     });
 
-    it('Email is required', function(done) {
+    it('Email is required (email case-sensitivity on)', function(done) {
       User.create({password: '123'}, function(err) {
         assert(err);
```
```js
if (err) return done(err);
```

I know this is not exactly related to your changes, so feel free to ignore this comment if you prefer to.

test/user.test.js
```diff
@@ -150,6 +150,19 @@ describe('User', function() {
       });
     });
 
+    it('Email is required (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      User.create({email: undefined, password: '123'}, function(err) {
+        assert(err);
```
```js
if (err) return done(err);
```

Since this is newly added code, let's do it the right way please.

When `assert(err)` fails, it reports an assertion error (a stack trace) pointing to this place, hiding any information about the original error that triggered the test failure. With my proposal, the test fails because of the original error, therefore Mocha will print the stack trace pointing to the real source of the problem.

test/user.test.js
```diff
@@ -590,6 +603,40 @@ describe('User', function() {
         });
       });
     });
+
+    it('Should be able to use mixed-case pattern (email case-sensitivity on)', function(done) {
```
bajtos: We avoid using `should` in test names (at least in the new code), see http://loopback.io/doc/en/contrib/style-guide-es6.html#test-naming

```js
it('supports email searches with "like" (email case-sensitivity on)', function(done) {
  // ...
});
```

test/user.test.js
```diff
@@ -590,6 +603,40 @@ describe('User', function() {
...
+        }}, function(err, result) {
+          if (err) done(err);
+          assert(result[0], 'The query did not find the user');
+          assert.equal(result[0].email, email);
+          done();
+        });
+      });
+    });
+
+    it('Should be able to use mixed-case pattern (email case-sensitivity off)', function(done) {
```
```js
it('supports email searches with "like" (email case-sensitivity off)', function(done) {
```

test/user.test.js
```diff
@@ -590,6 +603,40 @@ describe('User', function() {
...
       });
     });
+
+    it('Should be able to use mixed-case pattern (email case-sensitivity on)', function(done) {
+      User.settings.caseSensitiveEmail = true;
+      var insensitiveUser = {email: validMixedCaseEmailCredentials.email, password: 'abc'};
+      var email = validMixedCaseEmailCredentials.email;
+      var pattern = email.substring(0, email.length - 2) + '.*';
+      User.create(insensitiveUser, function(err, user) {
+        User.find({where: {email:
```
bajtos: http://loopback.io/doc/en/contrib/style-guide-es6.html#one-argument-per-line

```js
User.find(
  {where: {email: {like: pattern}}},
  function(err, result) {
    // ...
  }));
```

test/user.test.js
```diff
@@ -590,6 +603,40 @@ describe('User', function() {
...
+      });
+    });
+
+    it('Should be able to use mixed-case pattern (email case-sensitivity off)', function(done) {
+      User.settings.caseSensitiveEmail = false;
+      var insensitiveUser = {email: validMixedCaseEmailCredentials.email, password: 'abc'};
+      var email = validMixedCaseEmailCredentials.email;
+      var pattern = email.substring(0, email.length - 2) + '.*';
+      User.create(insensitiveUser, function(err, user) {
+        User.find({where: {email:
```
bajtos: ditto - http://loopback.io/doc/en/contrib/style-guide-es6.html#one-argument-per-line

bajtos:
@nitro404 the example I pointed out actually contains a bug because it does not return the Promise back to Mocha - see https://github.com/strongloop/loopback/pull/3787

nitro404:
Of course! Always happy to help. That sounds like a plan!

I still do however have concerns regarding the usage of patterns as per my reply to Raymond's concerns where you could have a situation like:
`{ like: "Test\\D+com" }`
Where the regular expression is in string form, and `\\D` gets changed to `\\d` which is a totally different character set. I think in this case `i` should be added to the options string to ensure case insensitive comparison while maintaining the actual pattern / regular expression integrity. I presume that `options` is directly mapped to the regular expression flags argument when the comparator is determined to be a regular expression? It can also be just a substring as well, or a regular expression object with case insensitivity specified implicitly too. There's a bunch of different cases here that need to be accounted for which are currently broken with my change. Do you have any thoughts on this?

Also thank you for the feedback, I will adjust the unit tests as soon as I have some time! I was only basing them on the existing unit tests, haha.

bajtos:
@nitro404 ping, what's the status of this pull request?

nitro404:
Sorry @bajtos! I totally forgot that this pull request was open. Things got a bit chaotic a few months back and I lost track of this. Updated as per your prior request. I decided to remove the changes to the like filter due to the apparent complexity of solving such an issue. I'd rather see a full fix than something partial, so this is just a fix to properly handle errors where the email is present, but undefined. Let me know if you require any other changes, thanks!

bajtos:
Landed and published in `loopback@3.19.2` 🎉 

Thank you for the contribution ❤️

PR #3782, opened 2018-01-31
update juggler dep

virkt25:
`loopback-datasource-juggler` was updated to have a security fix. Marking that as the minimum required version for `loopback`.


virkt25:
@slnode test please

Issue #3788, opened 2018-02-05
Help Wanted: loopback crashes at any attempt at creating "BelongTo" "HasMany" relation


rodolph66:
I have worked with loopback (v3) a few months back I did not encounter this before.
Any attempt at creating a BelongTo - HasMany relation between two models end up in the app/loopback failing. 

To reproduce simply create any basic two models such as Category and Product and attempt to create the described relation. 
Note there is an egghead Loopback introduction course (very concise) which creates such models and the relation described. Creating this scenario always worked in the past. This time (after many attempts) it simply does not go through.

One difference noticed from before was the addition in the generator of the two questions: "Allow relation to be nested in API" and "Disable relation from being included"

The following shows the generator steps:
![generator](https://user-images.githubusercontent.com/9017114/35818108-9832171c-0ab8-11e8-850d-1af7c340e125.jpg)

in separate terminal (win 10) the following occurs on completing the second relation steps:
![error](https://user-images.githubusercontent.com/9017114/35818151-b3779678-0ab8-11e8-8875-69bbb5dd81e3.jpg)

I have tried in so many combination to bypass this error but nothing worked. Again, this never happened during mid last year.
Finally, I am working under windows 10 and did not actually try this under macOS or Linux though both systems are available for me.

Thank You

rodolph66:
After reviewing the older files, in which this type of relation worked seamlessly, I noted that the inclusion of the "options" property in the "relations" section of the .json file might have been the cause, since after removing the 

`      "options": {
        "nestRemoting": true,
        "disableInclude": true
      }`

The app worked again with the relation behaving as intended. For now I will continue without the options part though I read that the options was added to address some security issues.

jannyHou:
@rodolph66 could you post the `loopback-datasource-juggler` and `loopback` version here?
And also the last version that doesn't have this problem if you know it.

rodolph66:
Did not use loopback since Aug 2017 at which I had...
loopback ver as per package.json: 3.0.0
loopback-datasource-juggler ver: 3.9.3
Now... respectively is 3.0.0 and 3.15.0

jannyHou:
@kjdelisle could you help take a look of this interrupt issue? Thanks!

kjdelisle:
I can confirm that the issue is related solely to the `nestRemoting` option (using `disableInclude` exclusively doesn't cause any problems).

I'm going to start tracking down the source of the fault now.

kjdelisle:
[...]

PR #3789, opened 2018-02-06
model: fix infinite loop on nestRemoting

kjdelisle:
### Description

Prevent endless recursion on nestRemoting calls for
two-way model links.

#### Related issues

fixes https://github.com/strongloop/loopback/issues/3788
### Checklist

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -873,7 +873,10 @@ module.exports = function(registry) {
 
     var regExp = /^__([^_]+)__([^_]+)$/;
     var relation = this.relations[relationName];
-    if (relation && relation.modelTo && relation.modelTo.sharedClass) {
+    if (relation && relation._nestRemotingProcessed) {
+      return; // Prevent unwanted circular traversals!
+    } else if (relation && relation.modelTo && relation.modelTo.sharedClass) {
+      relation._nestRemotingProcessed = true;
       var self = this;
       var sharedClass = this.sharedClass;
       var sharedToClass = relation.modelTo.sharedClass;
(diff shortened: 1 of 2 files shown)


kjdelisle:
Still needs unit tests!

kjdelisle:
This is proving to be a lot more difficult to reproduce in testing than it is in reality; I think there's something about the current integration test structure that's circumventing this problem...

@bajtos @raymondfeng Any thoughts on what a good test for this scenario would look like?

lib/model.js
```diff
@@ -1025,6 +1027,7 @@ module.exports = function(registry) {
           }
         });
       });
+      relation.visited = true;
```
raymondfeng: I don't think it's a good idea to change the global state of `relation`. I would rather have an `visited models` array local to this function.

raymondfeng:
@kjdelisle A similar test as https://github.com/strongloop/loopback/blob/master/test/relations.integration.js#L1441 can be used.

lib/model.js
kjdelisle: I'm not sure I understand what you mean; if I make it local scope, how will it survive between subsequent calls?

Or did you mean that the Model should keep a static array of visited relations?

raymondfeng: `var relation = this.relations[relationName];` is the metadata for the model and its state should not be mutated for remoting calculation unless it's one time charge for the whole life of the model def.

kjdelisle: To be honest, I'm not sure how to deal with this alternatively; I'm not familiar with this area of the codebase at a conceptual level, so if the level of effort to make the fix is similar and you know the area in which to do it, you might be better off making the PR instead.

kjdelisle:
@slnode test please

kjdelisle:
@raymondfeng I'm still not sure about my name choice for the flag, but feel free to change it in this PR if you'd prefer a different version.

raymondfeng:
I would suggest `_nestRemotingProcessed` or `_remotingHandled`

PR #3822, opened 2018-03-07
change format to double for prevalued JavaScript value

dmastag:
Since the default value is set as a JavaScript Number so the result should be double as per Swagger Specification "Floating-point numbers with double precision."

https://swagger.io/docs/specification/data-models/data-types/

#### Related issues

### Checklist

### DIFF
--- a/common/models/access-token.json
+++ b/common/models/access-token.json
@@ -7,6 +7,7 @@
     },
     "ttl": {
       "type": "number",
+      "format": "double",
       "ttl": true,
       "default": 1209600,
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
Hello, thank you for the pull request. Based on our discussion in https://github.com/strongloop/loopback-swagger/pull/116, the current implementation is correct, since loopback-swagger is converting LoopBack type `number` to swagger spec that represents `double`.

PR #3823, opened 2018-03-08
Correct PrincipalType for multiple extension of User models

akki-ng:
#3679 
Patch is created along with the test case. Now access context will have the correct PrincipalType and ```getUser``` will work for all extensions of ```User``` model

### DIFF
--- a/common/models/role.js
+++ b/common/models/role.js
@@ -273,6 +273,10 @@ module.exports = function(Role) {
         process.nextTick(function() {
           callback(null, matches(modelId, userId));
         });
+      }else {
+        process.nextTick(function() {
+          callback(null, false);
+        });
       }
       return callback.promise;
     }
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -86,8 +86,10 @@ function AccessContext(context) {
 
   var token = this.accessToken || {};
 
-  if (token.userId != null) {
-    this.addPrincipal(Principal.USER, token.userId);
+  if (token.userId) {
+    var userPrincipalType =
+      (this.accessToken && this.accessToken.principalType) || Principal.USER;
+    this.addPrincipal(userPrincipalType, token.userId);
   }
   if (token.appId != null) {
     this.addPrincipal(Principal.APPLICATION, token.appId);
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -87,8 +87,7 @@ function AccessContext(context) {
   var token = this.accessToken || {};
 
   if (token.userId) {
-    var userPrincipalType =
-      (this.accessToken && this.accessToken.principalType) || Principal.USER;
+    const userPrincipalType = token.principalType || Principal.USER;
     this.addPrincipal(userPrincipalType, token.userId);
   }
   if (token.appId != null) {
--- a/test/multiple-user-principal-types.test.js
+++ b/test/multiple-user-principal-types.test.js
@@ -22,7 +22,7 @@ describe('Multiple users with custom principalType', function() {
 
   var commonCredentials = {email: 'foo@bar.com', password: 'bar'};
   var app, OneUser, AnotherUser, AccessToken, Role,
-    userFromOneModel, userFromAnotherModel, userRole, userOneBaseContext;
+    userFromOneModel, userFromAnotherModel, accessTokenForUserFromOneModel, accessTokenForUserFromAnotherModel, userRole, userOneBaseContext;
 
   beforeEach(function setupAppAndModels() {
     // create a local app object that does not share state with other tests
@@ -213,6 +213,50 @@ describe('Multiple users with custom principalType', function() {
     });
 
     describe('getUser()', function() {
+      it("Check correct principalType for users belonging to different user models", function() {
+        Promise.all([
+          OneUser.login(commonCredentials),
+          AnotherUser.login(commonCredentials)
+          ]).spread(function(t1, t2) {
+            
+            accessTokenForUserFromOneModel = t1;
+            accessTokenForUserFromAnotherModel = t2;
+
+
+            const accessContextForUserFromOneModel = new AccessContext({registry: OneUser.registry, accessToken: accessTokenForUserFromOneModel});
+            const accessContextForUserFromAnotherModel = new AccessContext({registry: AnotherUser.registry, accessToken: accessTokenForUserFromAnotherModel});
+
+            var user1 = accessContextForUserFromOneModel.getUser();
+            expect(user1).to.eql({
+              id: userFromOneModel.id,
+              principalType: OneUser.modelName,
+            });
+
+            var user2 = accessContextForUserFromAnotherModel.getUser();
+            expect(user2).to.eql({
+              id: userFromAnotherModel.id,
+              principalType: AnotherUser.modelName,
+            });
+          
+          })
+
+        
+        return Promise.try(function() {
+            addToAccessContext([
+              {type: Principal.ROLE},
+              {type: Principal.APP},
+              {type: Principal.SCOPE},
+              {type: OneUser.modelName, id: userFromOneModel.id},
+            ]);
+            var user = accessContext.getUser();
+            expect(user).to.eql({
+              id: userFromOneModel.id,
+              principalType: OneUser.modelName,
+            });
+          });
+
+      })
+
       it('returns user although principals contain non USER principals',
         function() {
           return Promise.try(function() {


slnode:

ebarault:
Hi @akki-ng , 
does this [test](https://github.com/strongloop/loopback/blob/master/test/multiple-user-principal-types.test.js#L485) pass as well ?

akki-ng:
@ebarault [test](https://github.com/strongloop/loopback/blob/master/test/multiple-user-principal-types.test.js#L485) will not pass by this fix as this fix does not cover this test case

bajtos:
@akki-ng thank you for the new version. I'll try to review your changes next week.

bajtos:
@slnode ok to test

lib/access-context.js
```diff
@@ -86,8 +86,9 @@ function AccessContext(context) {
 
   var token = this.accessToken || {};
 
-  if (token.userId != null) {
-    this.addPrincipal(Principal.USER, token.userId);
+  if (token.userId) {
+    const userPrincipalType = token.principalType || Principal.USER;
+    this.addPrincipal(userPrincipalType, token.userId);
```
bajtos: I find this change suspicions - I would expect that `token.principalType` was already added on line 84 above.

Having said that, this change does fix certain bugs like e.g. #3829, so it's not without merit. I need to build a better understanding of this part of our codebase to be able to comprehend impacts of this change in a wider context.

bajtos: Never mind, the line 84 was added back in 2013, before support for multiple user models were introduced. I suspect that code path was actually never used.

akki-ng: @bajtos You are right about line 84, but this block of code is neglecting the principal type associated with the accessToken,

https://github.com/strongloop/loopback/blob/6ddf268cb626521971f5e5b63382bf08af78285a/lib/access-context.js#L47-L85

I think we will need to correct the whole block for a holistic solution

akki-ng: My initial thought is, if line 84 was supposed to set the correct principal type then line 47 is dumping the empty object in some corner case.

bajtos:
Related work: https://github.com/strongloop/loopback/pull/3835

bajtos:
Hello @akki-ng, I am taking a look at your proposed changes. I started by trying to run your tests against the current master branch (see f7744270) and they are passing for me! Does it mean the problem is already fixed?

I suspect the change you are proposing in [common/models/role.js](https://github.com/strongloop/loopback/pull/3823/files#diff-733982793c3bd5e61d1a4350451fb494R276) may be still needed. How can I reproduce the bug which requires such fix?

akki-ng:
Hey @bajtos, the test cases were to pointout the problem of setting correct principalType in access context and allowing the user to be read correctly. If tests are passing then essentially we solved the problem of getting the user correctly from access context irrespective of their principalType.

While the changes in role.js tries to solve a different problem but only specific use case. For ex, if we have many extensions of user model and other entities have only one belongsTo to any of custom user model, in that case $owner role was not working in ACL of those entities. The patch in role.js will allow users to use $owner in acls of a multi user environment considering if there is only one belongsTo relationship of a particular entity to any of custom user model to which $owner acl is specified.
This patch does not try to solve a problem where an entity may have many belongsTo relationship with users, which can be treated as a separate use case.

akki-ng:
@bajtos to solve the use case of $owner not working in multi user environment, the patch wlon accessContext was mandatory. Now that accessContext patch is merged, you can reproduce the issue without the patch in role.js and confirm if applying the patch works properly.

akki-ng:
Closed by mistake, reopening now.

slnode:

bajtos:
@akki-ng thank you for the response. 

> The patch in role.js will allow users to use $owner in acls of a multi user environment considering if there is only one belongsTo relationship of a particular entity to any of custom user model to which $owner acl is specified.
This patch does not try to solve a problem where an entity may have many belongsTo relationship with users, which can be treated as a separate use case.

We need to capture this in a test. Could you please explain little bit more the relation between different models, ideally as an example showing specific model names and the relations between them? I am happy to help you with writing the unit test, but I don't have bandwidth to research how to setup models and relations to trigger the problem.

akki-ng:
@bajtos thanks for writing the unit tests. I will create a sample app to demonstrate the issue by next day.

bajtos:
>  I will create a sample app to demonstrate the issue by next day.

That will be awesome!

akki-ng:
@bajtos Please follow the below example to understand why the code will fail in a corner case.

For this example we have 2 custom user models (Seller and Customer), one Building model and one CustomAccessToken model to support polymorpic relations of token and users.

```
{
  "name": "Seller",
  "base": "User",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "shopName": {
      "type": "string"
    }
  },
  "validations": [],
  "relations": {
    "accessTokens": {
      "type": "hasMany",
      "model": "CustomAccessToken",
      "polymorphic": {
        "foreignKey": "userId",
      },
      "options": {
        "disableInclude": true
      }
    },
    "buildings": {
      "type": "hasMany",
      "model": "Building",
      "foreignKey": ""
    }
  },
  "acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "property": ["__create__buildings", "__get__buildings"],
      "principalType": "ROLE",
      "principalId": "$owner",
      "permission": "ALLOW"
    }
  ],
  "methods": {}
}
```

```
{
  "name": "Customer",
  "base": "User",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "mobile": {
      "type": "number"
    }
  },
  "validations": [],
  "relations": {
    "accessTokens": {
      "type": "hasMany",
      "model": "CustomAccessToken",
      "polymorphic": {
        "foreignKey": "userId",
      },
      "options": {
        "disableInclude": true
      }
    }
  },
  "acls": [],
  "methods": {}
}
```
```
{
  "name": "Building",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "name": {
      "type": "string",
      "required": true
    }
  },
  "validations": [],
  "relations": {
    "seller": {
      "type": "belongsTo",
      "model": "Seller",
      "foreignKey": ""
    }
  },
  "acls": [
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$everyone",
      "permission": "DENY"
    },
    {
      "accessType": "*",
      "principalType": "ROLE",
      "principalId": "$owner",
      "permission": "ALLOW"
    }
  ],
  "methods": {}
}
```
```
{
  "name": "CustomAccessToken",
  "base": "AccessToken",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {},
  "validations": [],
  "relations": {
    "user": {
      "type": "belongsTo",
      "idName": "id",
      "polymorphic": {
        "idType": "string",
        "foreignKey": "userId",
      }
    }
  },
  "acls": [],
  "methods": {}
}
```

Now follow the below steps in order to have the records ready
1) Create `seller1`, `seller2`, `customer1`
2) Keep the `AccessToken` ready for all 3 users
3) Login as `seller1` and create a `building` for `seller1`

Set token of `seller1` and try the following API
1. PATCH /api/Sellers/1
2. PATCH /api/Sellers/2 .  <---401 as expected
3. GET /api/Buildings/1

Now set token of `seller2` and try the following API
1. PATCH /api/Sellers/1 .  <--- 401 work as expected
2. PATCH /api/Sellers/2 
3. GET /api/Buildings/1 <--- 401 work as expected

Till now things worked as expected
Now login as `customer1` and try the following
1. GET /api/Buildings/1 <--- 401 work as expected
**2. PATCH /api/Sellers/1  <--- Request Hangs --->**
**3. PATCH /api/Sellers/2  <--- Request Hangs --->**

Now make the changes in `role.js`
```
// Is the modelClass User or a subclass of User?
    if (isUserClass(modelClass)) {
      var userModelName = modelClass.modelName;
      // matching ids is enough if principalType is USER or matches given user model name
      if (principalType === Principal.USER || principalType === userModelName) {
        console.log("Going in")
        process.nextTick(function() {
          callback(null, matches(modelId, userId));
        });
      }else {
        process.nextTick(function() {
          callback(null, false);
        });
      }
      return callback.promise;
    }
```

After making changes retry the API which were hanging when customer was loggedIn.

The bug is if many extended user models are used and they have `$owner` in their `ACL` then if a user from different user model tries to fire an api of another user model, leading the callback to never retrun `trur` or `false` from `Role.isOwner`.

I hope this example will clear things up. Let me know if you are not able to reproduce.

bajtos:
@akki-ng thank you for the instructions on how to reproduce the problem you are experiencing and sorry for the long delay! I was two weeks on a sick leave and my inbox exploded by the time I got back. I'll try to take a look at this in the next week or two.

akki-ng:
@bajtos Did you get the time to reproduce the issue?

common/models/role.js
```diff
@@ -273,6 +273,10 @@ module.exports = function(Role) {
         process.nextTick(function() {
           callback(null, matches(modelId, userId));
         });
+      }else {
+        process.nextTick(function() {
+          callback(null, false);
+        });
```
bajtos: This will not work in the following case: one user model owns another user mode, e.g. a Seller owns the business account of Customer.

bajtos:
Closing in favour of https://github.com/strongloop/loopback/pull/3883

PR #3835, opened 2018-03-20
Fix role check in apps with multiple user models

### LINKED ISSUE #3829: ACL seems to not work when having multiple user models with a base UserModel

I'm having trouble using multiple user models in my app, here is my test case.

I create a new loopback app, follow the doc to add multiple user models, with custom AccesToken model (https://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#access-control-with-multiple-user-models).
I have 5 models:
1) BaseUser : that inherit from User and have common functions (not provided in the sample but needed for a DRY code).
2) Admin : inherits from BaseUser
3) Client: inherits from BaseUser
4) CustomAccessToken: inherits from AccessToken
5) TaskItem: a persisted model with these ACL
    ```
    "acls": [
        {
          "accessType": "*",
          "principalType": "ROLE",
          "principalId": "$everyone",
          "permission": "DENY"
        },
        {
          "accessType": "*",
          "principalType": "Admin",
          "principalId": "$everyone",
          "permission": "ALLOW",
          "property": "*"
        }
    ```
I want to forbid everyone from accessing the model except Admins.
I tried reversing the order, change principalType to "ROLE" and principalId to "Admin", tried by changing the case (admin, ADMIN, Admin, etc...)

But whatever I do, when accessing the end point GET TaskItems as an Admin user I have a forbidden response.

Of course if I remove all the acls I have the results.

# Link to reproduction sandbox
My code is here : https://github.com/abouroubi/loopbacktest

# Expected result

Empty Array when Admin and Unauthorized when Client.

# Additional information

bajtos:
Fix the bug where the access context was created with incorrect principal type when the application has multiple user models configured, and custom roles were not honored as a result.

Close #3829

@akki-ng This patch is inspired by your #3823 but fixes a different and smaller problem. Could you please take a look at the proposed code changes too?

@ebarault If you can, then please review my changes too, they are related to you work on multi-user-model support.

### DIFF
--- a/lib/access-context.js
+++ b/lib/access-context.js
@@ -84,10 +84,10 @@ function AccessContext(context) {
     this.addPrincipal(principalType, principalId, principalName);
   }
 
-  var token = this.accessToken || {};
+  const token = this.accessToken;
 
   if (token.userId != null) {
-    this.addPrincipal(Principal.USER, token.userId);
+    this.addPrincipal(token.principalType || Principal.USER, token.userId);
   }
   if (token.appId != null) {
     this.addPrincipal(Principal.APPLICATION, token.appId);
--- a/test/multiple-user-principal-types.test.js
+++ b/test/multiple-user-principal-types.test.js
@@ -5,7 +5,6 @@
 
 'use strict';
 var expect = require('./helpers/expect');
-var request = require('supertest');
 var loopback = require('../');
 var ctx = require('../lib/access-context');
 var extend = require('util')._extend;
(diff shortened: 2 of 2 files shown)


akki-ng:
LGTM.

akki-ng:
@bajtos I did not get any notification for CLA. Can this be done now?

bajtos:
@akki-ng the commit was authored by myself only, that's why no CLA check was triggered for you. We will deal with CLA in the other pull request you have sent yourself. For reference, this is the URL where you can sign: https://cla.strongloop.com/agreements/strongloop/loopback

akki-ng:
@bajtos Cool. Thanks for the update.

PR #3865, opened 2018-04-21
fix a multiuser owership problem

vanceeasleaf:
### Description
if principalType is 'Patient'  and userModelName is 'Doctor' then callback is never called

### Checklist


slnode:

zbarbuto:
Hi there, thanks for the PR!

I’m not seeing how this particular change fixed an issue. Could you please add a test case or two that fail without your change and pass with it?

Thanks.

PR #3883, opened 2018-05-18
Fix isOwner() bug in multiple-principal setup

bajtos:
Fix the owner role resolver to correctly handle the case when a user from one model (e.g. Seller) is accessing an instance of another user-like model (e.g. Customer).

This PR supersedes #3823, #3679 and possible other olderPRs.

@akki-ng @MartinCerny-awin could you please give this a try and let us know if the patch fixes the problem for you?

### Checklist


bajtos:
Published in `loopback@3.19.1`, enjoy 🎉

akki-ng:
@bajtos Great! Good work guys.

PR #3890, opened 2018-05-25
Add options.currentUserLiteral to loopback#rest

wolrajhti:
### Description

Internally, loopback.rest uses loopback.token.

I think it would be usefull for developpers to be able to set the loopback#token currentUserLiteral of their REST API when they use loopback#rest, instead of load an other instance of loopback#token in the middlewares chain.

Maybe it would be better to pass all loopback#token options with something, like :
```javascript
loopback.rest({
    tokenOptions: {
        cookies: ['foo-auth'],
        headers: ['foo-auth', 'X-Foo-Auth'],
        params: ['foo-auth', 'foo_auth'],
        currentUserLiteral: 'me',
        ...
    }
});
```

or putting these options in remoting options, like :
```json
// in config.json
{
    "remoting": {
        "rest": {
            "tokenOptions": {
                "cookies": ["foo-auth"],
                "headers": ["foo-auth", "X-Foo-Auth"],
                "params": ["foo-auth", "foo_auth"],
                "currentUserLiteral": "me",
                "...."
            }
        }
    }
}
```

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -29,7 +29,7 @@ module.exports = rest;
  * ```
  * For more information, see [Exposing models over a REST API](http://loopback.io/doc/en/lb2/Exposing-models-over-REST.html).
  * @header loopback.rest([options])
- * @property {String} [currentUserLiteral] String literal for the current user.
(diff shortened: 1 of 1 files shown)


slnode:

PR #3891, opened 2018-05-25
Add options.currentUserLiteral to loopback#rest

wolrajhti:
### Description

Internally, loopback.rest uses loopback.token.

I think it would be usefull for developpers to be able to set the loopback#token currentUserLiteral of their REST API when they use loopback#rest, instead of load an other instance of loopback#token in the middlewares chain.

Maybe it would be better to pass all loopback#token options with something, like :
```javascript
loopback.rest({
    tokenOptions: {
        cookies: ['foo-auth'],
        headers: ['foo-auth', 'X-Foo-Auth'],
        params: ['foo-auth', 'foo_auth'],
        currentUserLiteral: 'me',
        ...
    }
});
```

or putting these options in remoting options, like :
```json
// in config.json
{
    "remoting": {
        "rest": {
            "tokenOptions": {
                "cookies": ["foo-auth"],
                "headers": ["foo-auth", "X-Foo-Auth"],
                "params": ["foo-auth", "foo_auth"],
                "currentUserLiteral": "me",
                "...."
            }
        }
    }
}
```

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -23,13 +23,18 @@ module.exports = rest;
  *
  * For example:
  * ```js
- * app.use(loopback.rest());
+ * app.use(loopback.rest({
+ *   currentUserLiteral: 'me'
+ * }));
  * ```
  * For more information, see [Exposing models over a REST API](http://loopback.io/doc/en/lb2/Exposing-models-over-REST.html).
- * @header loopback.rest()
+ * @header loopback.rest([options])
+ * @property {String} [currentUserLiteral] String literal for the current user.
  */
 
-function rest() {
+function rest(options) {
+  options = options || {};
+  var currentUserLiteral = options.currentUserLiteral;
   var handlers; // Cached handlers
 
   return function restApiHandler(req, res, next) {
@@ -50,7 +55,7 @@ function rest() {
 
       if (app.isAuthEnabled) {
         var AccessToken = registry.getModelByType('AccessToken');
-        handlers.push(loopback.token({model: AccessToken, app: app}));
+        handlers.push(loopback.token({currentUserLiteral: currentUserLiteral, model: AccessToken, app: app}));
       }
 
       handlers.push(function(req, res, next) {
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
In general, `loopback.rest()` includes `loopback.token()` just as a convenience for people building LoopBack applications in code (e.g. in our unit tests). We are intentionally keeping it simple. Users needed a more complex configuration should mount `loopback.token()` middleware explicitly.

> instead of load an other instance of loopback#token in the middlewares chain.

Could you please explain why do you see this as an issue? The token middleware is prepared to handle the case where it's invoked multiple time for the same request - it will resolve to no-op on subsequent invocations. See here: https://github.com/strongloop/loopback/blob/3e3092c01732d9567b8859a9befbc90f4977bce8/server/middleware/token.js#L126-L131

I am inclined towards closing this pull request as rejected, but let's discuss it first - it's entirely possible I'm missing a valid concern/use-case.

wolrajhti:
I was thinking maybe, when you have multiple middlewares in phase `routes:before`, or multiple `loopback.rest()` instances mounted on different pathes, it could be great to access the `loopback.token()` used inside your `loopback.rest()` to avoid messing with duplicates pathes.

But finally i'm agree with you, maybe it's better to keep it like that for Separation of Concern, the rest middleware doesn't have to parse the `currentUserLiteral`.

currently i use :
```javascript
/* in middleware.json */
  "routes:before": {
    "loopback#token": {
      "paths": [
        "${restApiRoot}"
       ],
      "params": {
        "currentUserLiteral": "current"
      }
    }
  },
  "routes": {
    "loopback#rest": {
      "paths": [
         "${restApiRoot}"
      ]
    }
  }
```

Do you think it's the best way for doing it ? In the doc, it's said to use the following **before boot** :
```javascript
app.use(loopback.token({
    /*...*/
}));
```
Is it wrong to declare `loopback#token` in `middleware.json` ? If possible we doesn't want to edit `server.js`.

[making-authenticated-requests-with-access-tokens](https://loopback.io/doc/en/lb3/Making-authenticated-requests.html#making-authenticated-requests-with-access-tokens)

wolrajhti:
Ok thx, i'll update the doc !

Finally I put loopback#token in the auth middleware phase.

PR #3896, opened 2018-05-28
fix bug in User.verify when confirm is disabled

wolrajhti:
### Description

In `User.verify`, Loopback is looking for the `.http.path` of the `User.confirm` remoteMethod, even if you provide `verifyOptions.verifyHref`.

Then, when confirm remoteMethod is disabled (eg : `sharedMethods: {"*":false}` in `config.json`), Loopback can't findMethodByName('confirm') and cause an error.

IMO, it should not be a problem that confirm is disabled if you provide the verificationHref option.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -736,22 +736,32 @@ module.exports = function(User) {
       (verifyOptions.protocol === 'https' && verifyOptions.port == '443')
     ) ? '' : ':' + verifyOptions.port;
 
-    var urlPath = joinUrlPath(
-      verifyOptions.restApiRoot,
-      userModel.http.path,
-      userModel.sharedClass.findMethodByName('confirm').http.path
-    );
+    if (!verifyOptions.verifyHref) {
+      const confirmMethod = userModel.sharedClass.findMethodByName('confirm');
+      if (!confirmMethod) {
+        throw new Error(
+          'Cannot build user verification URL, ' +
+            'the default confirm method is not public. ' +
+            'Please provide the URL in verifyOptions.verifyHref.');
+      }
 
-    verifyOptions.verifyHref = verifyOptions.verifyHref ||
-      verifyOptions.protocol +
-      '://' +
-      verifyOptions.host +
-      displayPort +
-      urlPath +
-      '?' + qs.stringify({
-        uid: '' + verifyOptions.user[pkName],
-        redirect: verifyOptions.redirect,
-      });
+      const urlPath = joinUrlPath(
+        verifyOptions.restApiRoot,
+        userModel.http.path,
+        confirmMethod.http.path
+      );
+
+      verifyOptions.verifyHref =
+        verifyOptions.protocol +
+        '://' +
+        verifyOptions.host +
+        displayPort +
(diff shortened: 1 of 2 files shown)


bajtos:
Hello @wolrajhti, thank you for the pull request. Please add a unit-test to verify your fix and prevent regressions in the future. Also make sure that `npm test` is passing for you locally, I see that Travis CI is failing: https://travis-ci.org/strongloop/loopback/builds/384716707?utm_source=github_status&utm_medium=notification

bajtos:
@slnode ok to test

wolrajhti:
Sorry but I don't understand the travis error, and I can't access the [cis-jenkins] logs (website timeout). I simply add a ternary instruction to prevent accessing undefined object property and the error seems to be :

> Running "mochaTest:unit" (mochaTest) task
> Mocha exploded!
> Error: Cannot find module 'base64-js'

When i run `npm test` locally nothing goes wrong :/

For the unit test, forgive me but it's me first attempt : I understand that each `it(...)` statement correspond to a unit test. This current PR concern the remoting context, set at `test/user.test.js:47`, which is the root of all other tests. Should I clone the function `setupAppAndModels`, edit the concerned line adding `sharedMethods: {'*':false}` in `app.set('remoting', ...`, and add a basic test at the end of the file ?

I firstly try to clone the block `describe('Identity verification', ...` and invoke it after my custom `setupAppAndModels`, but a lot of tests expect a result based on the `urlPath`...

bajtos:
I have cleaned up the code for you and force-pushed into your branch. Let's wait for CI results before landing.

wolrajhti:
Thx for the cleanup. Just for info this bug isn't happening when you use `User.disableRemoteMethodByName('confirm');`, only when you use `app.set('remoting', {sharedMethods: 'confirm':false});`.

bajtos:
> Just for info this bug isn't happening when you use `User.disableRemoteMethodByName('confirm');`, only when you use `app.set('remoting', {sharedMethods: 'confirm':false});`.

That's weird. The test using `User.disableRemoteMethodByName('confirm')` I wrote is failing on `master`, because `userModel.sharedClass.findMethodByName('confirm')` returns `undefined` in such case. That's the problem we need to fix, right?

Note that under the hood, both `User.disableRemoteMethodByName('confirm')` and `app.set('remoting', {sharedMethods: 'confirm':false})` call the same strong-remoting method `User.sharedClass.disableMethodByName('confirm')`.

Is there perhaps some other problem that's causing the issue you are observing?

I am proposing to land this patch as it is now (cc4fc21), because it's a good fix on its own. If you later discover more problems to fix, then we can address them in follow-up pull request(s). Any objections, @wolrajhti ?

wolrajhti:
> Note that under the hood, both User.disableRemoteMethodByName('confirm') and app.set('remoting', {sharedMethods: 'confirm':false}) call the same strong-remoting method User.sharedClass.disableMethodByName('confirm').

that's what a though, perhaps I made a mistake in my test... I have no objections to merge this PR as it is, thx for your help !

PR #3907, opened 2018-06-08
chore: update nodemailer version

dhmlau:
### Description
To continue discussion from https://github.com/strongloop/loopback/pull/3585, PR #3585 was closed because at that time we cannot drop Node 4.  Since now Node 4 is EOL, I'm wondering if it makes sense to revisit this `nodemailer` upgrade.

#### Related issues
#3585


bajtos:
@dhmlau I agree we can (and should) upgrade to a newer nodemailer version now. We need more than just bump dependency versions, see https://github.com/strongloop/loopback/pull/3585/files.

Let's close this in favour of #3585, so that we can also preserve authorship of the original contributor.

Issue #398, opened 2014-07-24
Logout fails for model based on User (other than model called "user")


johnsoftek:
If you create a model, say Appuser, based on User, when you call Appuser.logout, it fails with:

```
.../server/node_modules/loopback/lib/models/user.js:232
  this.relations.accessTokens.modelTo.findById(tokenId, function (err, accessT
                             ^
TypeError: Cannot read property 'modelTo' of undefined
```

transedward:
this issue still exist, the new user model based on user will not inherit the relation
add relation by hand can bypass this issue

Silvercast:
I'm suffering the same issue. I have `Appuser` model which has a `hasMany` relation to `accessTokens` model. And it doesn't work , giving error 500 when logout :(

raymondfeng:
@Silvercast Did you pass in `access_token` during login?

Silvercast:
Thanks. I managed to fix it by manually removing the `hasMany` relation in my custom `Appuser` model. Work as expected.
I was confused by comments above.

sn-satyendra:
Is this issue fixed. I am still getting the 500 error when I try to logout the user with access_token. I have a `appuser` which is extended from `user`. `appuser` has a `hasMany` relation to `accessToken`. 
@Silvercast I tried removing the `hasMany` relation in `appuser` but then I got the below issue:

TypeError: Cannot read property 'create' of undefined
    at ModelConstructor.User.createAccessToken (D:\zv\node_modules\loopback\common\models\user.js:100: 22)
    at D:\zv\node_modules\loopback\common\models\user.js:256: 22
    at D:\zv\node_modules\loopback\common\models\user.js:312: 9
    at D:\zv\node_modules\loopback\node_modules\bcryptjs\dist\bcrypt.js:261: 17
    at D:\zv\node_modules\loopback\node_modules\bcryptjs\dist\bcrypt.js:1198: 21
    at Immediate.next (D:\zv\node_modules\loopback\node_modules\bcryptjs\dist\bcrypt.js:1078: 21)
    at Immediate._onImmediate (D:\zv\node_modules\loopback\node_modules\continuation-local-storage\node_modules\async-listener\glue.js:188: 31)
    at processImmediate [as _immediateCallback] (timers.js:358: 17)

superkhau:
Are you guys still running into this issue? I'm closing it for now, ping and I will reopen.

mcitdev:
same here, 

```
[...]

PR #3917, opened 2018-06-16
Update the "before delete" hook in the user.js model

mcitdev:
### Description
Update the "before delete" hook to take into account the case where the user is not logged in and therefore does not have any accessToken

#### Related issues

- connect to #398

### Checklist


slnode:

common/models/user.js
```diff
@@ -346,6 +346,10 @@ module.exports = function(User) {
   };
 
   User.observe('before delete', function(ctx, next) {
+    
+    // Do nothing and just go to next when the user is not logged in
```
bajtos: > when the user is not logged in

This description feels misleading to me. How about this?

_Do nothing and when the access control was disabled for user model._

bajtos:
Few more things to fix:

- We need you to sign our Contributor License Agreement, see https://cla.strongloop.com/agreements/strongloop/loopback

- Your commit message is not following our [50/72 rule](https://loopback.io/doc/en/contrib/git-commit-messages.html).

    ```
    **    532ac66 - Update the "before delete" hook in the user.js mode: First line should be 50 characters or less (saw 53)
    **    532ac66 - Update the "before delete" hook in the user.js mode: line 3 longer than 72 characters.
    ```

    Please use `git commit --amend && git push --force-with-lease` to fix it. I can fix it for you before landing if you prefer.

-  Your code have linting problems, please make sure that `npm test` is passing for you locally. See https://travis-ci.org/strongloop/loopback/jobs/393096317#L2689:

    ```
    /home/travis/build/strongloop/loopback/common/models/user.js
      348:53  error  Block must not be padded by blank lines  padded-blocks
      349:1   error  Trailing spaces not allowed              no-trailing-spaces
      351:5   error  Expected space(s) after "if"             keyword-spacing
      352:1   error  Trailing spaces not allowed              no-trailing-spaces
    ```

common/models/user.js
bajtos: Uh oh, my proposed description is not even proper English. Sorry for the typo! Here is what I meant:

_Do nothing when the access control was disabled for this user model._

mcitdev: Ok, I fixed all the linting problem, and excuse my english because I'm not very english. Feel free to modify anything you want.

bajtos:
@mcitdev hello, I have rebased the patch on top of the latest master, remove `package-lock.json`, fixed the commit message and added a test to verify the fix (see 682df35).

There is one thing I don't understand though. How do you disable "user has many access tokens" relation in your application? Is the solution used by my new test similar to what you use?

```js
     // Simulate the situation when "accessTokens" relation was not configured
     delete PrivateUser.definition.settings.relations.accessTokens;
     delete PrivateUser.definition.relations.accessTokens;
```

mcitdev:
I didn't disable access tokens for the user, I do just not use ACL at all by default. This was the scenarion I've tested - user with no access control -

mcitdev:
@bajtos , Ok I just see your test, this is it, it describes what I'm trying to do

bajtos:
@mcitdev Hmm. Somehow I cannot find the test you are talking about. Did you perhaps forget to push it to GitHub? It's also possible I have accidentally discarded it from this branch.

Could you please check your local code, make sure to commit any outstanding changes (run `git status` to check) and then `git push -f` your code? Just don't try to pull updates from GitHub before pushing your local copy.

> I create the user with no ACL configuration, So I can perform actions on it without needing to login.

When you say "perform actions", do you mean to invoke REST API (make HTTP requests), or just call model methods via JavaScript API?

> That's why I do not have the access token, because I'm not logged in. So, I think this is not the same thing, because I didn't disabled the access control manually, but it's disabled by default due to not using the access control.

Yes, I agree my test is a different scenario than yours.

The trouble is, I am not able to reproduce your scenario. When you take my test and modify the lines removing `accessTokens` relation, the test passes against the current master branch - i.e. there is no bug to fix! 

> What I've wrote in my test describes what I'm doing actually in the test app.

As I wrote above, somehow I lost the test you are talking about. Could you please add it to the pull request again?

mcitdev:
@bajtos Ok, it was my fault, I reviewed the user test file, I totally agree with your test, I'm with you to use your test instead of mine.

bajtos:
> Ok, it was my fault, I reviewed the user test file, I totally agree with your test, I'm with you to use your test instead of mine.

OK.

However, I'd still like to understand how the way how you are configuring your app. Do you really delete `relations.accessTokens` entries in your app, as I am doing in my test? Is there perhaps something else going on that's causing the problem?

It looks like you are not the only one running into this problem, #3921 opened by @ryanxwelch found a different bug that manifests when a user does not have any "accessTokens" relation.

It's a mystery to me which I'd like to fully understand before trying to fix the symptoms.

mcitdev:
> However, I'd still like to understand how the way how you are configuring your app. Do you really delete relations.accessTokens entries in your app, as I am doing in my test? Is there perhaps something else going on that's causing the problem?

I do not delete the accessTokens from relations, it's by default empty when you create a custom user based on the User model. But what seems odd to me, is that **"accessTokens"** relation should be inhereted from the User, thing that seems  not to happen here.
I create a repo that reproduces the error, I just follow the cli commands to generate the custom user: [https://github.com/mcitdev/CustomLoopbackUser](url)

> It looks like you are not the only one running into this problem, #3921 opened by @ryanxwelch found a different bug that manifests when a user does not have any "accessTokens" relation.

It may have a relation with the problem mentionned by @ryanxwelch in #3921 , in that if the user has no "accessTokens" relation configured, but that should be confirmed by @ryanxwelch with the scenario that causes the issue.

mcitdev:
I rectified my last comment, meanwhile I reviewed the doc, I still learning the framework, because I'm new on it.
So these are some details about my use case, because I think it is deferent from what @ryanxwelch mentionned in #3921:

- First, I create a new app based on the “empty-server” application type. So initially the access control is disabled in this case according to the doc [https://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#enabling-access-control](url) : 
> Applications created with the LoopBack application generator have access control enabled by default, except for the “empty-server” application type. To enable access control for an “empty-server” application, add a boot script that calls enableAuth()

That's why I do not get the warning message of the lib/application.js because the enableAuth function is not called in my case.

- Then I create a custom user based on the User model, wich is just inherit methods and properties from the base model according to the doc again [https://loopback.io/doc/en/lb3/Extending-built-in-models.html#extending-models-using-json](url) : 
> When you create a model with the model generator, you choose a base model, that is, the model that your model will “extend” and from which it will inherit methods and properties.

So Initially the custom user has no relation configured by default, thus the "accessTokens" does not exists yet. That's why I added the lines that handles this case in the "before delete" hook in the user.js model to prevent craching the app with the error : 

```
TypeError: Cannot read property 'modelTo' of undefined
    at node_modules/loopback/common/models/user.js:350:5
    ... .
```
As I said before, maybe this situation is not realistic, but it preserves the consistency of the use of ACL in the application.

mcitdev:
@bajtos I've modified the test a little bit, because the last version does not fail even if I remove the line added in the "before delete" hook.

This works for me : 

```
it('skips token invalidation when the relation is not configured', (done) => {
      // the test passed when the operation did not crash
      const app = loopback({localRegistry: true, loadBuiltinModels: true});
      app.set('remoting', {errorHandler: {debug: true, log: false}});

      const PrivateUser = app.registry.createModel({
        name: 'PrivateUser',
        base: 'User',
        // Speed up the password hashing algorithm for tests
        saltWorkFactor: 4,
      });
      app.model(PrivateUser, {dataSource: 'db'});

      var usersId;
      async.series([
        function(next) {
          PrivateUser.create({email: 'private@example.com', password: 'pass'}, function(err, user) {
            usersId = user.id;
            next(err);
          });
        },
        function(next) {
          PrivateUser.deleteById(usersId, function(err) {
            next(err);
          });
        },
      ], function(err) {
        if (err) return done(err);
        done();
      });
    });
```
As I said before, I do not need to delete the "accessTokens" relation, because the **_"PrivateUser.definition.settings.relations.accessTokens"_** from the base model will not be used, so it's unnessacary to delete that, instead it will use the child relations wich is by default empty **_(it can be checked by a simple console.log which will print "undefined" for "PrivateUser.definition.relations.accessTokens")_**

package.json
```diff
@@ -95,7 +95,7 @@
     "loopback-boot": "^2.7.0",
     "loopback-context": "^1.0.0",
     "mocha": "^3.0.0",
-    "nyc": "^10.1.2",
+    "nyc": "^10.3.2",
```
bajtos: Let's revert this change, it's not related to the fix.

test/user.test.js
```diff
@@ -299,6 +299,39 @@ describe('User', function() {
...
+      const PrivateUser = app.registry.createModel({
+        name: 'PrivateUser',
+        base: 'User',
+        // Speed up the password hashing algorithm for tests
+        saltWorkFactor: 4,
+      });
+      app.model(PrivateUser, {dataSource: 'db'});
+
+      var usersId;
+      async.series([
```
bajtos: Let's rewrite this block using promises.

test/user.test.js
```diff
@@ -299,6 +299,39 @@ describe('User', function() {
       });
     });
 
+    it('skips token invalidation when the relation is not configured', (done) => {
+      // the test passed when the operation did not crash
+      const app = loopback({localRegistry: true, loadBuiltinModels: true});
+      app.set('remoting', {errorHandler: {debug: true, log: false}});
```
bajtos: We don't need to change `remoting` config because we are not making any REST calls.

bajtos:
> As I said before, I do not need to delete the "accessTokens" relation, because the "PrivateUser.definition.settings.relations.accessTokens" from the base model will not be used, so it's unnessacary to delete that, instead it will use the child relations wich is by default empty (it can be checked by a simple console.log which will print "undefined" for "PrivateUser.definition.relations.accessTokens")

Interesting. I though I tried to create a new app for this new test without any luck. I confirm that your updated test works like a charm!

bajtos:
@slnode ok to test

bajtos:
Landed 🎉 Thank you for contributing the fix and helping me to better understand the issue at hand, @mcitdev ❤️

mcitdev:
@bajtos I'm glad to contribute to loopback, it was my pleasure.

PR #3934, opened 2018-06-29
fix: resolve customAccessToken warning error that causes app to crash

bajtos:
In the present implementation `verifyUserRelations(Model)` assumes that
`Model.relations.accessTokens` is always set, and as a result may
crash when trying to access `polymorphic` property of that relation.

It seems the intention is to check whether the the following conditions are met:
 1. a model has a hasMany accessTokens relation
 2. that relation is polymorphic

This commit fixes the problem by accounting for the case where the
accessTokens relation was not correctly set up.

This pull request supersedes https://github.com/strongloop/loopback/pull/3921, which was opened against a wrong branch (`3.x-latest`).

/cc @ryanxwelch

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -497,8 +497,10 @@ app._verifyAuthModelRelations = function() {
   function verifyUserRelations(Model) {
     const hasManyTokens = Model.relations && Model.relations.accessTokens;
 
+    const relationsConfig = Model.settings.relations || {};
+    const hasPolyMorphicTokens = (relationsConfig.accessTokens || {}).polymorphic;
     // display a temp warning message for users using multiple users config
-    if (hasManyTokens.polymorphic) {
+    if (hasPolyMorphicTokens) {
       console.warn(
         'The app configuration follows the multiple user models setup ' +
           'as described in http://ibm.biz/setup-loopback-auth',
@@ -508,7 +510,6 @@ app._verifyAuthModelRelations = function() {
 
     if (hasManyTokens) return;
 
-    const relationsConfig = Model.settings.relations || {};
     const accessTokenName = (relationsConfig.accessTokens || {}).model;
     if (accessTokenName) {
       console.warn(


bajtos:
Landed. Thank you @ryanxwelch for the contribution!

mcitdev:
Hi @bajtos, I think that the current fix of this issue only works because the parent and child settings merge strategy used by default is deep merge so the Model.settings.relations in the fix may not have the polymorphic relation if another merge strategy is used. But what is certain is that the polymorphic relation comes from the child model, so it makes more sense to test the relations of the child. **_(more info about merge strategies in mergeSettings function of loopback-datasource-juggler/lib/model-builder.js)_**

I think what should be used in this fix is ​​what @ryanwelch suggested in commit 778395e6d5ad0626dbc507e913c3bc834d93478c, I will just modify it a little bit and suggest this : 

```
const hasManyTokens = Model.relations && Model.relations.accessTokens;

if (hasManyTokens && hasManyTokens.polymorphic) {
```

instead of this : 

```
const hasManyTokens = Model.relations && Model.relations.accessTokens;

if (hasManyTokens && Model.settings.relations.accessTokens.polymorphic) {
```

I think that is the only thing that should be changed in the function to fix the issue.

mcitdev:
Hi @bajtos, I think that the current fix works because of the merge of the settings of the child and the parent, _more about merge settings in_ [http://apidocs.loopback.io/loopback-datasource-juggler/#modelbuilder](url) .

IMHO, the old version of the function was more accurate, because what is certain is that the polymorphic relation comes from the child model (the polymorphic relation will be configured in the child model), so it makes more sense to test the relations of the child.

I think what should be used in this fix is ​​:

```
function verifyUserRelations(Model) {
    const hasManyTokens = Model.relations && Model.relations.accessTokens;

    // display a temp warning message for users using multiple users config
    if (hasManyTokens && hasManyTokens.polymorphic) { //  <---- the only change that should be done
      console.warn(
        'The app configuration follows the multiple user models setup ' +
          'as described in http://ibm.biz/setup-loopback-auth',
        'The built-in role resolver $owner is not currently compatible ' +
          'with this configuration and should not be used in production.');
    }
```

I think that is the only thing that should be changed in the old verifyUserRelations function to fix the issue.

bajtos:
@mcitdev thank you for investigating this problem further. Could you please open a pull request making the change you are suggesting?

PR #3950, opened 2018-07-20
Issue with couchbase.js login

FerreroFacundo:
### Description

found this bug, when trying to login to couchbase data source, the data is not mapped inside "this", it's in "this.default".

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -369,18 +369,20 @@ module.exports = function(User) {
    * @promise
    */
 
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
Hell @FerreroFacundo, thank you for the pull request. From your description, this looks like a bug the Couchbase connector to me. LoopBack models don't support data to be stored in a `default` property. Please open an issue in the repository of the connector you are using.

PR #3954, opened 2018-07-22
Fix ACL rules with a wildcard match ('*') for model property

### LINKED ISSUE #3953: Adding ACL rule with a wildcard match ('*') for model property doesn't work
The ["Controlling data access" documentation page](https://loopback.io/doc/en/lb3/Controlling-data-access.html#acl-rule-precedence) explains how to apply a rule for all models, but adding this king of a rule doesn't work. Exploring the ACL.checkAccessForContext function, which in charge of checking if the relevant request has the permission to access, shows that there is no lookup for '*' and only for the specific model name.

Create an ACL rule on one of the boot scripts:
```javascript
module.exports = function(app) {
  app.models.ACL.create({
      model: '*',
      property: '*',
      accessType: '*',
      principalType: 'ROLE',
      principalId: '$everyone',
      permission: 'DENY',
    });
}
```

# Link to reproduction sandbox

# Expected result
Accessing to all models (which not includes other ACL rules) would be denied. 

# Additional information

moshemal:
fixes #3953

### DIFF
--- a/common/models/acl.js
+++ b/common/models/acl.js
@@ -478,8 +478,15 @@ module.exports = function(ACL) {
     var effectiveACLs = [];
     var staticACLs = self.getStaticACLs(model.modelName, property);
 
-    this.find({where: {model: model.modelName, property: propertyQuery,
-      accessType: accessTypeQuery}}, function(err, acls) {
+    const query = {
+      where: {
+        model: {inq: [model.modelName, ACL.ALL]},
+        property: propertyQuery,
+        accessType: accessTypeQuery,
+      },
+    };
+
+    this.find(query, function(err, acls) {
       if (err) return callback(err);
       var inRoleTasks = [];
 
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -15,16 +15,13 @@ var supertest = require('supertest');
 var Role = loopback.Role;
 var RoleMapping = loopback.RoleMapping;
 var User = loopback.User;
-var testModel;
 var async = require('async');
 
 // Speed up the password hashing algorithm for tests
 User.settings.saltWorkFactor = 4;
 
 var ds = null;
-before(function() {
-  ds = loopback.createDataSource({connector: loopback.Memory});
-});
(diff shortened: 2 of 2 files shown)


slnode:

bajtos:
Hello @moshemal, thank you for the pull request. Could you please add a unit test to verify your fix and sign our CLA here? https://cla.strongloop.com/agreements/strongloop/loopback

I am speaking at a conference this week, will try to review your pull request early next week.

moshemal:
@bajtos It seems that the assertions in the callback functions doesn't effect test results. For example, if I add assert(false) at line 239 the test still pass. 
As Mocha documentation explains one should use the method done()
https://mochajs.org/#asynchronous-code 

Do I miss something?

bajtos:
> It seems that the assertions in the callback functions doesn't effect test results. For example, if I add assert(false) at line 239 the test still pass.
As Mocha documentation explains one should use the method done()
https://mochajs.org/#asynchronous-code
>
> Do I miss something?

Good point! Our test suite for ACLs was written back in time when we don't fully understand how to correctly write async tests; and we did not have time to fix it since then. I think the test change you have described should still trigger a test failure if the test suite is run long enough for all async operations to complete; unfortunately that assertion error would manifest as being caused by a different test. Not pretty :(

The best solution is to split that test into multiple smaller one, so that each test has only one assertion. The second best solution is to use `async.parallel` to wait until both calls of `ACL.checkPermission` finish, and only then call `done()`.

Feel free to ignore the existing test cases in `test/acl.test.js` and write a new one in the style that makes most sense to you. For example, I think it should be possible to use Promise-style async control instead of the older callback (`done`) based.

Please let me know if you need any help with the tests

moshemal:
I fixed the tests in PR #3989
will update this PR with new tests

moshemal:
Yes

moshemal:
@bajtos test added

bajtos:
@moshemal landed, thank you for the contribution ❤️

PR #3965, opened 2018-08-08
fix: accessToken create default acl

virkt25:
### Description

Update the default ACL for the AccessToken model so we don't allow create.

#### Related issues

### Checklist


virkt25:
> Hmm, any security concern of allowing everyone to create the accesstoken?

@jannyHou Yes, that's why we're removing it. If the `AccessToken` API is exposed via REST, everyone can make an accessToken for any user and use it for accessing data / modifying data.

PR #3971, opened 2018-08-20
[GVT] Get the summary and description in the `lb export-api-def` output translatable

dhmlau:
### Description
In the original issue (https://github.com/strongloop/generator-loopback/issues/371), all the output from `lb export-api-def` are not translatable.   This PR will make the `summary` and `description` field translatable.

Output after fix:
screen shot 2018-08-20 at 12 59 43 pm

#### Related issues
 https://github.com/strongloop/generator-loopback/issues/371


lib/model.js
```diff
@@ -572,7 +572,7 @@ module.exports = function(registry) {
         {arg: 'refresh', type: 'boolean', http: {source: 'query'}},
         {arg: 'options', type: 'object', http: 'optionsFromRequest'},
       ],
-      description: format('Fetches hasOne relation %s.', relationName),
+      description: g.f('Fetches hasOne relation %s.', relationName),
```
jannyHou: I think `format` should be preserved?

dhmlau: `format` is to format the string to take the parameters, which `g.f` can do. 
Here is an example: 
https://github.com/strongloop/generator-loopback/blob/master/oracle/index.js#L78

dhmlau:
@jannyHou , I removed the one remaining `format` call.  Thanks for catching that!

Issue #3988, opened 2018-09-04
The assertions in the callback functions of test/acl.test.js doesn't affect test results 


moshemal:
It seems that the assertions in the callback functions doesn't effect test results. For example, if I add assert(false) at line 239 the test still pass.
As Mocha documentation explains one should use the method done()
https://mochajs.org/#asynchronous-code

# Description/Steps to reproduce
Add assert(false) to one of the callbacks, for example line 239 and run the tests. The tests are expected to fail on that assertion but they don't

# Expected result
fail

dhmlau:
Moving it to `In Progress` as there is PR #3989 to fix the issue.

dhmlau:
PR #3989 has been merged.  Closing it as resolved.
@moshemal , thanks for your contribution! 🎉

PR #3989, opened 2018-09-04
Use done function for multiple async assertions

moshemal:
### Description
fix acl.test.js tests 

#### Related issues
#3988

### DIFF
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -16,14 +16,10 @@ var Role = loopback.Role;
 var RoleMapping = loopback.RoleMapping;
 var User = loopback.User;
 var testModel;
+var async = require('async');
 
 // Speed up the password hashing algorithm for tests
 User.settings.saltWorkFactor = 4;
-
-function checkResult(err, result) {
-  // console.log(err, result);
-  assert(!err);
-}
 
 var ds = null;
 before(function() {
@@ -48,7 +44,7 @@ describe('security scopes', function() {
     testModel.attachTo(ds);
   });
 
-  it('should allow access to models for the given scope by wildcard', function() {
+  it('should allow access to models for the given scope by wildcard', function(done) {
     Scope.create({name: 'userScope', description: 'access user information'},
       function(err, scope) {
         ACL.create({
@@ -56,14 +52,19 @@ describe('security scopes', function() {
           model: 'User', property: ACL.ALL,
           accessType: ACL.ALL, permission: ACL.ALLOW,
         }, function(err, resource) {
-          Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.READ, checkResult);
+          async.parallel([
+            cb => Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, cb),
+            cb => Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, cb),
+            cb => Scope.checkPermission('userScope', 'User', 'name', ACL.READ, cb),
+          ], (err) => {
+            assert.ifError(err);
+            done();
+          });
         });
       });
   });
 
-  it('should allow access to models for the given scope', function() {
+  it('should allow access to models for the given scope', function(done) {
     Scope.create({name: 'testModelScope', description: 'access testModel information'},
       function(err, scope) {
         ACL.create({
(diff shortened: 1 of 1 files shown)


slnode:

test/acl.test.js
```diff
@@ -48,22 +44,28 @@ describe('security scopes', function() {
...
         ACL.create({
           principalType: ACL.SCOPE, principalId: scope.id,
           model: 'User', property: ACL.ALL,
           accessType: ACL.ALL, permission: ACL.ALLOW,
         }, function(err, resource) {
-          Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.READ, checkResult);
+          async.parallel([
+            clbk => Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, clbk),
```
bajtos: I find `clbk` as rather cryptic. Could you please use `cb`, `callback` or `next` instead?

test/acl.test.js
```diff
@@ -48,22 +44,28 @@ describe('security scopes', function() {
...
+            clbk => Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, clbk),
+            clbk => Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, clbk),
+            clbk => Scope.checkPermission('userScope', 'User', 'name', ACL.READ, clbk),
+          ], (err) => {
+            assert(!err);
```
bajtos: While you are changing this part: let's improve the assertion to produce a more helpful error message.

`assert(!err)` say something like `expected false to be true` and hide all details about the error.

Let's use the following:

```js
assert.ifError(err);
done();
```

test/acl.test.js
```diff
@@ -75,23 +77,22 @@ describe('security scopes', function() {
...
-              function(err, perm) {
-                assert(perm.permission === ACL.DENY);
-              });
+            async.parallel([
+              clbk => Scope.checkPermission('testModelScope', 'testModel', ACL.ALL, ACL.ALL, clbk),
+              clbk => Scope.checkPermission('testModelScope', 'testModel', 'name', ACL.ALL, clbk),
+              clbk => Scope.checkPermission('testModelScope', 'testModel', 'name', ACL.READ, clbk),
+              clbk => Scope.checkPermission('testModelScope', 'testModel', 'name', ACL.WRITE, clbk),
+            ], (err, perms) => {
+              assert(!err);
```
bajtos: Let's simplify this code and simply forward the error to `done` callback. It will also provide a more helpful error message.

```js
if (err) return done(err);
```

test/acl.test.js
```diff
@@ -75,23 +77,22 @@ describe('security scopes', function() {
...
+              assert(perms.length === 4);
```
bajtos: `assert.equal` please.

test/acl.test.js
```diff
@@ -75,23 +77,22 @@ describe('security scopes', function() {
...
+              if (perms.length === 4) {
+                assert(perms[0].permission === ACL.DENY); // because name.WRITE == DENY
+                assert(perms[1].permission === ACL.DENY); // because name.WRITE == DENY
+                assert(perms[2].permission === ACL.ALLOW);
+                assert(perms[3].permission === ACL.DENY);
```
bajtos: Please use `assert.equal` and add a custom message explaining which scenario is being asserted on, so that the test failure tells what exactly went wrong.

Alternatively, and I think this is a better solution, assert on the entire content of the array, for example:

```js
const actualPermissions = perms.map(p => p.permission);
assert.deepEqual(actualPermissions, [
  ACL.DENY, // because name.WRITE == DENY
  ACL.DENY, // // because name.WRITE == DENY
  ACL.ALLOW,
  ACL.DENY
]);
```

That way you don't need to assert on the length of the array. Instead of a not very helpful error "expected 3 to be 4", `deepEqual` will show the full array with permission values.

bajtos:
Changes I made after rebasing on top of the latest master: b47517

test/acl.test.js
```diff
@@ -48,22 +44,28 @@ describe('security scopes', function() {
...
           principalType: ACL.SCOPE, principalId: scope.id,
           model: 'User', property: ACL.ALL,
           accessType: ACL.ALL, permission: ACL.ALLOW,
         }, function(err, resource) {
-          Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, checkResult);
-          Scope.checkPermission('userScope', 'User', 'name', ACL.READ, checkResult);
+          async.parallel([
+            cb => Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, cb),
+            cb => Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, cb),
```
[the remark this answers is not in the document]
moshemal: I didn't change, add or remove any assertion. I just refactored it to work with done callback and async.parallel. I tried to be close to the original version as much as possible.
Fixing the logic of the tests should be, in my opinion, in another PR, but I am good with fixing it in the same PR too.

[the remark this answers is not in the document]
bajtos: Well the original code says this:

```js
           Scope.checkPermission('userScope', 'User', ACL.ALL, ACL.ALL, checkResult);
           Scope.checkPermission('userScope', 'User', 'name', ACL.ALL, checkResult);
           Scope.checkPermission('userScope', 'User', 'name', ACL.READ, checkResult);
```

While the code you have proposed contains the first line twice. I have removed the duplicated statement.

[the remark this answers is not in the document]
bajtos: > Fixing the logic of the tests should be, in my opinion, in another PR, but I am good with fixing it in the same PR too.

Agreed 👍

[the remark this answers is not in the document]
bajtos: > Fixing the logic of the tests should be, in my opinion, in another PR

Agreed 👍

bajtos:
Landed 🎉  Thank you @moshemal for your first contribution! ❤️

Issue #3993, opened 2018-09-05
`npm i` fails on linux/s390x architecture


virkt25:
# Description/Steps to reproduce

```
loopback$ npm i

> phantomjs-prebuilt@2.1.16 install /home/tleah/loopback/node_modules/phantomjs-prebuilt
> node install.js

PhantomJS not found on PATH
Unexpected platform or architecture: linux/s390x
It seems there is no binary available for your platform/architecture
Try to install PhantomJS globally
npm WARN ajv-keywords@3.2.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-plugin-mocha@4.12.1 requires a peer of eslint@^2.0.0 || ^3.0.0 || ^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN sinon-chai@3.2.0 requires a peer of chai@^4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"s390x"})

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! phantomjs-prebuilt@2.1.16 install: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the phantomjs-prebuilt@2.1.16 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tleah/.npm/_logs/2018-09-05T15_57_01_179Z-debug.log
```

CloudNativeJS tests projects against different architectures and currently we fail on `linux/s390x` because of `phantomjs` install failing. It's only needed for tests so isn't an accurate representation as LoopBack works find on this architecture. 

There was an attempt to address this issue by skipping this dependency in this PR https://github.com/strongloop/loopback/pull/3962 but it doesn't fully address the issue at hand. 

# Expected result

Test suite passes (excluding tests requiring phantomjs).

# Additional information

virkt25:
Released v 3.22.3 which only attempts to install `phantomjs` on supported architectures. Waiting on confirmation from cloudnativejs to see if the issue is now resolved.

[...]

PR #3996, opened 2018-09-07
chore: use grunt to optionally install the dependency

virkt25:
connected to #3993

#### Related issues

### Checklist


package.json
```diff
@@ -96,13 +96,15 @@
...
     "mocha": "^5.2.0",
     "nyc": "^10.1.2",
-    "phantomjs-prebuilt": "^2.1.7",
     "sinon": "^6.1.4",
     "sinon-chai": "^3.2.0",
     "strong-error-handler": "^3.0.0",
     "strong-task-emitter": "^0.0.8",
     "supertest": "^3.0.0"
   },
+  "optionalDependencies": {
```
bajtos: Please not that `optionalDependencies` are installed in production, e.g. when an app depends on loopback. With your change in place, every LoopBack application is going to install phantomjs! That would be a poor user experience IMO.

We need to mark `phantomjs-prebuilt` as an optional dev-dependency instead.

Unfortunately, `npm` does not support optional dev-dependencies out of the box, at least as far as I can tell from searching the internet and the discussion in https://github.com/npm/npm/issues/3870.

A possible workaround is offered by this npm package https://www.npmjs.com/package/optional-dev-dependency

Considering that we have a single optional dev-dependency only, it may be simpler to simply execute `npm install --no-save phantomjs-prebuilt` from our Gruntfile. (Maybe check if the package has not been already installed, to speed up repeated execution of `npm t` for framework developers practicing test-driven development.)

virkt25:
@slnode test please

PR #4025, opened 2018-10-11
Fix context propagation broken by async@2.x

### LINKED ISSUE #3966: LoopBack 2.x branch test failures

2 tests are failing on `2.x` branch. Run `npm it`.

## Failing Tests

```
  1) loopback loopback.getCurrentContext preserves callback domain:
      Uncaught AssertionError: expected 'context not available' to equal 'a value stored in context'
      + expected - actual
      -context not available
      +a value stored in context
      
      at Test.<anonymous> (test/loopback.test.js:629:36)
      at Test.assert (node_modules/supertest/lib/test.js:179:6)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at emitCloseNT (net.js:1659:8)
      at _combinedTickCallback (internal/process/next_tick.js:135:11)
      at process._tickDomainCallback [as _tickCallback] (internal/process/next_tick.js:218:9)
  2) loopback.rest context propagation should enable context with loopback.rest:
     Error: expected 200 "OK", got 500 "Internal Server Error"
      at Test._assertStatus (node_modules/supertest/lib/test.js:266:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:281:11)
      at Test.assert (node_modules/supertest/lib/test.js:171:18)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at emitCloseNT (net.js:1659:8)
      at _combinedTickCallback (internal/process/next_tick.js:135:11)
      at process._tickDomainCallback [as _tickCallback] (internal/process/next_tick.js:218:9)
```

# Expected result

All tests should pass.

bajtos:
Rework the REST middleware to use a hand-written version of "async.eachSeries". Before this change, we were loosing CLS context when the application was relying on the REST middleware to load the context middleware.

This is fixing a problem introduced by post-1.0 versions of async, which we upgraded to via fea3b78.

#### Related issues

- close  #3966

### Checklist

### DIFF
--- a/server/middleware/rest.js
+++ b/server/middleware/rest.js
@@ -73,8 +73,21 @@ function rest() {
     if (handlers.length === 1) {
       return handlers[0](req, res, next);
     }
-    async.eachSeries(handlers, function(handler, done) {
-      handler(req, res, done);
-    }, next);
+
+    executeHandlers(handlers, req, res, next);
   };
+}
+
+// A trimmed-down version of async.series that preserves current CLS context
+function executeHandlers(handlers, req, res, cb) {
+  var ix = -1;
+  next();
+
+  function next(err) {
+    if (err || ++ix >= handlers.length) {
+      cb(err);
+    } else {
+      handlers[ix](req, res, next);
+    }
+  }
 }


hacksparrow:
Why skip? Shouldn't we either make implementation changes to make it pass or remove the test itself with an explanation about its irrelevance?

bajtos:
> Why skip? Shouldn't we either make implementation changes to make it pass or remove the test itself with an explanation about its irrelevance?

You are right, I guess I should not have been lazy. The trouble is that continuous-local-storage is notoriously difficult to get right, but at the same time I don't know if it's ok to remove all CLS/context tests from LB 2.x. I'll give this more thoughts tomorrow.

bajtos:
@hacksparrow  I managed to track down the problem to a difference in behavior of `async` module between the version 0.9 which was used at the time we wrote the tests, and 1.x/2.x which we use now. I rewrote the relevant part of our code to use our own simplified version of `async.eachSeries`.

LGTY?

hacksparrow:
Looks good. 17 checks are failing, though.

bajtos:
> 17 checks are failing, though.

Most of them are outside of our control and have been failing for long time.

I checked `loopback-getting-started`, is has failed on `nsp check` (unrelated to this pull request). `loopback-connector-soap` is failing with "Connection fails: Error: connect ETIMEDOUT 139.59.176.153:80" (unrelated). And so on.

PR #4065, opened 2018-11-12
Set juggler configs in options by default

raymondfeng:
### Description

This PR allows set `prohibitHiddenPropertiesInQuery`, `maxDepthOfQuery`, and `maxDepthOfData` in `options` for CRUD operations.

#### Related issues

https://github.com/strongloop/loopback-datasource-juggler/pull/1662

### Checklist

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -470,14 +470,14 @@ module.exports = function(registry) {
 
     if (options.accepts) {
       options = extend({}, options);
-      options.accepts = setupOptionsArgs(options.accepts);
+      options.accepts = setupOptionsArgs(options.accepts, this);
     }
 
     this.sharedClass.defineMethod(name, options);
     this.emit('remoteMethodAdded', this.sharedClass);
   };
 
-  function setupOptionsArgs(accepts) {
+  function setupOptionsArgs(accepts, modelClass) {
     if (!Array.isArray(accepts))
       accepts = [accepts];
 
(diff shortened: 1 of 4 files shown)


raymondfeng:
I thought about setting default values for such flags in `createOptionsViaModelMethod` but found the following issues:

Method level `options.prohibitHiddenProperties` takes precedence over model/datasource level settings. As a result, we cannot override such defaults without customizing `createOptionsFromRemotingContext`. I think it's confusing that we cannot control such settings at model/datasource level.

> One of the big benefits I see in this approach: because security-related flags are configured via options, LB app developers can easily tweak them e.g. via custom options factory (see Customize the value provided to options in our docs).

With the PR, developers can still reset such flags by overriding `createOptionsFromRemotingContext`, such as:
```js
createOptionsFromRemotingContext(ctx) {
  return {prohibitHiddenProperties: true};
}
```

I'm open to a slightly different form, for example:

```js
const REMOTE_OPTIONS = {
  defaultJugglerSettings: {
    prohibitHiddenProperties: true
  }
}
```

raymondfeng:
@bajtos I reworked the patch to honor datasource/model settings and configure reasonable defaults in `options`.

lib/model.js
```diff
@@ -470,36 +470,48 @@ module.exports = function(registry) {
...
+     * Configure default values for juggler settings to protect user-supplied
+     * input from attacking juggler
+     */
+    var DEFAULT_OPTIONS = {
+      // Default to `true` so that hidden properties cannot be used in query
+      prohibitHiddenPropertiesInQuery: ModelCtor._getProhibitHiddenPropertiesInQuery({}, true),
+      // Default to `12` for the max depth of a query object
+      maxDepthOfQuery: ModelCtor._getMaxDepthOfQuery({}, 12),
+      // Default to `32` for the max depth of a data object
+      maxDepthOfData: ModelCtor._getMaxDepthOfData({}, 32),
```
bajtos: Lovely 👍

test/model.test.js
```diff
@@ -1028,7 +1062,6 @@ describe.onServer('Remote Methods', function() {
           return User.login(CREDENTIALS);
         }).then(function(token) {
           accessToken = token;
-          userId = token.userId;
```
bajtos: This change seems rather unrelated, could you please remove it (revert this line)?

PR #4108, opened 2019-01-08
Fix crash when modifying a user with an unknown id

horstmannmat:
Signed-off-by: Matheus Horstmann <mch15@inf.ufpr.br>

### Description
Fix the server crashes when try to PUT a user with a new  or a deleted id

#### Related issues

- fixes #4105

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

Issue #4105, opened 2019-01-07
The server crashes when PUT a extended user model with a new or deleted id 


horstmannmat:
# Description/Steps to reproduce
When I try to PUT extended model from User with new a id, or one that was previously deleted, the server crashes;

# Expected result

The server SHOULD NOT crash.

# Additional information
```
/app/node_modules/loopback/common/models/user.js:1361
        emailChanged = ctx.instance.email !== ctx.hookState.originalUserData[0].email;
                                                                                ^

TypeError: Cannot read property 'email' of undefined
    at /app/node_modules/loopback/common/models/user.js:1361:81
    at /app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/dao.js:2117:9
    at /app/node_modules/async/dist/async.js:1140:9
    at /app/node_modules/async/dist/async.js:473:16
    at eachOfArrayLike (/app/node_modules/async/dist/async.js:1057:9)
    at eachOf (/app/node_modules/async/dist/async.js:1117:5)
    at _asyncMap (/app/node_modules/async/dist/async.js:1133:5)
    at Object.map (/app/node_modules/async/dist/async.js:1122:16)
    at allCb (/app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/dao.js:2031:13)
    at /app/node_modules/loopback-connector/lib/sql.js:1452:7
    at /app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/observer.js:250:22
    at doNotify (/app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/observer.js:155:49)
    at PostgreSQL.ObserverMixin._notifyBaseObservers (/app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/observer.js:178:5)
    at PostgreSQL.ObserverMixin.notifyObserversOf (/app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/observer.js:153:8)
    at cbForWork (/app/node_modules/loopback/node_modules/loopback-datasource-juggler/lib/observer.js:240:14)
    at /app/node_modules/loopback-connector/lib/sql.js:643:9
    at Query.<anonymous> (/app/node_modules/loopback-connector-postgresql/lib/postgresql.js:152:7)
    at Query.handleReadyForQuery (/app/node_modules/pg/lib/query.js:125:10)
    at Connection.<anonymous> (/app/node_modules/pg/lib/client.js:191:19)
[...]

PR #4111, opened 2019-01-11
Fix crash when modifying an unknown user [2.x]

bajtos:
This is a back-port of #4108, see also #4105.

/cc @horstmannmat

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -897,7 +897,14 @@ module.exports = function(User) {
       });
       var emailChanged;
       if (ctx.instance) {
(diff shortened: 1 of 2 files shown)


bajtos:
The build failure on Windows is unrelated.

```
Building remotely on windows-64-0.10 (build-windows windows-0.10 x64 windows-64 0.10 windows) in workspace C:\jenkins\workspace\loopback\d60c4a15
Wiping out workspace first.
java.nio.file.FileSystemException: C:\jenkins\workspace\loopback\d60c4a15\xunit.xml: The process cannot access the file because it is being used by another process.
```

bajtos:
Published in `loopback@2.41.2`.

PR #4120, opened 2019-01-22
Fix Model.exists() working for remote connector

angfal:
### Description
The `GET /Models/:id/exists` service doesn't work for remote connector. It proxies to `GET /Models/:id/exists?id=${id}` endpoint except the required

### Checklist

### DIFF
--- a/lib/persisted-model.js
+++ b/lib/persisted-model.js
@@ -724,7 +724,8 @@ module.exports = function(registry) {
       description: 'Check whether a model instance exists in the data source.',
       accessType: 'READ',
       accepts: [
-        {arg: 'id', type: 'any', description: 'Model id', required: true},
+        {arg: 'id', type: 'any', description: 'Model id', required: true,
+          http: {source: 'path'}},
         {arg: 'options', type: 'object', http: 'optionsFromRequest'},
       ],
       returns: {arg: 'exists', type: 'boolean'},
--- a/test/util/model-tests.js
+++ b/test/util/model-tests.js
@@ -235,6 +235,27 @@ module.exports = function defineModelTestsWithDataSource(options) {
       });
     });
 
+    describe('Model.exists(id, [callback])', function() {
+      it('returns true when the model with the given id exists', function(done) {
+        User.create({first: 'max'}, function(err, user) {
+          if (err) return done(err);
+          User.exists(user.id, function(err, exist) {
+            if (err) return done(err);
+            assert.equal(exist, true);
+            done();
+          });
+        });
+      });
+
+      it('returns false when there is no model with the given id', function(done) {
+        User.exists('user-id-does-not-exist', function(err, exist) {
+          if (err) return done(err);
+          assert.equal(exist, false);
+          done();
+        });
+      });
+    });
+
     describe('Model.findById(id, callback)', function() {
       it('Find an instance by id', function(done) {
         User.create({first: 'michael', last: 'jordan', id: 23}, function() {


slnode:

bajtos:
Thank you for the pull request. I agree that `exists` method should work with the remote connector too.

But! IIUC, you are proposing a breaking (backwards-incompatible) change :(

- Currently, the method `exists` is exposed as `GET /Models/exists?id=:id`.
- Your patch is changing the URL to `GET /Models/:id/exists`.

Can we find a backwards-compatible fix please?

angfal:
Agreed, that the fix can be incompatible

But during this line:
https://github.com/strongloop/loopback/blob/b77907ffa59c7031fcb3bb6dbff96894bc597ba4/lib/persisted-model.js#L732

And loopback documentation:
https://loopback.io/doc/en/lb3/Exposing-models-over-REST.html

It looks like `/Models/:id/exists`

angfal:
@bajtos The fix looks like backward compatible, I made some tests based on mongodb connector:

```
it('should make some tests', async () => {
    const feature = await Feature.create({...});

    const res1 = await request(app)
      .get(`${apiPath}/Features/${feature.id}/exists`)
      .set('Authorization', accessTokenId);
    // res1: 200, exists = true

    const res2 = await request(app)
      .get(`${apiPath}/Features/exists?id=${feature.id}`)
      .set('Authorization', accessTokenId);
    // res2: 404, MODEL_NOT_FOUND

    const res3 = await request(app)
      .get(`${apiPath}/Features/:id/exists?id=${feature.id}`)
      .set('Authorization', accessTokenId);
    // res3: 200, exists = false
  });
```

angfal:
@bajtos When you send `GET /Models/exists?id=:id` loopback thinks that it's `Model.findById(...)` endpoint (`GET /Models/:id?id=:id`) and trying to find Model with `id = 'exists'`. As the result, `MODEL_NOT_FOUND` error

bajtos:
Thank you for investigating. I need to look into this myself too, to build a better understanding of how LB works now and what your proposed patch will change. I'll try to do that next week.

lib/persisted-model.js
```diff
@@ -724,7 +724,8 @@ module.exports = function(registry) {
       description: 'Check whether a model instance exists in the data source.',
       accessType: 'READ',
       accepts: [
-        {arg: 'id', type: 'any', description: 'Model id', required: true},
+        {arg: 'id', type: 'any', description: 'Model id', required: true,
+          http: {source: 'path'}},
```
bajtos: When building Swagger Spec describing LoopBack REST API, we are always converting arguments that are found in the URL path into `http: {source: 'path'}}` - see [lib/specgen/route-helper.js#L296-L29](https://github.com/strongloop/loopback-swagger/blob/fe5cfd7a11fb4cb0c9804db510d461a3d2ad53a6/lib/specgen/route-helper.js#L296-L298).

On strong-remoting's side, when extracting remote method arguments from the incoming requests:
 - If the http source is 'path', the value is read from `req.params`, see [HttpContext.prototype.buildArgs](https://github.com/strongloop/strong-remoting/blob/dde88ef2d05f6e60ca8ffa3295b35926b6424481/lib/http-context.js#L158-L162)
- If the argument does not have any http source set, we try to read it from `req.params` first, which is the same behavior. See [HttpContext.prototype.getArgByName](https://github.com/strongloop/strong-remoting/blob/dde88ef2d05f6e60ca8ffa3295b35926b6424481/lib/http-context.js#L226-L236)

My conclusion is that the proposed change is safe and should not break any existing applications 👍

test/util/model-tests.js
```diff
@@ -235,6 +235,19 @@ module.exports = function defineModelTestsWithDataSource(options) {
       });
     });
 
+    describe('Model.exists(id, [callback])', function() {
+      it('Check existing of a model instance in the attached data source', function(done) {
```
bajtos: Test names should read like English sentences. This one does not:

```
Model.exists(id, [callback]) Check existing of a model instance in the attached data source
```

To speed up the process, I'll edit the test name myself and force-push into your feature branch.

test/util/model-tests.js
```diff
@@ -235,6 +235,27 @@ module.exports = function defineModelTestsWithDataSource(options) {
...
+          if (err) return done(err);
+          User.exists(user.id, function(err, exist) {
+            if (err) return done(err);
+            assert.equal(exist, true);
+            done();
+          });
+        });
+      });
+
+      it('returns false when there is no model with the given id', function(done) {
```
bajtos: I have also added a test to verify what happens when the target model does not exist - we want to receive successful response with `false` value, not a `404 Not Found` error.

bajtos:
Landed, thank you @angfal for the contribution! ❤️

PR #4148, opened 2019-02-28
fix: `resetPasswordRequest` event is never emitted as callback is called Prematurely

gaurav21r:
### Description
Password reset depends on the User Model firing the `resetPasswordRequest` event with a temporary Access Token. The event is never fired as the callback `cb` is called _before_ the event is emitted.

This is a simple fix to just move the `cb()` after the event emit.

Need it urgently for a project! Thanks!

#### Related issues

None

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -992,13 +992,13 @@ module.exports = function(User) {
         if (err) {
           return cb(err);
         }
-        cb();
         UserModel.emit('resetPasswordRequest', {
           email: options.email,
           accessToken: accessToken,
           user: user,
           options: options,
         });
+        cb();
       }
     });
 


slnode:

bajtos:
@gaurav21r thank you for the pull request. I believe the current implementation is intentional. First we return back the HTTP response, then we fire the event

bajtos:
@gaurav21r What problem are you trying to solve? How can we reproduce it?

gaurav21r:
@bajtos it's simply not firing in my project, I just revered the order and I'm getting the HTTP response _and_ the event fired. Let me try to run the tests you mentioned and also the `user-management-example` and get back to you?

zbarbuto:
@gaurav21r It would be good if you could add a unit test that fails without your change and passes with it to give an idea of what might be going on.

Cheers

Issue #4180, opened 2019-04-12
Mark LB 2.x as EOL


bajtos:
At the end of April 2019, LoopBack 2.x goes end of life. We need to reflect this change in our documentation, project templates, etc.

## Acceptance criteria

- [x] Remove documentation for LB 2.x from our docs.

  _Alternative: keep the docs in, change the warning "LoopBack 2.x is the maintenance long-term support (LTS) release. " to mention that 2.x is no longer supported._

- [x] Remove 2.x from the list of LB versions offered by loopback-workspace. This list is used by generator-loopback and loopback-cli. 

  See [Workspace.getAvailableLBVersions()](https://github.com/strongloop/loopback-workspace/blob/6d6246777f5cb94b313d69793408a322f1d99564/common/models/workspace.js#L63-L69)

- [x] Review READMEs of all LB2/LB3 packages, update LTS section to clearly state that modules linked to LB 2.x is no longer supported. (For example, Explorer uses different major versions, 2.x hasn't been supported for ages.) 

  While changing READMEs, update LTS dates of the current Active LTS release to match the changes announced in https://strongloop.com/strongblog/lb3-extended-lts/

   Publish a new version of updated packages.

dhmlau:
> Remove documentation for LB 2.x from our docs.
> Alternative: keep the docs in, change the warning "LoopBack 2.x is the maintenance long-term support (LTS) release. " to mention that 2.x is no longer supported.

I'd prefer we mention in the docs that 2.x is no longer supported than removing it from the docs.  I think we won't stop users from using 2.x, even though it's highly recommended to move away from it.

dhmlau:
@bajtos, for `Review READMEs of all LB2/LB3 packages` (point 3 in the list), I realized that the connector repos don't have the LTS information.  So, I'm going to update the READMEs for the list below.  If you have a complete list or I miss any, please let me know. Thanks! 
- [ ] loopback-component-explorer
- [ ] strong-remoting
- [ ] loopback-workspace
- [ ] loopback-cli
- [ ] generator-loopback
- [ ] loopback-datasource-juggler

dhmlau:
Closing as done.

PR #4188, opened 2019-04-29
update README for 2.x EOL

dhmlau:
### Description
Related to https://github.com/strongloop/loopback/issues/4180
- Update the LTS schedule
- Update to state 2.x has already reached EOL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 [![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/strongloop/loopback?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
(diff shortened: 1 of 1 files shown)


bajtos:
I think it's even more important to update README in our `master` branch too, that's the content rendered when people arrive at https://github.com/strongloop/loopback and https://www.npmjs.com/package/loopback

dhmlau:
@bajtos, thanks for your review.

> If you prefer to keep the information about 2.x EOL at the top, then I am fine with that too, but please update the sentence "LoopBack 2.x is now on maintenance LTS".

Thanks. I'll update the sentence above the LTS table as well.  I put it at the top because the `master` branch has it at the top too. :) 

> I think it's even more important to update README in our master branch too

Yes, I'll do it in a separate PR for `master`.

Issue #4195, opened 2019-05-13
User model login endpoint: security issue with possible query injection


gabjauf:
Hello Loopback!

I found some kind of vulnerability on POST /Users/login route from the default User model:

# Description/Steps to reproduce

Send the following to /Users/login as credentials
```js
{
	"email": {"neq": "foo" },
	"password": "anything you want"
}
```

When inspecting the resulting user here https://github.com/strongloop/loopback/blob/4b3a3a347c9176f1a86b00d229fc942bdc60c794/common/models/user.js#L254 , this results in the first user in the database being returned, which means we now only need to find the password.

Found in loopback v3.25.1

# Expected result

We would expect the email to be forced as a string.

Feel free to reach me if my description is unclear 😄

gabjauf:
We may also use the "regexp" functionality to test multiple users without having to know their exact email:
{
	"email": {"regexp": "^a" },
	"password": "anything you want"
}

hacksparrow:
@gabjauf I tried both the payloads, I am getting 401. Can you elaborate the steps?

hacksparrow:
Oh, I got what you mean. Taking a look. Thanks for reporting.

fabien:
@hacksparrow I was seeing a 401 as well - could you elaborate on 'I got what you mean'?

bajtos:
The fix has been published last week in `loopback@3.26.0`.

gabjauf:
Great works guys, thanks 👍

PR #4208, opened 2019-05-29
fix: disallow queries in username and email fields

hacksparrow:
### Description

Disables queries in the username and email fields while logging in.

#### Related issues

https://github.com/strongloop/loopback/issues/4195

### Checklist

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -232,18 +232,39 @@ module.exports = function(User) {
     var query = self.normalizeCredentials(credentials, realmRequired,
       realmDelimiter);
 
-    if (realmRequired && !query.realm) {
-      var err1 = new Error(g.f('{{realm}} is required'));
-      err1.statusCode = 400;
-      err1.code = 'REALM_REQUIRED';
-      fn(err1);
-      return fn.promise;
+    if (realmRequired) {
+      if (!query.realm) {
+        var err1 = new Error(g.f('{{realm}} is required'));
+        err1.statusCode = 400;
+        err1.code = 'REALM_REQUIRED';
+        fn(err1);
+        return fn.promise;
+      } else if (typeof query.realm !== 'string') {
+        var err5 = new Error(g.f('Invalid realm'));
+        err5.statusCode = 400;
+        err5.code = 'INVALID_REALM';
+        fn(err5);
+        return fn.promise;
+      }
     }
     if (!query.email && !query.username) {
       var err2 = new Error(g.f('{{username}} or {{email}} is required'));
       err2.statusCode = 400;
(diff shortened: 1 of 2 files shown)


fabien:
How can we make sure this only affects the login method? Are there any other methods affected?

bajtos:
> How can we make sure this only affects the login method? Are there any other methods affected?

@fabien Long time no see, it's great to see you are still following LoopBack. Thanks for chiming in! 

IIUC, @hacksparrow is adding the additional validation directly into `User.login` method, alongside checks like `if (realmRequired && !query.realm)` and `if (!query.email && !query.username)`.

I don't see how other methods could be affected. The only possibility I see is that another method will be affected by this change if it calls `User.login` under the hood and constructs a query for the username instead of providing the string value only. IMO, that is not a valid call of `login`. What's your opinion? Do you see any other edge cases we should be concerned about?

fabien:
@bajtos I'm still very actively using Loopback, and try to keep an eye on things ...

It feels to me like this issue is very similar to what is described here: https://blog.websecurify.com/2014/08/hacking-nodejs-and-mongodb.html - since Loopback uses an object based query format, like MongoDB does natively, it might have similar caveats. That's why I was wondering about this being a problem on a wider scale or not. What do you think?

hacksparrow:
@bajtos added check for realm too. Good to merge?

hacksparrow:
@fabien the change is done within the `User.login` method, so other methods will not be affected.

fabien:
> @fabien the change is done within the `User.login` method, so other methods will not be affected.

@hacksparrow yes, I can see that. I was just wondering if the same issue might be popping elsewhere. Are we sure there aren't any other endpoints that allow arbitrary queries where it's not appropriate?

bajtos:
> It feels to me like this issue is very similar to what is described here: https://blog.websecurify.com/2014/08/hacking-nodejs-and-mongodb.html - since Loopback uses an object based query format, like MongoDB does natively, it might have similar caveats. That's why I was wondering about this being a problem on a wider scale or not. What do you think?

Thank you for sharing the link. I agree with you that there may be other places vulnerable to this attack. Let's not discuss such vulnerabilities in public please. I opened a new issue in the private repository accessible to project maintainers only, let's continue the discussion there: https://github.com/strongloop/private-loopback/issues/3

sujeshthekkepatt:
Can we get the patch any time soon?

hacksparrow:
Added to `user.test.js`.

hacksparrow:
@sujeshthekkepatt a new version with this fix will be released tomorrow, if not today.

sujeshthekkepatt:
Cool.  Looking forward to it

PR #4241, opened 2019-08-09
Updated "isemail" package to ver. 3.2.x

### LINKED ISSUE #4239: Update isemail dependency
Can you please update `isemail` to version **3.x.x** so we can use this lib also as an FE validation. Currently default User model can't disable validation for email, so at least update this lib to the new version, because versions below 3 can't be used in a browser.

Sarbinski:
### Description

isemail version 3.x.x is completely rewritten and not depend on NodeJS native modules. 

Resolve #4239

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "ejs": "^2.3.1",
     "express": "^4.14.0",
     "inflection": "^1.6.0",
-    "isemail": "^2.2.1",
+    "isemail": "^3.2.0",
     "loopback-connector-remote": "^3.0.0",
     "loopback-datasource-juggler": "^3.28.0",
     "loopback-filters": "^1.0.0",


slnode:

bajtos:
Thank you for the pull request. It looks is v3 of `isemail` does not work in the browser out of the box and thus it's breaking out tests verifying LoopBack functionality in browser (see the failed CI builds on Travis CI). In order to land this PR, we need to find a way how to make `isemail` work. It may be enough to tell `browserify` to ignore certain dependencies when creating the browser bundle, see the current config:

https://github.com/strongloop/loopback/blob/9b1d4885ab869d977bc7d464aa861b4f88691862/package.json#L109-L117

bajtos:
On the second thought, it's also possible that the new version of isemail is using modern JS constructs from ES6+ that are not supported by PhantomJS that we use to run our tests in the browser. This could be solved e.g. by adding Babel transpiler to our test setup or by switching from PhantomJS to Chrome Headless (or Firefox Headless). Either way, such change should be ideally made in a new pull request, or at least in a standalone commit clearly separated from the is-email update.

bajtos:
Rebased on top of the latest master to apply the CI fix from https://github.com/strongloop/loopback/pull/4262.

bajtos:
Landed, thank you for the contribution! ❤️

PR #4293, opened 2019-12-12
chore: update Juggler version

hacksparrow:
Updated Juggler to 4.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "inflection": "^1.6.0",
     "isemail": "^3.2.0",
     "loopback-connector-remote": "^3.0.0",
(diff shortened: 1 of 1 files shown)


bajtos:
Updating juggler to 4.x is a breaking change that's against our Long Term Support policy, see https://loopback.io/doc/en/contrib/Long-term-support.html

bajtos:
It was our intentional decision to keep `loopback@3` to depend on `loopback-datasource-juggler@3`, the same way as `loopback@2` depends on `loopback-datasource-juggler@2`.

I am open to discuss what would be the impact (both positive and negative) of upgrading LB3 to juggler v4. Unfortunately, your pull request description does not provide any clues.

PR #4119, opened 2019-01-21
Support middleware injected by AppDynamics

mschnee:
### Description
Much like the workaround for `NewRelic`, `appd` replaces the express handle with its own, and stores it in `handle.__appdynamicsProxyInfo__.orig`.  This causes `proto._findLayerByHandler` to fail.

This pull request was originally created by @mikewli and @gconaty - I have updated it to include unit tests.

### Checklist

- [x] New tests added or existing tests modified to cover all changes
  - Unit tests have been added
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
  - All lint tests passes

PR #4351, opened 2021-09-22
Support middleware injected by OpenTelemetry Express Instrumentation plugin

a-legrand:
[OpenTelemetry Express Instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-express) injects a proxy object into the router stack, which it
uses for its tracing.  This is similar to how NewRelic adds a sentinel handler to the router stack. This commit adds a similar workaround so that loopback can find the original layer.

I made the variable more generic to just test if shimmer is used to wrap the handler, as this library can be used by other libraries.

See also #4119

### Checklist
* [x]  New tests added or existing tests modified to cover all changes
  * Unit tests have been added
* [x]  Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
  * All lint tests passes


arousselotvi (approved):

adambusfy123:
Whats the status here?

